PR #10, opened 2015-02-05
Improve README.md

jamiecobbett:
I applied our styleguide: https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,48 @@
 # Publishing API
 
-An application to route requests between multiple content-store
-endpoints based on whether they're live or in other states.
+This is a Go application that proxies requests to multiple content-stores. Our
+use case for this is to keep two copies of the frontend of GOV.UK running: one
+which the public sees and another which is only used by people working on
+content to review work in progress.
+
+Publishing apps will talk to the publishing-api rather than the content-store.
+Normal publishing requests are forwarded to both the live and draft
+content-stores, whereas draft documents would only be forwarded to the draft
+content-store.
+
+### Dependencies
+
+- [alphagov/url-arbiter](https://github.com/alphagov/url-arbiter) - publishing-api will take over content-store's job of updating url-arbiter. This is to prevent race conditions as two content-stores try to register with the same url-arbiter.
+- [alphagov/content-store](https://github.com/alphagov/content-store) - publishing-api's function is to proxy requests to multiple content-stores (eg a draft and a live content-store)
+- [gom](https://github.com/mattn/gom) - to vendorise dependencies
+
+### Running the application
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,50 @@
...
+content to review work in progress.
+
+Publishing apps will talk to the publishing-api rather than the content-store.
+Normal publishing requests are forwarded to both the live and draft
+content-stores, whereas draft documents would only be forwarded to the draft
+content-store.
+
+### Dependencies
+
+- [alphagov/url-arbiter](https://github.com/alphagov/url-arbiter) - ?
```
jamiecobbett: It wasn't obvious to me why this app talks to `url-arbiter`. Can someone fill me in?

README.md
```diff
@@ -1,13 +1,50 @@
...
-You can run the tests locally by running the following: `make`. This
-will use the gom tool to vendorise dependencies into a folder within
-the project.
+You can run the tests locally with: `make`. This will use the gom tool to
+vendorise dependencies into a folder within the project.
 
 You can download the gom tool by running:
 `go get github.com/mattn/gom`.
+
+### Any deviations from idiomatic Rails/Go etc. (optional)
```
jamiecobbett: I'm not familiar enough with `Go` to tell whether what we've got is idiomatic or not.

README.md
alext: It seemed sensible for the url-arbiter integration to be done at this point.  Otherwise some content will be registered multiple times with the url-arbiter (once by each content-store).  This is redundant, and can also lead to race conditions with the url-arbiter uniqueness validation on path (this would only be the case for new items...).

Once this app is live and being used for all publishing to the content-store(s), we will remove the url-arbiter registration from content-store.

README.md
alext: I don't think there's anything that's not idiomatic, and if there is, we should fix it.

alext:
This looks good to me.  Good to merge once comments are addressed.

README.md
jamiecobbett: What happens if you want to draft a replacement piece of content in a different publishing app? I guess this is something that is not really supported anyway right now.

alext: That's deliberately blocked.  We want a publishing app to be able to create a draft of a piece of content under a path, and know that it will be allowed to publish it when the time comes.

What we will probably need at some point is an API endpoint in url-arbiter to allow a publishing app to release a path that it's no longer using.  We can deal with that when the actual need arises though.

jamiecobbett:
Thanks, rebased.

PR #15, opened 2015-02-11
Test cleanup

jamiecobbett:
Hopefully nothing controversial, but I am finding my way with Go as I... go. I recommend reading through each commits as the combined diff is noisy.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 
 	. "github.com/alphagov/publishing-api"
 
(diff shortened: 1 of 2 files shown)


integration_test.go
```diff
@@ -141,98 +165,83 @@ var _ = Describe("Integration Testing", func() {
...
+			Expect(response.StatusCode).To(Equal(http.StatusBadRequest))
 		})
 	})
+})
 
-	It("returns a 400 error if given invalid JSON", func() {
-		url := testPublishingAPI.URL + "/content" + "/foo/bar"
-		request, err := http.NewRequest("PUT", url, strings.NewReader("i'm not json"))
-		Expect(err).To(BeNil())
+func DoRequest(verb string, url string, body []byte) (resp *http.Response) {
```
alext: Minor style point: It's a little unclear that the return variable is named, but then that name isn't used.  Probably best to not name it in this case:

```go
func DoRequest(verb string, url string, body []byte) *http.Response {
```

alext:
Just one minor point, otherwise it looks good to me.

integration_test.go
jamiecobbett: Fixed and rebased

PR #17, opened 2015-02-11
gofmt'd integration_test.go for consistency

vinayvinay:


jamiecobbett:
I wonder if this should be part of the (CI) build? In the same way that we lint the puppet repo.

PR #18, opened 2015-02-12
Make builds fail when files are not gofmt'd

vinayvinay:
based on @jamiecobbett's [suggestion](https://github.com/alphagov/publishing-api/pull/17#issuecomment-73937716) to have CI tell us if files are not gofmt'd.

we're looking through all files with extension .go in the project directory, excluding the _vendor directory, and running `gofmt` against them. if there are files pending formatting, we return a non-zero exit status, which causes the make target to fail, and in turn the build to go red.

this branch build is failing because it is based on master which has am unfmt'd file, which will be fixed by https://github.com/alphagov/publishing-api/pull/17.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BINARY := publishing-api
 ORG_PATH := github.com/alphagov
 REPO_PATH := $(ORG_PATH)/$(BINARY)
 
-all: test build
+all: check_fmt test build
 
 deps:
 	gom install
@@ -25,3 +25,6 @@ run: build
 
 clean:
 	rm -rf bin $(BINARY) _vendor
+
+check_fmt:
+	./check_fmt.sh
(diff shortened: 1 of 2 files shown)


alext:
This is great...

We could improve the error output by running something like the following over all the failed files:
```
gofmt integration_test.go | diff -u integration_test.go -
```

vinayvinay:
@alext: thanks! i've used `gofmt -d` which gives exactly that. i thought it was enough to know which files were failing gofmt, because one could simply go back to your editor and run a few keybindings to fix the formatting without knowing what's unformatted. initially it'll be good to highlight what the differences are to familiarise ourselves with the nuances in Go syntax.

alext:
Good stuff, although this is revealing that we currently have lint.  This should be rebased after #17 is merged.

vinayvinay:
@alext: rebased. that was the idea, TDDing CI branch scripts :wink:.

PR #19, opened 2015-02-12
Publish intents

jamiecobbett:
publishing-api needs to support the existing API of content-store, which this PR would complete. The publishing-api vhost currently is an alias for content-store. Once this is merged, publishing-api should be capable of proxying content-store and taking over the publishing-api vhost.

Currently, content-store registers content with url-arbiter, but not publish-intents. This PR registers both content and publish-intents with url-arbiter. We will remove the url-arbiter integration from content-store to avoid them duplicating the functionality for content.

Further context:
Our intent is to have publishing-api handle url-arbiter registration to avoid two content-stores (one for `draft` and one for `live`) racing against each other.

### DIFF
--- a/Gomfile
+++ b/Gomfile
@@ -1,4 +1,6 @@
 gom 'github.com/codegangsta/negroni', :commit => 'c64702ca03d0f858ab7866638b69e85f0d4c6ee7'
+gom 'github.com/gorilla/mux', :commit => 'e444e69cbd2e2e3e0749a2f3c717cec491552bbf'
+gom 'github.com/gorilla/context', :commit => '215affda49addc4c8ef7e2534915df2c8c35c6cd'
 gom 'gopkg.in/unrolled/render.v1', :commit => '40ac5716c7e0bffcc440aac649ea4b880a5d7735'
 gom 'github.com/alext/tablecloth', :commit => '7f49e40'
 
--- a/content_store_handler.go
+++ b/content_store_handler.go
@@ -6,12 +6,13 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/gorilla/mux"
+
 	"github.com/alphagov/publishing-api/contentstore"
 	"github.com/alphagov/publishing-api/urlarbiter"
 )
 
 type ContentStoreRequest struct {
-	BasePath      string `json:"base_path"`
 	PublishingApp string `json:"publishing_app"`
 }
 
@@ -28,6 +29,8 @@ func NewContentStoreHandler(arbiterURL, contentStoreURL string) *ContentStoreHan
 }
 
 func (cs *ContentStoreHandler) PutContentStoreRequest(w http.ResponseWriter, r *http.Request) {
+	urlParameters := mux.Vars(r)
+
 	requestBody, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		renderer.JSON(w, http.StatusInternalServerError, err)
@@ -45,7 +48,7 @@ func (cs *ContentStoreHandler) PutContentStoreRequest(w http.ResponseWriter, r *
 		return
 	}
 
-	if !cs.registerWithURLArbiter(contentStoreRequest.BasePath, contentStoreRequest.PublishingApp, w) {
+	if !cs.registerWithURLArbiter(urlParameters["base_path"], contentStoreRequest.PublishingApp, w) {
 		// errors already written to ResponseWriter
 		return
 	}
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -25,11 +26,14 @@ func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
 	renderer.JSON(w, http.StatusOK, map[string]string{"status": "OK"})
 }
 
-func BuildHTTPMux(arbiterURL, contentStoreURL string) *http.ServeMux {
-	httpMux := http.NewServeMux()
-	httpMux.HandleFunc("/healthcheck", HealthCheckHandler)
+func BuildHTTPMux(arbiterURL, contentStoreURL string) *warp.ServeMux {
```
alext: It's probably cleaner for this to return a `http.Handler` as opposed to a pointer to a specific type.  All we need in the caller is something that implements that interface.

content_store_handler.go
```diff
@@ -52,20 +45,12 @@ func (cs *ContentStoreHandler) PutContentItem(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	if !cs.registerWithURLArbiter(path, contentStoreRequest.PublishingApp, w) {
+	if !cs.registerWithURLArbiter(contentStoreRequest.BasePath, contentStoreRequest.PublishingApp, w) {
```
alext: We don't want to take the path from the JSON.  In the case where the base_path in the JSON request body doesn't match the path in the request, we should use the path in the request (that's what content-store does at the moment).  

Alternatively, we could reject these requests as invalid.

jamiecobbett:
Rebased to use gorilla/mux, thanks @alext

jamiecobbett:
I've accidentally rebased lots of the commits together, so I'm going to go back over and fix them.

Gomfile
```diff
@@ -1,4 +1,5 @@
 gom 'github.com/codegangsta/negroni', :commit => 'c64702ca03d0f858ab7866638b69e85f0d4c6ee7'
+gom 'github.com/gorilla/mux', :commit => 'e444e69cbd2e2e3e0749a2f3c717cec491552bbf'
```
alext: `gorilla/mux` also depends on `gorilla/context`, so we should pin that here as well.

This sort of thing wouldn't be an issue if we were using godep instead of gom, but we need to figure out how to handle the `GOPATH` properly before we can switch.

jamiecobbett: Fixed and rebased.

main.go
jamiecobbett: Fixed and rebased.

content_store_handler.go
jamiecobbett: The first option would mean that a mismatched base_path would be registered with url-arbiter before being rejected by content-store. The second means adding extra knowledge and behaviour to this app.

I'm leaning towards the latter.

jamiecobbett:
Updated to use the base_path from the request URL, not the request body.

main.go
```diff
@@ -25,11 +26,14 @@ func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
...
-	httpMux.HandleFunc("/healthcheck", HealthCheckHandler)
+func BuildHTTPMux(arbiterURL, contentStoreURL string) http.Handler {
+	httpMux := mux.NewRouter()
+	httpMux.Methods("GET").Path("/healthcheck").HandlerFunc(HealthCheckHandler)
 	contentStoreHandler := NewContentStoreHandler(arbiterURL, contentStoreURL)
-	httpMux.HandleFunc("/content/", contentStoreHandler.PutContentItem)
+	httpMux.Methods("PUT").PathPrefix("/content{base_path:.*}").HandlerFunc(contentStoreHandler.PutContentStoreRequest)
+	httpMux.Methods("PUT").PathPrefix("/publish-intent{base_path:.*}").HandlerFunc(contentStoreHandler.PutContentStoreRequest)
+	httpMux.Methods("GET").PathPrefix("/publish-intent{base_path:.*}").HandlerFunc(contentStoreHandler.GetContentStoreRequest)
+	httpMux.Methods("DELETE").PathPrefix("/publish-intent{base_path:.*}").HandlerFunc(contentStoreHandler.DeleteContentStoreRequest)
```
alext: Now that these are capturing the `base_path`, there's no need for them to be `PathPrefix`'s.

Also it may be worth updating the regex used to require a leading `/` on the base_path.  This would match the routing in content-store etc.

alext:
Small comment about the routing, otherwise I think this is good to merge.

main.go
jamiecobbett: Rebased to address the first suggestion and added the second in cb1d86e.

jamiecobbett:
:fireworks:

PR #20, opened 2015-02-18
Renames for better understanding

vinayvinay:
renaming to convey intent better and not to confuse with: http://golang.org/pkg/net/http/#Handler

a Handler in Go implements the Handler interface and defines a `ServeHTTP` method. you should be able to pass a Handler to `http.Handle("/", ContentStoreHandler)` and allow it to handle requests at that path.

in our case, `ContentStoreHandler` is not a `Handler`, but a struct that has `func`s which are later typed to `HandlerFunc`s in main.go: http://git.io/AeDs

since we're using it as a controller with functions that are more or less actions, `ContentStoreController` is a suitable name.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,18 +22,19 @@ var (
 	renderer = render.New(render.Options{})
 )
 
-func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	renderer.JSON(w, http.StatusOK, map[string]string{"status": "OK"})
-}
-
 func BuildHTTPMux(arbiterURL, contentStoreURL string) http.Handler {
 	httpMux := mux.NewRouter()
-	httpMux.Methods("GET").Path("/healthcheck").HandlerFunc(HealthCheckHandler)
+
+	httpMux.Methods("GET").Path("/healthcheck").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -22,18 +22,19 @@ var (
...
-	httpMux.Methods("GET").Path("/healthcheck").HandlerFunc(HealthCheckHandler)
-	contentStoreHandler := NewContentStoreHandler(arbiterURL, contentStoreURL)
-	httpMux.Methods("PUT").Path("/content{base_path:/.*}").HandlerFunc(contentStoreHandler.PutContentStoreRequest)
-	httpMux.Methods("PUT").Path("/publish-intent{base_path:/.*}").HandlerFunc(contentStoreHandler.PutContentStoreRequest)
-	httpMux.Methods("GET").Path("/publish-intent{base_path:/.*}").HandlerFunc(contentStoreHandler.GetContentStoreRequest)
-	httpMux.Methods("DELETE").Path("/publish-intent{base_path:/.*}").HandlerFunc(contentStoreHandler.DeleteContentStoreRequest)
+
+	httpMux.Methods("GET").Path("/healthcheck").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		renderer.JSON(w, http.StatusOK, map[string]string{"status": "OK"})
+	})
```
alext: :+1: Good call to inline this.

PR #21, opened 2015-02-19
Reorganise test code

vinayvinay:
#### why?
- we need to introduce a draft content store in our specs, so
  - avoid repetition, which will only increase with a draft content store
  - extract testhelpers to make it easier to have more test servers
- `/publish-intent` and `/content` endpoint behaviours are about to diverge

#### what?

- split integration_test.go according to resources
- extracted `integration_test` package to make code dry
  - contains integration tests
  - contains common assertion code
  - keep `TestRequestTracker` code separate
  - use `structs` to establish a contract for what's needed to perform assertions, or setup stub responses 
- retain some test code to make sure test code looks complete, and not too DRY
- make test data meaningful
- using `ghttp.Server`, and its assertions helped realise that we don't set the appropriate content-type on requests sent to URL Arbiter, so added that.


content_item_requests_test.go
```diff
@@ -0,0 +1,85 @@
+package main_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+
+	. "github.com/alphagov/publishing-api"
+	. "github.com/alphagov/publishing-api/testhelpers"
```
alext: Using dot imports is generally not a recommended practice (it merges the namespaces, and can lead to interesting issues as a result).

IMO it's reasonable to use it for the test framework itself (ie ginkgo and gomega) as these define a DSL to use in the tests, and are designed to be used in this manner (although some purists would even object to that).

Regarding these imports, these tests are only using a single function from the `main` package (`BuildHTTPMux`), so that could easily not be a dot import.

I'm more on the fence over dot importing the test helpers.  In other places (eg the router test suite), I've build test helpers in the same package as the test suite (although in separate files), which is effectively the same as they're all in the same namespace.

testhelpers/server_factory.go
```diff
@@ -0,0 +1,45 @@
...
+
+		body, err := ReadHTTPBody(r.Body)
+		Expect(err).To(BeNil())
+		if request.Body != "" {
+			Expect(body).To(MatchJSON(request.Body))
+		}
+
+		w.WriteHeader(response.Code)
+		fmt.Fprintln(w, response.Body)
+	}))
```
alext: This feel a bit like it's coupling the construction of a test server with the expectations of what requests will hit it.  As it stands, it's not possible to construct a test server that handles multiple different requests.

It may be better to construct a test server that records the requests made, and then allows inspection on them (eg in the router test suite[1])

An alternative approach would be to construct a test server that could then have expectations set on it - as done in the [contentstore client tests](https://github.com/alphagov/publishing-api/blob/master/contentstore/client_test.go#L36-L43).

[1] https://github.com/alphagov/router/blob/master/integration_tests/proxy_function_test.go#L149-L160 and https://github.com/alphagov/router/blob/master/integration_tests/backend_helpers.go#L44-L49

content_item_requests_test.go
vinayvinay: on the lines of code organisation in router, i've added a `integration_test` package that includes testhelpers and integration tests both.

testhelpers/server_factory.go
vinayvinay: yeah, that's true, i too wasn't happy with that coupling, but thought of taking small steps to restructuring this. turns out, using `ghttp` package was an excellent suggestion, didn't know about that. :+1:

alext:
:+1: This looks good to me

integration_tests/publish_intent_requests_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/alphagov/publishing-api"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/ghttp"
+)
+
+var _ = Describe("Publish Intent Requests", func() {
+	var (
+		contentItemJSON = `{
```
jamiecobbett: This isn't really a `contentItemJSON`, it's a publishIntentJSON. The content should really change too to be like: https://github.com/alphagov/content-store/blob/master/doc/publish_intents.md#publish-intent-format

integration_tests/publish_intent_requests_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+					Expect(testURLArbiter.ReceivedRequests()).To(HaveLen(1))
+					Expect(testContentStore.ReceivedRequests()).To(BeEmpty())
+
+					expectedResponse = HTTPTestResponse{Code: 409, Body: urlArbiterResponseBody}
+					assertSameResponse(actualResponse, &expectedResponse)
+				})
+			})
+
+			It("registers a path with URL arbiter and then publishes the content to the content store", func() {
```
jamiecobbett: `then publishes the content` maybe it should be `then pushes the publish intent to content store`?

vinayvinay:
@jamiecobbett: fair points, corrected in https://github.com/alphagov/publishing-api/commit/85bb448059a8c83295e51f12869c91565d6deb1c.

PR #22, opened 2015-02-23
Allow publishing applications to store draft content

vinayvinay:
**merge first: https://github.com/alphagov/publishing-api/pull/23** Split content_store_controller.go

https://trello.com/c/PbCzGBIB

#### Need
Content designers need a way to preview content they're drafting. When draft content is saved, publishing applications should be able to push that draft content to a draft instance of the content store, from where it can be previewed.

#### Implementation
0. Publishing API exposes a `/draft-content` endpoint, where publishing applications can dump the draft content item JSON and expect it to appear on the draft site. [see discussion](https://gov-uk.atlassian.net/wiki/display/PUB/Routing+requests+between+draft+and+live+content+stores#).
0. Publishing API registers the base path of the draft content with url-arbiter (https://github.com/alphagov/publishing-api/commit/1949fcb041f91bee268ee5630aee1275c1aaf566), and then forwards the request only to the draft content store (https://github.com/alphagov/publishing-api/commit/877fed430bea2d308f6fe24b4374852839480a8d).
0. Publishing API also forwards a copy of the live content item received at `/content` endpoint to the draft content store: https://github.com/alphagov/publishing-api/commit/b01dfc3ade44008f240df7ba3c119b47a5d9b59f.


jamiecobbett:
I think we need to add a feature-flag to disable this functionality until we've got a stable draft content-store running.

Context:
publishing-api is currently deployed but not used. If we merge this, then we can't start using publishing-api until we have a stable draft content-store, otherwise live publishing requests will fail.

alext:
> I think we need to add a feature-flag to disable this functionality until we've got a stable draft content-store running.

Instead of a traditional feature-flag, you could switch this based on whether the `DRAFT_CONTENT_STORE` environment variable was populated.

content_store_controller.go
```diff
@@ -13,54 +14,69 @@ import (
...
+		controller.doDraftContentStoreRequest("PUT", strings.Replace(path, "draft-", "", -1), requestBody, w)
+	})
+}
 
-	requestBody, err := ioutil.ReadAll(r.Body)
-	if err != nil {
-		renderer.JSON(w, http.StatusInternalServerError, err)
-		return
-	}
+func (controller *ContentStoreController) PutContentStoreRequest(w http.ResponseWriter, r *http.Request) {
```
jamiecobbett: Given that this no longer handles publish-intents, does it make sense to name them `PutContentItemRequest` and `PutDraftContentItemRequest`?

content_store_controller.go
```diff
@@ -13,54 +14,69 @@ import (
...
+		afterRegister(r.URL.Path, requestBody)
+	}
 }
 
 // Register the given path and publishing app with the URL arbiter.  Returns
 // true on success.  On failure, writes an error to the ResponseWriter, and
 // returns false
-func (cs *ContentStoreController) registerWithURLArbiter(path, publishingApp string, w http.ResponseWriter) bool {
-	urlArbiterResponse, err := cs.arbiter.Register(path, publishingApp)
+func (controller *ContentStoreController) registerWithURLArbiter(path, publishingApp string, w http.ResponseWriter) bool {
```
alext: Is it worth calling the `path` parameter `basePath` to make it clear that it's the base path of the item without any prefix components etc.

content_store_controller.go
```diff
@@ -75,23 +91,52 @@ func (cs *ContentStoreController) registerWithURLArbiter(path, publishingApp str
...
+	if w != nil {
+		if err != nil {
+			renderer.JSON(w, http.StatusInternalServerError, err)
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(resp.StatusCode)
+		io.Copy(w, resp.Body)
+	}
+}
```
alext: This method is very similar to `doContentStoreRequest`.  The only difference being which content-store it writes to, and handling `w` being nil.  It's probably worth combining these 2 methods into a single one that takes the content store client instance as a parameter.  I don't think it would be a problem to have the `w` nil handling for both cases.

content_store_controller.go
```diff
@@ -75,23 +91,52 @@ func (cs *ContentStoreController) registerWithURLArbiter(path, publishingApp str
...
-func (cs *ContentStoreController) doContentStoreRequest(httpMethod string, path string, data []byte, w http.ResponseWriter) {
-	resp, err := cs.contentStore.DoRequest(httpMethod, path, data)
+func (controller *ContentStoreController) doContentStoreRequest(httpMethod string, path string, data []byte, w http.ResponseWriter) {
+	resp, err := controller.liveContentStore.DoRequest(httpMethod, path, data)
+	defer resp.Body.Close()
+
 	if err != nil {
 		renderer.JSON(w, http.StatusInternalServerError, err)
 	}
-	defer resp.Body.Close()
```
alext: Moving this above the error check could be problematic.  `resp` could be nil if an error is returned, and therefore calling `resp.Body` would trigger a panic.

content_store_controller.go
```diff
@@ -75,23 +91,52 @@ func (cs *ContentStoreController) registerWithURLArbiter(path, publishingApp str
...
 // data will be nil for requests without bodies
-func (cs *ContentStoreController) doContentStoreRequest(httpMethod string, path string, data []byte, w http.ResponseWriter) {
-	resp, err := cs.contentStore.DoRequest(httpMethod, path, data)
+func (controller *ContentStoreController) doContentStoreRequest(httpMethod string, path string, data []byte, w http.ResponseWriter) {
+	resp, err := controller.liveContentStore.DoRequest(httpMethod, path, data)
+	defer resp.Body.Close()
+
 	if err != nil {
 		renderer.JSON(w, http.StatusInternalServerError, err)
 	}
```
alext: This wasn't introduced in the PR, but this error handling needs a `return` statement, otherwise it will attempt to send a second response after this error response.

content_store_controller.go
vinayvinay: i have another PR coming-up to rename these methods and split them across 2 controllers, can we wait till then?

vinayvinay: @jamiecobbett: i've created https://github.com/alphagov/publishing-api/pull/23 to address the naming.

vinayvinay:
@alext, @jamiecobbett: for ease of reviewing, i've created a sub-PR: https://github.com/alphagov/publishing-api/pull/23, for splitting controllers so that we can have better naming and split of responsibilities.

vinayvinay:
@alext, @jamiecobbett: if `DRAFT_CONTENT_STORE` is not set in the environment, the content store client doesn't have a `rootURL`. in that case, how does [returning HEAD 200 OK](https://github.com/alphagov/publishing-api/compare/split-controllers...flag-draft-content-store) seem to you?

controllers/content_items_controller.go
```diff
@@ -0,0 +1,38 @@
...
+	return &ContentItemsController{
+		arbiter:           urlarbiter.NewURLArbiter(arbiterURL),
+		liveContentStore:  contentstore.NewClient(liveContentStoreURL),
+		draftContentStore: contentstore.NewClient(draftContentStoreURL),
+	}
+}
+
+func (c *ContentItemsController) PutDraftContentItem(w http.ResponseWriter, r *http.Request) {
+	registerWithURLArbiterAndForward(c.arbiter, w, r, func(basePath string, requestBody []byte) {
+		doContentStoreRequest(c.draftContentStore, "PUT", strings.Replace(basePath, "draft-", "", -1), requestBody, w)
```
alext: I'm not sure the `strings.Replace` call is 100% correct here.  It will remove all occurrences of "draft-" from the base path.  Using a limit of 1 would improve it, but it's still a little opaque.  It might be clearer if it did a replacement from "/draft-content/" to "/content/".

vinayvinay: @alext: good eye! fixed in https://github.com/alphagov/publishing-api/commit/41768c653859a030be541cd77d76dbbeb3c080dd.

alext:
:+1: Looks good to me.

jamiecobbett:
We'll do something along the lines of https://github.com/alphagov/publishing-api/compare/master...draft-content-store-unavailable in a separate PR

PR #23, opened 2015-02-24
Split content_store_controller.go

vinayvinay:
https://trello.com/c/PbCzGBIB

in response to [feedback](https://github.com/alphagov/publishing-api/pull/22#discussion_r25173958) on: https://github.com/alphagov/publishing-api/pull/22

why?

1. behaviour for publish intent and content item actions have diverged
2. publish intent controller need not know about draft content store
3. allows HandleFunc names to be verbs describing the actions
4. methods that need not be defined on a controller, are now in a
   separate file within the same package: helpers.go


controllers/helpers.go
```diff
@@ -0,0 +1,91 @@
...
+	httpMethod string, path string, data []byte, w http.ResponseWriter) {
+
+	resp, err := contentStoreClient.DoRequest(httpMethod, path, data)
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+
+	if w != nil {
+		if err != nil {
+			renderer.JSON(w, http.StatusInternalServerError, err)
```
heathd: do we need to write log output here as well?

controllers/helpers.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		renderer.JSON(w, http.StatusInternalServerError, err)
+		return nil, nil
+	}
+
+	var contentStoreRequest *ContentStoreRequest
+	if err := json.Unmarshal(requestBody, &contentStoreRequest); err != nil {
+		switch err.(type) {
+		case *json.SyntaxError:
+			renderer.JSON(w, http.StatusBadRequest, err)
```
heathd: likewise, if we are writing error responses via http, we should also write a log message. Perhaps this happens automatically, I may well have overlooked something?

vinayvinay:
@heathd: i suggest we review and add logging where ever necessary as a separate PR. i've also been thinking about how to log third-party error messages in our JSON logs. it seemed like a little more effort than just adding a log statement.

i understand that correct logging is a given in every PR. however in this case, i've mostly moved existing code to a new place and there's been quite a churn. https://github.com/alphagov/publishing-api/pull/22 as it is, had too many things to be reviewed, so for convenience i've pulled out this sub-PR, idea being we review and merge smaller changes. if you're happy, i'll add a card to Trello for reviewing and adding logging to publishing-api.

alext:
We already have standard request logging (implemented as [middleware](https://github.com/alphagov/publishing-api/blob/master/main.go#L50)).

jamiecobbett:
This is a really good change - makes it clearer and easier to read. It's always a smell when a code file gets really long. :+1:

PR #26, opened 2015-02-25
Stub draft content store response

vinayvinay:
https://trello.com/c/PbCzGBIB

this changes the client.go to return a `200 OK` response when its rootURL is not configured. by default it would generate an internal server error.

until we have a draft-content store running on all environments where publishing-api is installed, we return a `200 OK` response with body indicating it is a dummy response. this will ensure that live content items get stored to live content store, and there are no errors trying to save the same to a non-existent draft content store.

### DIFF
--- a/contentstore/client.go
+++ b/contentstore/client.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
 var (
@@ -25,7 +26,13 @@ func NewClient(rootURL string) *ContentStoreClient {
 }
 
 // data will be nil for requests without bodies
-func (p *ContentStoreClient) DoRequest(httpMethod string, path string, data []byte) (*http.Response, error) {
+func (p *ContentStoreClient) DoRequest(httpMethod, path string, data []byte) (*http.Response, error) {
+	if len(p.rootURL) == 0 {
+		// FIXME: remove once we have a draft content store wherever publishing-api is present.
+		// return a dummy OK response if the client doesn't know which server to call
+		return dummyOKResponse(), nil
+	}
+
 	url := p.rootURL + path
 	var reqBody io.Reader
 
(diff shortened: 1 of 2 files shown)


contentstore/client.go
```diff
@@ -45,3 +51,12 @@ func (p *ContentStoreClient) DoRequest(httpMethod string, path string, data []by
 
 	return p.client.Do(req)
 }
+
+// FIXME: remove once we have a draft content store wherever publishing-api is present.
+func emptyOKResponse() *http.Response {
+	return &http.Response{
+		Status:     "200 OK",
+		StatusCode: 200,
+		Body:       ioutil.NopCloser(bytes.NewBufferString("Draft content-store not configured. This is a dummy response")),
```
alext: Might be simpler to use a [`strings.Reader`](http://golang.org/pkg/strings/#Reader) here.

alext:
I do wonder whether this would be better handled in the controller.  That way we wouldn't have to generate dummy responses, and could just write to the `ResponseWriter`.

contentstore/client.go
```diff
@@ -45,3 +51,12 @@ func (p *ContentStoreClient) DoRequest(httpMethod string, path string, data []by
 
 	return p.client.Do(req)
 }
+
+// FIXME: remove once we have a draft content store wherever publishing-api is present.
+func emptyOKResponse() *http.Response {
```
jamiecobbett: It isn't `empty`, so perhaps `dummyOKResponse` would be better?

vinayvinay:
@alext: if i'm thinking correctly, to handle this in the controller we'll have ask the client whether it has a root url or not, and have to expose the root url for that. is there another way?

alext:
We could not construct a client if the url is empty, and then just check whether the draft client is nil or not.

vinayvinay:
@alext: i'm finding that there'll be fewer things to change if we return a dummy response. given that this is a temporary fix, till we roll out draft-content stores, i'd like to keep `FIXME`s as few as possible.

alext:
Fair enough.

:+1: Looks good to me.

PR #27, opened 2015-02-25
Make requests to content store in parallel

vinayvinay:
https://trello.com/c/PbCzGBIB

This will speed-up PUT operation for live content.

### DIFF
--- a/controllers/content_items_controller.go
+++ b/controllers/content_items_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"net/http"
 	"strings"
+	"sync"
 
 	"github.com/alphagov/publishing-api/contentstore"
 	"github.com/alphagov/publishing-api/urlarbiter"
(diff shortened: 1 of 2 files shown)


controllers/content_items_controller.go
```diff
@@ -30,9 +30,19 @@ func (c *ContentItemsController) PutDraftContentItem(w http.ResponseWriter, r *h
...
+		}()
+		go func() {
+			// for now, we ignore the response from draft content store for storing live content, hence `w` is nil
+			doContentStoreRequest(c.draftContentStore, "PUT", basePath, requestBody, nil)
+			done <- true
+		}()
+		// wait till both requests are forwarded
+		// https://gobyexample.com/channel-synchronization
+		<-done
+		<-done
```
alext: I think a [`sync.WaitGroup`](http://golang.org/pkg/sync/#WaitGroup) would be a better primative to use here...

You could also then also `defer` the call to `wg.Done()` at the top of the goroutines to prevent this deadlocking if something panics.

bradleywright:
Do we need to handle errors on either of the goroutines?

vinayvinay:
@brad: `doContentStoreRequest` has access to `http.ResponseWriter` where [it renders an error JSON when request to content store errors](https://github.com/alphagov/publishing-api/blob/master/controllers/helpers.go#L62-L65). to start simple, we've ignored responses from the draft-content-store when live content items are stored, so both won't write to the response stream.

controllers/content_items_controller.go
vinayvinay: `sync.WaitGroup` looks much much better. thanks @alext!

alext:
:+1: Looks good to me.

PR #29, opened 2015-02-27
Use Plek to find url-arbiter address

vinayvinay:


main.go
```diff
@@ -77,3 +78,11 @@ func getEnvDefault(key string, defaultVal string) string {
 
 	return val
 }
+
+func getURLUsingPlek(key string) string {
+	url, err := plek.Find(key)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return url.String()
```
alext: Oh that's annoying.  I wonder whether it's worth refactoring things so that urls are always instances of `url.URL`.

vinayvinay: @alext: can you help me understand how that would be beneficial? from our experience with plek.rb, do we think we need that?

alext: It just seemed like a sensible approach, given Go provides a type for holding a URL.  It would avoid us having to do URL parsing deep within the application.

Having said that, after digging further, `http.NewRequest` takes a string, not a `url.URL`, so that negates any benefit of doing this.

I therefore wonder whether it would be better for us to have Plek return strings as opposed to `url.URL` instances.  Thoughts?  Given this is the first thing using the Go version of Plek, now is the time to get this right...

vinayvinay: @alext: yes, i was expecting that to be discussed during review of this PR. plek.rb returns strings, and we've been happy with that. so we should be good to return strings from plek.go as well. i'll work on that.

vinayvinay: @alext: looks much better now, just `plek.Find`.

alext:
Nice.  The simplicity of that change is an indication that the Plek API is in good shape.

Issue #24, opened 2015-02-24
If dependent apps are not running, unhelpful error is returned


vinayvinay:
Actual:
```bash
// url-arbiter, content-store are not running
$ make run
$ curl http://localhost:3000/content/new-vat-rates -X PUT -H 'Content-type: application/json' -d '<content-item-json>'
> {"Offset":1}
```

Expected error message:
```bash
> FATAL: Couldn't connect to URL Arbiter at http://url-arbiter.dev.gov.uk.
```

jamiecobbett:
This is currently happening on preview.

jamiecobbett:
I don't think it's because the dependencies aren't running - you've sent invalid JSON. The response body: `{"Offset":1}` is the serialized error object because we're supplying it as the third argument (the body) here: https://github.com/alphagov/publishing-api/blob/master/content_store_controller.go#L44

You would have received the right HTTP status, but I think we should work on the response bodies.

I haven't figured out what's wrong on preview yet.

vinayvinay:
Hi Jamie: I confirmed again that this happens with one of the services being down as well. The same request returns the offset error, which it not the case when all dependent services are up:
```
$ curl http://localhost:3093/content/new-vat-rates -X PUT -H 'Content-type: application/json' -d '{"content_id": "a5d86183-7252-4426-aad2-3f732f7f2581", "title": "VAT Rates", "publishing_app":"publisher", "format": "guide", "update_type": "minor", "rendering_app":"frontend", "routes": [{"path":"/new-vat-rates", "type": "exact"}]}'
=> {"Offset":1}
```

alext:
I think this will be because we're doing things like 
```go
renderer.JSON(w, http.StatusInternalServerError, err)
```
in our error handling code, which is just asking renderer to serialise an error object to JSON.  This will probably just pull the fields out of the underlying struct, and they won't make much sense out of context (as we're seeing).

I'd suggest that we should either not return JSON in this case, and just return `text/plain` with the error message (`err.Error()`), or construct some meaningful JSON that includes this full error message.

jamiecobbett:
I made a start on returning saner errors, but got a bit bogged down trying to do it in Go. I've opened this PR in url-arbiter which I plan to apply to publishing-api: https://github.com/alphagov/url-arbiter/pull/19

PR #30, opened 2015-03-03
Improve error responses

jamiecobbett:
Currently, aside from errors relayed back from content-store, the errors returned from publishing-api are inscrutable. This is because the app is simply serialising the error object, which might contains an `Offset` key (in the case of a `json.SyntaxError`) or nothing serialisable.

~~I would like to include more detail by taking information out of the error objects but I am not sure how - are error objects in Go required to consistently implement anything?~~

This also fixes a potential source of bugs where we were assuming that any response from url-arbiter that wasn't 409 or 422 meant that it was OK to proceed.

This applies a pattern of including a top-level "message" key with a string containing a human-friendly summary of the response. This is based upon an RFC here: https://gov-uk.atlassian.net/wiki/display/PUB/RFC%3A+API+guidelines

Fixes https://github.com/alphagov/publishing-api/issues/24

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -66,7 +66,7 @@ func doContentStoreRequest(contentStoreClient *contentstore.ContentStoreClient,
 
 	if w != nil {
 		if err != nil {
-			renderer.JSON(w, http.StatusInternalServerError, err)
+			renderer.JSON(w, http.StatusInternalServerError, ErrorResponse{Message: "Unexpected error in request to content-store"})
 			return
 		}
 		w.Header().Set("Content-Type", "application/json")
@@ -78,7 +78,7 @@ func doContentStoreRequest(contentStoreClient *contentstore.ContentStoreClient,
 func readRequest(w http.ResponseWriter, r *http.Request) ([]byte, *ContentStoreRequest) {
 	requestBody, err := ioutil.ReadAll(r.Body)
 	if err != nil {
-		renderer.JSON(w, http.StatusInternalServerError, err)
+		renderer.JSON(w, http.StatusInternalServerError, ErrorResponse{Message: "Unexpected error in reading your request body"})
 		return nil, nil
 	}
 
@@ -88,7 +88,7 @@ func readRequest(w http.ResponseWriter, r *http.Request) ([]byte, *ContentStoreR
 		case *json.SyntaxError:
 			renderer.JSON(w, http.StatusBadRequest, ErrorResponse{Message: "Invalid JSON in request body"})
 		default:
-			renderer.JSON(w, http.StatusInternalServerError, err)
+			renderer.JSON(w, http.StatusInternalServerError, ErrorResponse{Message: "Unexpected error unmarshalling your request body to JSON"})
 		}
 		return nil, nil
 	}
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -46,7 +46,8 @@ func registerWithURLArbiter(urlArbiter *urlarbiter.URLArbiter, path, publishingA
 		case urlarbiter.UnprocessableEntity:
 			renderer.JSON(w, 422, urlArbiterResponse)
 		default:
-			renderer.JSON(w, http.StatusInternalServerError, err)
+			message := "Unexpected error whilst registering with url-arbiter"
+			renderer.JSON(w, http.StatusInternalServerError, ErrorResponse{Message: message})
 		}
 		return false
 	}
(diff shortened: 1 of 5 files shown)


urlarbiter/url_arbiter.go
```diff
@@ -57,14 +58,17 @@ func (u *URLArbiter) Register(path, publishingAppName string) (URLArbiterRespons
...
 
-	// Read the response body and then check the status code so we can
-	// return the errors from the response.
-	switch response.StatusCode {
-	case 422: // Unprocessable Entity.
-		return arbiterResponse, UnprocessableEntity
-	case http.StatusConflict:
-		return arbiterResponse, ConflictPathAlreadyReserved
+	if response.StatusCode < 300 && response.StatusCode >= 200 {
+		// 1xx and 2xx
```
vinayvinay: this comment doesn't match the condition; best to remove it, the conditional is easily readable. one more minor pedantic change, it'd be better to have the lower-bound (`>= 200`) condition first and upper-bound (`< 300`) next.

urlarbiter/url_arbiter.go
```diff
@@ -57,14 +58,17 @@ func (u *URLArbiter) Register(path, publishingAppName string) (URLArbiterRespons
...
+		return arbiterResponse, nil
+	} else {
+		switch response.StatusCode {
+		case 422: // Unprocessable Entity.
+			return arbiterResponse, UnprocessableEntity
+		case http.StatusConflict:
+			return arbiterResponse, ConflictPathAlreadyReserved
+		default:
+			return arbiterResponse, UnexpectedResponse
```
vinayvinay: as i read further, i wonder whether moving the `if` condition above into this `default` block would read better?

integration_tests/publish_intent_requests_test.go
```diff
@@ -110,6 +110,21 @@ var _ = Describe("Publish Intent Requests", func() {
...
+					actualResponse := doRequest("PUT", endpoint, publishIntentPayload)
+
+					Expect(testURLArbiter.ReceivedRequests()).To(HaveLen(1))
+					Expect(testDraftContentStore.ReceivedRequests()).To(BeEmpty())
+					Expect(testLiveContentStore.ReceivedRequests()).To(BeEmpty())
+
+					expectedResponseBody := `{"message": "Unexpected error whilst registering with url-arbiter"}`
+					expectedResponse = HTTPTestResponse{Code: 500, Body: expectedResponseBody}
+					assertSameResponse(actualResponse, &expectedResponse)
+				})
```
vinayvinay: this test makes me wonder whether there's a case for extracting out a `url_arbiter_error_response_test.go`, where publishing-api behaviour in response to url arbiter response is tested. i don't think we expect this behaviour to change irrespective of which endpoint the API consumer contacts.

maybe, i shouldn't have replicated these tests (url arbiter error response tests) across endpoints to begin with. i'm happy if we have to take care of this as a separate PR, just thought this was a nice example to discuss this change.

what do you think @jamiecobbett?

controllers/helpers.go
```diff
@@ -60,7 +66,7 @@ func doContentStoreRequest(contentStoreClient *contentstore.ContentStoreClient,
 
 	if w != nil {
 		if err != nil {
-			renderer.JSON(w, http.StatusInternalServerError, err)
+			renderer.JSON(w, http.StatusInternalServerError, ErrorResponse{Message: "Unexpected error in request to content-store"})
```
alext: This seems to be throwing away some of the error detail.  All errors in Go implement an `Error()` method that returns a string.  Would be good to include this in the response to give some additional context.

alext:
Sorry, I missed this point in your PR description when I commented above...

> I would like to include more detail by taking information out of the error objects but I am not sure how - are error objects in Go required to consistently implement anything?

All errors in Go implement an `Error()` method that returns a string.  See http://golang.org/ref/spec#Errors

controllers/helpers.go
jamiecobbett: Thanks - fixed.

urlarbiter/url_arbiter.go
jamiecobbett: Good spot - rebased to do both.

integration_tests/publish_intent_requests_test.go
jamiecobbett: I think you're right - it is worth exploring in a separate PR. One option is to have some simple integration tests and more detailed unit tests of the url-arbiter client.

PR #32, opened 2015-03-25
Improve url arbiter response handling

vinayvinay:
https://trello.com/c/8zCaPXNh

publishing-api fails to handle HTML error responses from Rails causing [errors](https://errbit.preview.alphagov.co.uk/apps/53020c2d0da11590e70000a0/problems/550fbf860da1152033002ba7). it fails because it tries to [parse HTML response as JSON](https://github.com/alphagov/publishing-api/blob/master/urlarbiter/url_arbiter.go#L57-L59).

this causes publishing-api to return JSON decoding errors to clients who're expecting to know what exactly went wrong with their requests. by setting an appropriate 'Accept' header we're ensuring Rails returns errors as a JSON document instead of HTML.

also, updating content-store client with the same change and passing the response content-type back to clients as-is.

hat-tip: @alext

### DIFF
--- a/contentstore/client.go
+++ b/contentstore/client.go
@@ -46,6 +46,7 @@ func (p *ContentStoreClient) DoRequest(httpMethod, path string, data []byte) (*h
 		return nil, err
 	}
 
+	req.Header.Set("Accept", "application/json")
 	if data != nil {
 		req.Header.Set("Content-Type", "application/json")
 	}
--- a/urlarbiter/url_arbiter.go
+++ b/urlarbiter/url_arbiter.go
@@ -42,11 +42,11 @@ func (u *URLArbiter) Register(path, publishingAppName string) (URLArbiterRespons
 	jsonRequestBody, _ := json.Marshal(requestBody)
 
 	request, err := http.NewRequest("PUT", url, bytes.NewBuffer(jsonRequestBody))
-	request.Header.Set("Content-Type", "application/json")
-
 	if err != nil {
 		return URLArbiterResponse{}, err
 	}
+	request.Header.Set("Content-Type", "application/json")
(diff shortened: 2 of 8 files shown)


urlarbiter/url_arbiter.go
```diff
@@ -42,20 +42,22 @@ func (u *URLArbiter) Register(path, publishingAppName string) (URLArbiterRespons
...
 	response, err := u.client.Do(request)
 	if err != nil {
 		return URLArbiterResponse{}, err
 	}
 
-	var arbiterResponse URLArbiterResponse
-	if err := json.NewDecoder(response.Body).Decode(&arbiterResponse); err != nil {
-		return URLArbiterResponse{}, err
+	arbiterResponse := URLArbiterResponse{}
+	if responseContentType := response.Header.Get("Content-Type"); responseContentType == "application/json" {
```
alext: `Header.Get()` only returns a single value, so there's no need to do the assignment here.  This can therefore be simplified to
```go
if response.Header.Get("Content-Type") == "application/json" {
```

contentstore/client.go
```diff
@@ -48,6 +48,7 @@ func (p *ContentStoreClient) DoRequest(httpMethod, path string, data []byte) (*h
 
 	if data != nil {
 		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Accept", "application/json")
```
alext: This wants to be outside the conditional.  We want to set an `Accept` header even if we're not sending a request body (eg for a GET request).

alext:
A couple of small points, otherwise looks good to me.

vinayvinay:
@alext: fair points, rebased. thanks!

PR #36, opened 2015-04-15
Notify errbit about internal server errors

vinayvinay:
## Prerequisites:

- [x] [puppet changes](https://github.gds/gds/puppet/pull/2725)
- [ ] make deployment repo changes mentioned on the Trello card
- [x] [upgrade errbit](https://github.com/alphagov/errbit/pull/29)

## Description

https://trello.com/c/HfNnOLFJ

this brings publishing-api in line with other applications
with respect to error reporting by integrating it with errbit.

to begin with, we notify errbit whenever a 500 is returned
to clients. we don't adequately handle internal server errors,
and send clients back an unhelpful message saying that an
unexpected error occurred. notifying errbit will inform us about
such occurrences, enabling us to figure-out what we should
be doing.

### gobrake

- we're using gobrake, which is airbrake's official Go client
- we've used a fork of that client [dwilkie/gobrake](https://github.com/dwilkie/gobrake), which allows the host to be configured, so that we can point it towards errbit ;)

### puppet

there is a related [puppet PR](https://github.gds/gds/puppet/pull/2725) which configures the host, api key
and environment name to be used by the gobrake client.

### refactoring

this required that some code in helper.go is moved back into
the controller, from where it was extracted. this enables us to
pass fewer parameters to those helper methods. especially,
prevents us from passing the `errorNotifier` endlessly.

this also allowed us to clean-up the `DraftStoreClient` value
on the content store client structs, which was needed only
by the helper to decide whether to suppress 502 errors
from draft content store.

### DIFF
--- a/Gomfile
+++ b/Gomfile
@@ -4,6 +4,7 @@ gom 'github.com/gorilla/context', :commit => '215affda49addc4c8ef7e2534915df2c8c
 gom 'gopkg.in/unrolled/render.v1', :commit => '40ac5716c7e0bffcc440aac649ea4b880a5d7735'
 gom 'github.com/alext/tablecloth', :commit => '7f49e40'
 gom 'github.com/alphagov/plek/go', :commit => 'f5da97e'
+gom 'github.com/dwilkie/gobrake', :commit => '716d910'
 
 # Testing
 gom 'github.com/onsi/ginkgo', :commit => '75d0cd9c7ee52e99b13e6ccfd9e58b4f92b7795e'
--- a/Gomfile
+++ b/Gomfile
@@ -3,7 +3,7 @@ gom 'github.com/gorilla/mux', :commit => 'e444e69cbd2e2e3e0749a2f3c717cec491552b
 gom 'github.com/gorilla/context', :commit => '215affda49addc4c8ef7e2534915df2c8c35c6cd'
 gom 'gopkg.in/unrolled/render.v1', :commit => '40ac5716c7e0bffcc440aac649ea4b880a5d7735'
 gom 'github.com/alext/tablecloth', :commit => '7f49e40'
-gom 'github.com/alphagov/plek/go', :commit => 'f5da97e'
+gom 'github.com/alphagov/plek/go', :commit => 'c0c56fb'
 gom 'github.com/dwilkie/gobrake', :commit => '716d910'
 
 # Testing
--- a/controllers/content_items_controller.go
+++ b/controllers/content_items_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 
@@ -18,8 +19,8 @@ type ContentItemsController struct {
 func NewContentItemsController(arbiterURL, liveContentStoreURL, draftContentStoreURL string) *ContentItemsController {
 	return &ContentItemsController{
 		arbiter:           urlarbiter.NewURLArbiter(arbiterURL),
-		liveContentStore:  contentstore.NewClient(liveContentStoreURL, false),
-		draftContentStore: contentstore.NewClient(draftContentStoreURL, true),
+		liveContentStore:  contentstore.NewClient(liveContentStoreURL),
+		draftContentStore: contentstore.NewClient(draftContentStoreURL),
 	}
 }
 
--- a/controllers/content_items_controller.go
+++ b/controllers/content_items_controller.go
@@ -30,7 +30,12 @@ func (c *ContentItemsController) PutDraftContentItem(w http.ResponseWriter, r *h
 			handleURLArbiterResponse(urlArbiterResponse, err, w)
 			return
 		}
-		doContentStoreRequest(c.draftContentStore, "PUT", strings.Replace(r.URL.Path, "/draft-content/", "/content/", 1), requestBody, w)
+		resp, err := c.draftContentStore.DoRequest("PUT", strings.Replace(r.URL.Path, "/draft-content/", "/content/", 1), requestBody)
+		if resp.StatusCode == http.StatusBadGateway && os.Getenv("SUPPRESS_DRAFT_STORE_502_ERROR") == "1" {
+			w.WriteHeader(http.StatusOK)
+		} else {
+			handleContentStoreResponse(resp, err, w)
+		}
 	}
 }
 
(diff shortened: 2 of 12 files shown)


controllers/content_items_controller.go
```diff
@@ -13,35 +15,59 @@ type ContentItemsController struct {
 	arbiter           *urlarbiter.URLArbiter
 	liveContentStore  *contentstore.ContentStoreClient
 	draftContentStore *contentstore.ContentStoreClient
+	errbitNotifier    *errornotifier.ErrbitNotifier
```
alext: It would be better for this to be a `errornotifier.Notifier` instead of a `*errornotifier.ErrbitNotifier`.  That way the controller doesn't need to know what sort of error notifier it's dealing with.

errornotifier/notifier.go
```diff
@@ -0,0 +1,28 @@
...
+func NewErrbitNotifier(host, apiKey, envName string) *ErrbitNotifier {
+	errbitClient := gobrake.NewNotifier(0, apiKey, host)
+	errbitClient.SetContext("environment", envName)
+
+	return &ErrbitNotifier{
+		errbitClient: errbitClient,
+	}
+}
+
+func (en ErrbitNotifier) Notify(err error, request *http.Request) {
```
alext: This doesn't match the definition in the `Notifier` interface type (it doesn't have an `error` return value).

main.go
```diff
@@ -12,32 +12,38 @@ import (
...
 	renderer = render.New(render.Options{})
+
+	errbitHost     = os.Getenv("ERRBIT_HOST")
+	errbitApiKey   = os.Getenv("ERRBIT_API_KEY")
+	errbitEnvName  = os.Getenv("ERRBIT_ENVIRONMENT_NAME")
+	errbitNotifier = errornotifier.NewErrbitNotifier(errbitHost, errbitApiKey, errbitEnvName)
 )
 
-func BuildHTTPMux(arbiterURL, liveContentStoreURL, draftContentStoreURL string) http.Handler {
+func BuildHTTPMux(arbiterURL, liveContentStoreURL, draftContentStoreURL string, errbitNotifier *errornotifier.ErrbitNotifier) http.Handler {
```
alext: Again, it would probably be better to use the interface type here...

main.go
```diff
@@ -12,32 +12,38 @@ import (
...
 
-	draftContentStoreHost = os.Getenv("DRAFT_CONTENT_STORE")
+	draftContentStoreURL = os.Getenv("DRAFT_CONTENT_STORE")
 
 	renderer = render.New(render.Options{})
+
+	errbitHost     = os.Getenv("ERRBIT_HOST")
+	errbitApiKey   = os.Getenv("ERRBIT_API_KEY")
+	errbitEnvName  = os.Getenv("ERRBIT_ENVIRONMENT_NAME")
+	errbitNotifier = errornotifier.NewErrbitNotifier(errbitHost, errbitApiKey, errbitEnvName)
```
alext: How will this behave if the ENV vars aren't defined (eg in dev).  Might it be better to move the construction of this into `main`?

main.go
```diff
@@ -46,11 +52,14 @@ func BuildHTTPMux(arbiterURL, liveContentStoreURL, draftContentStoreURL string)
...
 func main() {
-	httpMux := BuildHTTPMux(arbiterHost, liveContentStoreHost, draftContentStoreHost)
+	httpMux := BuildHTTPMux(arbiterURL, liveContentStoreURL, draftContentStoreURL, errbitNotifier)
 
 	requestLogger, err := request_logger.New(requestLogDest)
 	if err != nil {
 		log.Fatal(err)
+		if errbitNotifier != nil {
+			errbitNotifier.Notify(err, &http.Request{})
+		}
```
alext: [`log.Fatal` calls `os.Exit`](http://golang.org/pkg/log/#Fatal), so anything after it won't have any effect.  The errbit call will need to be moved before the log call.

alext:
A few comments, but the overall approach looks good to me.

vinayvinay:
@alext: good points, rebased. thanks for your inputs!

vinayvinay:
@alext: i was wondering if we could update plek.go to return just the host for a service (instead of the complete URL), and not store errbit host in an ENV var. that would work if plek uses `GOVUK_APP_ROOT` as `production.gov.uk` in staging, does it?

alext:
>@alext: i was wondering if we could update plek.go to return just the host for a service (instead of the complete URL), and not store errbit host in an ENV var. that would work if plek uses GOVUK_APP_ROOT as production.gov.uk in staging, does it?

If it just returned the host, we'd lose the logic that knows to use plain http in development.  An alternative approach would be to make it return a `url.URL`.  Then you could do `Plek.find("errbit").Host`.  Actually, we could make it match the ruby version, and have a FindURL method that does just that.  We haven't done that up to now because we haven't had the need...

controllers/content_items_controller.go
```diff
@@ -13,35 +15,59 @@ type ContentItemsController struct {
 	arbiter           *urlarbiter.URLArbiter
 	liveContentStore  *contentstore.ContentStoreClient
 	draftContentStore *contentstore.ContentStoreClient
+	errbitNotifier    errornotifier.Notifier
```
alext: Picky point: This property would probably be better named `errorNotifier`.

Related:  it's probably worth looking at the [gorename tool](https://godoc.org/golang.org/x/tools/cmd/gorename).  Having said that, it does require you to be working within a `GOPATH`.

danielroseman:
@alext do you think this is ready to merge?

Gomfile
```diff
@@ -3,7 +3,8 @@ gom 'github.com/gorilla/mux', :commit => 'e444e69cbd2e2e3e0749a2f3c717cec491552b
 gom 'github.com/gorilla/context', :commit => '215affda49addc4c8ef7e2534915df2c8c35c6cd'
 gom 'gopkg.in/unrolled/render.v1', :commit => '40ac5716c7e0bffcc440aac649ea4b880a5d7735'
 gom 'github.com/alext/tablecloth', :commit => '7f49e40'
-gom 'github.com/alphagov/plek/go', :commit => 'f5da97e'
+gom 'github.com/alphagov/plek/go', :commit => 'c0c56fb'
```
alext: A note.  gom does support pinning to tags (with `:tag => 'v1.8.1'`) - see https://github.com/mattn/gom#gomfile.

Not a blocker for merge, just worth noting...

alext:
:+1: Looks good to me...

PR #42, opened 2015-06-30
Strip access limiting

elliotcm:
In order to prevent the live content store protecting content which was sent with an access limited flag (which should only be going to the draft content store) the publishing API strips this information away if it is sent to the live endpoint.

There has been some debate about whether this should be a rejection of the request instead, but this approach was chosen in the short-medium term to avoid duplicating the logic around access limiting in clients which use the API.

This code was written by:
- Elliot Crosby-McCullough <elliot.cm@gmail.com>
- David Heath <david.heath@digital.cabinet-office.gov.uk>

https://trello.com/c/w1ltlWUm/232-prevent-access-limiting-from-being-enforced-for-published-content

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -101,3 +101,17 @@ func readRequest(w http.ResponseWriter, r *http.Request, errorNotifier errornoti
 
 	return requestBody, contentStoreRequest
 }
+
+func stripAccessLimitingMetadata(body []byte) []byte {
+	// This helper doesn't do any error handling as
+	// the request has already been passed through
+	// helpers#readRequest which does appropriate
+	// error handling.
+	var unmarshalled map[string]interface{}
+	json.Unmarshal(body, &unmarshalled)
+
+	delete(unmarshalled, "access_limited")
+
+	withoutAccessLimitedField, _ := json.Marshal(unmarshalled)
+	return withoutAccessLimitedField
+}
--- a/controllers/content_items_controller.go
+++ b/controllers/content_items_controller.go
@@ -51,16 +51,18 @@ func (c *ContentItemsController) PutLiveContentItem(w http.ResponseWriter, r *ht
 			return
 		}
 
+		requestBodyWithoutAccessLimiting := stripAccessLimitingMetadata(requestBody)
+
 		var wg sync.WaitGroup
 		wg.Add(2)
 		go func() {
 			defer wg.Done()
-			resp, err := c.liveContentStore.DoRequest("PUT", r.URL.Path, requestBody)
+			resp, err := c.liveContentStore.DoRequest("PUT", r.URL.Path, requestBodyWithoutAccessLimiting)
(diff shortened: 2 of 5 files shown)


alext:
> There has been some debate about whether this should be a rejection of the request instead, but this approach was chosen in the short-medium term to avoid duplicating the logic around access limiting in clients which use the API.

I'm not sure I follow the reasoning here.  In what cases would a client be requesting that a live document be access-limited.  That would seem to be an error that should be rejected by the publishing-api.  If we strip this out, we're effectively accepting the request to apply access-limiting to the document, but then not applying it, with no indication to the client that this is the case.

elliotcm:
@danielroseman could you jump in here?

danielroseman:
@alext there has been a lot of discussion about this. In the clients that we control, it is true that we should not be sending access-limited data in the first place. However, this is really a belt-and-braces approach; if a client does mistakenly describe a live document as access limited, we should allow it to be published.

controllers/helpers.go
```diff
@@ -101,3 +101,13 @@ func readRequest(w http.ResponseWriter, r *http.Request, errorNotifier errornoti
 
 	return requestBody, contentStoreRequest
 }
+
+func stripAccessLimitingMetadata(body []byte) []byte {
+	var unmarshalled map[string]interface{}
+	json.Unmarshal(body, &unmarshalled)
```
alext: We're ignoring any errors here.  I realise that if the request body is invalid JSON, this will already be caught by `readRequest`, so it should never error here.  I think it's probably worth a comment explaining that.

elliotcm: @alext Fair, updated to include a comment.

alext:
I may be missing something, but I can't see a test that covers this - ie I can't see a test that asserts that the "access_limited" key is removed from the request sent to the content-store.

elliotcm:
@alext Good spot.  In all the refactoring a key point was missed, the curse of testing the absence of something.  I've added it here: https://github.com/alphagov/publishing-api/commit/8e17427f0a9a2d2eb511a8a2978708c00a305e73

alext:
@elliotcm It feels that this should be a test of its own rather than just part of the existing test for the general case.  Otherwise the intent of the test isn't clear.

elliotcm:
@alext We could, but it would be a literal duplicate of the existing test with different text.  We added the point about access limiting to the equivalent one in draft, would you like me to do the same here?

alext:
I think we should do something.  As it is, it's very easy to miss the detail that this is testing that access limiting is stripped.

elliotcm:
@alext I've taken another look at the test file in question and because that endpoint could receive the document with or without access limiting I've changed it so that the generic test covers receipt of the document without access limiting and I've added another test specifically for when the endpoint receives an access limited document.

alext:
:+1: Looks good to me.

PR #44, opened 2015-07-02
Fix controller race condition.

alext:
When ignoring errors from the draft content-store, there's a race in the
controller because both the live and draft request goroutines will
attempt to call WriteHeader on the response.  The first one wins.

This updates the code to completely ignore the draft content-store
response in this scenario.

### DIFF
--- a/integration_tests/content_item_requests_test.go
+++ b/integration_tests/content_item_requests_test.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"os"
+	"time"
 
 	"github.com/alphagov/publishing-api"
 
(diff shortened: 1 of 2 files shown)


integration_tests/content_item_requests_test.go
```diff
@@ -190,5 +192,30 @@ var _ = Describe("Content Item Requests", func() {
 			expectedResponse = HTTPTestResponse{Code: http.StatusOK, Body: string(expectedBody[:])}
 			assertSameResponse(actualResponse, &expectedResponse)
 		})
+
+		Context("with SUPPRESS_DRAFT_STORE_502_ERROR set to 1", func() {
+			It("returns the live content-store response when the draft content store is not running", func() {
```
mattbostock: Is this strictly correct? Does is not return the live content-store response even if the draft content store is running, but its goroutine happens to complete after the live goroutine?

mattbostock: I guess this doesn't matter given that this is in the context of `SUPPRESS_DRAFT_STORE_502_ERROR` being enabled.

mattbostock:
:+1:

PR #45, opened 2015-07-02
Use a semaphore instead of sleeping in test

mattbostock:
Using sleep() in tests can be quite brittle, since the amount of time we
need to sleep is heavily dependent on the environment in which the tests
run.

Instead, use a semaphore by passing an anonymous struct through an
unbuffered channel, which will block in the live content store handler
until the draft handler has completed.


alext:
:+1: Good call

PR #46, opened 2015-07-03
Update Makefile to avoid unnecessary compiles

alext:
The existing make rules cause the app to be recompiles every time the
app is started via `make run`.  This is unnecessary, and slows things
down for people who just want to run the app via bowler etc.

Update the make rules so that the generated binary depends on all .go
source files so that it will be recompiles when one of them changes, but
not otherwise.


dsingleton:
:heart:

PR #47, opened 2015-07-28
Fix another case of leaking file descriptors.

danielroseman:
When draft content store is not present, and errors are being ignored, the handler was not closing the response.


mattbostock:
:+1:

PR #48, opened 2015-07-29
Convert to Rails 4

elliotcm:
This app will be growing and changing as happens to all apps in their life, and as such we've decided to port it to Rails as that is a more common knowledge set among our developers.

There will be a discussion about why Go ended up being replaced and what we might do to better accommodate it in the future.

There's lots of commits (the first few generated automatically by https://github.com/alphagov/govuk-rails-app-template but they're fairly focused, so you can review commit by commit or all in one go as you please.

https://trello.com/c/oDzUUPQT/281-convert-publishing-api-to-rails


elliotcm:
Sure I added a net 46 lines of code but one of the endpoints wasn't tested so I'm calling it a wash.

alext:
Some comments about the base Rails setup.  I haven't reviewed the app code though.

danielroseman:
Apart from a few mostly minor comments, :+1:

elliotcm:
Changes pushed.

@alext I've stripped a bunch of junk from the app and removed some railties.  We could possibly retrofit the controller base class but perhaps as another update later on.  I've also added the new unified jenkins.sh

@danielroseman I've addressed your comments regarding the `ContentStoreWriter` code-comment & inheritance, the access limited stripped metadata, and the webmock misplacement.

I've also added a `startup.sh` and updated the README.

PR #51, opened 2015-08-06
Apply route constraints to retain locale extension.

elliotcm:
This is otherwise stripped off by Rails and
treated as a format.

This now matches the routing in the content store:
https://github.com/alphagov/content-store/blob/master/config/routes.rb

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,10 +1,14 @@
 Rails.application.routes.draw do
-  get '/healthcheck', :to => proc { [200, {}, ['OK']] }
+  with_options format: false do |r|
+    r.get '/healthcheck', :to => proc { [200, {}, ['OK']] }
 
-  put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
-  put "/content/*base_path", to: "content_items#put_live_content_item"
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -1,10 +1,14 @@
 Rails.application.routes.draw do
-  get '/healthcheck', :to => proc { [200, {}, ['OK']] }
+  with_options format: false do |r|
+    r.get '/healthcheck', :to => proc { [200, {}, ['OK']] }
 
-  put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
-  put "/content/*base_path", to: "content_items#put_live_content_item"
+    r.constraints base_path: %r[/.*] do
+      put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
+      put "/content/*base_path", to: "content_items#put_live_content_item"
```
alext: You've now got an extra `/` in here.  Because `*base_path` is constrained to start with a `/`, the extra `/` in the route here isn't required.

Due to the logic that's now present here, it's probably worth adding some routing specs (eg https://github.com/alphagov/content-store/blob/master/spec/routing/content_item_routing_spec.rb)

PR #49, opened 2015-07-31
Port RabbitMQ push from Content Store

elliotcm:
This allows us to better prevent messages from being pushed onto the queue for draft content items.

https://trello.com/c/ldrFj0UN/282-move-queue-publishing-from-content-store-to-publishing-api

This may require some careful deploying in concert with the content store change.

PR #52, opened 2015-08-10
Port RabbitMQ push from Content Store

danielroseman:
This allows us to better prevent messages from being pushed onto the queue for draft content items.

https://trello.com/c/ldrFj0UN/282-move-queue-publishing-from-content-store-to-publishing-api

(Originally this was reviewed and merged as #49 which was against the convert-to-rails-4 branch - this is a new PR against master to allow merging directly.)


elliotcm:
:+1: for post-merge sanity check.

PR #54, opened 2015-08-12
Accept root path

elliotcm:
Because Rails, we have to duplicate the routes to
accept a base path of `/`.  The write endpoints
will be at some point changed to use content IDs
removing this duplication, though the read endpoints
will need to stay.

Required for https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,14 +1,12 @@
 Rails.application.routes.draw do
   with_options format: false do |r|
-    r.get '/healthcheck', :to => proc { [200, {}, ['OK']] }
+    r.put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
+    r.put "/content/*base_path", to: "content_items#put_live_content_item"
 
-    r.constraints base_path: %r[/.*] do
-      put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
-      put "/content/*base_path", to: "content_items#put_live_content_item"
-
-      put "/publish-intent/*base_path", to: "publish_intents#create_or_update"
-      get "/publish-intent/*base_path", to: "publish_intents#show"
-      delete "/publish-intent/*base_path", to: "publish_intents#destroy"
-    end
+    r.put "/publish-intent/*base_path", to: "publish_intents#create_or_update"
+    r.get "/publish-intent/*base_path", to: "publish_intents#show"
+    r.delete "/publish-intent/*base_path", to: "publish_intents#destroy"
   end
+
+  get '/healthcheck', :to => proc { [200, {}, ['OK']] }
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,11 +1,11 @@
 Rails.application.routes.draw do
-  with_options format: false do |r|
-    r.put "/draft-content/*base_path", to: "content_items#put_draft_content_item"
-    r.put "/content/*base_path", to: "content_items#put_live_content_item"
+  scope format: false do |r|
+    put "/draft-content(/*base_path)", to: "content_items#put_draft_content_item"
+    put "/content(/*base_path)", to: "content_items#put_live_content_item"
 
-    r.put "/publish-intent/*base_path", to: "publish_intents#create_or_update"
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -1,14 +1,22 @@
...
-      put "/publish-intent/*base_path", to: "publish_intents#create_or_update"
-      get "/publish-intent/*base_path", to: "publish_intents#show"
-      delete "/publish-intent/*base_path", to: "publish_intents#destroy"
-    end
+    r.put "/publish-intent/*base_path", to: "publish_intents#create_or_update"
+    r.get "/publish-intent/*base_path", to: "publish_intents#show"
+    r.delete "/publish-intent/*base_path", to: "publish_intents#destroy"
   end
+
+  get '/healthcheck', :to => proc { [200, {}, ['OK']] }
```
bradleywright: Did you intentionally move these outside the `format` block?

elliotcm: Yes, regardless of the format or locale that might be tagged on to the end of `/healthcheck` it's always going to return "OK".

alext: The ones below want to be inside the format constraint though.  We don't want `/content.json` to be a route for example.

elliotcm: Right, updated.

spec/support/request_helpers.rb
```diff
@@ -135,4 +135,17 @@ def check_forwards_locale_extension
...
+
+  def check_accepts_root_path
+    context "with the root path as a base_path" do
+      let(:base_path) { "/" }
+
+      it "creates the content item" do
+        put_content_item
+
+        expect(response.status).to eq(200)
+        expect(a_request(:put, %r{.*/(content|publish-intent)/$})).to have_been_made.at_least_once
```
heathd: I guess this regexp is to accomodate the possibility that it's either a draft or a live put request. I was a bit confused by this at first. I wonder if it would be sufficient to check the assigns in the controller given that other tests verify that the correct thing is passed through to content store?

elliotcm: Draft, live, content item or publish intent.

I want to confirm the actual path requested downstream because we already encountered a situation where it was PUTing downstream to `/content//`

heathd:
The change in routing behaviour was due to https://github.com/rails/rails/pull/17220 which was raised in https://github.com/rails/rails/issues/17212 ([discussion of the issue](https://github.com/rails/rails/issues/17212#issuecomment-130269604)).

pixeltrix:
You make the `base_path` optional in the route and that would remove the need to duplicate them.

pixeltrix:
You also don't need the `with_options`, e.g:

``` ruby
Rails.application.routes.draw do
  scope format: false do
    defaults base_path: '/' do
      put "/draft-content(/*base_path)", to: "content_items#put_draft_content_item"
      put "/content(/*base_path)", to: "content_items#put_live_content_item"

      put "/publish-intent(/*base_path)", to: "publish_intents#create_or_update"
      get "/publish-intent(/*base_path"), to: "publish_intents#show"
      delete "/publish-intent(/*base_path)", to: "publish_intents#destroy"
    end
  end

  get '/healthcheck', :to => proc { [200, {}, ['OK']] }
end
```

elliotcm:
@pixeltrix That would lead to inconsistencies as `base_path` would be either `/` or `something` not `/something`, right?  But the optional part is very helpful.

pixeltrix:
@elliotcm sure, I added the default to make sure that `base_path` was always set in `params` - you could either normalise the path or just call `to_s` and prefix it.

elliotcm:
@pixeltrix aye, we've this in place which should cover it. Thanks! https://github.com/alphagov/publishing-api/blob/master/app/controllers/application_controller.rb#L6

elliotcm:
@heathd Updated to remove the route duplication.  Should be good to go.

heathd:
:+1: looks good. Using the `(`, `)` to declare optional parts of the route makes it much cleaner

PR #55, opened 2015-09-09
Add event log

danielroseman:
Add database, with a bare-bones EventLog model which will form the basis for the transaction log.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -4,7 +4,7 @@ require "rails"
 # Pick the frameworks you want:
 # require "active_model/railtie"
 # require "active_job/railtie"
-# require "active_record/railtie"
+require "active_record/railtie"
 require "action_controller/railtie"
 # require "action_mailer/railtie"
 # require "action_view/railtie"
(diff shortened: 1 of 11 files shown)


config/application.rb
```diff
@@ -2,9 +2,9 @@
 
 require "rails"
 # Pick the frameworks you want:
-# require "active_model/railtie"
+require "active_model/railtie"
```
elliotcm: Fairly sure you don't need this when you have active record, and in any case this app has no forms.

spec/lib/event_logger_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  it "records an event, given the name, payload and user id" do
+    EventLogger.new.log(action, user_uid, payload)
+    expect(Event.count).to eq(1)
+    expect(Event.first.action).to eq(action)
+    expect(Event.first.user_uid).to eq(user_uid)
+    expect(Event.first.payload).to eq(payload)
+  end
+
```
elliotcm: Whitespace.

elliotcm:
Couple of minor things but otherwise :+1:

config/application.rb
danielroseman: Removed.

spec/lib/event_logger_spec.rb
danielroseman: Fixed.

PR #58, opened 2015-09-14
Modify existing commands to log events

danielroseman:
Creates a new Command class which logs API calls to the event log before performing the action, and convert all existing calls to use it.

## Todo

- [x] review error handling edge cases and ensure they are covered by tests (in particular 500 errors from content store)


app/command/base_command.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def queue_publisher
+    PublishingAPI.services(:queue_publisher)
+  end
+
+  def url_arbiter
+    PublishingAPI.services(:url_arbiter)
+  end
+
```
elliotcm: Whitespace.

elliotcm:
The different commands return different types of things, sometimes consistently, sometimes not.  Sometimes it's returning a hash of a content item, sometimes it's returning the last response.

app/command/put_publish_intent.rb
```diff
@@ -0,0 +1,17 @@
...
+    url_arbiter.reserve_path(
+      base_path,
+      publishing_app: content_item[:publishing_app]
+    )
+
+    live_content_store.put_publish_intent(
+      base_path: base_path,
+      publish_intent: content_item
+    )
+  rescue GOVUK::Client::Errors::HTTPError => e
```
elliotcm: It's a little unclear that this specific type of HTTP error can only (currently) come from the URL arbiter client, but I'm not sure how we'd make that clearer right now.

app/lib/command_processor.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :base_path, :user_id, :item
+
+  def initialize(base_path, user_id, item)
+    @base_path = base_path
+    @user_id = user_id
+    @item = item
+  end
+
+  def put_content_with_links
+    event = log_event('PutContentWithLinks')
```
elliotcm: Could we use the command's class name here?

spec/support/request_helpers.rb
```diff
@@ -89,23 +89,11 @@ def check_400_on_invalid_json
     end
   end
 
-  def check_content_type_header
```
elliotcm: Are we removing this behaviour?  If so why?

elliotcm:
General approach :+1:, a couple of minor tweaks requested and some things I'm not sure about.

app/command/put_publish_intent.rb
alext: We shouldn't assume that that's the case. The plan was (still might be) to migrate all the api clients to separate gems with a shared base class.  The url-arbiter client was a first pass at this, and as such, these error classes were intended to be returned by all clients in time.

heathd: I agree. I'll address this.

spec/support/request_helpers.rb
heathd: we attempted to explain the reasoning in [this commit message](https://github.com/alphagov/publishing-api/commit/4ff4e2486fe34351837e83cac9e7ca2268ebeb06)

Neither Daniel nor I could understand the reason for doing this given that we're either:

a) propagating a successful response from content store, which is always json (because we json parse it and a parse failure would be an error)
b) returning an error, in which case we construct a json response body and response code, rather than propagating the content stores response

we couldn't think of a scenario in which propagating the response header made sense, but we may have missed something?

elliotcm: Based on [this commit](https://github.com/alphagov/publishing-api/commit/4a63a91fc95d61463894988e34d2649f7ef6d3d9) it looks like we were doing it to better handle Rails HTML errors, see [this ticket](https://trello.com/c/8zCaPXNh/96-improve-url-arbiter-response-handling-in-publishing-api).  It may be that this is no longer relevant?

heathd: Summary of in-person discussion with Elliot just now:

Looks as though the golang publishing api used to propagate the body of the response from content-store and that's why it was modified to also propagate the headers. The ruby version does not propagate the response body from content-store so it does not make sense to propagate the content type header.

There may still be some error handling edge cases not properly handled, I'm going to review that.

heathd:
regarding the (in)consistency of return values from commands, I think we should probably have an explicit response object to represent success and error responses. Using exceptions for this doesn't feel right.

PR #59, opened 2015-09-18
Add contract testing

heathd:
This configures pact contract testing in publishing-api.

When running pact tests locally using `rake pact:verify` the pact file will be read from the local filesystem at `../gds-api-adapters/spec/pacts/gds_api_adapters-publishing_api.json`.

A rake task `pact:verify:master` is added which will fetch the pact file from CI and use that instead. This requires a password to be specified in the `PACT_CI_API_KEY` environment variable.

To run on jenkins, as well as specifying the API key we must also configure the `TEST_TASK` variable to specify that we want to run the contract tests as well (we may make this the default in the future).

Here's a sample output:

```
~/alphagov/publishing-api(add-contract-testing)$ bundle exec rake pact:verify:master
Reading pact at https://pactcontract:*****@ci-new.alphagov.co.uk/job/govuk_gds_api_adapters/lastSuccessfulBuild/artifact/spec/pacts/gds_api_adapters-publishing_api.json

Verifying a pact between GDS API Adapters and Publishing API
  Given a publish intent exists at /test-intent in the live content store
    a request to delete a publish intent
      with DELETE /publish-intent/test-intent
        returns a response which
          has status code 200
          has a matching body
          includes headers
            "Content-Type" with value "application/json; charset=utf-8"
  Given both content stores and url-arbiter empty
    a request to delete a publish intent
      with DELETE /publish-intent/test-intent
        returns a response which
          has status code 404
          has a matching body
          includes headers
            "Content-Type" with value "application/json; charset=utf-8"
  Given both content stores and url-arbiter empty
    a request to create a draft content item
      with PUT /draft-content/test-draft-content-item
        returns a response which
          has status code 200
          has a matching body
          includes headers
            "Content-Type" with value "application/json; charset=utf-8"
  Given both content stores and url-arbiter empty
    a request to create a publish intent
      with PUT /publish-intent/test-intent
        returns a response which
          has status code 200
          has a matching body
          includes headers
            "Content-Type" with value "application/json; charset=utf-8"
  Given both content stores and url-arbiter empty
    a request to create a content item
      with PUT /content/test-content-item
        returns a response which
          has status code 200
          has a matching body
          includes headers
            "Content-Type" with value "application/json; charset=utf-8"

5 interactions, 0 failures

```

You can see a build on ci here:

https://ci-new.alphagov.co.uk/job/govuk_publishing-api_branches/303/console

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,3 +6,15 @@ require File.expand_path('../config/application', __FILE__)
 require 'pact/tasks'
 
 Rails.application.load_tasks
+
+Pact::VerificationTask.new("master") do | pact |
+  pact.uri "https://pactcontract:#{ENV['PACT_CI_API_KEY']}@ci-new.alphagov.co.uk/job/govuk_gds_api_adapters/lastSuccessfulBuild/artifact/spec/pacts/gds_api_adapters-publishing_api.json"
+end
+
+task :require_pact_ci_api_key do
+  unless ENV['PACT_CI_API_KEY']
+    raise "Environment variable 'PACT_CI_API_KEY' required. See https://ci-new.alphagov.co.uk/user/pactcontract/configure"
+  end
+end
+
+task "pact:verify:master" => :require_pact_ci_api_key
--- a/Rakefile
+++ b/Rakefile
@@ -3,18 +3,4 @@
 
 require File.expand_path('../config/application', __FILE__)
 
-require 'pact/tasks'
-
 Rails.application.load_tasks
-
-Pact::VerificationTask.new("master") do | pact |
-  pact.uri "https://pactcontract:#{ENV['PACT_CI_API_KEY']}@ci-new.alphagov.co.uk/job/govuk_gds_api_adapters/lastSuccessfulBuild/artifact/spec/pacts/gds_api_adapters-publishing_api.json"
-end
-
-task :require_pact_ci_api_key do
-  unless ENV['PACT_CI_API_KEY']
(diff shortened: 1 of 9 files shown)


Rakefile
```diff
@@ -3,4 +3,18 @@
 
 require File.expand_path('../config/application', __FILE__)
 
+require 'pact/tasks'
+
 Rails.application.load_tasks
+
+Pact::VerificationTask.new("master") do | pact |
```
elliotcm: Can we move these to their own file?

elliotcm:
The business with reading from other directories and/or pulling from Jenkins is a bit weird to me, possibly just because it's a new-ish way of working which we've only used in a couple of places to date.

My concern is that it's a bit hard to know how to run them, and if to run them.

Should these be running as part of the normal "rake" run?

Also we should document this in the README.

elliotcm:
By and large though it looks like a good idea, just some questions on how we get used to something new.

heathd:
Thanks for your comments, good questions :-)

> The business with reading from other directories and/or pulling from Jenkins is a bit weird to me, possibly just because it's a new-ish way of working which we've only used in a couple of places to date.

yes, it's new, we're experimenting with a new way of doing things. Please bear with it :-)

> My concern is that it's a bit hard to know how to run them, and if to run them.
> Should these be running as part of the normal "rake" run?

I didn't put them in for now as I want to get all the pieces set up and make sure it all makes sense.

I think we probably do eventually want them to be run as part of the default rake task.

> Also we should document this in the README.

I'll add a note about it

heathd:
Added readme notes

PR #60, opened 2015-09-22
Add derived representations

heathd:
Make the `PUT /content/:base_path` endpoint update derived
representations

Trello: https://trello.com/c/AlN6TwoN/341-design-and-implement-model-of-derived-representations

## Todo

- [x] make `PUT /draft-content/:base_path` update derived representations
- [x] review whether to allow putting content items without a content id to update derived representations or not


db/schema.rb
```diff
@@ -24,4 +42,30 @@
     t.datetime "updated_at"
   end
 
+  create_table "links", force: :cascade do |t|
+    t.string  "content_id"
+    t.integer "version",    null: false
+    t.json    "links"
```
elliotcm: `link.links` is a confusing structure, especially since a "link" is actually a representation of multiple links.

jenkins.sh
```diff
@@ -40,7 +40,7 @@ export RAILS_ENV=test
 bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment --without development
 
 # Uncomment if this app uses a database
-bundle exec rake db:drop db:create db:schema:load
+bundle exec rake db:drop db:create db:schema:load db:test:prepare
```
elliotcm: If we're in RAILS_ENV test when we schema load is this not a no-op?

app/command/put_content_with_links.rb
```diff
@@ -1,5 +1,11 @@
 class Command::PutContentWithLinks < Command::BaseCommand
   def call
+    if content_item[:content_id]
+      create_or_update_live_content_item!
+      create_or_update_draft_content_item!
+      create_or_update_links!
```
elliotcm: Does this sequential set of calls maintain the single transaction model we need?

jenkins.sh
elliotcm: Conclusion was ¯\_(ツ)_/¯ rails

alext: We shouldn't need to add this.  We haven't had to for any other repos. `db:schema:load` will load the schema defined in `db/schema.rb` into the database (assuming it's been checked in - it could well do nothing if `db/scheam/rb` doesn't exist).

db/migrate/20150922125847_add_derived_representations.rb
```diff
@@ -0,0 +1,50 @@
...
+      t.string :base_path
+      t.string :title
+      t.string :description
+      t.string :format
+      t.datetime :public_updated_at
+
+      t.json :access_limited
+      t.json :metadata
+      t.json :details
+      t.json :routes
```
alext: This is missing `redirects`

elliotcm:
We've decided to defer the business of locks and transactions to another ticket because it is hell of complicated to do cleanly.

PR #69, opened 2015-10-01
Create or replace

danielroseman:
When updating an existing live or draft content item, the entity should be completely replaced by the new data, including blanking any fields which are not supplied.

Adds a create_or_replace method, intended to be called within a transaction, that queries for an existing item by content id and locale, locks it, and completely replaces it by merging the supplied data with
the column defaults for that class, updating the version number. This method is used in the put_draft_content_with_links and put_content_with_links commands.

In addition, this attempts to retry a command if it resulted in a unique constraint violation. These are being caused by a request being retried before the original insert has been committed, retrying the transaction should ensure that the second request sees the original insert.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,4 +35,5 @@ group :development, :test do
   gem "rspec-rails", "~> 3.3"
   gem "simplecov", "0.10.0", require: false
   gem "simplecov-rcov", "0.2.3", require: false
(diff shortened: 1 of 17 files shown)


heathd:
:+1: for intent

:+1: like the Replacable mixin and use of shared examples to test it

some issues on the retry implementation

PR #70, opened 2015-10-02
Make tests more generic

heathd:
In order to be able to re-use the test helpers for testing the new v2 API endpoints, they need to be made more generic.

Did the following:

* rename `put_content_item` to `do_request`. This is used by the test helpers to trigger the action under test
* rename `content_item` to `request_payload`. This was used throughout the test helpers both when making the request and when making assertions about stored values or values propagated to external services. Renaming it to request_payload again makes it more generic. Test helpers which used `content_item` for assertions were parameterised to make explicit what checks they perform.

I haven't done the refactoring to shift to using the rspec shared fixtures. I think these changes are sufficient to unblock further work. We can revisit the move to shared fixtures later.

### DIFF
--- a/spec/requests/publish_intent_requests_spec.rb
+++ b/spec/requests/publish_intent_requests_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe "Publish intent requests", type: :request do
     suppresses_draft_content_store_502s
     accepts_root_path
 
-    def put_content_item(body: content_item.to_json)
+    def do_request(body: content_item.to_json)
       put "/publish-intent#{base_path}", body
     end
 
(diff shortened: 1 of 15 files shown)


spec/requests/publish_intent_requests_spec.rb
```diff
@@ -27,7 +30,7 @@
     suppresses_draft_content_store_502s
     accepts_root_path
 
-    def put_content_item(body: content_item.to_json)
+    def do_request(body: request_body)
```
danielroseman: Is this method necessary, since Actions#do_request uses `request_path` - can we not just define that value in this context?

danielroseman:
:+1: apart from one question.

PR #71, opened 2015-10-05
Refactor create content item test helper

heathd:
Some more refactorings to make tests helpers more generic/reusable

### DIFF
--- a/spec/support/request_helpers/derived_representations.rb
+++ b/spec/support/request_helpers/derived_representations.rb
@@ -37,7 +37,11 @@ module RequestHelpers
       end
     end
 
-    def creates_a_content_item_representation(representation_class, expected_attributes:, access_limited: false, immutable_base_path: false)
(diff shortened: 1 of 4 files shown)


spec/support/request_helpers/derived_representations.rb
```diff
@@ -37,7 +37,11 @@ def creates_a_link_representation(expected_attributes: )
       end
     end
 
-    def creates_a_content_item_representation(representation_class, expected_attributes:, access_limited: false, immutable_base_path: false)
+    def creates_a_content_item_representation(representation_class, expected_attributes_proc:, access_limited: false, expectation: nil)
```
tuzz: I don't think the 'expectation' param is used here. Can we remove it?

tuzz:
One comment - otherwise fine. :+1:

spec/support/request_helpers/derived_representations.rb
heathd: yes, good spot!

PR #72, opened 2015-10-06
Use Replaceable mixin in LinkSets.

danielroseman:
LinkSets need to behave the same as live and draft content items, in that they
need to fully replace existing entities on update, and retry correctly on
unique key violations.

This has involved some refactoring of the Replaceable mixin and specs, to
parameterise the keys used for the unique query and some of the expectations
in the tests themselves.

### DIFF
--- a/spec/support/replaceable.rb
+++ b/spec/support/replaceable.rb
@@ -1,6 +1,6 @@
 RSpec.shared_examples Replaceable do
   let(:payload) {
-    build(described_class).as_json.symbolize_keys.except(:format, :routes).merge(content_id: content_id, title: "New title")
+    build(described_class).as_json.except("format", "routes").merge(new_attributes)
   }
 
   context "an item exists with that content_id" do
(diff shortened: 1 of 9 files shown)


spec/support/replaceable.rb
```diff
@@ -49,7 +48,7 @@
       expect(described_class.first.version).to eq(1)
     end
 
-    it "returns the created item" do
+    it "returns the updated item" do
```
elliotcm: Are you sure about this change?

alext:
> LinkSets need to behave the same as live and draft content items, in that they need to fully replace existing entities on update,

Is that true? I thought we'd designed the API so that it's possible to submit a list of eg topic links, and that would replace the existing topic links, but leave all other links intact. Or is that handled at a higher layer?

spec/support/replaceable.rb
danielroseman: Sorry, cut and paste fail, fixed.

danielroseman:
@alext This is on the old endpoints, which IMO need to behave the same way as they always have; a `POST /content/:path` will completely replace all existing content. When we come to the `PUT /links/:content_id` endpoint we will need to do something different.

alext:
Ah, fair enough.  Missed that detail.

PR #77, opened 2015-10-07
Publishing api v2 endpoints

tuzz:
Depends on this being merged: https://github.com/alphagov/gds-api-adapters/pull/360

Jenkins won't pass until that branch is merged into master because this project depends on the PACT json file it generates. Everything is passing locally.

## Todo

- [x] create the ticket for "validate that base path may not change if this item has been previously published"
- [x] create ticket for separate task of combining content body with links (done [here](https://trello.com/c/oSM40z9b/357-combine-content-and-links-when-writing-to-the-content-store))

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "plek", "~> 1.10"
 gem "airbrake", "~> 4.2.1"
 gem "pg"
 
-gem "govuk-client-url_arbiter", "0.0.2"
+gem "govuk-client-url_arbiter", "0.0.3"
 gem "gds-api-adapters", "22.0.0"
 
 gem 'bunny', '2.0.0'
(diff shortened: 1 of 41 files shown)


danielroseman:
General comment: we prefer commit messages to have a specific summary of what's changed, rather than eg "This isn't used". See https://github.com/alphagov/styleguides/blob/master/git.md

tuzz:
Thanks @danielroseman - I have reworded.

heathd:
looks good.

I agree it's good to standardise on symbol or string keys for hashes.

There's one important piece of functionality missing from this which is to combine the links with content when sending to draft content store. I'd suggest that we should make a separate ticket for that and go ahead and merge this pull request so that we can make progress with integrating it against #78

tuzz:
Agreed - we should write something up in Trello. I'll do that now.

tuzz:
We're going to just check this validation:

"base path may not change if this item has been previously published"

Don't merge yet.

PR #78, opened 2015-10-07
v2 Publish command

heathd:
Support for the v2 publish command

The PR for the corresponding changes to gds-api-adapters is [here](https://github.com/alphagov/gds-api-adapters/pull/361)

During the implementation we realised that the versioning behaviour of the v1 api needs to be different than that of the v2 api. See 5f5859d for an discussion of the reasons and possible issues.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,5 +13,7 @@ Rails.application.routes.draw do
     end
   end
 
+  post '/v2/content/:content_id/publish', to: 'content_items#publish'
+
   get '/healthcheck', :to => proc { [200, {}, ['OK']] }
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,9 +11,9 @@ Rails.application.routes.draw do
       put "/content/:content_id", to: "content_items#put_content"
       get "/content/:content_id", to: "content_items#show"
     end
-  end
 
-  post '/v2/content/:content_id/publish', to: 'content_items#publish'
+    post '/v2/content/:content_id/publish', to: 'content_items#publish'
+  end
 
   get '/healthcheck', :to => proc { [200, {}, ['OK']] }
 end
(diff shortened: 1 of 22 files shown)


config/routes.rb
```diff
@@ -8,5 +8,7 @@
     delete "/publish-intent(/*base_path)", to: "publish_intents#destroy"
   end
 
+  post '/v2/content/:content_id/publish', to: 'content_items#publish'
```
alext: This should probably be inside the above `scope` block to disable the `.json` variant of this route.

spec/support/request_helpers/event_logging.rb
```diff
@@ -8,8 +8,7 @@ def logs_event(event_class_name, expected_payload:)
         expect(Event.first.action).to eq(event_class_name)
         expect(Event.first.user_uid).to eq(nil)
 
-        expected_payload = expected_payload.deep_stringify_keys.merge("base_path" => base_path)
-        expect(Event.first.payload).to eq(expected_payload)
+        expect(Event.first.payload).to eq(instance_exec(&expected_payload_proc))
```
alext: Why does this need to use `instance_exec` here as opposed to just calling the proc?

spec/support/shared_context/message_queue_test_mode.rb
```diff
@@ -0,0 +1,24 @@
+RSpec.shared_context "using the message queue in test mode" do
+  before :all do
+    @config = YAML.load_file(Rails.root.join("config", "rabbitmq.yml"))[Rails.env].symbolize_keys
```
alext: Now that this has been extracted, it's probably a good idea to give these variables a more specific name to reduce the change they conflict with variables used in a test.

spec/support/request_helpers/event_logging.rb
heathd: so that it can access the `let` blocks defined in the test

alext:
:+1: Looks good to me.  Holding off merging until we've resolved deployment blockage.

PR #81, opened 2015-10-09
Add links to payload sent to draft content store

danielroseman:
We need to ensure that the object sent to draft content store after PUT /v2/content also includes the links from the LinkSet object for that content_id, if it exists.

### DIFF
--- a/spec/requests/content_item_requests/downstream_requests_spec.rb
+++ b/spec/requests/content_item_requests/downstream_requests_spec.rb
@@ -104,5 +104,32 @@ RSpec.describe "Downstream requests", type: :request do
 
       do_request(body: content_item.to_json)
     end
+
+    context "when a link set exists for the content item" do
+      it "includes links in the payload sent to draft content store" do
+        link_set = create(:link_set, content_id: content_item[:content_id])
+        expect(PublishingAPI.service(:draft_content_store)).to receive(:put_content_item)
+          .with(
+            base_path: base_path,
+            content_item: content_item.merge(links: link_set.links),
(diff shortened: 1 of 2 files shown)


app/command/v2/put_content.rb
```diff
@@ -30,6 +30,22 @@ def metadata
     content_item.except(*content_item_top_level_fields)
   end
 
+  def link_set
+    @link_set ||= LinkSet.find_by(content_id: content_id)
+  end
+
+  def link_set_hash
+    if link_set.present?
+      {links: link_set.links.deep_symbolize_keys}
```
tuzz: We might not need to `#deep_symbolize_keys` here because `include SymbolizeJSON` appears in the LinkSet model.

This will symbolize keys automatically when they're ready from the model.

tuzz:
We're going to have to combine the content item with links in a few places, so we might want to consider how we do this generally (service layer?) - but I think this is the first step.

Other than one nitpick, looks good to me. :+1:

spec/requests/content_item_requests/downstream_requests_spec.rb
```diff
@@ -104,5 +104,30 @@
 
       do_request(body: content_item.to_json)
     end
+
+    context "handles links" do
```
tuzz: Maybe consider rewording to:

```ruby
context "when a link set exists for the content item"
context "when a link set does not exists for the content item"
```

danielroseman:
Changes made as suggested, also rebased onto master and resolved conflicts.

PR #82, opened 2015-10-09
General refactoring

tuzz:
It's probably easier to go through this pull request commit-by-commit.

@elliotcm and I did some general refactoring on what we deemed as low hanging fruit. Some of these may be controversial. Where possible, we tried to use idiomatic Ruby/Rails features and follow standards.


danielroseman:
Apart from one comment on naming, :+1:

elliotcm:
Error name changed, checked with @heathd.

PR #84, opened 2015-10-09
Add an architecture decision record documenting use of event sourcing

heathd:
The registers project [used this pattern](https://github.com/openregister/presentation/pull/71) to document their architecture decisions, and Daniel and I thought that including these decisions in the repository is more valuable than having them on the wiki.

This is a draft attempt to capture the background and reasoning behind the choice to use the event sourcing pattern.


rboulton:
:100: for intent.  Not read the text yet.

PR #86, opened 2015-10-12
Put links endpoint

elliotcm:
Part of https://trello.com/c/SPLM0ANz/318-support-links-endpoints-in-publishing-api

Adds all the database related code needed for the PUT links endpoint.

### DIFF
--- a/app/commands/v2/put_link_set.rb
+++ b/app/commands/v2/put_link_set.rb
@@ -6,10 +6,7 @@ module Commands
 
         link_set = LinkSet.create_or_replace(link_params.except(:links)) do |link_set|
           link_set.version += 1
-
-          link_set.links = link_set.links
-            .merge(link_params.fetch(:links))
-            .reject {|_, links| links.empty? }
+          link_set.links = merge_links(link_set.links, link_params.fetch(:links))
         end
 
         Success.new(links: link_set.links)
@@ -34,6 +31,12 @@ module Commands
 
       def link_params
         payload
+      end
+
+      def merge_links(base_links, new_links)
+        base_links
+          .merge(new_links)
+          .reject {|_, links| links.empty? }
       end
     end
   end
--- a/app/commands/v2/put_link_set.rb
+++ b/app/commands/v2/put_link_set.rb
@@ -9,6 +9,14 @@ module Commands
           link_set.links = merge_links(link_set.links, link_params.fetch(:links))
         end
 
+        if (draft_content_item = DraftContentItem.find_by(content_id: link_params.fetch(:content_id)))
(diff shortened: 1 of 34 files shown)


app/commands/v2/put_link_set.rb
```diff
@@ -0,0 +1,44 @@
+module Commands
+  module V2
+    class PutLinkSet < BaseCommand
+      def call
+        validate!
+
+        content_id = link_params.fetch(:content_id)
+
+        link_set = LinkSet.find_or_initialize_by(content_id: content_id)
```
danielroseman: I realise you can't use create_or_replace here because of the need to merge, but I'm still nervous about contention. At the very least we need to use `LinkSet.lock. find_or_initialize_by` and rescue from `ActiveRecord::RecordNotUnique`.

app/commands/v2/put_link_set.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        link_set.version += 1
+
+        link_set.links = link_set.links
```
danielroseman: I think this is sufficiently complex that it needs a comment, or extracting into a method that describes what it does. Or both.

spec/commands/v2/put_link_set_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  it "validates the link params" do
+    link_params_with_missing_links = {}
+
+    expect {
+      described_class.call(link_params_with_missing_links)
+    }.to raise_error(CommandError, "Links are required")
+  end
+
+  it "updates the LinkSet on disk" do
```
danielroseman: Maybe "updates links when a LinkSet already exists for that content_id".

spec/commands/v2/put_link_set_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    described_class.call(link_params)
+
+    expect(@existing_link_set.reload.links).to eq(
```
danielroseman: Think this needs splitting out into separate specs to fully describe the behaviour: 
* when a key is supplied that already exists, fully replaces that key;
* when a supplied key is not present in the existing hash, adds it;
* when a key is supplied with a blank value, remove it completely;
* preserves any keys that are not specified in the payload.

danielroseman:
:+1: generally and for changes to db defaults, some comments on the put_links implementation.

spec/commands/v2/put_link_set_spec.rb
elliotcm: The detail on this is handled by the pact tests since the endpoint returns the result of the operation.  This spec is just to prove that the database gets used.

PR #87, opened 2015-10-12
Combine content item with links

elliotcm:
- Creates derived representations, replacing links on a per-type basis.
- Sends to content stores and message bus as needed.

https://trello.com/c/SPLM0ANz/318-support-links-endpoints-in-publishing-api


danielroseman:
Merging this into parent branch before reviewing.

PR #90, opened 2015-10-14
Validate base path

tuzz:
https://trello.com/c/SQ7uHbIc/358-v2-endpoints-don-t-allow-base-path-of-live-items-to-change

### DIFF
--- a/app/validators/base_path_validator.rb
+++ b/app/validators/base_path_validator.rb
@@ -1,0 +1,11 @@
+class BasePathValidator < ActiveModel::Validator
+  def validate(record)
+    unless record.mutable_base_path?
+      live_item = LiveContentItem.find_by(content_id: record.content_id)
+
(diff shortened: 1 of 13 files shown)


app/validators/base_path_validator.rb
```diff
@@ -0,0 +1,11 @@
+class BasePathValidator < ActiveModel::Validator
+  def validate(record)
+    unless record.mutable_base_path?
+      live_item = LiveContentItem.find_by(content_id: record.content_id)
+
+      if live_item.present? && live_item.base_path != record.base_path
+        record.errors.add(:base_path, 'is immutable for published items')
```
danielroseman: At some point this message will be exposed to the user, so we probably want to avoid words like "immutable". Perhaps "cannot be changed for published items"?

tuzz: Agreed, I'll do that now.

PR #91, opened 2015-10-14
Import initial data

elliotcm:
This requires a file with one JSON document per line
with each document having the structure:
```
{
  updated_at: "2015-10-14T15:07:09+01:00",
  content_item: {
    content_id: "2e2d3551-4147-4136-8c0d-f0904b69c8c6",
    etc..
  }
}
```

This file can be generated by the content store
using `rake data_hygiene:export_content_items:all`.

You also need to inform the task as to whether you
exported the data from the "live" or "draft"
content stores.

Example usage:
```
$ rake import_content_items['../content-store/tmp/content_items_2015-10-14_13-58-32.json','draft']
Progress [=====>...............................................................................................] (5%)
```

All downstream actions (sending to content stores, sending emails) are skipped.

Requires the output from https://github.com/alphagov/content-store/pull/155 to run but can be merged without.

### DIFF
--- a/lib/tasks/import_data.rb
+++ b/lib/tasks/import_data.rb
@@ -1,0 +1,72 @@
+require "json"
+
+module Tasks
+  class ImportData
+    def initialize(file:, total_lines:, stdout:, draft:)
+      @file = file
+      @total_lines = total_lines
+      @stdout = stdout
+      @draft = draft
+    end
+
+    def import_all
+      file.each.with_index(1) do |json, index|
+        parsed_json = JSON.parse(json).deep_symbolize_keys
+
+        updated_at = parsed_json.fetch(:updated_at)
+        content_item_hash = parsed_json.fetch(:content_item)
+        content_id = content_item_hash.fetch(:content_id)
+        locale = content_item_hash.fetch(:locale)
+
+        unless derived_content_class.where(content_id: content_id, locale: locale).exists?
+          Event.create!(
+            action: action,
+            payload: content_item_hash,
+            created_at: updated_at,
+            updated_at: updated_at,
+          )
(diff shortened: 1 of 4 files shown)


lib/tasks/import_data.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def import_all
+      file.each.with_index(1) do |json, index|
+        parsed_json = JSON.parse(json).deep_symbolize_keys
+
+        updated_at = parsed_json.fetch(:updated_at)
+        content_item_hash = parsed_json.fetch(:content_item)
+        content_id = content_item_hash.fetch(:content_id)
+
+        unless Event.where("payload ->> 'content_id' = '#{content_id}'").exists?
```
danielroseman: This isn't quite right; content_ids are not unique by themselves, it is the combination of (content_id, locale) that is unique, so if we do this subsequent locales for an item won't be imported. Also, this is likely to get very slow as the import proceeds. Might it be better to query the derived table?

rboulton: Isn't there a constraint on the index to ensure this property, anyway?  If the data being imported isn't valid, I think you need to fix the data, rather than just silently ignoring the later duplicates.

elliotcm: Updated.

elliotcm: @rboulton The case this is protected against isn't bad data, but avoiding overwriting newer data.

This import will be run after these event logs and derived representations are being created via the API.  Our data exported from the content store should be by definition equal to or older than any data in the publishing API's databases, so we at best don't need to save it, and at worst really don't want to.

PR #92, opened 2015-10-16
Version invariants (part 1)

tuzz:
This pull request implements these two invariants from the https://trello.com/c/Uv9ZgQs8/356-ensure-consistent-version-numbering story. We thought it would make sense to pull request this separately as there's already a sizeable chunk of work in here and it can go in as a standalone piece of functionality.

```
1) There must always be a draft content item for every live content with the same content id.

2) The version number for every live content item must always be less than, or equal to, the version number of its associated draft content item.
```

Note: Given a LiveContentItem can only ever be created from a DraftContentItem, we added an association connecting the two and a validation on this column. We ran this past @danielroseman first.

### DIFF
--- a/app/lib/presenters/content_item_presenter.rb
+++ b/app/lib/presenters/content_item_presenter.rb
@@ -5,7 +5,7 @@ module Presenters
       public_updated_at = content_item_hash.fetch(:public_updated_at).iso8601
 
       content_item_hash
-        .except(:metadata, :id, :version)
+        .except(:metadata, :id, :version, :draft_content_item_id)
         .merge(metadata)
         .merge(public_updated_at: public_updated_at)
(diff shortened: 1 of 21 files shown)


app/lib/presenters/content_item_presenter.rb
```diff
@@ -5,7 +5,7 @@ def self.present(content_item_hash)
       public_updated_at = content_item_hash.fetch(:public_updated_at).iso8601
 
       content_item_hash
-        .except(:metadata, :id, :version)
+        .except(:metadata, :id, :version, :draft_content_item_id)
```
elliotcm: We should consider switching this to a `slice` if this list is looking to grow.

PR #99, opened 2015-10-22
Versioning locale fix

tuzz:
We found a few problems with the scoping down of content items with regards to locale throughout the application. This includes the recent versioning work, but also some other endpoints such as publish and get_content. The base path validation also needed a change.

We've included a data migration in this pull request which should correct any data that has gone awry. We've pulled down a copy of preview data and verified that this works against that set.

Once this pull request has been merged, we can merge the adapters pull request here:
https://github.com/alphagov/gds-api-adapters/pull/373

### DIFF
--- a/spec/models/live_content_item_spec.rb
+++ b/spec/models/live_content_item_spec.rb
@@ -63,6 +63,21 @@ RSpec.describe LiveContentItem do
     end
   end
 
+  describe "#refreshed_draft_item" do
+    let(:content_id) { SecureRandom.uuid }
+
+    let!(:arabic_live) { FactoryGirl.create(:live_content_item, locale: "ar", content_id: content_id) }
+    let!(:arabic_draft) { arabic_live.draft_content_item }
+
+    let!(:english_live) { FactoryGirl.create(:live_content_item, locale: "en", content_id: content_id) }
+    let!(:english_draft) { english_live.draft_content_item }
+
+    it "finds the corresponding draft item scoped correctly to locale" do
+      expect(english_live.refreshed_draft_item).to eq(english_draft)
+      expect(arabic_live.refreshed_draft_item).to eq(arabic_draft)
+    end
+  end
+
   let(:existing) { FactoryGirl.create(:live_content_item) }
 
   let(:draft) { existing.draft_content_item }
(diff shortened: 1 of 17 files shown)


spec/models/live_content_item_spec.rb
```diff
@@ -63,6 +63,21 @@ def verify_new_attributes_set
...
+  describe "#refreshed_draft_item" do
+    let(:content_id) { SecureRandom.uuid }
+
+    let!(:arabic_live) { FactoryGirl.create(:live_content_item, locale: "ar", content_id: content_id) }
+    let!(:arabic_draft) { arabic_live.draft_content_item }
+
+    let!(:english_live) { FactoryGirl.create(:live_content_item, locale: "en", content_id: content_id) }
+    let!(:english_draft) { english_live.draft_content_item }
+
+    it "finds the corresponding live item scoped correctly to locale" do
```
danielroseman: draft

spec/service_consumers/pact_helper.rb
```diff
@@ -161,4 +161,23 @@
...
+  provider_state "a draft content item exists with content_id: bed722e6-db68-43e5-9079-063f623335a7 and locale: fr" do
+    set_up do
+      DatabaseCleaner.clean_with :truncation
+
+      FactoryGirl.create(:draft_content_item, content_id: "bed722e6-db68-43e5-9079-063f623335a7", locale: "fr")
+      stub_request(:put, Regexp.new('\A' + Regexp.escape(Plek.find('content-store')) + "/content"))
+    end
+  end
+
+  provider_state "a content item exists with content_id: bed722e6-db68-43e5-9079-063f623335a7 with locale: fr" do
```
danielroseman: "draft content item" and "and locale", for consistency. Also it seems a bit odd that the description is only referencing fr, but the code creates en and ar as well.

tuzz: Agreed - I'll update this provider_state description to make this clearer.

tuzz:
@danielroseman also enquired about whether this addresses the auto-increment of DraftContentItem when calling #valid?.

This pull request does not address this issue as this is not the reason for the validation errors we have been seeing recently. In the command objects, we call #save once and therefore, the auto-increment should have only been called once per record - which is correct behaviour.

After running the data migration fix that appears in this pull request, I also ran the following command:

```ruby
LiveContentItem.includes(:draft_content_item).all? do |item|
  item.version <= item.draft_content_item.version
end
#=> true

LiveContentItem.last.id # To check the freshness of the data
#=> 3821
```

This tells us that all LiveContentItems have a version that is less than or equal to the corresponding DraftContentItem. This satisfies the invariant.

PR #101, opened 2015-10-23
Url arbitration

steventux:
This is the initial step to internalizing url arbitration in the publishing API. 
The aim of this work is to continue to write downstream url arbiter records whilst populating internal publishing API path reservations.
There are changes to gds-api-adapters which relate to this pull request.
Pact tests are expected to break with these changes because of unexpected status codes. The update to the pact is here https://github.com/alphagov/gds-api-adapters/pull/374 The failing pact tests are an expected side effect because of changing error codes returned by the API.


app/commands/reserve_url.rb
```diff
@@ -0,0 +1,8 @@
+module Commands
+  class ReserveUrl < BaseCommand
+    def call
+      UrlReservation.reserve_base_path!(base_path, payload[:publishing_app])
```
danielroseman: This needs to call `Adapters::UrlArbiter` as well.

app/controllers/v2/url_reservations_controller.rb
```diff
@@ -0,0 +1,15 @@
+class UrlReservationsController < ApplicationController
```
danielroseman: Don't think we need this controller. It's not linked from the routes, in any case.

spec/commands/reserve_url_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+RSpec.describe Commands::ReserveUrl do
+  describe "call" do
+    context "with a new base_path" do
+      let(:payload) {
+        { base_path: "/foo", publishing_app: "Foo" }
+      }
+
+      it "successfully reserves the path" do
+        expect(described_class.call(payload)).to be_a Commands::Success
```
danielroseman: needs expectation on calling url-arbiter.

steventux:
@danielroseman  thanks for reviewing, updated accordingly.

spec/support/request_helpers/endpoint_behaviour.rb
```diff
@@ -87,6 +87,38 @@ def accepts_root_path
...
+
+          expect(PathReservation.count).to eq(1)
+          expect(PathReservation.first.base_path).to eq(base_path)
+          expect(PathReservation.first.publishing_app).to eq(content_item[:publishing_app])
+        end
+      end
+
+      context "base_path has already been registered" do
+        it "should be successful if the publishing app matches" do
+          expect{ do_request }.to change(PathReservation, :count).by(1)
```
danielroseman: This should be testing against an existing PathReservation, and the count should not change.

PR #102, opened 2015-10-23
Task to verify pacts for a gds-api-adapters branch

alext:
This will be used in a separate CI job to run as a downstram from the
gds-api-adapters branch builds.

This also includes some refactoring to the jenkins.sh script to make it more general. See 097311c for details.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -1,14 +1,13 @@
 #!/bin/bash
 
-REPO_NAME=${REPO_NAME:-"alphagov/publishing-api"}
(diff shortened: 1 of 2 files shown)


alext:
This will be used by the job configured here: https://ci-new.alphagov.co.uk/job/govuk_gds_api_adapters_pact_branches/configure

tuzz:
I don't understand everything in here as bash is quite hard to read, but from what I can tell it looks ok.

steventux:
Likewise, LGTM but I don't totally understand the context

PR #103, opened 2015-10-23
Version extract spike

tuzz:
This pull request is the conclusion of the Version extraction spike to refactor our application to be more decoupled and extensible. Please review and comment on this and we will discuss next week whether this is the approach we'd like to take going forwards.

In summary, this pull request introduces a Version model that externalises the responsibility of "versioning" from ContentItems as well as LinkSets. Note how these models no longer contain any reference to a version. We intentionally made the relationship one-sided (i.e. there's no bi-directionality to the association) to keep versioning logic out of these models.

At present, we've left the version columns on these models to give this code a grace period before we subsequently remove these columns in a follow-up pull request.

Reminder: **do not merge** until this has been discussed by the team. Thanks.

Note: We found a potential problem whilst working on this story. Currently, when you push content into the system, the content store receives the payload that was made in the request to our API. This payload has things like `access_limited` removed where appropriate, but it leaves version in as it is.

When we discussed this, we said that this should probably change and the payload that gets sent to the content store should be created from the derived representation models rather than simply forwarding the request payload through. We haven't addressed that issue in this pull request (given this is just a spike), but we should look at this - probably when we deal with the asynchronous content store story.


danielroseman:
I think this approach is really great and simplifies things nicely, :+1:

My only comment would be to wonder whether we're making things harder for ourselves when we come to implementing the optimistic locking. Now that version is no longer part of the model, it is rejected as part of the payload. Not sure if this is a concern.

elliotcm:
I worked with @tuzz on this so I'm largely on board.  My only qualms during the process were the overhead of extracting version into a complete activerecord model with database table, given that it carries all of one integer and probably will never be more complicated.

That's really commentary on that specific use case, however, rather than the approach in general and I wouldn't block the merge.

I would like some thought put into the effects on debugging when lots of joins are required to get a full picture of the current state of the system.

bradwright:
The code stuff looks good, but:

As I raised in the RFC, I'm not entirely on board with the database table separation here - once we set that precedent we're likely to see a proliferation of database tables for managing the history of a single document, and this leads to difficult debugging and potentially bad performance.

This might be an issue with implementation, as @elliotcm says - is there any reason we need the persistence layer to mirror our object layout?

app/models/version.rb
```diff
@@ -0,0 +1,62 @@
+class Version < ActiveRecord::Base
+  belongs_to :target, polymorphic: true
+
+  validate :numbers_must_increase
+  validate :draft_cannot_be_behind_live
+  validate :live_is_equal_to_draft
```
alext: This raises a flag for me because this will only necessarily be true at the point the live version is saved. If this record is loaded from the database in future, a newer draft could have been saved, and then this record won't be valid any more.

tuzz:
```
is there any reason we need the persistence layer to mirror our object layout?
```

My reason for doing this was to reinforce the separation of concerns between content items and versioning. I think that versioning is a concept in its own right and so it felt more natural to mirror this in the persistence layer. That being said, if people feel strongly about it - I'm happy to change this.

I would point out that we also have versions for link sets and could have versions for more things in the future. If we went back to having the version number on the content item table, we'd also need to store the version number on the link sets table and potentially more things in the future.

Maybe it's too early to be thinking about the future – after all; we aspire to be agile. I don't think that means we shouldn't be thinking about how the system might change, though. I'm concerned that we could get ourselves into a corner if we let versions bleed throughout our persistence layer too much.

I was thinking about cases such as:

1) What happens if we want to store a "major" and "minor" version number?

2) What happens if we need multiple versions for an item? (e.g. LockVersion / UserFacingVersion)

3) What happens if we decide to extract versioning into a separate service?

I appreciate that these are contrived examples, but in all of these cases, I think that it would be easier to apply the change if the versioning persistence was consolidated in one place. It follows that this would be the case for less contrived, real examples.

**TL;DR**

I'd prefer to persist versions separate - but happy to change if people think otherwise.

app/models/version.rb
tuzz: :+1: I think this is a good spot.

It's perfectly valid for the Version for live item to differ from the draft. This invariant should only really apply at the point of publish. Maybe this is something we move into the publish command or conditionalise it on the model.

What do people think?

tuzz:
> @elliotcm: I would like some thought put into the effects on debugging when lots of joins are required to get a full picture of the current state of a given piece of content.

I think this is a fair point and I'm not really sure how it would affect debugging. I guess that I'd argue it would be easier to debug an issue if your context is leaner - i.e. you don't have all of these other irrelevant concepts floating around.

Maybe this is a bit of a wooly argument – so feel free to shoot it down. :sheep: :gun: 

**Real world example:**

Let's say that I have a problem with my debit card. In order to debug this problem, I'd probably gather information about my bank account, a telephone number for customer services, my pin number, etc. I wouldn't gather information about holidays I've been on or things from my Twitter stream. These are probably irrelevant and distract from the issue at hand.

What do people think?

tuzz:
> @danielroseman: My only comment would be to wonder whether we're making things harder for ourselves when we come to implementing the optimistic locking. Now that version is no longer part of the model, it is rejected as part of the payload. Not sure if this is a concern.

I'm not sure I understand *"it is rejected as part of the payload"* - please could you elaborate a bit further.

My understanding of how optimistic locking should work within the context of this app is:

1) Request comes in to create a draft piece of content

2) The version number in the request body is checked against the version in the database

3) If someone else has incremented this version, the decline the request

Is this correct?

elliotcm:
@tuzz Slightly more real world example, content store rejects pushed documents because of version mismatches (put in place due to async out of order issues), and now I need to go look up a specific document and see what version it's at in the live state.  Probably quite a common check.

If version and state were in different tables I've got 2 joins to do every time.

It's all about what information we have to work with from our logs and bug reports, and how many steps are required to get from that to a full view of the document.

elliotcm:
I'm not necessarily saying this implementation will cause concrete issues in these areas, but that's the sort of thing I meant when I said putting some thought in, to make sure our common tasks are easy to do.

tuzz:
@elliotcm you're right – it would be two joins to do this. That does add a bit of an overhead.

steventux:
I paired with @tuzz on the original version validation story and I think this approach simplifies some of the problems we experienced when adding validations.

As a concept, it cleans up a lot of the mixed in concerns, something we've fought with in Edition models in Whitehall and Content Models.

I was wary of persistence per model in this case, possibly because there's not much to `Version`.
There's a trade off with keeping the central content item model clean and table proliferation, complexity and performance.
Having discussed this a few times though I'd prefer to see separate persistence, it's notionally cleaner, it's not irreversible if it proves too complex or slow.

steventux:
@tuzz @elliotcm perhaps materialized views could help dealing with the data directly. Would it be possible to have a read-only object sitting on top of a view? Perhaps something like [this](https://medium.com/@WoloxEngineering/activerecord-postgresql-materialized-views-9a9ef56b5a8d#.v6lslb8yi)
Presumably something like this could also produce the derived representation?
I admit complexity is on the increase in this suggestion...

danielroseman:
@tuzz re optimistic locking, that sounds fine, I just wanted to be sure that we hadn't written ourselves into a corner.

tuzz:
rebased

PR #106, opened 2015-10-26
Avoid setting LinkSet#links to nils

tadast:
I've encountered a bug while migrating github.com/alphagov/service-manual-publisher from V1 to V2. After some investigation I've uncovered that the V1 API sets LinkSet#links to nil values in some cases (see commit notes). V2 code does not expect that and as a result accidentally posts `links: nil` to the content store, that then breaks because #empty? is not defined on nils ([see here](https://github.com/alphagov/content-store/blob/0a263ada809ae5064bd73dcc0dc91314917cb7bc/app/models/content_item.rb#L205))

This is my approach at fixing the problem in this repo, but once this is merged we could make the code in content-store more robust too, also, perhaps we can consider validating against nil values in LinkSet#links?

I haven't worked on this repo and appreciate there could be better ways to solve this problem, so I'm fine if you want to throw this away and provide a better fix :) Otherwise I'm happy to modify this PR based on your comments.

/cc @alext

### DIFF
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -69,7 +69,7 @@ module Commands
     end
 
     def create_or_update_links!
-      LinkSet.create_or_replace(content_id: content_id, links: content_item[:links]) do |link_set|
+      LinkSet.create_or_replace(content_id: content_id, links: content_item[:links] || {}) do |link_set|
         link_set.version += 1
       end
     end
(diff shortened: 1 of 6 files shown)


tijmenb:
Should we add `null: true` to the database field too? https://github.com/alphagov/publishing-api/blob/master/db/schema.rb#L50

tadast:
I assume you mean `null: false`. I think it would make more sense than a validation, because all code responsible for setting this value lives in this application and an errbit notification would make more sense than an error message to the client application.

danielroseman:
Thanks for spotting and fixing this, Tadas.

tadast:
@danielroseman do you want me to do any of the extra things mentioned above i.e. `null: false` or a fix in content-store?

PR #107, opened 2015-10-27
Add import_path_reservations task.

danielroseman:
This imports path reservations previously exported from URL Arbiter. Note that
all existing path reservations will be deleted before import; this is because
URL Arbiter will always contain the canonical data.

### DIFF
--- a/lib/tasks/import_data.rake
+++ b/lib/tasks/import_data.rake
@@ -1,4 +1,5 @@
 require "tasks/import_data"
+require "tasks/import_path_reservations"
 
 desc "Imports all content items from the provided JSON file (one doc per line)"
 task :import_content_items, [:json_file_path, :state] => :environment do |_, args|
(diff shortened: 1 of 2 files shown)


steventux:
I ran this successfully against preview data:

```bash
vagrant@development:/var/govuk/publishing-api$ wc -l ../url-arbiter/tmp/reservations_2015-10-27_16-53-41.json 
219235 ../url-arbiter/tmp/reservations_2015-10-27_16-53-41.json
irb(main):001:0> PathReservation.count
   (84.8ms)  SELECT COUNT(*) FROM "path_reservations"
=> 219235
```

steventux:
@danielroseman  An attempt to import data in preview failed with a base path uniqueness validation error.
Given we clear all `PathReservation` records at the start of the task it's possible this is the result of Panopticon putting a path reservation to Publishing API while the import was running?

PR #108, opened 2015-10-29
Publishing API validations

tuzz:
https://trello.com/c/PIuXfDjy/339-make-publishing-api-validate-content

I've noticed that the migration `20151026112807_make_base_path_unique.rb` fails locally after pulling down data from preview. After speaking with @danielroseman – we're going to rebuild the derived representations from @elliotcm's script. We'll need to do this before this can be merged.

- [x] Preview data has been validated locally against the new validations
- [ ] Derived representations have been rebuilt from @elliotcm's script


danielroseman:
Generally :+1:

There's clearly a lot of duplication between the validations in Draft and Live content items. Is this something we should be tidying up at this point?

tuzz:
> @danielroseman: There's clearly a lot of duplication between the validations in Draft and Live content items. Is this something we should be tidying up at this point?

I have addressed this in the refactor ticket, see [this commit](https://github.com/alphagov/publishing-api/commit/346b6d6b570967688d46847868356de2a6899634)

tuzz:
I have addressed your previous comment re. include/exclude @danielroseman.

Please could you take another look. Thanks.

PR #111, opened 2015-11-02
Routes validator

tuzz:
This is a refactor of the RegisterableRouteSet object graph to consolidate these validations into a custom activerecord validator as this is more conventional.

In the content store, these models were also being used for registering the routes with Panopticon(?) and so it made more sense to have these as ActiveModels. Publishing API doesn't need to do this (yet) so we decided to simplify this.

I also combed through the tests and made sure that all paths through the validator are covered as there were a few bits missed.


spec/support/routes_and_redirects_validator.rb
```diff
@@ -0,0 +1,226 @@
...
+      expect(subject).to be_invalid
+      expect(subject.errors[:routes]).to eq(["is not a valid absolute URL path"])
+    end
+
+    it "does not throw an error when routes is nil rather than an empty array" do
+      subject.routes = nil
+      expect { subject.valid? }.to_not raise_error
+    end
+
+    context "for a redirect item" do
```
alext: I don't think this is quite correct.  All formats can include redirects along with the routes. redirect items are special in that they only include redirects (and therefore require the redirects to include the `base_path` instead of the routes).

PR #113, opened 2015-11-02
Support optimistic locking on v2 endpoints.

elliotcm:
This allows publisher applications to detect conflicts
when (for example) a user has left their edit page
open for a long time and someone else has updated
the item in the meantime.

If the version the app thinks it's updating does
not match the version in the Publishing API, a 409
Conflict is returned.

If the `previous_version` key is not provided, no
checking is performed.  This makes this locking
opt-in on the part of the publisher application.

https://trello.com/c/s3dN4eDP/336-optimistic-locking

Client changes here: https://github.com/alphagov/gds-api-adapters/pull/383

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -12,6 +12,11 @@ module Commands
 
     private
       def validate!
+        validate_update_type!
+        validate_version_lock!
+      end
+
+      def validate_update_type!
         raise CommandError.new(
           code: 422,
           message: "update_type is required",
@@ -25,6 +30,10 @@ module Commands
             }
           }
         ) unless update_type.present?
+      end
+
+      def validate_version_lock!
+        super(DraftContentItem, content_id, payload[:previous_version])
       end
 
       def content_id
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -233,4 +233,36 @@ Pact.provider_states_for "GDS API Adapters" do
       FactoryGirl.create(:version, target: arabic_draft, number: 1)
     end
   end
+
+  provider_state "the content item bed722e6-db68-43e5-9079-063f623335a7 is at version 3" do
+    set_up do
+      DatabaseCleaner.clean_with :truncation
+
+      draft = FactoryGirl.create(:draft_content_item, content_id: "bed722e6-db68-43e5-9079-063f623335a7")
+      FactoryGirl.create(:version, target: draft, number: 3)
+
+      stub_default_url_arbiter_responses
+      stub_request(:put, Regexp.new('\A' + Regexp.escape(Plek.find("content-store")) + "/content"))
+      stub_request(:put, Regexp.new('\A' + Regexp.escape(Plek.find("draft-content-store")) + "/content"))
+    end
+  end
+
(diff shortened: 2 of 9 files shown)


spec/service_consumers/pact_helper.rb
```diff
@@ -204,4 +204,34 @@
...
+
+  provider_state "the content item bed722e6-db68-43e5-9079-063f623335a7 is at version 3" do
+    set_up do
+      DatabaseCleaner.clean_with :truncation
+
+      draft = FactoryGirl.create(:draft_content_item, content_id: "bed722e6-db68-43e5-9079-063f623335a7")
+      FactoryGirl.create(:version, target: draft, number: 3)
+
+      stub_default_url_arbiter_responses
+      stub_request(:put, /content-store/)
```
alext: I wonder whether this has the potential to match unexpected thigns in future.  It would probably be better to match this against an anchored base URL (like on line 190).  That also means that stubs that have the same intent are expressed in the same way.

elliotcm: There's a lot of inconsistency in these different setups.  Some don't stub them at all but don't have any indication as to why.

What's the likelihood of this matching something unexpected?

elliotcm: I guess what I'm saying is they don't feel right here at all since the stubs are related to the behaviour which is related to the request more than the setup, so I'd rather get this merged and then talk about how we can do it better.

app/commands/v2/publish.rb
```diff
@@ -27,6 +32,28 @@ def validate!
...
+          error_details: {
+            error: {
+              code: 409,
+              message: "Version conflict",
+              fields: {
+                previous_version: ["does not match"],
+              }
+            }
+          }
+        ) unless current_version.number == previous_version_number
```
alext: Having a multi-line block with an `unless` on the end isn't the easiest to follow.  Would be easier to follow if this used the block form of `unless`.

elliotcm: I was being consistent with the other validations.

app/commands/v2/publish.rb
```diff
@@ -27,6 +32,28 @@ def validate!
         ) unless update_type.present?
       end
 
+      def validate_version_lock!
+        current_content_item = DraftContentItem.find_by(content_id: content_id)
+        current_version = Version.find_by(target: current_content_item)
+        previous_version_number = payload[:previous_version]
```
tuzz: Given this business logic is duplicated in the various command objects, would it be possible to move it to the model layer?

```ruby
Version.conflict?(current_content_item, payload[:previous_version]) # Something like this?
```

tuzz: Or maybe even:

```ruby
Version.about_to_override_someone_elses_changes?
```

elliotcm:
@alext @tuzz I've refactored the command objects.

alext:
I feel this could do with some integration-level tests that cover things like nothing should be sent to the content-store etc when a version conflict is detected.

tuzz:
Thanks @elliotcm – looks good!

elliotcm:
This is now updated to include request specs.

tuzz:
:+1:

PR #114, opened 2015-11-02
Set up pry to aid debugging…

tuzz:
I think that [pry](https://github.com/pry/pry) is a much better debugging tool
than the one provided by byebug.

It provides syntax highlighting and should be a
superset of all of byebugs features.

You invoke it by calling `binding.pry` from
anywhere in the suite. The existing `debugger`
command still continues to work.

Any objections?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,8 +25,8 @@ group :development do
 end
 
 group :development, :test do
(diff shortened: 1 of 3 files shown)


mattbostock:
I'm a fan of Pry, but I don't work on this application. :+1: from me.

PR #119, opened 2015-11-04
[RFC - WIP]Add get linked endpoint

jackscotti:
#### This is a WIP branch, only opened for obtaining some feedback on whether I am approaching this problem correctly. The code requires refactoring.

We need an endpoint to get a list of the content which has a link of the given type to the given content-id.  This is needed to support the https://trello.com/c/DLakGcP0/351-add-a-way-to-monitor-that-the-taggings-in-panopticon-and-content-store-are-the-same story.

It is the `publisher-get-linked` endpoint [described in the RFC](https://gov-uk.atlassian.net/wiki/display/FS/RFC+31%3A+Requirements+for+tagging+architecture).

What the output of the endpoint looks like:
screen shot 2015-11-04 at 10 27 46

### DIFF
--- a/app/queries/get_linked.rb
+++ b/app/queries/get_linked.rb
@@ -1,0 +1,20 @@
+module Queries
+  module GetLinked
+    def self.call(content_id, link_type)
+      content_item_links = ContentItemLink.where(target: content_id, link_type: link_type)
+
+      if content_item_links.empty?
+        error_details = {
+          error: {
+            code: 404,
+            message: "Could not find any item with a link of type '#{link_type}' linked to this item: '#{content_id}'"
+          }
+        }
+
+        raise CommandError.new(code: 404, error_details: error_details)
+      else
+        content_item_links
+      end
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151029144554) do
+ActiveRecord::Schema.define(version: 20151102110431) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -16,6 +16,14 @@ ActiveRecord::Schema.define(version: 20151029144554) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "content_item_links", force: :cascade do |t|
+    t.string "source",    null: false
+    t.string "link_type"
+    t.string "target",    null: false
+  end
+
+  add_index "content_item_links", ["source", "target"], name: "index_content_item_links_on_source_and_target", using: :btree
+
   create_table "draft_content_items", force: :cascade do |t|
     t.string   "content_id"
     t.string   "locale",               default: "en"
--- a/spec/commands/v2/put_link_set_spec.rb
+++ b/spec/commands/v2/put_link_set_spec.rb
@@ -42,4 +42,15 @@ RSpec.describe Commands::V2::PutLinkSet do
       topics: [topic_uuid],
     )
   end
+
+  it "creates related ContentItemLinks" do
+    described_class.call(link_params)
+
+    first_link = ContentItemLink.first
+    second_link = ContentItemLink.last
+
+    expect(first_link.source).to eq content_id
+    expect(first_link.link_type).to eq "organisations"
+    expect(first_link.target).to eq new_uuid
+  end
 end
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -5,6 +5,7 @@ module Commands
         draft_content_item = create_or_update_draft_content_item!
         create_or_update_live_content_item!(draft_content_item)
         create_or_update_links!
+        create_or_update_content_item_links!
       end
 
       PathReservation.reserve_base_path!(base_path, content_item[:publishing_app])
@@ -76,6 +77,10 @@ module Commands
         version.increment
         version.save! if link_set.valid?
       end
+    end
+
+    def create_or_update_content_item_links!
+      ContentItemLinkPopulator.create_or_replace(content_id, content_item[:links] || {})
     end
   end
 end
--- a/app/commands/v2/put_link_set.rb
+++ b/app/commands/v2/put_link_set.rb
@@ -21,6 +21,8 @@ module Commands
           PublishingAPI.service(:queue_publisher).send_message(message_bus_payload(live_content_item))
         end
 
+        create_or_update_content_item_links!
+
         Success.new(links: link_set.links)
       end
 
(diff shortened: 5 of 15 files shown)


app/queries/get_linked.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      if linked_items.empty?
+        error_details = {
+          error: {
+            code: 404,
+            message: "Could not find any item with a link of type '#{link_type}' linked to this item: '#{content_id}'"
+          }
+        }
+
+        raise CommandError.new(code: 404, error_details: error_details)
```
alext: This doesn't look quite correct. It's not differentiating between the case where no content item exists with the given content_id, and the content item exists, but has no links of the given type. The latter should probably return empty array, not a 404.

spec/commands/v2/put_link_set_spec.rb
```diff
@@ -42,4 +42,15 @@
       topics: [topic_uuid],
     )
   end
+
+  it "creates related LinkedItems" do
+    described_class.call(link_params)
+
+    first_link = LinkedItem.first
+    second_link = LinkedItem.last
```
alext: These are relying on an ordering that's not guaranteed - there's nothing that guarantees that the first linked item will be the one you're interested in.  It would be better to make these queries specific to the item you're looking for.

Additionally, you're looking up `second_link`, but not then asserting against it below.

db/migrate/20151102110431_populate_linked_items.rb
```diff
@@ -0,0 +1,8 @@
+class PopulateLinkedItems < ActiveRecord::Migration
+  def change
+    LinkSet.all.each do |link_set|
```
alext: It would be more efficient to use `find_each` here (http://apidock.com/rails/ActiveRecord/Batches/find_each)

elliotcm:
Hi @jackscotti.

Small details aside, this looks like a good approach to me.  The only significant thing I'd change is the naming.  The things in that table aren't really linked items, they represent the links between items.

I would also question whether you need that table at all.  From the ticket it looks like this won't be an especially heavy-use endpoint, and we already have a table which contains the links between content items.  You could use postgresql's [JSON queries](http://www.postgresql.org/docs/9.3/static/functions-json.html) to achieve the same thing.

alext:
This could do with some integration level tests that cover the external behaviour of this feature.

jackscotti:
Thank you all for the great feedback.

@elliotcm this is an example of the simplest query that does what is required by the endpoint:
```
SELECT * FROM (SELECT content_id,links->>'organisations' link_type FROM link_sets) linkedItem where linkedItem.link_type LIKE '%96ae61d6-c2a1-48cb-8e67-da9d105ae381%’
````

It pulls all the items out of links as a string and then uses the '%' wildcards to look whether the wanted content id is in the string containing all the links. It's not really elegant and I spent more than a day to write it. :disappointed: 

From the chat I've had with @danielroseman it seemed that your team wanted to create this table anyway. What do you think is the better approach?

**Comment by elliotcm** [comment]
@jackscotti There's a bunch of JSON support functions that would probably make it easier, but if we need the table anyway let's go ahead with it.

**Comment by jackscotti** [comment]
@elliotcm from what I've understood browsing the Internet, there is a bunch of support for JSON only from Postgres 9.3. Anyway, I'll ask the team.

File: app/queries/get_linked.rb (hunk 1 above)
Comment (jackscotti) [inline app/queries/get_linked.rb]: you're completely right. I've rushed through this and missed or wrongly covered these cases.

**Comment by elliotcm** [comment]
@jackscotti yes, that's correct.  All our apps are on the 9.3 cluster now.

**Comment by jackscotti** [comment]
@elliotcm got confused, I meant to say 9.4.

File: app/queries/get_linked.rb (hunk 1 above)
Comment (jackscotti) [inline app/queries/get_linked.rb]: Re-reading this. I think both cases should return an empty array. @rboulton ?

File: db/schema.rb [hunk 4]
```diff
@@ -57,6 +57,14 @@
 
   add_index "link_sets", ["content_id"], name: "index_link_sets_on_content_id", unique: true, using: :btree
 
+  create_table "linked_items", force: :cascade do |t|
+    t.string "source",    null: false
+    t.string "link_type"
+    t.string "target",    null: false
+  end
+
+  add_index "linked_items", ["source", "target"], name: "index_linked_items_on_source_and_target", using: :btree
```
tijmenb: Don't forget foreign keys here.

tuzz:
To echo some of what's already been said – we already have a table that stores this information. We should probably use that for serving this request.

I think that if we want to start querying our data as opposed to just plumbing it through to the content store, we should be thinking about persisting it in a structure that's conducive to this. I think it would be better if we had a `links` table that is many-to-one with `content_items`, as opposed to `link_sets`.

This takes the responsibility of handling the multiplicity of this relation out of the application and captures it semantically in the database. That's kind of how relational databases are intended to be used.

jackscotti:
We were just discussing about the very same thing and we all seem to agree on it. Tomorrow I'll pop over and talk to your team to have a further discussion.

jackscotti:
Closing this PR. We agreed this work will be split in 3 different PRs:

First:
 - add a `ContentItemLink` table that is the derived representation of the `links` in `LinkSet`, AND
 - populate `ContentItemLink`

Second:
 - modify the logic of the application to use `ContentItemLink`instead of `LinkSet`, AND
 - delete `LinkSet`

Third:
 - Add the `GetLinked` endpoint

jackscotti:
EDIT. Further update:

`LinkSet` table, would keep to exist, but would drop its `link` and `version` columns. `link` column will be represented by the new `Link` table and the `version` column can be dropped because it's redundant.

The reason of this change is that we want to keep versioning set of links and not links singularly.

First:
 - add a `ContentItemLink` table that is the derived representation of the `links` in `LinkSet`, AND
 - populate `ContentItemLink` table

Second:
 - modify the logic of the application to use `ContentItemLink`instead of `LinkSet`, AND
 - drop `links` and `version` form the `LinkSet` table

Third:
 - Add the `GetLinked` endpoint

tuzz:
I think we should have:

```
ContentItem <---------------------|
        | 1                               Version
        | 1                                    |
  LinkSet  <-------------------------|
        | 1
        | *
     Link
```

I think that we should think of these tables/models as the domain model for the application. When they're sent downstream to other applications, they are presented to adhere to the contract of the service.

tijmenb:
@tuzz Is that the same schema @elliotcm told us about?

tuzz:
Yes – I think so. My comment was intended to clarify to ensure we're on the same page.

elliotcm:
:+1:

tijmenb:
Cool! We are

tuzz:
Hey @jackscotti – we had a quick play around with writing the query we think you need. This is untested, but it might help:

```sql
select l.source
from link_sets ls
join links l on
l.link_set_id = ls.id and
ls.content_id = "some-content-id" and
l.type = "organisations"
```

In Rails, we think this would translate to:

```ruby
link_set = LinkSet.find_by!(content_id: "some-content-id")
content_ids = link_set.links.where(type: "organisations").pluck(:source)
```

Shout if you have any questions.

Thanks

PR #123, opened 2015-11-05
Force text error template when exceptions occur

tijmenb:
Rails by default shows nice error pages, with a lot of markup, css and javascript. This is cool when you visit the page in the browser, but not helpful when you consume the app as an API. Rails' HTML response then makes it really hard find the error when it's displayed as part of the exception message.

But Rails will [render a text-based exception](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L93-L99) when doing AJAX requests. By setting the `X-Requested-With` header we can make Rails show the text template for all requests.

Note that there's still some markup in the error. This is a Rails bug, addressed in https://github.com/rails/rails/pull/22172.

This is the result saving something in Panopticon that triggers an exception in publishing-api (in this case because I hadn't run the migrations). 

## Before

screen shot 2015-11-05 at 14 34 27
(and much more scrolling)

## After

screen shot 2015-11-05 at 14 36 27

I feel this is not a great solution. I'm considering a PR on Rails to show the text-template for all non-HTML requests, or make it configurable.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,11 @@ class ApplicationController < ActionController::Base
     head :bad_request
   end
 
+  before_action do
+    # Force Rails to show text-error pages
+    request.env['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest'
+  end
+
 private
   def respond_with_command_error(error)
     render status: error.code, json: error


AndrewVos:
lgtm!

tuzz:
Is there a way to make this a little less intrusive than something in `ApplicationController`?

Is there a `Rails.application.config` for it?

tijmenb:
Unfortunately the text template is only triggered for `request.xhr?` ([code](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L93-L99)), so no config yet. Might try making a PR to Rails to do this.

The only thing I can think of is inserting a middleware that sets the `X-Requested-With` header.

h-lame:
You could set `action_dispatch.show_detailed_exceptions` to `false` and write your own exception middleware that has a more reasonable behaviour than querying `request.xhr?` for showing the text version.

tijmenb:
@h-lame that would involve re-implementing https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb, maybe a ambitious for now.

Setting:

```
request.env['action_dispatch.show_exceptions'] = false
request.env['action_dispatch.show_detailed_exceptions'] = false
```

Shows this:

screen shot 2015-11-10 at 12 18 52

Which isn't nice, but maybe it's better than having to wade through 2800 lines of HTML to find the your error?

rboulton:
Merging this - it's an improvement, and there's no clear way to do it more nicely.

tuzz:
:+1:

PR #126, opened 2015-11-06
Implement the "publisher-bulk-get" endpoint

tijmenb:
This commit adds an endpoint to fetch all content-items for a given format. It will be used to populate the select-boxes in the tagging interfaces.

For example:

```
/v2/content?format=topic&fields=base_path,title
```

will return:

```json
[
  {
    "base_path": "/browse/benefits/entitlement",
    "title": "Benefits entitlement"
  }
]
```

This endpoint is described in https://gov-uk.atlassian.net/wiki/display/FS/RFC+31%3A+Requirements+for+tagging+architecture.

> Example: GET http://publishing-api/bulk format=topic&fields=title,content_id,base_path,draft,links.parent.title
> Get information about all content items of a particular format, including draft items.
> Only a limited set of fields need to be returned, but these may include:
> - fields from the top-level of the content item
> - whether the item is in draft state or not
> - fields from within the details hash
> - fields from within the expanded links hash
> The endpoint may also need to support pagination / filtering for efficiency.

Not all features described here have been implemented, because we might not need those yet/at all.

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -1,5 +1,11 @@
 module V2
   class ContentItemsController < ApplicationController
+    def index
+      content_format = params.fetch(:content_format)
+      fields = params.fetch(:fields)
+      render json: Queries::GetContentCollection.new(content_format: content_format, fields: fields).call
+    end
+
     def show
       render json: Queries::GetContent.call(params[:content_id], params[:locale])
     end
--- a/spec/factories/live_content_item.rb
+++ b/spec/factories/live_content_item.rb
@@ -33,10 +33,10 @@ FactoryGirl.define do
 
       draft = FactoryGirl.build(
         :draft_content_item,
-        content_id: live_content_item.content_id,
-        locale: live_content_item.locale,
-        base_path: live_content_item.base_path,
+        live_content_item.as_json(only: %i[content_id locale base_path format routes redirects]),
       )
+
+      raise "Draft is not valid: #{draft.errors.full_messages}" unless draft.valid?
 
       live_content_item.draft_content_item = draft
     end
--- a/spec/queries/get_content_collection_spec.rb
+++ b/spec/queries/get_content_collection_spec.rb
@@ -1,0 +1,63 @@
+require "rails_helper"
+
+RSpec.describe Queries::GetContentCollection do
+  it "returns the content items of the given format" do
+    create(:draft_content_item, base_path: '/a', format: 'topic')
+    create(:draft_content_item, base_path: '/b',  format: 'topic')
+    create(:draft_content_item, base_path: '/c',  format: 'mainstream_browse_page')
+
+    expect(Queries::GetContentCollection.new(
+      content_format: 'topic',
+      fields: ['base_path'],
+    ).call).to eq([
+      { "base_path" => "/a", "publication_state" => "draft" },
+      { "base_path" => "/b", "publication_state" => "draft" },
+    ])
+  end
+
+  it "returns the content items of the given format, and placeholder_format" do
+    create(:draft_content_item, base_path: '/a', format: 'topic')
+    create(:draft_content_item, base_path: '/b', format: 'placeholder_topic')
+
+    expect(Queries::GetContentCollection.new(
+      content_format: 'topic',
+      fields: ['base_path'],
+    ).call).to eq([
+      { "base_path" => "/a", "publication_state" => "draft" },
+      { "base_path" => "/b", "publication_state" => "draft" },
+    ])
+  end
+
+  it "includes the publishing state of the item" do
+    create(:draft_content_item, base_path: '/draft', format: 'topic')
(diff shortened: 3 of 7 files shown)


elliotcm:
This is something that the content register was designed to do, but as @tijmenb points out it only contains live content.

@danielroseman @rboulton Do we have a steer on whether we want to change the way the content register works?  e.g. policy publisher uses it to populate these sorts of fields currently.

rboulton:
I had assumed that content register was being deprecated, but that was just a personal assumption.

Seems like this needs more face-to-face discussion.

tijmenb:
Closed until meeting today at 16:00.

tijmenb:
To recap: we've decided to retire content-register in favour of this new endpoint. 

I'll update this PR:

- Make it api-compatible with content-register to provide an easy migration path
- Also return the `placeholder_#{format}` entries for a format

I guess we'll have to wait to migrate apps until all content-store data is in publishing-api?

/cc @jennyd @rboulton @elliotcm @danielroseman @tuzz @steventux

elliotcm:
@tijmenb yep

tijmenb:
I've added support for querying the placeholder format and added `publication_state` return field.

This endpoint can now return the same data as the `content-register`, except for `links`.

The links hash is only used by Whitehall (see https://github.com/alphagov/content-register/pull/21 and https://github.com/alphagov/whitehall/pull/2198) and I'm quite hesitant to add it now because it will add a bunch of complexity and will degrade performance. My proposal is to merge this first, and see what to do about the links when migrating Whitehall to use this endpoint.

tijmenb:
@elliotcm ready for review again.

app/controllers/v2/content_items_controller.rb
```diff
@@ -1,5 +1,11 @@
 module V2
   class ContentItemsController < ApplicationController
+    def index
+      format = params.fetch(:format)
```
elliotcm: Does this resolve the clash with rails params?

app/models/draft_content_item.rb
```diff
@@ -39,6 +39,10 @@ def viewable_by?(user_uid)
     !access_limited? || authorised_user_uids.include?(user_uid)
   end
 
+  def publication_state
+    live_content_item.present? ? 'live' : 'draft'
```
elliotcm: I'm going to say we should check whether there's a live item when rendering this in the presenter.

spec/factories/live_content_item.rb
```diff
@@ -33,11 +33,11 @@
 
       draft = FactoryGirl.build(
         :draft_content_item,
-        content_id: live_content_item.content_id,
-        locale: live_content_item.locale,
-        base_path: live_content_item.base_path,
+        live_content_item.as_json(only: %i[content_id locale base_path format routes redirects]),
       )
 
+      raise "Draft is not valid: #{draft.errors.full_messages}" unless draft.valid?
```
tuzz: :+1:

spec/queries/get_content_collection_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    create(:draft_content_item, base_path: '/c',  format: 'mainstream_browse_page')
+
+    expect(Queries::GetContentCollection.new(
+      format: 'topic',
+      fields: ['base_path'],
+    ).call).to eq([
+      { "base_path" => "/a", "publication_state" => "draft" },
+      { "base_path" => "/b", "publication_state" => "draft" },
+    ])
+  end
```
tuzz: It might make sense to include the content_id in the response as well since we're keying content items by them. What do you think?

app/models/draft_content_item.rb
tijmenb: You mean moving this into `Queries::GetContentCollection`?

app/controllers/v2/content_items_controller.rb
tijmenb: I've been back and forth on this. I've changed it all to `content_format` again to avoid any confusion.

rboulton:
:sparkles:

rboulton:
Attempting to access this in preview (using `curl -I 'http://publishing-api.preview.alphagov.co.uk/v2/content?content_format=topic&fields=base_path&fields=title'`) gives an HTTP 500 error (https://errbit.preview.alphagov.co.uk/apps/552637940da1157bfa000399/problems/5649a38265786306aa1f0600)

elliotcm:
@rboulton The example given in this PR uses a comma separated list

tijmenb:
The correct syntax is `/v2/content?format=topic&fields[]=base_path&fields[]=title`. I've amended the description.

tijmenb:
This is done so that this just works:

```ruby
publishing_api.get_content_items(
  content_format: 'topic',
  fields: [:title, :base_path]
)
```

https://github.com/alphagov/gds-api-adapters/blob/master/test/publishing_api_v2_test.rb#L917-L920

PR #127, opened 2015-11-09
Protect against out of order messages

tuzz:
This work makes it so that the publishing api sends a transmitted_at timestamp (in nanoseconds) to the content store for content item /v1 and /v2 incoming requests. We discussed this with @danielroseman and @bradwright and we think this is the best approach for now (without worrying too much about content item vs. link set race conditions). If we find that race conditions are an issue (we don't think they will be), we've got an upgrade path in mind for the future.

This work also sets up a pact for this specific piece of work between the publishing api and the content store. There's probably a job to back fill the pacts for the other requests that are made between these applications at some point in the future. As part of this set up work, we push the pact json file up to the pact broker. When we spoke to Alex he said that this was the right thing to do. We've more-or-less copied the rake task implementations for this from gds-adapters/publishing-api.

**Note:** The content store pull request will not pass on Jenkins until the Publishing API pull request in merged.

The sister pull request in the content store is here: https://github.com/alphagov/content-store/pull/166


app/commands/put_content_with_links.rb
```diff
@@ -10,10 +10,20 @@ def call(downstream: true)
...
+          access_limited: false,
+          update_type: false
+        )
+
+        Adapters::DraftContentStore.call(base_path, content_store_payload)
+        Adapters::ContentStore.call(base_path, content_store_payload)
+
+        message_bus_payload = Presenters::ContentStorePresenter::V1.present(
+          payload,
+          access_limited: false,
```
elliotcm: Feels to me that having different presenters for each downstream representation would be clearer, even if they just wrap some shared functionality.  For example, removing access limiting and update type are used here, but the removal of the base path is left to the command to do.  Encapsulating all that behaviour would make life easier for this and for reuse around the app.

app/commands/put_draft_content_with_links.rb
```diff
@@ -9,16 +9,18 @@ def call(downstream: true)
...
       end
 
       Success.new(content_item)
     end
 
-  private
-    def content_item_for_content_store
-      content_item.except(:update_type)
-    end
+    private
```
elliotcm: Outdented further than styleguide recommends.

app/commands/put_publish_intent.rb
```diff
@@ -3,14 +3,20 @@ class PutPublishIntent < BaseCommand
...
         base_path: base_path,
-        publish_intent: publish_intent
+        publish_intent: payload
       )
 
-      Success.new(publish_intent)
+      Success.new(payload)
+    end
+
+    private
```
elliotcm: Indented further than styleguide recommends.

app/commands/v2/publish.rb
```diff
@@ -69,21 +69,20 @@ def publish_content_item(draft_content_item)
           if live_version.number == draft_version.number
             raise CommandError.new(code: 400, message: "This item is already published")
           else
-            version = Version.find_or_initialize_by(target: live_item)
```
elliotcm: :+1:

app/commands/v2/put_content.rb
```diff
@@ -4,14 +4,15 @@ class PutContent < BaseCommand
       def call
         validate_version_lock!
 
-        content_item = create_or_update_draft_content_item!
+        content_item, version = create_or_update_draft_content_item!
```
elliotcm: `version` is unused here.

app/commands/v2/put_content.rb
```diff
@@ -27,26 +28,22 @@ def content_id
...
+
+        content_item = DraftContentItem.create_or_replace(content_item_attributes) do |item|
           version = Version.find_or_initialize_by(target: item)
           version.increment
           version.save! if item.valid?
 
           item.assign_attributes_with_defaults(content_item_attributes)
         end
+
+        [content_item, version]
```
elliotcm: We don't use the version returned here.

app/commands/v2/put_link_set.rb
```diff
@@ -13,21 +13,24 @@ def call
...
+          live_payload = Presenters::ContentStorePresenter.present(live_content_item)
           ContentStoreWorker.perform_async(
             content_store: Adapters::ContentStore,
             base_path: live_content_item.base_path,
-            payload: live_content_store_payload(live_content_item),
+            payload: live_payload,
           )
 
-          PublishingAPI.service(:queue_publisher).send_message(message_bus_payload(live_content_item))
+          message_bus_payload = message_bus_payload(live_payload)
```
elliotcm: If we're consolidating things through presenters we should probably have a message bus presenter.

app/presenters/content_store_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    def present
+      symbolized_attributes
+        .slice(*content_item.class::TOP_LEVEL_FIELDS)
+        .except(:id, :update_type)
+        .merge(public_updated_at)
+        .merge(links)
+        .merge(transmitted_at)
+    end
+
+    private
```
elliotcm: Indenting.

app/presenters/content_store_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    attr_accessor :content_item, :link_set
+
+    def symbolized_attributes
+      content_item.as_json.deep_symbolize_keys
```
elliotcm: Not sure the deep symbolize is necessary, we do that in the model.

app/presenters/content_store_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def initialize(content_item, link_set)
+      self.content_item = content_item
+      self.link_set = link_set
+    end
+
+    def present
+      symbolized_attributes
+        .slice(*content_item.class::TOP_LEVEL_FIELDS)
+        .except(:id, :update_type)
```
elliotcm: The ID isn't in the top level fields.

app/presenters/content_store_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+        {}
+      end
+    end
+
+    def public_updated_at
+      { public_updated_at: content_item.public_updated_at.iso8601 }
+    end
+
+    def transmitted_at
+      { transmitted_at: Time.new.to_f }
```
elliotcm: Should we be using `Time.zone.now.to_f`?

app/presenters/content_store_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    class V1
+      def self.present(attributes, access_limited: true, update_type: true, transmitted_at: true)
+        unless access_limited
```
elliotcm: We could inline these `if`s and `else`s.

lib/tasks/pact.rake
```diff
@@ -40,4 +40,18 @@ unless Rails.env.production?
   ensure
     ENV[key] = original_value
   end
+
+  require 'pact_broker/client/tasks'
+
+  def configure_pact_broker_location(task)
+    task.pact_broker_base_url = ENV.fetch("PACT_BROKER_BASE_URL")
+    if ENV['PACT_BROKER_USERNAME']
+      task.pact_broker_basic_auth =  { username: ENV['PACT_BROKER_USERNAME'], password: ENV['PACT_BROKER_PASSWORD']}
```
elliotcm: Mismatched whitespace in hugged braces.

spec/pacts/content_store_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+  let(:body) { Presenters::ContentStorePresenter.present(content_item) }
+  let(:request_time) { Time.at(1000000000.0000001) }
+
+  around do |example|
+    Timecop.freeze(request_time) { example.run }
+  end
+
+  it "accepts in-order messages to the content store" do
+    content_store
+    .given("a content item exists with base_path /vat-rates and transmitted_at 1000000000.0000000")
```
elliotcm: We've been indenting these elsewhere.

elliotcm:
What's the risk of accuracy issues when using floats for the timestamps?

spec/pacts/content_store_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+    )
+    .will_respond_with(
+      status: 200,
+      body: {},
+      headers: {
+        "Content-Type" => "application/json; charset=utf-8"
+      },
+    )
+
+    response = client.put_content_item(base_path: "/vat-rates", content_item: body)
```
elliotcm: Would be nice to split the setup (`given` etc) into a context.

spec/pacts/content_store_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+    let(:body) do
+      Presenters::ContentStorePresenter::V1.present(
+        attributes, access_limited: false, update_type: false
+      )
+    end
+
+    it "accepts in-order messages to the content store" do
+      content_store
+      .given("a content item exists with base_path /vat-rates and transmitted_at 1000000000.0000000")
+      .upon_receiving("a request to create a content item originating from v1 endpoint")
```
elliotcm: How is the implementation of this different from the previous two tests?  All three use the v1 endpoint.

spec/presenters/content_store_presenter_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      result = described_class.present(attributes, transmitted_at: false)
+
+      expect(result).to eq(
+        content_id: "content_id",
+        access_limited: "access_limited",
+        update_type: "update_type",
+      )
+    end
+  end
+end
```
elliotcm: Would be good to break out these tests so they test distinct presenters for the purpose.  Also there's a lot of uses of `:something => "a thing"` where `something: "a thing"` would do.

spec/spec_helper.rb
```diff
@@ -58,3 +60,12 @@
...
 end
+
+Pact.service_consumer "Publishing API" do
+  has_pact_with "Content Store" do
+    mock_service :content_store do
+      port 3093
+    end
+  end
+end
+
```
elliotcm: Bonus newline.

spec/requests/content_item_requests/downstream_requests_spec.rb
```diff
@@ -19,23 +20,13 @@
...
     sends_to_draft_content_store
-
-    it "sends to live content store" do
-      expect(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
-        .with(
-          base_path: base_path,
-          content_item: content_item_for_live_content_store,
-        )
-        .and_return(json_response)
-        .ordered
```
elliotcm: This has been removed and not restored elsewhere.

elliotcm:
By and large looks good.  A number of small-ish things to look at.

tuzz:
@elliotcm – thanks for the review. Will address first thing tomorrow. :-)

app/presenters/content_store_presenter.rb
tuzz: What's your thinking on this?

spec/pacts/content_store_spec.rb
tuzz: These requests have a different body. It's created as a result of the `Presenters::ContentStorePresenter::V1`.

We wanted to use a single presenter for both V1 and V2 that takes a content item and presents its attributes. However, the V1 commands do this:
https://github.com/alphagov/publishing-api/blob/master/app/commands/put_content_with_links.rb#L4

No content item is persisted if the content id was not in the request payload and so the presenter doesn't have a model to present.

spec/requests/content_item_requests/downstream_requests_spec.rb
tuzz: We're calling `sends_to_live_content_store` further down. There was already a shared support method for this.

elliotcm: `sends_to_live_content_store` doesn't check order.

tuzz: Ahh – will fix. Good spot.

tuzz:
> @elliotcm: What's the risk of accuracy issues when using floats for the timestamps?

`Time.new.to_f` reports the time in microseconds (millionth of a second). When we mapped over a list of items, we get a unique result for each call:

```
100.times.map { Time.new.to_f }
```

Is this what you mean? Or am I misunderstanding.

elliotcm:
@tuzz We're taking floats from one system, putting them into JSON, then retrieving them in another system.  Given that they're prone to inaccuracy at very high precision (which is what we're using) what's the risk of us losing that microsecond granularity?

spec/requests/content_item_requests/downstream_requests_spec.rb
tuzz: After discussing with @elliotcm – we realised that the `.ordered` call was only there because of the URL Arbiter, which no longer exists, so it's fine to do this.

PR #128, opened 2015-11-10
Expose version number in get links and get content endpoints...

steventux:
https://trello.com/c/GiuptR4Q/398-expose-version-in-v2-get-endpoints

Versioning is no longer a concern of the `DraftContentItem`, `LiveContentItem` or `LinkSet` models.
A side effect of this change is that the link set version number is no longer exposed via the get endpoints. 
This version number will be used to ensure optimistic locking of these models.
This change exposes the version number on both endpoints.

I've opted to use presenters to merge the version number into the command response, to achieve this I've added some presenters specific to the `Presenters::Queries` module, this was a workaround to a namespace collision with the current `Presenters::ContentItemPresenter` which should go away with this PR: https://github.com/alphagov/publishing-api/pull/127

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -12,11 +12,12 @@ module Presenters
       end
 
       def present
-        content_item.as_json.deep_symbolize_keys
+        content_item.as_json
+          .symbolize_keys
           .merge(version: version.number)
       end
 
-      private
+    private
 
       attr_accessor :content_item, :version
     end
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -1,0 +1,24 @@
+module Presenters
+  module Queries
+    class ContentItemPresenter
+      def self.present(content_item)
+        version = Version.find_by(target: content_item)
+        new(content_item, version).present
+      end
+
+      def initialize(content_item, version)
+        self.content_item = content_item
+        self.version = version
+      end
+
+      def present
+        content_item.as_json.deep_symbolize_keys
+          .merge(version: version.number)
+      end
+
+      private
+
+      attr_accessor :content_item, :version
+    end
+  end
+end
--- a/app/presenters/queries/link_set_presenter.rb
+++ b/app/presenters/queries/link_set_presenter.rb
@@ -12,11 +12,12 @@ module Presenters
       end
 
       def present
-        link_set.as_json.deep_symbolize_keys
+        link_set.as_json
+          .symbolize_keys
           .merge(version: version.number)
       end
 
-      private
+    private
 
       attr_accessor :link_set, :version
     end
--- a/app/presenters/queries/link_set_presenter.rb
+++ b/app/presenters/queries/link_set_presenter.rb
@@ -1,0 +1,24 @@
+module Presenters
+  module Queries
+    class LinkSetPresenter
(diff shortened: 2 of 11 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -0,0 +1,24 @@
...
+        self.content_item = content_item
+        self.version = version
+      end
+
+      def present
+        content_item.as_json.deep_symbolize_keys
+          .merge(version: version.number)
+      end
+
+      private
```
elliotcm: Indented too far for styleguide.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -0,0 +1,24 @@
...
+        new(content_item, version).present
+      end
+
+      def initialize(content_item, version)
+        self.content_item = content_item
+        self.version = version
+      end
+
+      def present
+        content_item.as_json.deep_symbolize_keys
```
elliotcm: I think the model already deep symbolizes its json fields.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -0,0 +1,24 @@
...
+          .merge(version: version.number)
```
elliotcm: Is it possible to not have a version?

app/presenters/queries/link_set_presenter.rb
```diff
@@ -0,0 +1,24 @@
...
+        link_set.as_json.deep_symbolize_keys
+          .merge(version: version.number)
+      end
+
+      private
+
+      attr_accessor :link_set, :version
+    end
+  end
+end
```
elliotcm: Issues for this are the same as the content presenter.

spec/presenters/queries/content_item_presenter_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe Presenters::Queries::ContentItemPresenter do
+  describe "present" do
+
```
elliotcm: Whitespace.

elliotcm:
:+1: Some small changes/questions.

app/presenters/queries/content_item_presenter.rb
steventux: @elliotcm from what I can see the concern `SymbolizeJSON` will call `deep_symbolize_keys` on Hash and Array values but doesn't symbolize top level keys, is my reading of this right? I've removed the `deep_symbolize_keys` and replaced with `symbolize_keys`. Without this top level keys are strings.

app/presenters/queries/content_item_presenter.rb
steventux: I've discussed this with @tuzz to be certain this should not happen. Both GET commands do a `find_by(content_id: content_id)` and respond with a `CommandError` in the absence of a record.
We enforce version population on the PUT endpoints so retrieved records should all have a version. From what I remember of the version work, legacy data has been migrated to have a version too.

steventux:
@elliotcm thanks for review, I've updated and commented where approriate.

app/presenters/queries/content_item_presenter.rb
elliotcm: I had assumed that `as_json` would provide top level symbols but you're right, it doesn't.

PR #129, opened 2015-11-11
Improve error message when sending a string as `previous_version`

tijmenb:
The current implementation of the version conflict checker assumes that `previous_version` is a integer. This isn't always the case, as clients will often send their version number as a string.

The following will currently work in client apps, because the integer type is preserved in the JSON:

```ruby
publishing_api.put_links(content_id, previous_version: 12, links: [])
```

The following will raise a HTTPConflict because `"12" != 12`:

```ruby
publishing_api.put_links(content_id, previous_version: "12", links: [])
```

This PR adds a better error message so that consumers can debug this class of errors more easily.

### DIFF
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -40,13 +40,17 @@ module Commands
 
       return unless current_versioned_item && current_version
 
+      friendly_message = "A version conflict occurred. The version you've sent " +
+        "(#{previous_version_number.inspect}) is not the same as the current " +
+        "version of the content item (#{current_version.number.inspect})."
+
       conflict_error = CommandError.new(
         code: 409,
         message: "Conflict",
         error_details: {
           error: {
             code: 409,
-            message: "Version conflict",
+            message: friendly_message,
             fields: {
               previous_version: ["does not match"],
             }


elliotcm:
Hi @tijmenb.

You're right that we need to fix this (and the friendlier error message is good), but I don't know if this is the right place for it.

It would be nice to have all publishing applications send numbers as numbers instead of strings, but in the spirit of accepting as much as we can I'm OK with the publishing API doing the conversion, but not in the version model.

The controllers should be the gatekeepers of converting incoming params to the right types.  This is where we do string to symbol conversion on the keys, for example.  The models are responsible for doing the same for data coming from the database.

The crux of it is doing these conversions at the boundaries of the system and being consistent internally, rather than at each point of use.

tijmenb:
@elliotcm I've removed the first commit as I can't decide between validating (`raise unless previous_version.nil? || previous_version.is_a?(Integer)`) and coercing `previous_version = previous_version.to_i if previous_version`. Both feel bad.

The improved error message should give the developer a hint that the types are off (by using `inspect`):

> A version conflict occurred. The version you've sent ("7") is not the same as the current version of the item (7).

tadast:
@tijmenb not sure how useful this is, but you could use `Integer(x)` for validation, it is less forgiving than `#to_i`.
```
irb(main):001:0> Integer(12)
=> 12
irb(main):002:0> Integer("12")
=> 12
irb(main):003:0> Integer("12.23")
ArgumentError: invalid value for Integer(): "12.23"
irb(main):004:0> Integer("123hello")
ArgumentError: invalid value for Integer(): "123hello"
irb(main):005:0> Integer(nil)
TypeError: can't convert nil into Integer
```

elliotcm:
@tijmenb, @tadast's mechanism looks top to me.

tijmenb:
Thanks @tadast, that's good to know!

@elliotcm the main issue still stands right? There's no obvious place to to this?

elliotcm:
@tijmenb I'd be fine merging it once all the changes to the tests go away (except a test checking it handles strings)

tijmenb:
@elliotcm I removed that commit yesterday, this PR is only the error message now. You kinda convinced me yesterday that making this a concern of the clients would be best.

PR #130, opened 2015-11-11
Mention Redis dependency in the docs

AndrewVos:


danielroseman:
I'm not sure this is really useful; we should absolutely mention the need for Redis to run sidekiq but we don't generally give instructions on installing third-party deps (and anyway, most people will run this inside the dev VM which has Redis installed already). In any case, there is much more that is out of date in this README than just this point. Closing in favour of a full rewrite.

PR #132, opened 2015-11-12
Add V2::ContentItemsController spec

steventux:
While documenting the v2 API endpoints I found a few responses which
weren't adequately covered and one scenario where a validation class raised an uncaught error.
This change adds some coverage specific to the V2 content items controller and remedies the uncaught error scenario.

### DIFF
--- a/app/validators/dns_hostname_validator.rb
+++ b/app/validators/dns_hostname_validator.rb
@@ -2,7 +2,7 @@ class DnsHostnameValidator < ActiveModel::EachValidator
   DNS_HOSTNAME_PATTERN = /\A[a-z0-9\-_]*\z/
 
   def validate_each(record, attribute, value)
-    unless value.match(DNS_HOSTNAME_PATTERN)
+    unless value && value.match(DNS_HOSTNAME_PATTERN)
       record.errors[attribute] << "is not a valid dns hostname"
     end
   end
--- a/spec/controllers/v2/content_items_controller_spec.rb
+++ b/spec/controllers/v2/content_items_controller_spec.rb
@@ -1,0 +1,119 @@
+require "rails_helper"
+
+RSpec.describe V2::ContentItemsController do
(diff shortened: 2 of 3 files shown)


app/validators/dns_hostname_validator.rb
```diff
@@ -2,8 +2,10 @@ class DnsHostnameValidator < ActiveModel::EachValidator
   DNS_HOSTNAME_PATTERN = /\A[a-z0-9\-_]*\z/
 
   def validate_each(record, attribute, value)
-    unless value.match(DNS_HOSTNAME_PATTERN)
-      record.errors[attribute] << "is not a valid dns hostname"
+    if value
+      unless value.match(DNS_HOSTNAME_PATTERN)
```
elliotcm: This could be `unless value && value.match..`

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+        expect(response.status).to eq(200)
+      end
+
+      it "responds with the content item as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.fetch("content_id")).to eq("#{content_id}")
+      end
+    end
+
+    context "for a non-existant content item" do
```
elliotcm: "non-existent"

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      it "responds with 404" do
+        get :show, content_id: "missing"
+
+        expect(response.status).to eq(404)
+      end
+    end
+  end
+
+  describe "put_content" do
+
```
elliotcm: Whitespace.

elliotcm:
Generally :+1:

steventux:
Thanks @elliotcm, updated.

PR #134, opened 2015-11-13
Add publication state to the content item GET presenter.

elliotcm:
This matches the info provided by the index endpoint.


tadast:
:trophy: :+1:

PR #135, opened 2015-11-16
[WIP-DO NOT MERGE]Add `Links` table & get linked endpoint

jackscotti:
Why:

- We want to introduce a [`get-linked-endpoint`](https://gov-uk.atlassian.net/wiki/display/FS/RFC+31%3A+Requirements+for+tagging+architecture). And in order to do that we need to be able to query the links easily. The current version of Postgres (9.4) does not allow us to do that, we therefore need to extract the `links` column into its own table.
- This work will also ease any other future query on the links data

DONE:

- Introduced a `Links` table
- Updated the logic of the app to use that able instead of the `links` column in the `LinkSets` table
- Populated the `Links` table with a migraiton
- Dropped the `links` column from `LinkSet` table with a migration

TODO:

- [x] Validate links somewhere (possibly within the `Link` model)
- [x] Address comments
- [ ] Rename this branch
- [ ] Add `get-linked` endpoint (in another PR)

Questions:

- Shuould `LinkSet` behave like a `Replaceable`?

Blockers:
- This work will make the app stop preserving the order in which the links are stored in it. 
Luckily enough there is an RFC just about that:
[RFC 36](https://gov-uk.atlassian.net/wiki/display/FS/RFC+36%3A+Stop+preserving+order+for+links).

You should only merge this PR if the RFC is accepted. We shoul also check whether some applications will not be able to handle the this new behaviour .

Trello Ticket:
https://trello.com/c/zppxFP6p/345-implement-the-publisher-get-linked-endpoint


app/commands/v2/put_link_set.rb
```diff
@@ -9,7 +9,29 @@ def call
           version.increment
           version.save! if link_set.valid?
 
-          link_set.links = merge_links(link_set.links, link_params.fetch(:links))
+          links_hash = link_params.fetch(:links)
+
+          links_hash.each do |link_type, target_content_ids|
+            if target_content_ids.empty?
+              Link.where(link_set: link_set, link_type: link_type).delete_all
```
tijmenb: This could probably be `link_set.links.where(link_type: link_type).delete_all`

app/commands/v2/put_link_set.rb
```diff
@@ -9,7 +9,29 @@ def call
...
+            else
+              old_target_ids = Link.where(link_type: link_type, link_set: link_set).map(&:target_content_id)
```
tijmenb: You could use `link_set.links.where(link_type: link_type).pluck(: target_content_id)`. 

`pluck` avoids loading all records and is faster.

app/commands/v2/put_link_set.rb
```diff
@@ -9,7 +9,29 @@ def call
...
+
+              ids_to_be_deleted = old_target_ids - target_content_ids
+              ids_to_be_created = target_content_ids - old_target_ids
+
+              ids_to_be_deleted.each do |id|
```
tijmenb: You can give active record queries arrays in a `where` clause, so this would do the same as these three lines, and would be one query:

```
link_set.links.where(link_type: link_type, target_content_id: ids_to_be_deleted).delete_all
```

app/models/link_set.rb
```diff
@@ -1,42 +1,56 @@
...
   include Replaceable
   include DefaultAttributes
   include SymbolizeJSON
 
-  validate :links_are_valid
-
-  def links=(links_hash)
-    super(links_hash || {})
-  end
+  # validate :links_are_valid
```
tijmenb: Yes, I think these migrations should go to the `Link` model.

app/presenters/queries/link_set_presenter.rb
```diff
@@ -14,7 +14,9 @@ def initialize(link_set, version)
       def present
         link_set.as_json
```
tijmenb: You have to check this, but I think we need a `includes(:links)` here now to avoid a N+1 query.

db/migrate/20151116114008_drop_links_column_from_links_sets.rb
```diff
@@ -0,0 +1,5 @@
+class DropLinksColumnFromLinksSets < ActiveRecord::Migration
+  def change
+    remove_column :link_sets, :links
```
tijmenb: It might be safer to rename this column to something like `legacy_links` until we've deployed all this, to avoid removing data accidently.

spec/integration/tagging_things_spec.rb
```diff
@@ -0,0 +1,179 @@
+require "rails_helper"
```
tijmenb: Before merging this PR we should probably remove this file. It's a good thing to have during this migration, but it's (hopefully) not testing anything new.

tijmenb:
:+1: 

Some minor comments, but this all seems really sound to me! Well done.

spec/integration/tagging_things_spec.rb
jackscotti: Yes, agree.

jackscotti:
@tijmenb thaks for the great feedback. Do you think I should implement the `get-linked` query in another PR? I will be on 2nd line from Wednesday so I assume it's better to split this work into different PRs.

tijmenb:
Yep, that's better in a separate one.

app/commands/v2/put_link_set.rb
jackscotti: wow, thanks!

app/presenters/queries/link_set_presenter.rb
jackscotti: Checked, this is not the case and there is no need for that. Thank you though.

jackscotti:
I renamed the branch and therefore created a new one. Closing this.

PR #136, opened 2015-11-16
Update rendering and publishing app factory attributes...

steventux:
The pacts for publishing API contain example publishing and rendering app values which aren't representative of real world app names. This has been misleading teams integrating publishing apps with the publishing API.


elliotcm:
:+1: but needs rebase

steventux:
@elliotcm thanks, rebased.

PR #137, opened 2015-11-17
Extract LinkSet's links column into its own table

jackscotti:
Why:

- We want to introduce a [`get-linked-endpoint`](https://gov-uk.atlassian.net/wiki/display/FS/RFC+31%3A+Requirements+for+tagging+architecture). And in order to do that we need to be able to query the links easily. The current version of Postgres (9.4) does not allow us to do that, we therefore need to extract the `links` column into its own table.
- This work will also ease any other future query on the links data

DONE:

- Introduced a `Links` table
- Updated the logic of the app to use that able instead of the `links` column in the `LinkSets` table
- Populated the `Links` table with a migraiton
- Dropped the `links` column from `LinkSet` table with a migration

TODO:

- [x] Validate links somewhere (possibly within the `Link` model)
- [x] Address comments
- [x] Rename this branch
- [x] Make `LinkSet` behave like a `Replaceable`

Blockers:
- ~~This work will make the app stop preserving the order in which the links are stored in it. 
Luckily enough there is an RFC just about that:
[RFC 36](https://gov-uk.atlassian.net/wiki/display/FS/RFC+36%3A+Stop+preserving+order+for+links).~~
- [Finder Frontend](https://github.com/alphagov/finder-frontend/pull/242) changes must be deployed first
- [Collections](https://github.com/alphagov/collections/pull/172) changes must be deployed first

You should only merge this PR if the RFC is accepted. We shoul also check whether some applications will not be able to handle the this new behaviour .

Trello Ticket:
https://trello.com/c/zppxFP6p/345-implement-the-publisher-get-linked-endpoint


app/commands/put_content_with_links.rb
```diff
@@ -73,10 +73,21 @@ def content_item_attributes
...
+      LinkSet.create_or_replace(content_id: content_id) do |link_set|
         version = Version.find_or_initialize_by(target: link_set)
         version.increment
         version.save! if link_set.valid?
+
+        link_set.links.delete_all
+
+        content_item[:links].each do |link_type, links|
+          links.each do |link|
+            link_set.links.create(
```
tuzz: Please could we use `#create!`

app/commands/v2/put_link_set.rb
```diff
@@ -9,7 +9,21 @@ def call
...
+            if target_content_ids.empty?
+              link_set.links.where(link_type: link_type).delete_all
+            else
+              old_target_ids = link_set.links.where(link_type: link_type).pluck(:target_content_id)
+
+              ids_to_be_deleted = old_target_ids - target_content_ids
+              link_set.links.where(link_type: link_type, target_content_id: ids_to_be_deleted).delete_all
+
+              ids_to_be_created = target_content_ids - old_target_ids
+              link_set.links.where(link_type: link_type, target_content_id: ids_to_be_created).create
```
tuzz: `#create!`

app/commands/v2/put_link_set.rb
```diff
@@ -9,7 +9,21 @@ def call
...
+            end
+          end
```
tuzz: I wonder if we could move this diff'ing logic to the model layer.

I think you could probably make `#links` a virtual attribute on the model that this method sets. The model then performs the diff itself and creates/deletes records. I'm not sure exactly how this would look.

app/models/link.rb
```diff
@@ -0,0 +1,27 @@
...
+  include Replaceable
+  include DefaultAttributes
+
+  validate :link_is_valid
+
+private
+  def link_is_valid
+    # Test that:
+    # - the `link_type` attribute is valid
+    # - the `target_content_id` attribute is a UUID
```
tuzz: Missing tests?

app/models/link_set.rb
```diff
@@ -1,42 +1,26 @@
...
 
-    bad_values = links.values.reject { |value|
-      value.is_a?(Array) && value.all? { |content_id|
-        UuidValidator.valid?(content_id)
-      }
-    }
-    if bad_values.any?
-      errors[:links] = "must map to lists of UUIDs"
-    end
+    links_hash.symbolize_keys
```
tuzz: `each.with_object` might be a little nicer here

app/models/link.rb
jackscotti: I think I covered everything in `spec/models/link_spec.rb`

app/models/link_set.rb
```diff
@@ -1,42 +1,26 @@
...
-  def links=(links_hash)
-    super(links_hash || {})
-  end
+  include Replaceable
 
-private
-  def self.query_keys
-    [:content_id]
-  end
+  def hashed_links
```
tuzz: I find this method name quite confusing. I was thinking MD5 hashes.

My understanding is that this method is building a hash of the links. Would `#as_json` suffice? If not, `#to_hash` might be better?

db/migrate/20151110102119_add_links_table.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :links do |t|
+      t.references :link_set, index: true, foreign_key: true
+      t.string :target_content_id, null: false
+      t.string :link_type, null: false
+
+      t.timestamps null: false
+    end
+
+    add_index :links, [:link_set_id, :target_content_id]
```
tuzz: Good thinking

db/migrate/20151116114008_retire_links_column_from_links_sets.rb
```diff
@@ -0,0 +1,5 @@
+class RetireLinksColumnFromLinksSets < ActiveRecord::Migration
+  def change
+    rename_column :link_sets, :links, :legacy_links
+  end
+end
```
tuzz: We normally leave the existing fields as is. We then add a story to trello to remove redundant fields.

Therefore, I'd leave this migration out for now (if possible).

tuzz:
A few comments. General approach looks good. I think this change will make things a lot nicer. Standard AR models keep things simple.

I'll push on getting the RFC through.

app/models/link_set.rb
jackscotti: This method builds a hash of the links and groups them by their link_type.

ie 
```
{
    "organisations" => ["a-uuid-xxx-xxx", "some-other-uuid"]
}
```
In order to do this, the methods you suggested wouldn't suffice but might simplify this method.

db/migrate/20151116114008_retire_links_column_from_links_sets.rb
jackscotti: Leaving this out to be picked up later would mean also removing the alias removed here:
https://github.com/alphagov/publishing-api/commit/d4b174dab8ab7369521c1f1f5c8091f800ccd1fa

I will also need to refer to `links` as `content_item_links`. I leaving this migration and the commit mentioned above in this PR would make everything easier and less confusing for the next person that will work on this.

app/models/link_set.rb
jackscotti: This method has been introduced to display links as was done before by the json blob column that has been removed. Maybe I should rename it?

jackscotti:
Thanks! I will address the feedback tomorrow.

app/models/link_set.rb
jackscotti: What would this do?

app/models/link.rb
tuzz: So do we need this comment?

app/models/link_set.rb
tuzz: each.with_object is a way of enumerating through a collection with a privately scoped object that you can add values to. In this case, you're creating a hash before the each, adding some values to it, then returning the hash. This is what each.with_object is intended to be used for.

db/migrate/20151116114008_retire_links_column_from_links_sets.rb
tuzz: I see – I guess that's reasonable. We just need to be extra careful that the migration works correctly.

app/models/link_set.rb
tuzz: I understand now. Maybe something like `#grouped_by_type` would be a better name.

I'd question why this needs to be in the model at all? Isn't this only a presentational concern when the links are sent downstream? We have some presenters for that. That might be a better place for this code.

jackscotti: Perfect reasoning, I'll move and rename this method.

app/models/link.rb
jackscotti: I'd say we do not need it, but the original version had a similar comment so I thought I should have left it there:
https://github.com/alphagov/publishing-api/blob/master/app/models/link_set.rb#L18

I am happy to remove it.

app/commands/v2/put_link_set.rb
jackscotti: Rather than moving only the diff'ing logic to the model layer, I'd move the whole link creation/update/delete within it. Maybe inside a method like:

```
def create_or_update_link_items(links, delete_old=false)
      ...
end
```

The method could then be called with `delete_all = true` here:
https://github.com/alphagov/publishing-api/pull/137/files#diff-f618486a57af99708e95cf5c3c2bae66R81

And with `delete_old=false` here in `put_link_set.rb`

app/models/link.rb
tijmenb: Maybe splitting up this method in `validate :link_type_is_valid` and `validate  :content_id_is_valid` would make sense.

app/models/link_set.rb
```diff
@@ -1,42 +1,32 @@
...
-      errors[:links] = "must map to lists of UUIDs"
-    end
+    links_hash.symbolize_keys
   end
 
-  def link_key_is_valid?(link_key)
-    link_key.is_a?(Symbol) &&
-      link_key.to_s.match(/\A[a-z0-9_]+\z/) &&
-      link_key != :available_translations
+private
```
tijmenb: Fun Ruby fact: the `private` call here doesn't work like you think it would.

See http://domon.cc/2013/12/25/private-class-methods-in-ruby/ or https://github.com/bbatsov/rubocop/issues/1367#issuecomment-58933163 for details.

spec/models/link_set_spec.rb
```diff
@@ -1,134 +1,50 @@
...
-
-    it "allows string keys that are underscored alphanumeric" do
-      [
-        'word',
-        'word2word',
-        'word_word',
-      ].each do |key|
-        subject.links = {key => []}
-        expect(subject).to be_valid, "expected item to be valid with links key '#{key}'"
+  context "#hashed_links" do
```
tijmenb: `describe` is better than `context` here.

app/models/link_set.rb
jackscotti: Learnt something new. Thanks.

app/models/link_set.rb
jackscotti: But this is not a class method, right?

tijmenb: Yup, `self.query_keys` is a class method.

jackscotti: I'm stoopid.

app/models/link.rb
tuzz: :+1:

db/migrate/20151116114008_retire_links_column_from_links_sets.rb
jackscotti: Mmmh, I could check that the migration worked with a rake task that compares the old links with the new ones. Do you think it's necessary?

app/models/link_set.rb
tijmenb: @tuzz https://github.com/alphagov/publishing-api/pull/149 is specifically made to address this comment. 

Once that is merged, we can than move the logic to `LinkSetPresenter#links` without having to change other parts of the app.

jackscotti:
@tuzz, I have refactored most of the code and rebased on master. Would you mind giving a look?

app/commands/put_content_with_links.rb
```diff
@@ -77,10 +77,23 @@ def content_item_attributes
     end
 
     def create_or_update_links!
-      LinkSet.create_or_replace(content_id: content_id, links: content_item[:links]) do |link_set|
+      LinkSet.create_or_replace(content_id: content_id) do |link_set|
         version = Version.find_or_initialize_by(target: link_set)
         version.increment
         version.save! if link_set.valid?
+
```
tadast: Did you consider wrapping this whole block of code in a database transaction?

```
ActiveRecord::Base.transaction do
  link_set.links.delete_all
  # link creation logic
end
```

This way if a link creation fails for some reason your database will remain in the before the transaction.

tijmenb: If I'm understanding it all correctly, everything called from a controller will be called within `EventLogger.log_command`, which is wrapped in a transaction, for this reason.

https://github.com/alphagov/publishing-api/blob/7ecc62dbbdd1c4d47c193bc93c2602ce9e0f22d9/app/event_logger.rb#L7

tijmenb:
:+1: This looks good as far as I can see, but 

Note that this shouldn't be merged before `collections` and `finder-frontend` have been deployed. This will happen tomorrow morning.

app/presenters/queries/link_set_presenter.rb
```diff
@@ -22,7 +25,21 @@ def present
...
+        # }
+
+        link_set.links.map.with_object({}) do |link, hash|
+          if hash.has_key?(link.link_type)
+            hash[link.link_type] << link.target_content_id
+          else
+            hash[link.link_type] = [link.target_content_id]
+          end
+        end.symbolize_keys
       end
```
tuzz: Could this be private?

tuzz: Something link this might be a bit more readable:

```ruby
link_set.links.each.with_object({}) do |link, hash|
  type = link.link_type.to_sym

  hash[type] ||= []
  hash[type] << link.target_content_id
end
```

(I haven't tested this)

tuzz:
Looks good – one minor point. :+1:

jackscotti:
@tuzz updated, the method to look clearer, thanks! Sadly, it can't be private because it's used outside ot that module.

tuzz:
Thanks @jackscotti – looks good to merge :tada:

tijmenb:
Dependencies have been deployed for this change. Good to go!

PR #139, opened 2015-11-17
Semantics docs

tuzz:
This is now complete and has been reviewed by @tadast and @AndrewVos.

I have addressed their comments.

https://trello.com/c/Zz8AXqBr/369-docs-semantics-of-publishing-api

### DIFF
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -33,6 +33,10 @@ a content item share a `content_id`
 `content_ids` are [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier)
 and will not be accepted by the Publishing API otherwise.
 
+Note: Previously, the `base_path` was a content item's main identifier. This is
+no longer the case. It has been changed to `content_id` because `base_paths` had
+a tendency to change.
+
 ---
 
 ### base_path
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -162,8 +162,8 @@ Example: *government-frontend*
 Required: Conditionally
 
 The `rendering_app` identifies the front-end application that is responsible for
-rendering the content item. Front-end applications will provide this string when
-querying the content store.
+rendering the content item. The router will use this information to direct
+users' requests to the appropriate front-end application.
 
 The `rendering_app` is required except in cases where the content item is
 non-renderable (see [**format**](#format)).
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -12,6 +12,10 @@ used. That document can be found
 
 ---
 
+### **Required Fields**
+
+---
+
 ### content_id
 
 Example: *d296ea8e-31ad-4e0b-9deb-026da695bb65*
@@ -78,6 +82,30 @@ for more information.
 
 ---
 
+### publishing_app
+
+Example: *collections-publisher*
+
+Required: Yes
+
+The `publishing_app` identifies the application that is responsible for
+publishing the content item. When a content item is created, its `base_path` is
+registered to the `publishing_app`. The path may not be used by a content item
+that was created with a different `publishing_app`.
+
+The `publishing_app` can then be used to filter content items appropriately when
+requests are made to the Publishing API. The `publishing_app` is also used as a
+means of auditing which applications are making use of the publishing pipeline.
+
+Note: `publishing_app` should be hyphenated as this is the convention used in
+the Router API.
+
+---
+
+### **Conditionally Required**
+
+---
+
 ### title
 
 Example: *VAT rates*
@@ -86,32 +114,6 @@ Required: Conditionally
 
 The `title` names the content item. It is required except in cases where the
 content item is non-renderable (see [**format**](#format)).
-
----
-
-### description
-
-Example: *VAT rates for goods and services*
-
-Required: No
-
-The `description` captures a short explanation of the content item. It is sent
-downstream to the content store and may be displayed at the discretion of the
-front-end application, when the content item is rendered.
-
----
-
-### need_ids
-
-Example: *["1234", "1235"]*
-
-Required: No
-
-The `needs_ids` are the identifiers of [user needs](https://www.gov.uk/design-principles)
-that are entered through the [Maslow Application](https://github.com/alphagov/maslow).
-They are passed through to the content store, untouched by the pipeline. The
-front-end applications can then use the `needs_ids` to present pages to users,
-such as [this one](https://www.gov.uk/info/overseas-passports).
 
 ---
 
@@ -135,26 +137,6 @@ speak to the Publishing Platform team if you have questions about this.
 
 ---
 
-### publishing_app
-
-Example: *collections-publisher*
-
-Required: Yes
-
-The `publishing_app` identifies the application that is responsible for
-publishing the content item. When a content item is created, its `base_path` is
-registered to the `publising_app`. The path may not be used by a content item
-that was created with a different `publishing_app`.
-
-The `publishing_app` can then be used to filter content items appropriately when
-requests are made to the Publishing API. The `publishing_app` is also used as a
-means of auditing which applications are making use of the publishing pipeline.
-
-Note: `publishing_app` should be hyphenated as this is the convention used in
-the Router API.
-
----
-
 ### rendering_app
 
 Example: *government-frontend*
@@ -167,42 +149,6 @@ users' requests to the appropriate front-end application.
 
 The `rendering_app` is required except in cases where the content item is
 non-renderable (see [**format**](#format)).
-
----
-
-### locale
-
-Example: *en*
-
-Required: No
-
-The `locale` is the language in which the content item is written. Front-end
-applications will optionally provide this string when querying the content store
-in order to retrieve content items in a given `locale`.
-
-A list of valid locales can be viewed in the Publishing API [here](https://github.com/alphagov/publishing-api/blob/master/config/application.rb#L32-L37).
-This field is not required and if a `locale` is not provided, it will be set to
-*en* automatically.
-
----
-
-### phase
-
-Examples: *alpha, beta, live*
-
-Required: No
-
-The `phase` is an optional field that may be used to indicate the ‘Service
-Design Phase’ of the content item. The phase must be one of either *alpha*,
-*beta* or *live*. If the `phase` is not specified, the Publishing API will
-default it to *live*.
-
-Content items will be published to the content store regardless of their
-`phase`.  If a content item has a `phase` of either *alpha* or *beta*, a visual
-component will be added to the front-end to show this (assuming the front-end
-application for that format supports this).
-
-There is more information on what each of the phases mean [here](https://www.gov.uk/service-manual/phases).
 
 ---
 
@@ -325,6 +271,72 @@ together with the `format` of the content item (e.g. *policy.major*)
 
 ---
 
+### **Optional Fields**
+
+---
+
+### description
+
+Example: *VAT rates for goods and services*
+
+Required: No
+
+The `description` captures a short explanation of the content item. It is sent
+downstream to the content store and may be displayed at the discretion of the
+front-end application, when the content item is rendered.
+
+---
+
+### need_ids
+
+Example: *["1234", "1235"]*
+
+Required: No
+
+The `needs_ids` are the identifiers of [user needs](https://www.gov.uk/design-principles)
+that are entered through the [Maslow Application](https://github.com/alphagov/maslow).
+They are passed through to the content store, untouched by the pipeline. The
+front-end applications can then use the `needs_ids` to present pages to users,
+such as [this one](https://www.gov.uk/info/overseas-passports).
+
+---
+
+### locale
+
+Example: *en*
+
+Required: No
+
+The `locale` is the language in which the content item is written. Front-end
+applications will optionally provide this string when querying the content store
+in order to retrieve content items in a given `locale`.
+
+A list of valid locales can be viewed in the Publishing API [here](https://github.com/alphagov/publishing-api/blob/master/config/application.rb#L32-L37).
+This field is not required and if a `locale` is not provided, it will be set to
+*en* automatically.
+
+---
+
+### phase
+
+Examples: *alpha, beta, live*
+
+Required: No
+
+The `phase` is an optional field that may be used to indicate the ‘Service
+Design Phase’ of the content item. The phase must be one of either *alpha*,
+*beta* or *live*. If the `phase` is not specified, the Publishing API will
+default it to *live*.
+
+Content items will be published to the content store regardless of their
+`phase`.  If a content item has a `phase` of either *alpha* or *beta*, a visual
+component will be added to the front-end to show this (assuming the front-end
+application for that format supports this).
+
+There is more information on what each of the phases mean [here](https://www.gov.uk/service-manual/phases).
+
+---
+
 ### access_limited
 
 Example: *{ users: ["bf3e4b4f-f02d-4658-95a7-df7c74cd0f50"] }*
@@ -367,6 +379,8 @@ that information about user activity can be tracked for the content item.
 ### links
 
 Example: *{ “related”: [“8242a29f-8ad1-4fbe-9f71-f9e57ea5f1ea”] }*
+
+Required: No
 
 The `links` contain the [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier)
 of other content items that this content item links to. For example, if a
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -270,7 +270,7 @@ element in this array is a JSON object that contains the properties *path*,
 *type* and *destination*. No other properties are supported. The *type* must be
 set to one of either *exact* or *prefix*.
 
-An *exact type* denotes that the *path* should checked for an exact match
+An *exact type* denotes that the *path* should be checked for an exact match
 against the user’s request URL when determining whether a redirect should occur.
 A *prefix type* denotes that any subpath under the specified path should be
 redirected to the destination. You can think of these as “exact” and “fuzzy”
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -24,21 +24,21 @@ Required: Yes
 
 The `content_id` is the content item’s main identifier and it forms its identity
 as it travels through the pipeline. This is why requests to create and query
-content items tend to be keyed by `content_id` in the URL of the request.
+content items are keyed by `content_id` in the URL of the request.
 
 Each `content_id` refers to a single piece of content, with a couple of caveats:
 
-- `content_ids` are shared across locales – the English and French versions of a
+- Content IDs are shared across locales – the English and French versions of a
 content item share a `content_id`
 
 - `content_ids` are shared across publish states – the draft and live versions of
 a content item share a `content_id`
 
-`content_ids` are [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+`content_ids` are [UUIDs](https://github.com/alphagov/govuk-content-schemas/blob/44dfad0cc241b7bd9576f0a7cf7f3fdeac8ddfce/formats/metadata.json#L94-L97)
 and will not be accepted by the Publishing API otherwise.
 
 Note: Previously, the `base_path` was a content item's main identifier. This is
-no longer the case. It has been changed to `content_id` because `base_paths` had
+no longer the case. It has been changed to `content_id` because base paths had
 a tendency to change.
 
 ---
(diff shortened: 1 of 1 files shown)


doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+### publishing_app
+
+Example: *collections-publisher*
+
+Required: Yes
+
+The `publishing_app` identifies the application that is responsible for
+publishing the content item. When a content item is created, its `base_path` is
+registered to the `publising_app`. The path may not be used by a content item
```
boffbowsh: `publising_app`

boffbowsh: (Sorry)

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+### content_id
+
+Example: *d296ea8e-31ad-4e0b-9deb-026da695bb65*
+
+Required: Yes
+
+The `content_id` is the content item’s main identifier and it forms its identity
+as it travels through the pipeline. This is why requests to create and query
+content items tend to be keyed by `content_id` in the URL of the request.
```
elliotcm: Tend to be?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Each `content_id` refers to a single piece of content, with a couple of caveats:
+
+- `content_ids` are shared across locales – the English and French versions of a
```
elliotcm: Technically "`content_id`s"

elliotcm: Or just "Content IDs"

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+content item share a `content_id`
+
+- `content_ids` are shared across publish states – the draft and live versions of
+a content item share a `content_id`
+
+`content_ids` are [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+and will not be accepted by the Publishing API otherwise.
```
elliotcm: Do we need to specify a format for the content ID?  Do we validate it anywhere other than local tests?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Note: Previously, the `base_path` was a content item's main identifier. This is
+no longer the case. It has been changed to `content_id` because `base_paths` had
```
elliotcm: Like with content IDs, we can just say "base paths" when referring to the concept rather than the actual field.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+a tendency to change.
+
+---
+
+### base_path
+
+Example: */vat-rates*
```
elliotcm: :D

doc/semantics-of-the-publishing-api.md
bradwright: govuk-content-schemas [validates that they're GUIDs](https://github.com/alphagov/govuk-content-schemas/blob/44dfad0cc241b7bd9576f0a7cf7f3fdeac8ddfce/formats/metadata.json#L19-L21)

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+### format
+
+Examples: *manual, policy, redirect*
+
+Required: Yes
+
+The `format` specifies the data format of the content item as per the
+[GOV.UK content schemas](https://github.com/alphagov/govuk-content-schemas).
+It is used downstream to render the content item in the manner
+it was intended.
```
elliotcm: "in the manner it was intended" is a bit vague.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+---
+
+### publishing_app
+
+Example: *collections-publisher*
+
+Required: Yes
+
+The `publishing_app` identifies the application that is responsible for
+publishing the content item. When a content item is created, its `base_path` is
```
elliotcm: "the application that published the content item"?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+registered to the `publishing_app`. The path may not be used by a content item
+that was created with a different `publishing_app`.
```
elliotcm: This has now been restricted further.  The path can't be used by any other content ID unless the content item is deleted first or if the new content ID is listed as a `can_be_replaced_by` link.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+The `publishing_app` can then be used to filter content items appropriately when
```
elliotcm: "appropriately"?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+requests are made to the Publishing API. The `publishing_app` is also used as a
+means of auditing which applications are making use of the publishing pipeline.
+
+Note: `publishing_app` should be hyphenated as this is the convention used in
```
elliotcm: Should make it clear that the value of the field should be hyphenated, not the field itself.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+Required: Conditionally
+
+The `title` names the content item. It is required except in cases where the
+content item is non-renderable (see [**format**](#format)).
+
+---
+
+### public_updated_at
+
+Example: *2015-01-01T12:00:00Z*
```
elliotcm: Do we need to specify a format?  e.g. ISO 8601?

bradwright: We should enforce ISO8601 (if we're not already).

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+Required: Conditionally
+
+The `details` (sometimes referred to as “details hash”) contains content and
+other attributes that are specific to the `format` of the content item. The
+[GOV.UK content schemas](https://github.com/alphagov/govuk-content-schemas)
+determine which fields appear in the details and which are required. The details
+can contain arbitrary JSON that will be stored against the content item.
+
+Not all `formats` have required fields and so details is not required unless the
+`format` demands it. If it is not set, it will default to an empty JSON object.
```
elliotcm: .. as specified in the content schema

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Example: *[{ path: “/vat-rates”, type: "exact" }]*
+
+Required: Conditionally
+
+The `routes` are used to configure the GOV.UK router for the content item. The
+`routes` appear as a JSON array and each element in this array is a JSON object
+that contains the properties *path* and *type*. No other properties are
+supported.  The *type* must be set to *exact* to denote that the route maps to
+an exact path on the GOV.UK website.
```
elliotcm: Do we still support "prefix"?  If not we should mention that the type is legacy and/or remove it.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+The `routes` are generally required, except for the case when the content item
```
elliotcm: Remove "generally"

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+has a `format` of *redirect*. In this case, the `routes` must not be present as
+it doesn’t make sense to have routes for a redirect. When the `format` is
+anything except *redirect*, the routes must include the `base_path` of the
+content item, at the very least.
```
elliotcm: "must include" makes "at the very least" redundant.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+The `redirects` are used to configure the GOV.UK router to specify redirects
+related to the content item. The `redirects` appear as a JSON array and each
+element in this array is a JSON object that contains the properties *path*,
+*type* and *destination*. No other properties are supported. The *type* must be
+set to one of either *exact* or *prefix*.
+
+An *exact type* denotes that the *path* should be checked for an exact match
+against the user’s request URL when determining whether a redirect should occur.
+A *prefix type* denotes that any subpath under the specified path should be
+redirected to the destination. You can think of these as “exact” and “fuzzy”
```
elliotcm: Maybe wildcard is a more widely understood term than fuzzy?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+matches, respectively.
+
+The `redirects` are generally optional, except for the case when the content
```
elliotcm: Remove "generally"

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+item has a `format` of *redirect*. In this case, the redirects must be present
+and they must include the `base_path` of content_item in the *path* property, at
+the very least.
```
elliotcm: - "at the very least"

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+---
+
+### update_type
+
+Example: *major*
+
+Required: Conditionally
+
+The `update_type` is an indicator of how much the content item has changed since
+the last time it was published. This field is required when publishing the
```
elliotcm: It's more about the importance of the change than the amount.  It could be two words but if it has a significant impact on the readers it would be a major change.  The litmus test is whether the readers would want to be notified about the change or not.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+This document explains the semantics of Publishing API V2. This includes:
+
+- What do each of the fields mean
+- What values are allowed in each field
+- What is the effect of choosing one value over another
+
+There is another document that explains the syntax for how the API should be
+used. That document can be found
+[here](https://docs.google.com/document/d/1zYFZFZ2TLzasgX4SZTxuQhLzqRYCk4hWKBTcb8QFIRA/edit).
```
danielroseman: Is this staying on Google Docs, rather than going into the wiki?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Example: *major*
+
+Required: Conditionally
+
+The `update_type` is an indicator of how much the content item has changed since
+the last time it was published. This field is required when publishing the
+content item.
+
+There’s no restriction on what the `update_type` should be, but at present we
```
elliotcm: Should we restrict this?  The original docs for the content store stated it must be one of the known types which might be added to later.  Three options I see:
1. We place no restrictions and make it a publisher/frontend/bus service concern
2. We place no programmatic restrictions but ask people to talk with us before adding a new type
3. We place programmatic restrictions and deal with type requests as a team

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+use:
+
+- *minor* - for minor edits that probably won’t be of interest to users
+
+- *major* - for major edits that probably will be of interest to users
```
elliotcm: Worth mentioning that these trigger email alerts.

doc/semantics-of-the-publishing-api.md
elliotcm: UPDATE

This is being backed out for gone and redirect items, making `can_be_replaced_by` irrelevant

elliotcm:
The fields are grouped well but ordered seemingly randomly.  Alphabetize?

Should we be indicating which fields support govspeak?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+---
+
+### need_ids
+
+Example: *["1234", "1235"]*
+
+Required: No
+
+The `needs_ids` are the identifiers of [user needs](https://www.gov.uk/design-principles)
+that are entered through the [Maslow Application](https://github.com/alphagov/maslow).
```
elliotcm: "Maslow application"

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+They are passed through to the content store, untouched by the pipeline. The
+front-end applications can then use the `needs_ids` to present pages to users,
```
elliotcm: This sentence doesn't provide any actual information about what type of pages are presented without clicking the link.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Example: *en*
+
+Required: No
+
+The `locale` is the language in which the content item is written. Front-end
+applications will optionally provide this string when querying the content store
+in order to retrieve content items in a given `locale`.
+
+A list of valid locales can be viewed in the Publishing API [here](https://github.com/alphagov/publishing-api/blob/master/config/application.rb#L32-L37).
```
elliotcm: Recommend linking to a blob on a commit rather than master as this may change.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+### access_limited
+
+Example: *{ users: ["bf3e4b4f-f02d-4658-95a7-df7c74cd0f50"] }*
+
+Required: No
+
+The `access_limited` field determines who should be granted access to the
+content item before it has been published. At the point of publish the content
+item is public and everyone has access to it. This field should be used when a
+content item needs to be drafted, but it should not be visible on
```
elliotcm: Drop the comma, "it", and mention that it shouldn't be visible on content preview except for authorised users (typically members of the organisation that created the content).

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+component will be added to the front-end to show this (assuming the front-end
+application for that format supports this).
+
+There is more information on what each of the phases mean [here](https://www.gov.uk/service-manual/phases).
+
+---
+
+### access_limited
+
+Example: *{ users: ["bf3e4b4f-f02d-4658-95a7-df7c74cd0f50"] }*
```
elliotcm: Not 100% sure that user UUIDs follow this format.

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+content-preview (and become public knowledge) until a publish occurs.
+
+The value of this field should be set to a JSON object that contains the key
+*users* and a value that is an array of [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier).
+These are the users that should be granted access to the content item. At
+present, the only supported key is *users*.  If `access_limited` is not set, no
+access restriction will be placed on the content item.
+
+When front-end applications make requests to the content store, they must supply
+the user they are making the request on behalf of if the content item is
```
elliotcm: How?  Details of the HTTP header?

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+items should appear in the `links` of the content item.
+
+The `links` for content items are managed separately to the content item itself.
+When changing `links`, it is not necessary to republish the content item.
+Instead, these changes will immediately be sent downstream to the draft content
+store and also to the live content store if the content item has previously been
+published.
+
+An update to the `links` causes a message to be placed on the message queue.
+This message will have a special update_type of links (see [**update_type**](#update_type)).
```
elliotcm: NB no emails sent

doc/semantics-of-the-publishing-api.md
```diff
@@ -0,0 +1,406 @@
...
+This queue is consumed by the Rummager application in order to reindex the
```
elliotcm: Is it?

doc/semantics-of-the-publishing-api.md
tuzz: :D

doc/semantics-of-the-publishing-api.md
tuzz: Ok – thanks for the update. I'll leave this as-is for now then.

doc/semantics-of-the-publishing-api.md
tuzz: The code that was ported over from Content store supports 'prefix' types. I don't know whether this is supported in the router, though. @danielroseman ?

doc/semantics-of-the-publishing-api.md
tuzz: Not sure, @steventux ?

doc/semantics-of-the-publishing-api.md
tuzz: I think option 2 sounds the most sensible. I don't think we should be too prescriptive about how update_type should be used as we don't want to be exposed to too much knowledge about downstream consumers.

That being said, I think it's important that our team knows what that field is being used for.

doc/semantics-of-the-publishing-api.md
steventux: The syntax docs are appended to @elliotcm's overview [on the atlassian wiki](https://gov-uk.atlassian.net/wiki/display/TECH/Publishing+Platform). IMO these would be better placed in GH along with examples and semantics, any objections?

doc/semantics-of-the-publishing-api.md
tuzz: I took this from the PACT: https://pact-broker.dev.publishing.service.gov.uk/pacts/provider/Content%20Store/consumer/Publishing%20API/latest

doc/semantics-of-the-publishing-api.md
tuzz: I'm not sure. @danielroseman, do you know what this HTTP header looks like?

doc/semantics-of-the-publishing-api.md
tuzz: Not sure. @danielroseman, @steventux ?

doc/semantics-of-the-publishing-api.md
tuzz: Addressed.

tuzz:
> @elliotcm: The fields are grouped well but ordered seemingly randomly. Alphabetize?

Done. I decided to keep routes/redirects together rather than wedging rendering_app in between. I also thought that it makes more sense to put routes before redirects.

doc/semantics-of-the-publishing-api.md
tuzz: Addressed.

tuzz:
> @elliotcm: Should we be indicating which fields support govspeak?

I've added this as a "Post merge" step on [that pull request](https://github.com/alphagov/publishing-api/pull/142). I'll document it then.

doc/semantics-of-the-publishing-api.md
tuzz: No objections. Please could you let me know if you move it so I can update this pull request. Thanks

tuzz:
I have addressed your comments as much as possible @elliotcm. There are a few questions that I have CC'ed other people into (@danielroseman, @bradwright, @steventux).

Thanks

doc/semantics-of-the-publishing-api.md
rboulton: That may well not remain true in future - in particular, we probably need to send email alerts when links are changed shortly after publication, even when the link change isn't sent as a major update.  I'd suggest adding the word "currently" to this, so that it doesn't look like this is setting a contract that can be relied on.

doc/semantics-of-the-publishing-api.md
bradwright: https://github.com/alphagov/content-store/pull/129

bradwright: Given that is implemented in the content store rather than in the Publishing API, I'd be tempted to either link to the content store OR say that "items with access_limited are limited to only the user IDs appearing in the list" (since the details of the content-store implementation may change).

tuzz: :+1: I decided to link to that pull request. I don't think it's likely to change any time soon and we'll just have to be mindful to update these docs if they do.

doc/semantics-of-the-publishing-api.md
elliotcm: I was talking about the ones used by the user model in signon specifically, but it turns out I was mistaken.

doc/semantics-of-the-publishing-api.md
bradwright:
```
irb(main):005:0> Route.where(route_type: 'prefix').count
=> 3462
```

elliotcm: @bradwright I guess my original question was are we supporting the creation of new ones generally rather than technically.  If so we should update the docs to include them as an option, if not we should mention that they used to be a thing but are no longer.

bradwright: Smart Answers still need prefix routes, as does Trade Tariff.

Confusing Smart Answers [says it needs an exact route](https://github.com/alphagov/smart-answers/blob/9f29165a85dadbd3f905acfb7e6e7fbe5ee9a55e/app/presenters/flow_content_item.rb#L20), but in the router they're all prefixes.

elliotcm: @tuzz Let's update these docs to allow for prefix routes then, but perhaps say they're discouraged in favour of exact routes.

tuzz:
I have addressed all comments again. @elliotcm @danielroseman @steventux @bradwright 

Thanks

bradwright:
For my part this looks good :+1:

PR #141, opened 2015-11-19
Allow gone/redirect replacement.

elliotcm:
This allows gones and redirects to replace content
and also allows content to replace gones and redirects
for republishing.

It does not allow normal content to overwrite other
normal content, reducing the chance of accidentally
blatting something important.

Half of https://trello.com/c/MtnVvpID/420-allow-overwriting-of-redirects-and-gones

### DIFF
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -48,6 +48,8 @@ module Commands
       )
 
       LiveContentItem.create_or_replace(attributes) do |item|
+        clear_space_for(item) if item.new_record?
+
         version = Version.find_or_initialize_by(target: item)
         version.copy_version_from(draft_content_item)
         version.save! if item.valid?
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -1,7 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe Commands::V2::PutContent do
-
   describe 'call' do
     before do
       stub_request(:put, %r{.*content-store.*/content/.*})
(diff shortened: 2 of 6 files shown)


app/commands/put_content_with_links.rb
```diff
@@ -48,6 +48,8 @@ def create_or_update_live_content_item!(draft_content_item)
       )
 
       LiveContentItem.create_or_replace(attributes) do |item|
+        clear_space_for(item)
```
danielroseman: Seems to me that we only need to do this when item itself is new, so maybe we can do `clear_space_for(item) if item.new_record?`

elliotcm: True.  Updated.

spec/commands/v2/put_content_spec.rb
```diff
@@ -59,5 +58,83 @@
...
+
+    context "when a content item wants to replace a content item" do
+      before do
+        create(:draft_content_item, base_path: base_path)
+      end
+
+      it "does not replace the existing content" do
+        expect {
+          described_class.call(payload)
+        }.to raise_error(CommandRetryableError)
```
danielroseman: Not part of this change necessarily, but we do probably want to return a different kind of error here; there's no point in retrying this call, it will never succeed.

danielroseman:
We need to do this in v1/put_content_with_links as well I think.

spec/commands/v2/put_content_spec.rb
elliotcm: Yeah, we can update this when it's changed.  The test will pick up the change.

elliotcm:
@danielroseman Already is.  I opted not to do specific tests for the v1 endpoints because the business logic is shared with the (fully tested) v2 endpoint, and the v1s are not long for this world.

danielroseman:
@elliotcm yes sorry, realised that straight after I made the comment and deleted it, but that doesn't unsend the notification...

PR #142, opened 2015-11-19
Forward html content type only

tuzz:
See [here](https://github.com/alphagov/publishing-api/blob/b532ef8c883598cc5faa91e45d4bba53a173bca2/doc/representation-for-multiple-content-types.md) for a decision record regarding the representation.

**Post deploy:**

- [ ] Document multiple representations in the content schemas
- [ ] Document multiple representations in the syntax / semantics docs
- [ ] Create a pull request to remove old_description columns

https://trello.com/c/nkofgTdj/372-add-raw-govspeak-to-publishing-api-payload

### DIFF
--- a/app/models/draft_content_item.rb
+++ b/app/models/draft_content_item.rb
@@ -38,6 +38,26 @@ class DraftContentItem < ActiveRecord::Base
   validates :description, well_formed_content_types: { must_include: "text/html" }
   validates :details, well_formed_content_types: { must_include: "text/html" }
 
+  # Postgres's JSON columns have problems storing literal strings, so these
+  # getter/setter overrides wrap and unwrap those strings in hashes.
+  def description=(value)
+    if value.is_a?(String)
+      super(string: value)
+    else
+      super
+    end
+  end
+
+  def description
+    value = super
+
+    if value.is_a?(Hash) && value.key?(:string)
+      value.fetch(:string)
+    else
+      value
+    end
+  end
+
   def viewable_by?(user_uid)
     !access_limited? || authorised_user_uids.include?(user_uid)
   end
--- a/app/models/draft_content_item.rb
+++ b/app/models/draft_content_item.rb
@@ -35,6 +35,9 @@ class DraftContentItem < ActiveRecord::Base
   }
   validates_with RoutesAndRedirectsValidator
 
+  validates :description, well_formed_content_types: { must_include: "text/html" }
+  validates :details, well_formed_content_types: { must_include: "text/html" }
+
   def viewable_by?(user_uid)
     !access_limited? || authorised_user_uids.include?(user_uid)
   end
--- a/app/models/draft_content_item.rb
+++ b/app/models/draft_content_item.rb
@@ -39,22 +39,23 @@ class DraftContentItem < ActiveRecord::Base
   validates :details, well_formed_content_types: { must_include: "text/html" }
 
   # Postgres's JSON columns have problems storing literal strings, so these
-  # getter/setter overrides wrap and unwrap those strings in hashes.
+  # getter/setter wrap the description in a JSON object.
   def description=(value)
-    if value.is_a?(String)
-      super(string: value)
-    else
-      super
-    end
+    super(value: value)
   end
 
   def description
-    value = super
+    super.fetch(:value)
+  end
 
-    if value.is_a?(Hash) && value.key?(:string)
-      value.fetch(:string)
+  def attributes
+    attributes = super
+    description = attributes.delete("description")
+
+    if description
+      attributes.merge("description" => description.fetch("value"))
     else
-      value
+      attributes
     end
   end
 
--- a/app/presenters/content_type_resolver.rb
+++ b/app/presenters/content_type_resolver.rb
@@ -1,0 +1,47 @@
+module Presenters
+  class ContentTypeResolver
+    def initialize(content_type)
+      self.content_type = content_type
+    end
+
+    def resolve(object)
+      if object.is_a?(Hash)
+        resolve_hash(object)
+      elsif object.is_a?(Array)
+        resolve_array(object)
+      else
+        object
+      end
+    end
+
+    private
+
+    def resolve_hash(hash)
+      hash.inject({}) do |hash, (key, value)|
+        hash.merge(key => resolve(value))
+      end
+    end
+
+    def resolve_array(array)
+      if has_content_types?(array)
+        extract_content(array)
+      else
+        array.map { |e| resolve(e) }
+      end
+    end
+
+    def has_content_types?(array)
+      return false unless array.all? { |v| v.is_a?(Hash) }
+
+      hash_keys = array.flat_map(&:keys).map(&:to_sym)
+      hash_keys.include?(:content_type)
+    end
+
+    def extract_content(array)
+      array = array.map(&:symbolize_keys)
+      array.detect { |h| h[:content_type] == content_type }[:content]
+    end
+
+    attr_accessor :content_type
+  end
+end
--- a/app/validators/well_formed_content_types_validator.rb
+++ b/app/validators/well_formed_content_types_validator.rb
@@ -1,0 +1,78 @@
+class WellFormedContentTypesValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    @error_messages = []
+
+    validate!(value)
+
+    if @error_messages.present?
+      record.errors[attribute] ||= []
+      record.errors[attribute].concat(@error_messages)
+    end
+  end
+
+  private
+
+  def validate!(object)
+    if object.is_a?(Hash)
+      validate_hash!(object)
+    elsif object.is_a?(Array)
+      validate_array!(object)
+    end
+  end
+
+  def validate_hash!(hash)
+    validate_array!(hash.values)
+  end
+
+  def validate_array!(array)
+    if has_content_types?(array)
+      validate_content!(array)
+    else
+      array.all? { |e| validate!(e) }
+    end
+  end
+
+  def has_content_types?(array)
+    return false unless array.all? { |v| v.is_a?(Hash) }
+
+    hash_keys = array.flat_map(&:keys).map(&:to_sym)
+    hash_keys.include?(:content_type)
+  end
+
+  def validate_content!(array)
+    array = array.map(&:symbolize_keys)
+
+    validate_that_content_is_present!(array) &&
+      validate_that_there_are_no_duplicates!(array) &&
+      validate_that_mandatory_content_type_is_present!(array)
+  end
+
+  def validate_that_content_is_present!(array)
+    array.each do |hash|
+      unless hash.key?(:content)
+        content_type = hash.fetch(:content_type)
+        @error_messages << "the '#{content_type}' content type does not contain content"
+      end
+    end
+  end
+
+  def validate_that_there_are_no_duplicates!(array)
+    groups = array.group_by { |hash| hash[:content_type] }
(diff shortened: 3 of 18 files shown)


app/models/draft_content_item.rb
```diff
@@ -35,6 +35,29 @@ class DraftContentItem < ActiveRecord::Base
...
+
+  def description
+    value = super
+
+    if value.is_a?(Hash) && value.key?(:string)
+      value.fetch(:string)
+    else
+      value
+    end
+  end
```
tuzz: If anyone has ideas on how to make this better – I'd appreciate it.

I had a quick poke around [in Rails](https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activerecord/lib/active_record/type/internal/abstract_json.rb#L13), but stopped as it began to spiral into a rabbit hole.

I had a quick go at upgrading Rails to 4.2.5 (we're currently 4.2.4), but this didn't help. We should probably do this upgrade anyway. I may put in a pull request.

app/presenters/content_type_resolver.rb
```diff
@@ -0,0 +1,47 @@
...
+        array.map { |e| resolve(e) }
+      end
+    end
+
+    def has_content_types?(array)
+      return false unless array.all? { |v| v.is_a?(Hash) }
+
+      hash_keys = array.flat_map(&:keys).map(&:to_sym)
+      hash_keys.include?(:content_type)
+    end
```
tuzz: This method is duplicated in the validator. Ideas for DRYing this up would be appreciated.

app/validators/well_formed_content_types_validator.rb
```diff
@@ -0,0 +1,78 @@
...
+      array.all? { |e| validate!(e) }
+    end
+  end
+
+  def has_content_types?(array)
+    return false unless array.all? { |v| v.is_a?(Hash) }
+
+    hash_keys = array.flat_map(&:keys).map(&:to_sym)
+    hash_keys.include?(:content_type)
+  end
```
tuzz: This method is duplicated in the resolver. Ideas for DRYing this up would be appreciated.

db/schema.rb
```diff
@@ -59,7 +60,7 @@
     t.string   "locale",                default: "en"
     t.string   "base_path"
     t.string   "title"
-    t.string   "description"
+    t.string   "old_description"
```
tuzz: I'm keeping the `old_description` fields for now (just in case). We can remove in another pull request.

spec/requests/content_item_requests/message_bus_spec.rb
```diff
@@ -141,4 +141,73 @@
...
+          transmitted_at: DateTime.now.to_s(:nanoseconds),
+        ).to_json
+
+        delivery_info, _, payload = wait_for_message_on(@queue)
+
+        expect(delivery_info.routing_key).to eq("guide.major")
+        expect(JSON.parse(payload)).to eq(JSON.parse(expected_payload))
+      end
+    end
+  end
```
tuzz: This test coverage was missing previously.

app/models/draft_content_item.rb
h-lame: I might be inclined to remove the type check:

```ruby
def description=(value)
  super(string: value.to_s)
end

def description
  super.fetch(:string)
end
```

You lose some correctness for cases where the description is being set as a properly constructed hash already, or it was somehow stored as an incorrect hash.  But, the api of the object is that the description attribute is a string and the content item will store that appropriately - if you give it anything else it'll get massaged into a shape we expect (e.g. what AR would do with any other field that you give it the wrong type of object - it'll do a simple coercion and serialize away).

doc/representation-for-multiple-content-types.md
```diff
@@ -0,0 +1,81 @@
+## Decision Record: Representation for Multiple Content Types
```
rboulton: This file looks great.  Would it be better in doc/arch/ to live with the existing ADR that's there?

app/models/draft_content_item.rb
tuzz: Good plan: https://github.com/alphagov/publishing-api/commit/dafa1de03d8d4b4eab63b9eaa9c886f2816a078f

That field can hold other things, that aren't just strings, so I called the key `value`.

I also overrode `#attributes`. See the commit message for more info.

What do you think?

doc/representation-for-multiple-content-types.md
tuzz: Good plan. Addressed.

app/models/draft_content_item.rb
danielroseman: My only concern with this is that it could make it harder to index this data if we ever need to search it. This may not be a problem in practice though.

db/migrate/20151119100750_change_description_column_types.rb
```diff
@@ -0,0 +1,19 @@
+class ChangeDescriptionColumnTypes < ActiveRecord::Migration
+  def change
+    [
+      [:draft_content_items, DraftContentItem],
+      [:live_content_items, LiveContentItem],
+    ].each do |table, model|
+      add_column table, :new_description, :json, default: { value: nil }
```
danielroseman: Given the discussions we've had about migrations which lock tables, and the advice here: https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/ we should probably split this into two actions.

doc/arch/representation-for-multiple-content-types.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+## Decision
+
+The decision was made to choose Option 1. This representation is not very
+compact, but it means that JSON objects could store additional information if
+required, such as a Govspeak version. It was noted that this option may not
+be the easiest for Publishing Applications to work with.
+
+Option 2 was refuted on the basis that it doesn't clearly demonstrate which
```
danielroseman: quibble: refuted is the wrong word here (it means "disproved"); maybe "rejected"?

app/presenters/content_type_resolver.rb
```diff
@@ -0,0 +1,47 @@
+module Presenters
+  class ContentTypeResolver
+    def initialize(content_type)
+      self.content_type = content_type
```
danielroseman: Why `self.` rather than `@`?

spec/requests/content_item_requests/message_bus_spec.rb
```diff
@@ -141,4 +141,73 @@
...
+
+  describe "content type inlining" do
```
danielroseman: Are we certain that this is what we want, ie only the HTML version goes on the queue?

danielroseman:
A few mostly minor comments, plus needs rebasing.

app/presenters/content_type_resolver.rb
tuzz: Two reasons:

1) I'm `nil` averse. If you misspell an instance variable, it's set to `nil` at runtime, which can lead to some nasty / hard-to-find bugs.

2) The object better encapsulates those fields. Contrived example:

Say we wanted to restrict it such that `content_type` must be a valid MIME type from a whitelist. By using a setter method on the object, we can put that logic in a single place rather than have to check it whenever we set the `content_type` on the object.

I try not to use instance variables anywhere if I can help it (except perhaps memoization).

spec/requests/content_item_requests/message_bus_spec.rb
tuzz: Not certain. Currently, Rummager strips tags from HTML prior to indexing. I spoke to @elliotcm about this and we thought that it's probably ok for it to only send the HTML content type forward, but happy to discuss further.

It could cause problems if we're "throwing information away" and then later find that we need it for some reason.

rboulton: Quick correction - rummager doesn't do any HTML tag stripping - but I think some publisher apps which are sending content to rummager do before they send it.

Regardless, I believe having only the HTML is fine for indexing - but we definitely need to know if a field is expected to contain HTML markup, so we can strip it out (or possibly make use of it to index some bits of text as being more important - eg, headings).

I think the current design is that we'd be able to find that out by reading the schema, but not simply by inspecting the document itself without reference to the schema. This is probably fine, though does make it a bit hard to manage if the schema changes from a field being a plain text field to being a govspeak + html field - there's no way for our indexing code to know about the change, and thus to ensure that it's stripping out HTML tags for all HTML fields.  This is probably something we'll just have to live with, though.

spec/presenters/content_type_resolver_spec.rb
```diff
@@ -0,0 +1,97 @@
+require "rails_helper"
+
+RSpec.describe Presenters::ContentTypeResolver do
+  subject { described_class.new("html") }
+
+  it "inlines content of the specified content type" do
```
rboulton: Could you add a test of what happens if the `html` version isn't present?  ie, if it gets
`body: [ { content_type: "text", content: "some content" } ]`

I'm not sure if it should return `"body": null`, or omit `body` entirely.  It definitely shouldn't fall back to any other representation (because then downstream apps won't know what content type the data they get is).

tuzz: I designed it such that the system fails early on receiving content that isn't "well formed" through a custom validator. The scenario you describe is tested here:

https://github.com/alphagov/publishing-api/blob/forward-html-content-type-only/spec/validators/well_formed_content_types_validator_spec.rb#L67-L76

My thinking was that I'd fail early on receiving illy formed content and relieve the ContentTypeResolver of this responsibility. I'm happy to double test this though – but I'd be a little concerned that people would think it's acceptable to send illy formed content to the resolver.

What do you think?

spec/requests/content_item_requests/message_bus_spec.rb
elliotcm: @rboulton Is there any harm to just stripping HTML on all fields?

elliotcm:
@tuzz Looks fine, just need @rboulton's sign-off on a couple of things and needs the migration splitting as per @danielroseman's comment.

tuzz:
All comments addressed, waiting on feedback from @rboulton on his points and confirmation from @danielroseman that the migration split is acceptable.

On this point, I've tested it against preview data with:

```ruby
DraftContentItem.find_each.map(&:description)
#=> [ ... ]

LiveContentItem.find_each.map(&:description)
#=> [ ... ]
```

Working as expected.

danielroseman:
:+1:, needs rebasing though.

tuzz:
Rebased.

spec/requests/content_item_requests/message_bus_spec.rb
rboulton: Stripping all HTML would be a problem, because we are quite likely to want to be able to use things like <h1> tags to mark some parts of the text as more important.  Plus, some fields may contain identifiers with characters like < which are significant.

spec/presenters/content_type_resolver_spec.rb
tuzz: Any thoughts @rboulton ?

spec/requests/content_item_requests/message_bus_spec.rb
tuzz: I see. We will have to bear that in mind when we tackle the consumer of those messages on the queue. Thanks for clearing that up.

Given this point is somewhat orthogonal to the code in this pull request – do you have any objections if we merge this? It's becoming stale and we're keen to get this through.

tuzz:
Following a discussion with @danielroseman, @bradwright and @elliotcm – we're going to close this and take a slightly different approach. We will send all representations downstream and resolve the 'html' content type in a presenter in the content store.

PR #144, opened 2015-11-20
Update public_updated_at on publish

steventux:
https://trello.com/c/BU5p5lG6/418-v2-publish-content-id-should-update-the-public-updated-at

Updates the `public_updated_at` timestamp on major updates.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -90,6 +90,12 @@ module Commands
           .attributes
           .except("access_limited", "version")
           .merge(draft_content_item: draft_content_item)
+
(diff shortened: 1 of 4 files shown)


tijmenb:
Are we sure there are no apps that want to override this? 

It may also be surprising `public_updated_at` can be set on minor/republish updates, and not in a `major` update.

tuzz:
I remember discussing this with @steventux and @danielroseman earlier this week / last week. I don't recall what the conclusion of that discussion was.

danielroseman:
As discussed, we should allow the value to be supplied in a publish action for any update_type, but if not supplied we should use the current time for major change and the existing persisted value for minor/republish.

PR #148, opened 2015-11-25
Accommodate LiveContentItems that do not have an associated DraftContentItem

tuzz:
https://trello.com/c/l2Fddq9e/425-relax-the-constraint-that-livecontentitems-must-have-an-associated-draftcontentitem


app/substitution_helper.rb
```diff
@@ -0,0 +1,42 @@
+module SubstitutionHelper
+  class << self
+    def clear_draft!(draft_content_item)
+      if clear!(draft_content_item, DraftContentItem)
+        clear_foreign_key!(draft_content_item)
+      end
+    end
+
+    def clear_live!(draft_content_item)
+      clear!(draft_content_item, LiveContentItem)
```
elliotcm: This is a live content item, not a draft content item.

app/substitution_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+  private
+    def clear!(draft_content_item, klass)
```
elliotcm: This could be either a live or draft content item.

app/substitution_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+      existing_content_item = klass.find_by(base_path: draft_content_item.base_path)
+      return unless existing_content_item
+
+      mismatch = existing_content_item.content_id != draft_content_item.content_id
+      allowed_to_substitute = substitute?(draft_content_item) || substitute?(existing_content_item)
```
elliotcm: Do you want to wrap these in parens?  I know it's not strictly required but reads a little easier.

elliotcm:
Some minor notes, otherwise looks good.  Would suggest someone else look at it too since I paired with you on chunks of it.

app/substitution_helper.rb
tuzz: Confusingly, it's not. We always lookup by the draft_content_item.

app/substitution_helper.rb
tuzz: See above.

spec/commands/v2/reallocating_base_path_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+  end
+
+  let(:substitute_payload) do
+    FactoryGirl.build(:redirect_draft_content_item,
+      content_id: content_id,
+      base_path: base_path
+    ).as_json.deep_symbolize_keys
+  end
+
+  context "when a base path is reserved by a not-yet-published regular content item" do
```
danielroseman: Talking about reserving base paths is a bit misleading because that implies the PathReservation stuff. Maybe "when a not-yet-published regular content item exists at a base path".

app/substitution_helper.rb
elliotcm: https://github.com/alphagov/publishing-api/pull/148/files#diff-a8f5f38617f3679840b8740e6627ab92R66

???

tuzz: Oh right – sorry, I was thinking of another thing. Will fix.

danielroseman:
The original code short-circuited the substitution checker if the item is not a new record, on the basis that there's no point going through all the complex querying etc since you know it's not trying to replace an existing item. Is that not still valid?

spec/commands/v2/reallocating_base_path_spec.rb
elliotcm: :+1: or "occupied by"

elliotcm:
@danielroseman if it *is* a new record

elliotcm:
but yes, we should keep that

tuzz:
> @danielroseman: The original code short-circuited the substitution checker if the item is not a new record, on the basis that there's no point going through all the complex querying etc since you know it's not trying to replace an existing item. Is that not still valid?

> @elliotcm: but yes, we should keep that

We kept that short-circuit. It's here:

https://github.com/alphagov/publishing-api/pull/148/files#diff-65b7caf2f046e9c7f6167f5abc5d7f46R16

We moved it inside of the SubstitutionHelper because knowledge of whether a substitution can be short-circuited or not is the responsibility of that helper.

tuzz:
All comments addressed: @elliotcm @danielroseman

elliotcm:
@tuzz That's not quite the same short cut.  The original checked whether the content item in the payload already existed in the DB, because if it does exist then the base path index will have already made sure they don't clash.

elliotcm:
@tuzz That would avoid the lookup of the "existing_content_item", though if we allow changing base path via the current v2 endpoint maybe this logic doesn't quite hold up.

tuzz:
> @elliotcm: That would avoid the lookup of the "existing_content_item", though if we allow changing base path via the current v2 endpoint maybe this logic doesn't quite hold up.

Hmm, I see what you mean – but I'm hesitant to change it because I don't want this logic to seep into the commands if we can avoid it. It's relatively cheap to do that lookup. Cohesion trumps performance?

What do you think?

elliotcm:
Merged because it's fine and also safer.

PR #149, opened 2015-11-25
Refactor LinkSetPresenter

tijmenb:
This is a change to make future changes easier.

Currently the `LinkSetPresenter` is only used when presenting the links for the `GET /v2/links/:content_id` endpoint. This commit makes it so that it is used in all instances where we send the links hash to the user:

- When setting the links (PUT /v2/links/:content_id)
- When presenting a document (/v2/content/:content_id)

This will make it easier to change what we present as `links` when we change the way links are stored in the database (#137).

Paired with @jackscotti on this.

### DIFF
--- a/spec/presenters/queries/link_set_presenter_spec.rb
+++ b/spec/presenters/queries/link_set_presenter_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Presenters::Queries::LinkSetPresenter do
-  describe "present" do
+  describe ".present" do
     before do
       FactoryGirl.create(:version, target: link_set, number: 101)
       @result = Presenters::Queries::LinkSetPresenter.present(link_set)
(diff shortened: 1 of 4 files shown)


steventux:
@tijmenb @jackscotti  :+1: makes sense to do this.

spec/presenters/queries/link_set_presenter_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Presenters::Queries::LinkSetPresenter do
-  describe "present" do
+  describe ".present" do
```
elliotcm: Technically this is `#present`.

tijmenb: Nope :) It's calling `LinkSetPresenter.present` in the test.

elliotcm: You are correct.

PR #151, opened 2015-11-27
Add discard draft endpoint

elliotcm:
This allows publishing applications to discard
drafts.  The main use cases for this in whitehall are:

1. A document is created in draft but never published,
   and then deleted, wherein the draft should be
   deleted entirely from the publishing API and the
   draft content store.
2. A document is created, published, and redrafted.
   The draft should be reset to be the same as the
   live content item and resent to the content store.

https://trello.com/c/d6trl2sb/431-add-discard-endpoint-to-publishing-api-for-throwing-away-non-published-drafts

### DIFF
--- a/app/errors/command_error.rb
+++ b/app/errors/command_error.rb
@@ -1,6 +1,23 @@
 class CommandError < StandardError
   attr_reader :code, :error_details
 
+  def self.with_error_handling(&block)
+    block.call
+  rescue GdsApi::HTTPServerError => e
+    should_suppress = (PublishingAPI.swallow_connection_errors && e.code == 502)
+    raise CommandError.new(code: e.code, message: e.message) unless should_suppress
+  rescue GdsApi::HTTPClientError => e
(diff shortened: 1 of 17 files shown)


app/errors/command_error.rb
```diff
@@ -1,6 +1,23 @@
 class CommandError < StandardError
   attr_reader :code, :error_details
 
+  def self.with_error_handling(&block)
+    block.call
+  rescue GdsApi::HTTPServerError => e
+    should_suppress = (PublishingAPI.swallow_draft_connection_errors && e.code == 502)
```
danielroseman: The existing code only suppresses errors in draft, not live; this does both.

Although I'm wondering how much we care. We originally had this before the draft content store was running in production. Nowadays it is only useful for preventing errors in dev, where you typically only run one content store at a time; maybe we should be a bit more explicit that this is what it is doing.

elliotcm: Can we get rid of it?

PR #156, opened 2015-11-30
Support multiple content types in request payloads

tuzz:
The old pull request is here: #142 (unable to reopen)

https://trello.com/c/nkofgTdj/372-govspeak-part-2-move-contenttyperesolver-to-a-presenter-in-the-contentstore

This is the same as the previous pull request which had been reviewed and was ready to ship, but we decided on a slightly different approach.

Instead of performing the content type resolution in the publishing api, we instead, resolve the html content type in content store's presentation layer. 

That pull request is here: https://github.com/alphagov/content-store/pull/172

**DO NOT MERGE** until the content store pull request is merged ^


elliotcm:
:+1: though the title isn't quite right anymore.

PR #157, opened 2015-11-30
Add GetLinked query

jackscotti:
Get a list of the content which has a link of the given type to the given
content-id. This is the reverse lookup from the "get-links" lookup - find the
things which link to a piece of content.

Ticket:
https://trello.com/c/zppxFP6p/345-implement-the-publisher-get-linked-endpoint-and-refactor-how-links-are-stored-in-publishing-api

### DIFF
--- a/app/controllers/v2/link_sets_controller.rb
+++ b/app/controllers/v2/link_sets_controller.rb
@@ -1,5 +1,8 @@
 module V2
   class LinkSetsController < ApplicationController
+
+    rescue_from ActionController::ParameterMissing, with: :parameter_missing_error
+
     def get_links
       render json: Queries::GetLinkSet.call(content_id)
     end
@@ -12,6 +15,14 @@ module V2
       render status: response.code, json: response
     end
 
+    def get_linked
+      render json: Queries::GetLinked.new(
+        content_id: content_id,
+        link_type: params.fetch(:link_type),
+        fields: params.fetch(:fields),
+      ).call
+    end
+
   private
     def links_params
       payload.merge(content_id: content_id)
@@ -19,6 +30,17 @@ module V2
 
     def content_id
       params.fetch(:content_id)
+    end
+
+    def parameter_missing_error(e)
+      error = CommandError.new(code: 422, error_details: {
+        error: {
+          code: 422,
+          message: e.message
+        }
+      })
+
+      respond_with_command_error(error)
     end
   end
 end
--- a/app/queries/get_linked.rb
+++ b/app/queries/get_linked.rb
@@ -1,0 +1,66 @@
+module Queries
+  class GetLinked
+    attr_reader :target_content_id, :link_type, :fields
+
+    def initialize(content_id:, link_type:, fields:)
+      @target_content_id = content_id
+      @link_type = link_type
+      @fields = fields
+    end
+
+    def call
+      validate_presence_of_item!
+      validate_fields!
+
+      content_items.map do |content_item|
+        hash = content_item.as_json(only: fields)
+        publication_state = content_item.published? ? 'live' : 'draft'
+        hash['publication_state'] = publication_state if fields.any?
+        hash
+      end
+    end
+
+  private
+
+    def validate_presence_of_item!
+      return if DraftContentItem.where(content_id: target_content_id).any?
+
+      raise CommandError.new(code: 404, error_details: {
+        error: {
+          code: 404,
+          message: "No item with content_id: '#{target_content_id}'"
+        }
+      })
+    end
+
+    def validate_fields!
+      invalid_fields = fields - permitted_fields
+      return unless invalid_fields.any?
+
+      raise CommandError.new(code: 400, error_details: {
+        error: {
+          code: 400,
+          message: "Invalid column name(s): #{invalid_fields.to_sentence}"
+        }
+      })
+    end
+
+    def content_items
+      content_ids = Link.select("link_sets.content_id")
+                      .joins(:link_set)
+                      .where(target_content_id: target_content_id, link_type: link_type)
+
+      draft_items = DraftContentItem.includes(:live_content_item).where(content_id: content_ids)
+
+      live_items_without_draft = LiveContentItem
+                                  .where(content_id: content_ids)
+                                  .where.not(content_id: draft_items.map(&:content_id))
+
+      draft_items + live_items_without_draft
+    end
+
+    def permitted_fields
+      DraftContentItem.column_names
+    end
+  end
+end
--- a/spec/controllers/v2/link_sets_controller_spec.rb
+++ b/spec/controllers/v2/link_sets_controller_spec.rb
@@ -1,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe V2::LinkSetsController do
+  let(:content_id) { SecureRandom.uuid }
+
+  before do
+    create(:draft_content_item, content_id: content_id)
+    stub_request(:any, /content-store/)
+  end
+
+  describe "get_linked" do
+    context "called without providing fields parameter" do
+      it "is unsuccessful" do
+        get :get_linked, {
+          content_id: content_id,
+          link_type: "topic",
+        }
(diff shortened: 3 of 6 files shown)


app/controllers/v2/link_sets_controller.rb
```diff
@@ -20,5 +28,9 @@ def links_params
     def content_id
       params.fetch(:content_id)
     end
+
+    def fields
```
tijmenb: Would it be clearer to inline this?

app/queries/get_linked.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def call
+      validate_presence_of_item!
+      validate_fields!
+
+      content_ids = Link.select("link_sets.content_id")
+                      .joins(:link_set)
+                      .where(target_content_id: target_content_id, link_type: link_type)
+
+      content_items = DraftContentItem.where(content_id: content_ids)
```
tijmenb: This doesn't `select` the columns yet.

app/controllers/v2/link_sets_controller.rb
```diff
@@ -12,6 +12,14 @@ def put_links
       render status: response.code, json: response
     end
 
+    def get_linked
+      render json: Queries::GetLinked.new(
+        content_id: content_id,
+        link_type: link_type,
```
tijmenb: This `link_type` isn't defined anywhere so the endpoint crashes. We should add integration tests that cover the controller.

jackscotti:
@tijmenb updated

app/queries/get_linked.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def content_items
+      content_ids = Link.select("link_sets.content_id")
+                      .joins(:link_set)
+                      .where(target_content_id: target_content_id, link_type: link_type)
+
+      draft_items = DraftContentItem.where(content_id: content_ids)
+
+      live_items_without_draft = LiveContentItem
+                                  .where(content_id: content_ids)
```
tijmenb: I think https://github.com/alphagov/publishing-api/pull/158 applies here too - could you check?

app/queries/get_linked.rb
```diff
@@ -0,0 +1,66 @@
...
+        }
+      })
+    end
+
+    def content_items
+      content_ids = Link.select("link_sets.content_id")
+                      .joins(:link_set)
+                      .where(target_content_id: target_content_id, link_type: link_type)
+
+      draft_items = DraftContentItem.where(content_id: content_ids)
```
tijmenb: I think https://github.com/alphagov/publishing-api/pull/158 applies here too - could you check?

tijmenb:
Looks good broadly. Last things:

- We need to add docs to https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md
- I'm sure sure about `/linked/:link_type/:content_id`. Maybe `/linked/:content_id/:link_type` is more in line with the other methods? Or perhaps make move `type` to a parameter (`GET /linked/:content_id?link_type=topics&fields[]=blabla`). That way it also looks more like the others.
- We have to add the endpoint to `gds-api-adapters`. You can check https://github.com/alphagov/gds-api-adapters/commit/3aa6816a3846bac6d0c9b5d8799b102766b1c1a3 for an example, it shouldn't be too far off. You'll have to add some "provider state" too in `pact_helper` ([see]( https://github.com/alphagov/publishing-api/blob/8c43a6ef8bd660ecbb7e9c796ca3b6fcfaa8eee5/spec/service_consumers/pact_helper.rb#L252-L257)).

app/queries/get_linked.rb
jackscotti: I ran `get_linked_spec` checking the logs with and without preloading the `LiveContentItems`. The queries to the database are exactly the same.

jackscotti:
`link_type` is compulsory, I have the feeling it should be present in the query url, maybe as:

`GET /linked/:content_id/link_type/:link_type?fields[]=balblabla`

What do you think?

jackscotti:
I've updated the query as you suggested. 

I will now proceed in this way:
- get this PR merged
- update gds-api-adapters
- add the documentation to the endpoint

I need to proceed in this order because when writing the doc I will need to link to the "Request/Response detail" that will be added when updating gds-api-adapters.

app/queries/get_linked.rb
tijmenb: It still does, unfortunately:

https://github.com/alphagov/publishing-api/blob/add-get-linked-endpoint/app/queries/get_linked.rb#L17

You are requesting `content_item.published?` for each draft content item you're requesting here, which will load a LiveContentItem each time. An `includes(:live_content_item)` will fix that.

spec/controllers/v2/link_sets_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe V2::LinkSetsController do
+  let(:content_id) { SecureRandom.uuid }
+  let(:link_type) { "topic" }
```
tijmenb: These `let`'s don't add much, because they're only used once per test. Maybe clearer to inline them?

jackscotti: I removed the `link_type` let. The other one is actually used.

jackscotti:
Comments addressed. Thanks for the feedback and for showing me how to find out a N+1 query!

app/controllers/v2/link_sets_controller.rb
```diff
@@ -12,6 +12,14 @@ def put_links
       render status: response.code, json: response
     end
 
+    def get_linked
+      render json: Queries::GetLinked.new(
+        content_id: content_id,
+        link_type: params.fetch(:link_type),
+        fields: params.fetch(:fields),
+      ).call
+    end
```
tuzz: If you don't provide a `link_type` or `fields` parameter, would you get a 500? Can we make this a 4xx response with a descriptive message?

app/queries/get_linked.rb
```diff
@@ -0,0 +1,66 @@
...
+                      .where(target_content_id: target_content_id, link_type: link_type)
+
+      draft_items = DraftContentItem.includes(:live_content_item).where(content_id: content_ids)
+
+      live_items_without_draft = LiveContentItem
+                                  .where(content_id: content_ids)
+                                  .where.not(content_id: draft_items.map(&:content_id))
+
+      draft_items + live_items_without_draft
+    end
```
tuzz: I wonder if some of this could be pushed down to methods/scopes in the models. We might want to look up all link sets that have links that reference some content_id, for example, in other places in the app.

If you did this, there'd be an opportunity to give some descriptive names to these operations, too – rather than having to make sense of these queries to decipher what their "meaning" is.

It's totally up to you and we can refactor these into the models if/when we need to reuse them. Just a thought.

spec/queries/get_linked_spec.rb
```diff
@@ -0,0 +1,99 @@
+require "rails_helper"
+
+RSpec.describe Queries::GetLinked do
+  content_id = SecureRandom.uuid
+  target_content_id = SecureRandom.uuid
```
tuzz: Please could `#let` these. It helps to mitigate the risk of tests interfering with each other.

tuzz:
Generally looks good :+1: – a few minor points that are probably more suggestions rather than directives.

app/controllers/v2/link_sets_controller.rb
jackscotti: you are right, going to do it. Can you advise where I should return the 4xx response? Here in the controller or in the query itself?

I'd say here.

jackscotti:
@tuzz, updated the controller to return a meaningful message when `link_type` or `fields` parameters are not provided. I've also updated the `let`s.

I would like to get this merged first and do the refactoring subsequently in a separate PR. I'd also like to refactor the app to rescue the missing parameters in the application controller rather than only in the linkset controller.

PR #158, opened 2015-12-02
Fix N+1 query in `GetContentCollection`

tijmenb:
To discover the `publication_state` of items returned by this query, `content_item.published?` is called. 

This in turn kicks off another query for the `live_content_item`. This means an extra query is performed for each item in the collection. We can avoid that by preloading the `live_content_item`.

cc @tuzz


tuzz:
:+1: good spot

PR #159, opened 2015-12-02
Protect alpha_taxons from being overwritten

tijmenb:
The "alpha_taxonomy" is a prototype for the new taxonomy on GOV.UK. It is being managed exclusively via the V2 API. While in the migration, we need to prevent these tags from being overwritten by applications using the V1 endpoint.

It should be noted that this is a temporary "hack", because this code will be removed with the `/v1` endpoint once all apps have been upgraded.

Trello: https://trello.com/c/bsZAwZ35

### DIFF
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -82,7 +82,11 @@ module Commands
         version.increment
         version.save! if link_set.valid?
 
-        link_set.links.delete_all
+        # The "alpha_taxonomy" is a prototype for the new taxonomy on GOV.UK. It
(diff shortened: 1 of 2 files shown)


bradwright:
Is this a thing that's actually happened? As far as I can tell, no apps are using this key that your team doesn't own.

tijmenb:
Maybe I wasn't clear enough. 

The current behaviour in V1 is that `links` of a content item are completely replaced (`link_set.links.delete_all`, then re-create any links provided in the payload). This also happens when the content items doesn't have `links`.

bradwright:
I see. In that case :+1: for intent.

PR #160, opened 2015-12-07
Provide a mechanism to export/import events

tuzz:
We will need this for the data cleanup task to replay events that come into the publishing api while the other rake tasks are running.

### DIFF
--- a/lib/tasks/events.rake
+++ b/lib/tasks/events.rake
@@ -1,0 +1,44 @@
+namespace :events do
+  desc "Exports events from a given timestamp to tmp/events.json"
+  task export: :environment do
+    timestamp = ENV.fetch("TIMESTAMP")
+    datetime = DateTime.parse(timestamp)
+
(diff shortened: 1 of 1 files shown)


lib/tasks/events.rake
```diff
@@ -0,0 +1,43 @@
...
+    hash_events = json_events.map { |json| JSON.parse(json) }
+
+    hash_events.each do |hash|
+      action = hash.fetch("action")
+      payload = hash.fetch("payload")
+      command = "Commands::#{action}".constantize
+      response = nil
+
+      begin
+        response = command.call(payload)
```
danielroseman: This doesn't recreate the event log entry itself. Think we need to wrap it in `with_event_logging`.

tuzz:
@danielroseman I've wrapped it with `EventLogger.log_command`. The `with_event_logging` method is just a convenience method for this.

PR #161, opened 2015-12-07
Populate new links table

tijmenb:
https://github.com/alphagov/publishing-api/pull/137 added a new data database structure for the links. However, the data has not been migrated. This PR fixes that omission.

We don't copy the links for content items that already have `links`. These must have been added after #137 was deployed, and are therefore up-to-date.

Thanks @danielroseman for noticing this.

### DIFF
--- a/db/migrate/20151207154727_populate_links.rb
+++ b/db/migrate/20151207154727_populate_links.rb
@@ -1,0 +1,15 @@
+class PopulateLinks < ActiveRecord::Migration
+  def up
+    link_sets_with_populated_links = LinkSet.includes(:links).where("legacy_links::text <> '{}'::text")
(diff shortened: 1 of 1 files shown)


db/migrate/20151207154727_populate_links.rb
```diff
@@ -0,0 +1,15 @@
+class PopulateLinks < ActiveRecord::Migration
+  def up
+    link_sets_with_populated_links = LinkSet.includes(:links).where("legacy_links::text <> '{}'::text")
+    link_sets_with_populated_links.each do |link_set|
```
danielroseman: Should this use `find_each` for efficiency?

tijmenb:
@danielroseman updated PR.

PR #164, opened 2015-12-08
Write up the proposed data cleanup plan

tuzz:
I thought it would be sensible to write up the proposed data cleanup plan as it's quite complicated and involves a lot of steps and different tasks. Please comment on this pull request if anything's not clear.


tuzz:
Going to merge ourselves. Steve wants to make reference to this in another writeup.

PR #168, opened 2015-12-09
Filter content collection by publishing app 

danielroseman:


danielroseman:
@tijmenb We're hoping to avoid adding too many filtering options here. Our long-term plan is to move the get_content_items endpoint to elasticsearch, at which point we'll implement a much richer set of filters.

Fair point on the "all" value: I've removed that, omitting the value is the way to get all data.

tijmenb:
@danielroseman okay, great. What's the immediate use case for this by the way? Would `content_format` ever become optional?

danielroseman:
I don't think so. The idea is that gds-api-adapters will automatically pass the publishing_app parameter (based on the app making the request) so that apps only see content they own.

tijmenb:
Ok, sounds good! Wouldn't do it automatically though, I think content-tagger, specialist-publisher & whitehall tagging prototype now depend on it returning the content from all apps, and policy-publisher will do soon.

PR #170, opened 2015-12-11
Fix replaying events import...

steventux:
Ensure that events import will not attempt to call downstream services
by stubbing these. Also logs newly created events to file.

This is necessary for the successful import of publishing api data from content store.


danielroseman:
The put_(draft)_content_with_links commands have a `downstream` bool parameter which bypasses the downstream requests. It would be nicer to generalise this to the other commands rather than stub them out.

steventux:
@danielroseman updated as advised.

PR #171, opened 2015-12-11
Allow content items to move

danielroseman:
Currently items can't be moved to a new path once they have been published. This change allows them to move, automatically creating a redirect from the old path to the new one.

Updating the path of an item that has never been published triggers a delete of the old item from draft content-store.

https://trello.com/c/CoS7IYEh


steventux:
This looks fine to me, will let @tuzz review in case I've missed something...

tuzz:
Looks good (needs rebase).

danielroseman:
@tuzz rebased

app/commands/v2/publish.rb
```diff
@@ -101,6 +102,18 @@ def build_live_attributes(draft_content_item)
...
       end
+
+      def handle_path_change(live_content_item)
+        path_change = live_content_item.previous_changes[:base_path]
+        if path_change.present? && path_change[0].present?
+          draft_redirect = DraftContentItem.find_by(format: "redirect", base_path: path_change[0])
+          self.class.call(
+            content_id: draft_redirect.content_id,
+            locale: draft_redirect.locale,
+            update_type: "major"
```
jamiecobbett: Why does this happen as a "major" change? Shouldn't it use whatever `update_type` was used by the client?

app/commands/v2/put_content.rb
```diff
@@ -45,6 +47,26 @@ def create_or_update_draft_content_item!
       def content_item_attributes
         payload.slice(*DraftContentItem::TOP_LEVEL_FIELDS)
       end
+
+      def handle_path_change(content_item)
+        path_change = content_item.previous_changes[:base_path]
```
jamiecobbett: I'd have been tempted to use destructuring assignment:
`old_path, new_path = content_item.previous_changes[:base_path]`

app/commands/v2/publish.rb
tuzz: This is the `update_type` of the newly created `redirect` content item. It's different from the `update_type` of the content item that has moved. Given this is a brand new content item, it seems reasonable to use an `update_type` of `major`, although technically, it's not an update at all.

jamiecobbett: Aha, thanks!

jamiecobbett:
This doesn't take care of updating any other routes that the item has - for example lots of Whitehall content has `.atom` feeds. They might not be being sent in the placeholders yet, but they will be.

bradwright:
We don't have a nice way of linking URIs together yet (eg `.json`, `.atom`)

PR #172, opened 2015-12-14
Update get linked endpoint

jackscotti:
I missed a couple of points when I've added this endpoint:

- Content items do not have always a draft version
- The query should not allow requests with an empty `fields` parameter

This PR is updating the endpoint to take care of them.

### DIFF
--- a/app/queries/get_linked.rb
+++ b/app/queries/get_linked.rb
@@ -35,12 +35,20 @@ module Queries
 
     def validate_fields!
       invalid_fields = fields - permitted_fields
-      return unless invalid_fields.any?
+      return if invalid_fields.empty? && fields.any?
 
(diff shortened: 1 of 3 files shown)


app/queries/get_linked.rb
```diff
@@ -23,7 +23,10 @@ def call
   private
 
     def validate_presence_of_item!
-      return if DraftContentItem.where(content_id: target_content_id).any?
+      return if (
```
tijmenb: I think this is faster:

```ruby
DraftContentItem.exists?(content_id: target_content_id) || LiveContentItem.exists?(content_id: target_content_id)
```

http://apidock.com/rails/ActiveRecord/Base/exists%3F/class

jackscotti:
thanks @tijmenb, updated.

PR #174, opened 2015-12-14
Add provider state for get-linked query

jackscotti:
This provider state is required in order to make pass the pact tests in gds-api-adapters: https://github.com/alphagov/gds-api-adapters/pull/408

Part of: https://trello.com/c/zppxFP6p/345-implement-the-publisher-get-linked-endpoint-and-refactor-how-links-are-stored-in-publishing-api

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -209,4 +209,22 @@ Pact.provider_states_for "GDS API Adapters" do
       FactoryGirl.create(:draft_content_item, title: 'Content Item C', base_path: '/yet-another-base-path', format: 'topic', publishing_app: 'whitehall')
     end
   end
+
+  provider_state "there are two documents with a 'topic' link to another document" do
+    set_up do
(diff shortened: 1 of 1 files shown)


spec/service_consumers/pact_helper.rb
```diff
@@ -209,4 +209,22 @@
       FactoryGirl.create(:draft_content_item, title: 'Content Item C', base_path: '/yet-another-base-path', format: 'topic', publishing_app: 'whitehall')
     end
   end
+
+  provider_state "there are two documents that link to the wantend document" do
```
tuzz: What's `wantend`?

jackscotti: You're right. Let me rewrite this.

jackscotti: Would _"there are two documents with a 'topic' link to another document"_ be descriptive enough?

tuzz: Makes more sense to me. I think specifics are also worth including, but maybe quote them to show that they're data points.

PR #178, opened 2015-12-15
Expose redrafted status and live_version

danielroseman:
If an item has been published then subsequently edited - ie it has a live
version number less than the draft version number - then publication_state is
reported as "redrafted".

Also expose live_version attribute if the item has been published.

### DIFF
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -14,8 +14,8 @@ module Queries
       content_items.map do |content_item|
         hash = Presenters::Queries::ContentItemPresenter.new(
           content_item,
-          draft_versions[content_item.id],
-          live_versions[content_item.live_content_item.try(:id)]
+          draft_version(content_item),
+          live_version(content_item)
         )
         hash.present.as_json(only: output_fields)
       end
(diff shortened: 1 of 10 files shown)


tijmenb:
Note that the states for this `publication_state` are now different from https://github.com/alphagov/publishing-api/blob/cdc28b4580ee6ce1407a58afeb408b1580a0e27c/app/queries/get_content_collection.rb#L16-L17 and https://github.com/alphagov/publishing-api/blob/bd0a6f3b13c26e1129e9842bb5552e1d6a7985d4/app/queries/get_linked.rb#L17-L18.

danielroseman:
Thanks @tijmenb - added. This needed some fairly complex work to avoid the n+1 queries issue.

app/queries/get_content_collection.rb
```diff
@@ -36,6 +40,10 @@ def content_items
 
       live_items = live_items.where(publishing_app: @publishing_app) if @publishing_app.present?
 
+      @draft_versions = Version.in_bulk(draft_items, DraftContentItem)
```
tijmenb: @danielroseman I am not 100% sure, but would it be possible to add `belongs_to :version` on `LiveContentItem`/`DraftContentItem`, so that we can use `includes(:version)` and `item.version` instead of `Version.in_bulk`?

danielroseman: That would indeed make things quite a lot simpler. See the comment to 3c332f3 though - we're being quite careful to keep Versions and ContentItems separate, at least for now.

PR #181, opened 2015-12-31
Add rake task for sanitizing access limited drafts.

elliotcm:
Changes the draft discard code so access limiting is also reset.  @danielroseman I remember us having a conversation about this but not the conclusion.  Did we say we needed to keep these access limits around after a reset?


danielroseman:
No there wouldn't be any point preserving access_limited status for items copied from live.

PR #182, opened 2016-01-04
Extracts access limiting into its own model.

elliotcm:
This is done in a similar manner to versioning.

Some points:
- The `access_limited` field remains and needs to
  be removed in a later ticket.
- Some references to the field exist in the tests
  and will also need removing when the column is
  gone.
- The `viewable?` method which I ported over from
  the draft content item model is not used anywhere.
  Neither was the `access_limited?` method which
  was tested but unused.
  Do we know why?

https://trello.com/c/m0TgNj9i/379-refactor-extract-accesslimitation-from-contentitems

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -11,6 +11,8 @@ module Commands
           delete_draft
         end
 
+        delete_access_limit
+
         Success.new(content_id: content_id)
       end
 
@@ -29,12 +31,7 @@ module Commands
       end
 
       def update_draft_from_live
-        live_attributes = live.attributes.except("id", "draft_content_item_id")
-        safe_attributes = live_attributes.merge(
-          "access_limited" => {}
-        )
-
-        draft.update_attributes(safe_attributes)
+        draft.update_attributes(live.attributes.except("id", "draft_content_item_id"))
 
         if downstream
           ContentStoreWorker.perform_async(
@@ -55,6 +52,10 @@ module Commands
             delete: true,
           )
         end
+      end
+
+      def delete_access_limit
+        AccessLimit.find_by(target: draft).try(:destroy)
       end
 
       def draft
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -5,13 +5,13 @@ module Commands
         validate_version_lock!
         raise_error_if_missing_draft!
 
+        delete_access_limit
+
         if live
           update_draft_from_live
         else
           delete_draft
         end
-
-        delete_access_limit
 
         Success.new(content_id: content_id)
       end
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -93,7 +93,6 @@ module Commands
       def build_live_attributes(draft_content_item)
         attributes = draft_content_item
           .attributes
-          .except("version")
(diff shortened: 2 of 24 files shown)


app/commands/v2/discard_draft.rb
```diff
@@ -11,6 +11,8 @@ def call
           delete_draft
         end
 
+        delete_access_limit
```
tuzz: Could potentially be moved before the update/delete of the draft. Seems more intuitive that way.

app/commands/v2/publish.rb
```diff
@@ -93,7 +93,7 @@ def publish_content_item(draft_content_item)
       def build_live_attributes(draft_content_item)
         attributes = draft_content_item
           .attributes
-          .except("access_limited", "version")
+          .except("version")
```
tuzz: Given we still have the access limited column on draft content item, should we still strip this field out?

app/models/replaceable.rb
```diff
@@ -14,7 +14,7 @@ def create_or_replace(payload, &block)
       end
 
       retry_strategy.call do
-        item.assign_attributes(payload.except(:id, :version))
+        item.assign_attributes(payload.except(:id, :version, :access_limited))
```
tuzz: We should get rid of version here as that column no longer exists on content item models.

db/migrate/20151231125222_extract_access_limiting.rb
```diff
@@ -0,0 +1,21 @@
+class ExtractAccessLimiting < ActiveRecord::Migration
+  def up
+    create_table :access_limits do |t|
+      t.integer :target_id, null: false
+      t.string :target_type, null: false
+      t.json :users, null: false, default: []
+      t.timestamps null: false
+    end
```
tuzz: We could probably do with a unique index across a composite of the target columns.

db/migrate/20151231125222_extract_access_limiting.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    DraftContentItem.where("access_limited::text <> '{}'::text").each do |limited_draft|
+      AccessLimit.create(
```
tuzz: #create! might be a good idea here incase we add validations later.

Printing as well?

tuzz:
Looks good – it's great to see the Draft and Live content item models fall closer inline. Looking forward to these models being merged together.

A few minor comments (some of which were @elliotcm's as we reviewed it together)

tuzz:
One thing that we discussed whilst reviewing this ticket was whether the access limited record should be destroyed when the draft content item is published. This would affect the user interface/behaviour for publishing apps.

This pull request keeps the behaviour the same as it was previously and is a refactor only, so we deferred this discussion/decision to later.

jamiecobbett:
> The access_limited field remains and needs to be removed in a later ticket.

Could you use the pattern we've got in Whitehall and Signon to deprecate the column? https://rubygems.org/gems/deprecated_columns

tuzz:
I've pushed some changes based on my prior comments.

I've also verified that the pact still passes between this branch and the content store master branch. :+1:

tuzz:
I have addressed all comments including @jamiecobbett's (see 1f843b87b78d3186148b2104c2ae9f6611ce2ace)

Ready for another round of reviewing.

Thanks

elliotcm:
:+1:

PR #183, opened 2016-01-06
Allow coming soon items to be overwritten

jamiecobbett:
Coming Soon pages are displayed at base_paths that will soon have the actual
content. Whitehall sends the coming soon with a different content-id to the
item that will be published, so until now, publishing-api has been rejecting
the request that tries to publish the actual document.

An alternative might be to change Whitehall to use the same content-id, but
this would mean we'd have to reconcile the existing data in a fairly manual
way.

Once deployed, I think it would be appropriate to republish all documents in
Whitehall that have been rejected by this issue.

/cc @elliotcm

### DIFF
--- a/spec/factories/live_content_item.rb
+++ b/spec/factories/live_content_item.rb
@@ -63,4 +63,10 @@ FactoryGirl.define do
     sequence(:base_path) {|n| "/dodo-sanctuary-#{n}" }
     format "gone"
   end
+
+  factory :coming_soon_live_content_item, parent: :live_content_item do
+    format "coming_soon"
+    title "Coming soon"
+    description "This item will be published soon"
+  end
 end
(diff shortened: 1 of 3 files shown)


elliotcm:
@jamiecobbett Looks like we could do with a unit test for https://github.com/alphagov/publishing-api/blob/master/app/substitution_helper.rb (which was a refactor from the original code).

jamiecobbett:
@elliotcm I've added a test.

spec/factories/live_content_item.rb
```diff
@@ -63,4 +63,10 @@
     sequence(:base_path) {|n| "/dodo-sanctuary-#{n}" }
     format "gone"
   end
+
+  factory :coming_soon_content_item, parent: :live_content_item do
```
elliotcm: We've been naming these derived factories `x_live_content_item` as we frequently have draft equivalents.

jamiecobbett: Good point - rebased.

PR #184, opened 2016-01-11
Fix a bug with nil descriptions

tuzz:
@jamiecobbett noticed a problem with `nil` descriptions appearing as a JSON object in the content store.

e.g. https://www.gov.uk/api/content/government/organisations/hm-revenue-customs

The above link has a description of: `{ value: nil }` rather than just `nil`

This fixes that.

I've also extracted the description overrides into a concern to keep it all in one place and to DRY up the models.

Note: This fixes the problem in Publishing API. For changes to be reflected in ContentStore, we'd need to republish these content items (~800).


jamiecobbett:
:100:

PR #188, opened 2016-01-18
Bearer token authentication

danielroseman:
Implement authentication via gds-sso. Adds a User model and enforces signon permission for all actions.

By default the get_content_collection is now filtered to return items whose publishing_app matches the authenticated app's app_name; applications with the additional "view_all" permission will see unfiltered responses.

All apps have been updated to pass the token on requests, but this PR should not be merged until they have all been deployed.


steventux:
:+1: This looks ready to merge when other PR dependencies are met.

spec/requests/content_item_requests/derived_representations_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 RSpec.describe "Derived representations", type: :request do
+  before do
+    login_as_stub_user
```
elliotcm: This is needed on all request specs, so it can be configured in the spec helper:
```
  config.before :each, type: :request do
    login_as_stub_user
  end
```

elliotcm: If it's needed in other types too, maybe just do it always.

app/controllers/application_controller.rb
```diff
@@ -11,6 +13,12 @@ class BadRequest < StandardError; end
     request.env['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest'
   end
 
+  before_action :require_signin_permission!
+
+  Warden::Manager.after_authentication do |user,auth,opts|
```
elliotcm: We only use `user` here, so can do `|user, _, _|` for clarity.

spec/requests/content_item_requests/derived_representations_spec.rb
danielroseman: Changed to run on request and controller specs.

app/controllers/application_controller.rb
danielroseman: Done.

elliotcm:
:+1: Good to self-merge when the timing is right.

PR #189, opened 2016-01-18
Content item index endpoint pagination

Rosa-Fox:
https://trello.com/c/OZF9rKeX/487-linked-items-endpoint-should-support-pagination

Added pagination so that content items can be retrieved in batches by publishing apps.

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -7,7 +7,7 @@ module V2
       render json: Queries::GetContentCollection.new(content_format: content_format,
                                                     fields: fields,
                                                     publishing_app: publishing_app,
-                                                    pagination: pagination_params).call
+                                                    pagination: Pagination.new(params)).call
     end
 
     def show
@@ -41,13 +41,6 @@ module V2
   private
     def content_item
       payload.merge(content_id: params[:content_id])
-    end
-
-    def pagination_params
-      {
-        start: params.fetch(:start, 0).to_i,
-        count: params.fetch(:count, 50).to_i,
-      }
     end
   end
 end
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -4,7 +4,10 @@ module V2
       content_format = params.fetch(:content_format)
       fields = params.fetch(:fields)
       publishing_app = params[:publishing_app]  # can be blank
-      render json: Queries::GetContentCollection.new(content_format: content_format, fields: fields, publishing_app: publishing_app).call
+      render json: Queries::GetContentCollection.new(content_format: content_format,
+                                                    fields: fields,
+                                                    publishing_app: publishing_app,
+                                                    pagination: pagination_params).call
     end
 
     def show
@@ -38,6 +41,13 @@ module V2
   private
     def content_item
       payload.merge(content_id: params[:content_id])
+    end
+
+    def pagination_params
+      {
+        start: params.fetch(:start, 0).to_i,
+        count: params.fetch(:count, 50).to_i,
+      }
     end
   end
 end
--- a/app/pagination.rb
+++ b/app/pagination.rb
@@ -1,0 +1,8 @@
+class Pagination
+  attr_reader :start, :count
+
+  def initialize(options = {})
+    @start = options[:start] ? options[:start].to_i : 0
+    @count = options[:count] ? options[:count].to_i : 50
+  end
+end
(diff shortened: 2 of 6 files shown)


app/controllers/v2/content_items_controller.rb
```diff
@@ -4,7 +4,10 @@ def index
       content_format = params.fetch(:content_format)
       fields = params.fetch(:fields)
       publishing_app = params[:publishing_app]  # can be blank
-      render json: Queries::GetContentCollection.new(content_format: content_format, fields: fields, publishing_app: publishing_app).call
+      render json: Queries::GetContentCollection.new(content_format: content_format,
```
danielroseman: style niggle: move `content_format:` onto a new line indented by only four spaces, line up the following lines to match.

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -5,10 +5,37 @@
...
+    context "with pagination params" do
+      before do
+        get :index, content_format: 'topic', fields: ['content_id'], start: "0", count: "20"
+      end
+      it "is successful" do
+        expect(response.status).to eq(200)
+      end
+      it "responds with the content item as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.first.fetch("content_id")).to eq("#{content_id}")
```
danielroseman: No need to use string substitution here (or below), `content_id` is already a string.

spec/queries/get_content_collection_spec.rb
```diff
@@ -90,4 +90,88 @@
...
+            start: 0,
+            count: 5,
+          })
+        ).call
+
+        expect(content_items.first['base_path']).to eq('/a')
+        expect(content_items.last['base_path']).to eq('/live1')
+      end
+    end
+    context "with only live items" do
```
danielroseman: need blank line before

spec/queries/get_content_collection_spec.rb
```diff
@@ -90,4 +90,88 @@
...
+      before do
+        create(:live_content_item, :with_version, base_path: '/live1',  format: 'topic')
+        create(:live_content_item, :with_version, base_path: '/live2',  format: 'topic')
+      end
+      it "returns expected number of live items" do
```
danielroseman: need another blank line before here

danielroseman:
A few mostly stylistic issues, otherwise :+1:

danielroseman:
It occurs to me that we need to support an "all items" parameter; policy-publisher is now using this endpoint to get all organisations, persons and working groups, which shouldn't be paginated.

rboulton:
This looks good.

I'm wondering what order the items are returned in?  It looks from a cursory reading of the code like they might be in database order, which isn't guaranteed to be the same between calls to the database.  There may well be a default order that I'm missing, though.  Also, do the callers to this api that need pagination need any particular ordering (eg, publication date?)

Regarding @danielroseman's comment about all items; yes, policy-publisher needs to get all organisations, etc, for the moment, in order to populate a select box.  content-tagger will have a similar requirement.  Having a way to express this in the api would be ideal.  One thing to think about, based on our experience with the search API, is that you might want to make the "count" parameter a required parameter, rather than having a default value (or make the default value be to return everything).  We've had a few bugs caused by assuming that the search API returns everything matching, which is much harder to do by mistake if you have to specify an explicit "count" parameter.

app/pagination.rb
```diff
@@ -0,0 +1,8 @@
+class Pagination
+  attr_reader :start, :count
+
+  def initialize(options = {})
+    @start = options[:start] ? options[:start].to_i : 0
+    @count = options[:count] ? options[:count].to_i : 50
+  end
+end
```
tuzz: Currently, this Pagination class holds the state for pagination. I wonder if it could also hold it's behaviour?

e.g.

```ruby
paginator = Paginator.new(page_size: 50)

items = paginator.paginate(DraftContentItem.all, page: 3)
#=> ActiveRecord::Relation

items.count
#=> 50
```

What do you think?

tuzz:
Looks good. One small comment with an idea for how to make Pagination a bit more reusable.

:+1:

app/pagination.rb
```diff
@@ -0,0 +1,8 @@
+class Pagination
+  attr_reader :start, :count
+
+  def initialize(options = {})
+    @start = options[:start] ? options[:start].to_i : 0
+    @count = options[:count] ? options[:count].to_i : 50
```
jamiecobbett: I think `page_size` would be more explicit name for this parameter.

jamiecobbett:
From a quick scan of the code I couldn't see anything which is ensuring that there is a sort order on the pagination. This means that you can't guarantee that you're actually paginating over the set correctly. In other words, I think a sort order should be required, but it could have a default value.

PR #192, opened 2016-01-21
Scope items by locale

mgrassotti:
Add a new `locale` parameter to the contents index path `/v2/content`, in order to:
* filter by the provided `locale` value
* filter by `locale == 'en'` if no `locale` parameter is provided
* do not filter if `locale == 'all'` parameter is provided (this was the previous behaviour)

https://trello.com/c/xAitVd24/505-scope-content-items-index-results-by-locale

### DIFF
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -1,11 +1,12 @@
 module Queries
   class GetContentCollection
-    attr_reader :content_format, :fields
+    attr_reader :content_format, :fields, :locale
 
-    def initialize(content_format:, fields:, publishing_app: nil)
+    def initialize(content_format:, fields:, publishing_app: nil, locale: nil)
       @content_format = content_format
       @fields = fields
       @publishing_app = publishing_app
+      @locale = locale || "en"
     end
 
     def call
@@ -32,6 +33,7 @@ module Queries
         .select(*fields + %i[id content_id])
 
       draft_items = draft_items.where(publishing_app: @publishing_app) if @publishing_app.present?
+      draft_items = draft_items.where(locale: locale) unless @locale == 'all'
 
       live_items = LiveContentItem
         .where("draft_content_item_id IS NULL")
@@ -39,6 +41,7 @@ module Queries
         .select(*fields + %i[id])
 
       live_items = live_items.where(publishing_app: @publishing_app) if @publishing_app.present?
+      live_items = live_items.where(locale: locale) unless @locale == 'all'
 
       @draft_versions = Version.in_bulk(draft_items, DraftContentItem)
       @live_versions = Version.in_bulk(
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -33,7 +33,7 @@ module Queries
         .select(*fields + %i[id content_id])
 
       draft_items = draft_items.where(publishing_app: @publishing_app) if @publishing_app.present?
-      draft_items = draft_items.where(locale: locale) unless @locale == 'all'
+      draft_items = draft_items.where(locale: locale) unless locale == 'all'
 
       live_items = LiveContentItem
         .where("draft_content_item_id IS NULL")
@@ -41,7 +41,7 @@ module Queries
         .select(*fields + %i[id])
 
       live_items = live_items.where(publishing_app: @publishing_app) if @publishing_app.present?
-      live_items = live_items.where(locale: locale) unless @locale == 'all'
+      live_items = live_items.where(locale: locale) unless locale == 'all'
 
       @draft_versions = Version.in_bulk(draft_items, DraftContentItem)
       @live_versions = Version.in_bulk(
--- a/spec/controllers/v2/content_items_controller_spec.rb
+++ b/spec/controllers/v2/content_items_controller_spec.rb
@@ -5,17 +5,16 @@ RSpec.describe V2::ContentItemsController do
 
   before do
     stub_request(:any, /content-store/)
-    @draft = FactoryGirl.create(:draft_content_item, content_id: content_id)
+    @draft = FactoryGirl.create(:draft_content_item,
+        content_id: content_id, locale: "en",
+        base_path: "/content.en",
+        format: "topic")
     FactoryGirl.create(:version, target: @draft, number: 2)
   end
 
   describe "index" do
     before do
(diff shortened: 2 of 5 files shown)


spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -9,6 +9,93 @@
     FactoryGirl.create(:version, target: @draft, number: 2)
   end
 
+  describe "index" do
+    let(:en_draft_content_id) { SecureRandom.uuid }
```
danielroseman: This test is fine as far as it goes, but it doesn't really reflect the expected data. We will have items in multiple locales with the *same* content_id, and we need to test that particular case.

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -9,6 +9,93 @@
...
+      end
+
+      it "is successful" do
+        expect(response.status).to eq(200)
+      end
+
+      it "responds with the english content item as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.length == 2)
+        expect(parsed_response_body.first.fetch("content_id")).to eq("#{en_draft_content_id}")
```
danielroseman: No need for string substitution here, the content_ids are already strings.

app/queries/get_content_collection.rb
```diff
@@ -1,11 +1,12 @@
 module Queries
   class GetContentCollection
-    attr_reader :content_format, :fields
+    attr_reader :content_format, :fields, :locale
 
-    def initialize(content_format:, fields:, publishing_app: nil)
+    def initialize(content_format:, fields:, publishing_app: nil, locale: nil)
```
danielroseman: Why not `locale: "en"`, then you don't need the `||` below.

app/queries/get_content_collection.rb
```diff
@@ -32,6 +33,7 @@ def content_items
         .select(*fields + %i[id content_id])
 
       draft_items = draft_items.where(publishing_app: @publishing_app) if @publishing_app.present?
+      draft_items = draft_items.where(locale: locale) unless @locale == 'all'
```
danielroseman: You've defined an attr_reader for locale, so don't need the `@`

app/queries/get_content_collection.rb
tuzz: Technically those aren't equivalent, e.g.

```ruby
def foo(bar: "en")
  puts bar.inspect
end

foo(bar: nil)
# nil
```

Named parameter defaults are ignored if you pass in `nil` explicitly, which is the case here because that's coming from `params[:locale]` in the controller.

spec/controllers/v2/content_items_controller_spec.rb
tuzz: :+1:

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -9,6 +9,93 @@
...
+
+      it "is successful" do
+        expect(response.status).to eq(200)
+      end
+
+      it "responds with the specific locale content item as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.length == 2)
+        expect(parsed_response_body.first.fetch("content_id")).to eq("#{ar_draft_content_id}")
+        expect(parsed_response_body.second.fetch("content_id")).to eq("#{ar_live_content_id}")
```
tuzz: same here

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -9,6 +9,93 @@
...
+      end
+
+      it "responds with all the localised content items as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.length == 4)
+        expect(parsed_response_body[0].fetch("content_id")).to eq("#{en_draft_content_id}")
+        expect(parsed_response_body[1].fetch("content_id")).to eq("#{ar_draft_content_id}")
+        expect(parsed_response_body[2].fetch("content_id")).to eq("#{ar_live_content_id}")
+        expect(parsed_response_body[3].fetch("content_id")).to eq("#{en_live_content_id}")
+      end
```
tuzz: I wonder if the following would be clearer:

```ruby
content_ids = parsed_response_body.map { |item| item.fetch("content_id") }
expect(content_ids). to eq [en_draft_content_id, ar_draft_content_id, ar_live_content_id, en_live_content_id]
```

If the test failed, you'd see the total ordering for the list, which might be more useful.

spec/queries/get_content_collection_spec.rb
```diff
@@ -90,4 +90,48 @@
...
+        fields: ['base_path'],
+        locale: 'all',
+      ).call).to eq([
+        { "base_path" => "/content.en", "publication_state" => "draft" },
+        { "base_path" => "/content.ar", "publication_state" => "draft" },
+        { "base_path" => "/content.ar", "publication_state" => "live" },
+        { "base_path" => "/content.en", "publication_state" => "live" },
+      ])
+    end
+  end
```
tuzz: :+1:

tuzz:
Looks good. Just needs some pact tests and a few minor tweaks and this can be merged.

spec/controllers/v2/content_items_controller_spec.rb
mgrassotti: :+1:

spec/controllers/v2/content_items_controller_spec.rb
mgrassotti: :+1:

spec/controllers/v2/content_items_controller_spec.rb
mgrassotti: :+1:

app/queries/get_content_collection.rb
mgrassotti: Yes @tuzz, that's exactly why I used a "double" default value. 
I didn't make it mandatory in order to keep the former calls of the `initialize` method valid.

app/queries/get_content_collection.rb
mgrassotti: :+1:

tuzz:
:+1: from me – I worked on this a bit, so I'll refrain from merging myself.

PR #194, opened 2016-01-25
Allow empty links to be sent

jamiecobbett:
We were sending an empty links hash for pieces of content which have no links
of any kind. This should make it so that those requests don't get rejected.

Additionally, it seems that if the list of links is empty, then all the
existing links should be removed. How else would you remove them?

This is breaking Whitehall in Integration right now following some work to change how it sends data to publishing-api.

### DIFF
--- a/app/commands/v2/put_link_set.rb
+++ b/app/commands/v2/put_link_set.rb
@@ -11,6 +11,8 @@ module Commands
 
           links_hash = link_params.fetch(:links)
 
+          link_set.links.delete_all if links_hash.empty?
+
           links_hash.each do |link_type, target_content_ids|
             if target_content_ids.empty?
               link_set.links.where(link_type: link_type).delete_all
@@ -74,7 +76,7 @@ module Commands
               }
             }
           }
-        ) unless link_params[:links].present?
+        ) unless link_params.has_key?(:links)
       end
 
       def validate_version_lock!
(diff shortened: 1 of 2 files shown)


app/commands/v2/put_link_set.rb
```diff
@@ -11,6 +11,8 @@ def call
 
           links_hash = link_params.fetch(:links)
 
+          link_set.links.delete_all if links_hash.empty?
```
danielroseman: I don't think this is the right behaviour. The links endpoint is meant to have patch semantics, only updating the link types that are actually sent. In particular we need to support the world when only some of the links are managed by Whitehall, and the rest are managed by content-tagger.

elliotcm:
Deletion issue aside, if the content has no links, why send a request to the links endpoint at all?

rboulton:
To answer the "How else would you remove them" question - if you actively wanted to remove the links for a content item, you would need to request the current links, and then send empty lists of links for each of the link types that existed.  However, I'm not sure why you'd need to do this.

app/commands/v2/put_link_set.rb
jamiecobbett: If it is meant to have patch semantics, why doesn't it use HTTP PATCH?

jamiecobbett:
> Deletion issue aside, if the content has no links, why send a request to the links endpoint at all?

Because then you need a conditional in your code to see if there any links to send. It didn't occur to us to code for this behaviour. Should we expect all API users to take this special case into account?

elliotcm:
If we're treating it as a restful resource then yes, because sending empty data to one of those would result in a 422.  If it's considered a different kind of endpoint then I don't know.

Just seems strange to send an empty PUT to an endpoint without it complaining.

jamiecobbett:
Closing this to reopen with a different scope.

PR #196, opened 2016-01-26
Index endpoint details hash

Rosa-Fox:
https://trello.com/c/pegK7Wtw/485-index-endpoint-should-return-details-hash

### DIFF
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -11,14 +11,13 @@ module Queries
 
     def call
       validate_fields!
-
       content_items.map do |content_item|
-        hash = Presenters::Queries::ContentItemPresenter.new(
+        presenter = Presenters::Queries::ContentItemPresenter.new(
           content_item,
           draft_version(content_item),
           live_version(content_item)
         )
-        hash.present.as_json(only: output_fields)
+        select_output_fields_only(presenter)
       end
     end
 
@@ -86,6 +85,10 @@ module Queries
       when LiveContentItem
         @live_versions[item.id]
       end
+    end
+
+    def select_output_fields_only(presenter)
+      presenter.present.slice(*output_fields).as_json
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/queries/get_content_collection.rb
```diff
@@ -11,14 +11,13 @@ def initialize(content_format:, fields:, publishing_app: nil, locale: nil)
 
     def call
       validate_fields!
-
       content_items.map do |content_item|
         hash = Presenters::Queries::ContentItemPresenter.new(
```
tuzz: I think we should rename the `hash` variable to `presenter`.

Alternatively, we could call chain a call to `#present` onto the presenter, e.g.

```ruby
hash = Presenters::Queries::ContentItemPresenter.new( ... ).present
```

app/queries/get_content_collection.rb
```diff
@@ -87,5 +86,9 @@ def live_version(item)
         @live_versions[item.id]
       end
     end
+
+    def select_output_fields_only(hash)
+      hash.present.as_json.select { |k, v| output_fields.map(&:to_s).include?(k) }
+    end
```
tuzz: You can probably use [Hash#slice](http://api.rubyonrails.org/classes/Hash.html#method-i-slice) for this. You won't need to `#to_s` the fields if you slice before calling `#as_json`.

spec/queries/get_content_collection_spec.rb
```diff
@@ -134,4 +134,18 @@
...
+        expect(Queries::GetContentCollection.new(
+          content_format: 'topic',
+          fields: ['details'],
+          publishing_app: 'publisher'
+        ).call).to eq([
+          { "details" => {"foo" => "bar"}, "publication_state" => "draft" },
+          { "details" => {"baz" => "bat"}, "publication_state" => "draft" }
+        ])
+    end
+  end
```
tuzz: :+1:

Rosa-Fox:
@tuzz thanks for the feedback :) I have altered. Let me know if you think there should be any further changes.

tuzz:
:+1: :rocket:

PR #198, opened 2016-01-26
Rename PutLinkSet to PatchLinkSet

jamiecobbett:
After speaking to @danielroseman and @rboulton we agreed that this endpoint's behaviour and the original intent was for it to be a PATCH action, not an overwrite (PUT). 

This is a stab at changing the method.

Related PRs:
* cleanup data migration: https://github.com/alphagov/publishing-api/pull/199
* update gds-api-adapters to use PATCH for links: https://github.com/alphagov/gds-api-adapters/pull/428

### DIFF
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -5,7 +5,7 @@ module Commands
         delete_existing_links
 
         V2::PutContent.call(v2_put_content_payload, downstream: downstream)
-        V2::PutLinkSet.call(v2_put_link_set_payload, downstream: downstream)
+        V2::PatchLinkSet.call(v2_put_link_set_payload, downstream: downstream)
         V2::Publish.call(v2_publish_payload, downstream: downstream)
       else
         base_path = payload.fetch(:base_path)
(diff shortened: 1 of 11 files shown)


danielroseman:
Looks good.
As you mentioned, it makes sense to change the name but keep supporting PUT for now. 

For the event log, we could probably do a migration to change the `action` value.
The pact tests - and in fact the api-adapters client - would need to change so that they call the new methods, there probably wouldn't need to be any other changes.

jamiecobbett:
@danielroseman I've added a PR to do the data migration after this is deployed and I've opened a PR to switch over gds-api-adapters.

jamiecobbett:
I've just successfully tested Whitehall master against these changes, both with using PUT (ie Whitehall master) and switching it to use PATCH.

boffbowsh:
LGTM but will let Publishing Platform team merge

danielroseman:
@jamiecobbett ready to merge this at last, but needs a rebase first.

jamiecobbett:
This is now rebased on top of master and is ready for a final review/merge.

PR #201, opened 2016-02-10
Refactor object model

tuzz:
**do not merge yet**

Opening early for visibility and to use a discussion point.

[Note from @elliotcm: please see my comment below for the best way to contribute feedback at this stage]

Depends on: https://github.com/alphagov/gds-api-adapters/tree/publishing-api-refactor

Requires a rebase/merge from master. Some of the pact tests won't work until this is done.

Check https://trello.com/c/qV540aFr/380-refactor-extract-publishstate-from-contentitems for a list of changes still to come.


elliotcm:
I've had a chat with @boffbowsh about how much value there is in people reviewing this as is, and we've concluded that the high level questions (like the semver one) are still useful but maybe not in the PR itself.  If you're looking to review this and have those sorts of questions, first check the trello ticket to see if it's something we're going to work on and then drop me and patuzzo an email so we can either address them or include the answers in the docs.

boffbowsh:
We're running the New Object Model migration at the moment, it's taking a while to run. What's the plan for deployment of this? It feels like the sort of thing that needs doing carefully - as far as I can tell there's no provision to keep the new and old object models updated while the migration is running, and to sync them up afterwards?

bradwright:
@boffbowsh we're going to talk about that soon, probably with a whiteboard.

boffbowsh:
Cool, mind involving me in that please?

tuzz:
Superseded by https://github.com/alphagov/publishing-api/tree/refactor-object-model-merged-master

PR #202, opened 2016-02-11
RFC: Change automatic filtering by publishing-app

tijmenb:
Hi publishing team,

Since https://github.com/alphagov/publishing-api/pull/188 the publishing-api only returns content that is "owned" by the authenticating user. This is done by inferring the requesting app from the email address belonging to the authenticating user (via the bearer token). It is possible to make publishing-api return all content by giving the user the `view_all` permission.

This is done so that if `foo-publisher` calls `get_content_items` (the `/v2/content` endpoint), it will only see its own content. This is is good because content should only be managed by one app. 

But there are some problems with this approach.

- In development, `publishing-api` uses a [`MockBearerToken`](https://github.com/alphagov/gds-sso/blob/0c4a05e8b6eaf0f14e28ac831d7635c097e24c14/lib/gds-sso/bearer_token.rb#L48-L61). This causes making the publishing-api to think that the request has come from the app `dummyapiuser`. By default, publisher apps in development will therefore not be able to see their own content, or the content synced from production. The same for `content-tagger`, which currently doesn't get data back in development. This is only fixed by changing the permissions of the dummy user to include `view_all` (which defeats the purpose and causes dev-prod disparity).
- It prevents querying the URL of the publishing-api in the browser. This silently returns data for the currently signed in user (which is not the `dummyapiuser` mentioned above, but rather `User.first` - which currently is `whitehall`). The only way to discover what is being returned is looking at the Rails SQL logs.
- The `view_all` permission is set on the application-level. This means that a publisher application will not be able to query it's own content items (for the admin index page) and _all_ content items (for tagging). 

In addition, the method of automatically filtering items could be surprising. It may be confusing that you can't see from the code what data is returned (all or filtered). For example, a developer has no way to know what [any of these calls](https://github.com/search?q=user%3Aalphagov+get_content_items&ref=searchresults&type=Code&utf8=%E2%9C%93) is *supposed* to return. There's no way to specify the intention.

I'm looking for a way to fix the problems and make it more explicit. Here's what I'm thinking of:

1. Clients must explicitly ask for the content of their `publishing_app` in the URL
2. Clients can specify `publishing_app=all` to get all content
3. Requests without `publishing_app` are rejected

I think that the added security of enforcing this query with the `view_all` permissions may be overkill (apps have to explicitly request `all`, which is easily caught in code review). If we do want this I'd suggest returning HTTP 403 if the client requests `all` but doesn't have `view_all` permissions.

Thoughts?


tuzz:
Hi @tijmenb – it might be best to raise this as an RFC as I think that would get a bit more traction.

Is that ok?

tijmenb:
Hey @tuzz. The only traction I need is from your team on this issue - and you all seemed busy with plague & refactor this week so I didn't chase this :) 

So now you're here: any thoughts on the issues? Does the refactor change the way auth/filtering is done?

tuzz:
I've made a ticket so we can track this: https://trello.com/c/cw6gknmC/610-review-tijmen-s-rfc-change-automatic-filtering-by-publishing-app

tijmenb:
Great! Thanks.

mobaig:
Big :+1: to opening this up for discussion - has caught me out a few times when working with content-tagger.

bradwright:
We're going to be working on filtering and other features next as part of a bigger piece of work - we'll come back to you all to discuss your needs then. Until that happens I'm closing this to keep it off our radar.

PR #203, opened 2016-02-11
`ContentStoreWorker` accepts content item `id`

gpeng:
To avoid the potential race when links and content are sent to the Publishing API almost simultaneously by Whitehall we have moved the presentation of the item into the worker. This will ensure that even if a newer version of the item is received by Publishing API it will not cause `transmitted_at` conflict on the Content Store.

[Trello](https://trello.com/c/a7qp3eOx)

There is a refactor going on of Publishing API so please don't merge until we've had a 👍from the Publishing Platform team.

cc: @danielroseman @tuzz

In response to @jamiecobbett comment -

Where the publishing API receives two requests in quick succession (as it does when Whitehall publishes content and links). With the existing approach request one gets a `transmitted_at` timestamp of 100 (for simplicity) and the subsequent request gets `transmitted_at` timestamp of 110. These are both queued on an async queue. If request two gets processed before one then content store will return and error when one is processed. Admittedly it 'should' have the latest representation at this point anyway but the amount of errors when doing a 'bulk' republish could be masking other errors.

With this PR when request two is 'worked' it is sent to the content store no problem. Request one is then processed and retrieves the latest representation from the Publishing API database (which was stored when request two was sent to the Publishing API). So... it gets the latest version and that is sent to the content store. 

This does ultimately cause two identical requests to be sent to the content store but that is preferable to the error and could be fixed in an optimisation in the future should it cause any performance issues (which is unlikely).

This does beg the question as to whether `transmitted_at` is required any longer? I'm happy to remove it but would appreciate some feedback.

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -36,8 +36,7 @@ module Commands
         if downstream
           ContentStoreWorker.perform_async(
             content_store: Adapters::DraftContentStore,
-            base_path: draft.base_path,
-            payload: Presenters::ContentStorePresenter.present(live),
+            live_content_item_id: live.id,
           )
         end
       end
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -77,11 +77,9 @@ module Commands
         end
 
         if downstream
-          live_payload = Presenters::ContentStorePresenter.present(live_content_item)
           ContentStoreWorker.perform_async(
             content_store: Adapters::ContentStore,
-            base_path: live_content_item.base_path,
-            payload: live_payload,
+            live_content_item_id: live_content_item.id,
           )
 
           queue_payload = Presenters::MessageQueuePresenter.present(live_content_item, update_type: update_type)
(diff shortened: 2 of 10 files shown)


jamiecobbett:
Would it be possible to add a (hypothetical) timeline of a bug that this change prevents?

boffbowsh:
This looks good to me :+1:

tuzz:
Looks good. :+1: 

> This does beg the question as to whether transmitted_at is required any longer? I'm happy to remove it but would appreciate some feedback.

I was wondering this, too.

boffbowsh:
It does seem to me that it could be removed. Could we merge this as-is though and look at removal of it in a different PR?

gpeng:
It looks like we could remove all that UpdateLock malarkey from content store as a result. I'll do a PR for that and get feedback. Just delete a few tings...

tuzz:
Happy for this to be merged.

We'll fix the refactor merge with this later. These changes don't look too complicated that we could reasonably merge them with the refactor once it's ready.

PR #204, opened 2016-02-15
Remove `transmitted_at` from V2 Content Store calls

gpeng:
This commit removes `transmitted_at` from the V2 Content Store calls and
related tests. This supports PR https://github.com/alphagov/content-store/pull/189

Some tests have been updated to allow V1 to still support
`transmitted_at`

### DIFF
--- a/spec/pacts/content_store/put_endpoint_spec.rb
+++ b/spec/pacts/content_store/put_endpoint_spec.rb
@@ -9,16 +9,11 @@ RSpec.describe "PUT endpoint pact with the Content Store", pact: true do
 
   let(:client) { ContentStoreWriter.new("http://localhost:3093") }
   let(:body) { Presenters::ContentStorePresenter.present(content_item) }
-  let(:request_time) { Time.at(2000000000) }
 
-  around do |example|
-    Timecop.freeze(request_time) { example.run }
-  end
-
-  context "when a content item exists that has an older transmitted_at than the request" do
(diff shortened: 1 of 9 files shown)


spec/pacts/content_store/put_endpoint_spec.rb
```diff
@@ -9,132 +9,32 @@
...
         .upon_receiving("a request to create a content item")
         .with(
-          method: :put,
-          path: "/content/vat-rates",
-          body: body,
-          headers: {
-            "Content-Type" => "application/json"
-          },
-        )
+      method: :put,
```
boffbowsh: I think the indentation is incorrect here

boffbowsh:
This LGTM apart from indentation issue, but would like someone from Publishing Platform team to look at it too.

spec/pacts/content_store/put_endpoint_spec.rb
gpeng: Fixed in 7b558fa

gpeng:
`DownstreamPresenter::V1` still has the `transmitted_at` attribute. I did have a go at removing it but there is a fair amount of dependent code in the Commands (the non-V2 ones... bit confusing.) If this version is used once `transmitted_at` is removed from Content Store then the calls will fail. Is all that code still being used by something?

boffbowsh:
Closing in favour of a new atomic counter

PR #207, opened 2016-02-17
Make content store push on discard draft more obvious

benilovj:
This change doesn't functionally change anything, because at the point
when the publishing action is called, the live and draft content items
are exactly the same.

However, this change makes the code slightly clearer because it no longer breaks
the pattern "live items go to the live store, draft items to the draft store".


gpeng:
I think with this change we no longer need to make a distinction between `live_content_id` and `draft_content_id` as this can now be inferred from the `content_store` argument. We could just pass in `content_id`.

benilovj:
I'm going to close this change, because the object model re-engineering (#206) is going to change how this works anyway, and this isn't a big issue at the moment.

PR #208, opened 2016-02-18
Add delay to `ContentStoreWorker` enqueuing

gpeng:
We are currently experiencing a fair number of `ActiveRecord::RecordNotFound` errors from the `ContentStoreWorker`. These are successfully retried so don't impact functionality as such but
do cause a lot of error 'noise'. This commit enqueues the jobs to be run in `1.second` as a temporary measure to reduce error noise. This is not meant as a permanent fix. Once the new object model is implemented and deployed we can revisit if this is still a problem so I don't think this needs to be merged into that branch.

I discussed this yesterday and it was mentioned that delaying scheduled publishings might be an issue but I don't feel that 1 second is a problem as there is no guarantee that they'd have been 'worked' from the queue within 1 second anyway.

I've run this branch on integration and so far have not seen any errors so it appears to have 'fixed' the problem for now.

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -34,7 +34,8 @@ module Commands
         draft.update_attributes(live.attributes.except("id", "draft_content_item_id"))
 
         if downstream
-          ContentStoreWorker.perform_async(
+          ContentStoreWorker.perform_in(
+            1.second,
             content_store: Adapters::DraftContentStore,
             live_content_item_id: live.id,
           )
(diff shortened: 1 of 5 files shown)


benlovell:
I believe it's spelt _enqueuing_.

gpeng:
@benlovell Noted and fixed (with no small amount of annoyance)

benlovell:
Looks good to me. My first thought was maybe one second is too low, but after speaking in person I'm assured everything works as expected in integration.

Probably best to wait for commentary from the publishing folks before :ship:

PR #210, opened 2016-02-22
Add `receipt_order` to `DraftContentItem` and `LiveContentItem`

gpeng:
This PR adds an integer field `receipt_order` to each of `DraftContentItem` and `LiveContentItem` that is incremented at the database to avoid the risk of incorrect ordering caused by race conditions or the time based ordering that we currently employ.

In this PR the content items are stamped with the `receipt_order` field within each command on their way to the content store. The field is not currently 'presented` to the content store though. There will be subsequent PRs to add support for both the `transmitted_at` and `receipt_order` fields concurrently on content store following which we'll be able to go back and start sending `receipt_order` before removing `transmitted_at` altogether.

[Trello](https://trello.com/c/dBmI96gI/304-change-locking-criteria-on-content-store)

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -33,6 +33,8 @@ module Commands
       def update_draft_from_live
         draft.update_attributes(live.attributes.except("id", "draft_content_item_id"))
 
+        draft.increment_receipt_order
+
         if downstream
           ContentStoreWorker.perform_in(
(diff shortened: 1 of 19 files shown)


gpeng:
I've close this pending the merge of https://github.com/alphagov/publishing-api/pull/206. There is general agreement that this approach is sound it can be replicated on top of the new version of Publishing API.

PR #211, opened 2016-02-24
Paginate index with the refactor

Rosa-Fox:
Performs pagination on content items. (Trello card [here!](https://trello.com/c/OZF9rKeX/487-index-endpoint-should-support-pagination)).

I began working on this a few weeks ago but we decided to park it until the refactor was complete so that the pagination would be performed with true database order. 

The previous work was undertaken on a [separate branch](https://github.com/alphagov/publishing-api/pull/189) which I plan to delete with this as replacement (basically due to difficulties with the rebase/to save time!). This is now branched from `refactor-object-model-merged-master`. I have taken on board the comments made on the previous branch from @tuzz, @jamiecobbett and @rboulton in this implementation of pagination with the refactor.

Please let me know of any changes that could be made to improve it.


app/pagination.rb
```diff
@@ -0,0 +1,16 @@
+class Pagination
+  attr_reader :start, :page_size, :all_items
+
+  def initialize(options={})
+    if options[:start] || options[:page_size]
+      @start = options[:start] ? options[:start].to_i : 0
```
benlovell: You could reduce some of this using `fetch`. For example: 
```ruby
@start     = Integer(options.fetch(:start, 0))
@page_size = Integer(options.fetch(:page_size, 50))
```

danielroseman:
Looks good so far. The refactor branch has been merged so this can be rebased onto master.

The main thing missing is ordering; as noted on the previous PR, paginating without a specific sort order could result in inconsistent data between separate calls. It was suggested that we provide a order parameter, with a sensible default such as reverse updated_at; it should be much easier to implement now that there is only one ContentItem model.

steventux:
I've borked this branch when rebasing, I'll cherry-pick @Rosa-Fox's relevant commits along with my own and open a new PR

PR #212, opened 2016-02-25
Get content collection tuning

tuzz:
https://trello.com/c/3z7pDHmO/616-performance-tuning-getcontentcollection

Solves the immediate performance issues we've been seeing on the GetContentCollection endpoint. We're going to be looking at extracting some of this logic from the presenter in this ticket:
https://trello.com/c/jzI3JlLH/515-remove-content-register

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -22,8 +22,8 @@ module Presenters
         scope = join_supporting_objects(content_item_scope)
         scope = select_fields(scope)
 
-        items = scope.as_json.map(&:symbolize_keys)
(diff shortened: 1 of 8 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -10,20 +10,21 @@ def self.present(content_item)
...
+      def initialize(content_item_scope, fields = nil)
         self.content_item_scope = content_item_scope
+        self.fields = fields
       end
 
       def present
         scope = join_supporting_objects(content_item_scope)
         scope = select_fields(scope)
 
-        items = scope.as_json.map(&:symbolize_keys)
```
tuzz: This was the main source of the performance issues. `#as_json` is incredibly slow.

PR #213, opened 2016-02-25
Add ContentStorePayloadVersion

gpeng:
This PR replaces `transmitted_at` with an incrementing integer `payload_version`. Since we started 'presenting' the content items at the time the Sidekiq worker actually runs we should now be able to safely ignore payloads that have a lower or equal `payload_version`. Using an integer, incremented atomically at the DB should protect us from race conditions and negate the time variance we have between machines that was causing issues with the time based `transmitted_at`.

[This PR](https://github.com/alphagov/content-store/pull/193) on Content Store added support at the 'other end'

[Trello](https://trello.com/c/dBmI96gI/304-change-locking-criteria-on-content-store)

### DIFF
--- a/app/models/content_store_payload_version.rb
+++ b/app/models/content_store_payload_version.rb
@@ -1,0 +1,39 @@
+class ContentStorePayloadVersion < ActiveRecord::Base
+  def self.increment(content_item_id)
+    sql = <<-SQL
+        INSERT INTO content_store_payload_versions (content_item_id)
+        SELECT #{content_item_id || 'NULL'}
+        WHERE NOT EXISTS(
+        SELECT *
+        FROM content_store_payload_versions
+        WHERE content_item_id #{content_item_id_sql(content_item_id)});
+
+        UPDATE content_store_payload_versions
+        SET current = COALESCE(current, 0) + 1
+        WHERE content_item_id #{content_item_id_sql(content_item_id)}
+        RETURNING current;
+      SQL
+
+    ContentStorePayloadVersion.connection
+      .execute(sql)
+      .first["current"].to_i
+  end
+
+  def self.current_for(content_item_id)
+    ContentStorePayloadVersion.find_by(content_item_id: content_item_id).current
+  end
+
+  class V1
+    def self.current
+      ContentStorePayloadVersion.current_for(nil)
+    end
+
+    def self.increment
+      ContentStorePayloadVersion.increment(nil)
+    end
+  end
+
+  def self.content_item_id_sql(content_item_id)
+    content_item_id.nil? ? "IS NULL" : " = #{content_item_id}"
+  end
+end
--- a/db/migrate/20160225113309_create_content_store_payload_versions.rb
+++ b/db/migrate/20160225113309_create_content_store_payload_versions.rb
@@ -1,0 +1,10 @@
+class CreateContentStorePayloadVersions < ActiveRecord::Migration
+  def change
+    create_table :content_store_payload_versions do |t|
+      t.integer :content_item_id
+      t.integer :current, default: 0
+    end
+
+    add_index :content_store_payload_versions, :content_item_id, unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -54,6 +54,13 @@ ActiveRecord::Schema.define(version: 20160229133026) do
   add_index "content_items", ["publishing_app"], name: "index_content_items_on_publishing_app", using: :btree
   add_index "content_items", ["rendering_app"], name: "index_content_items_on_rendering_app", using: :btree
 
+  create_table "content_store_payload_versions", force: :cascade do |t|
+    t.integer "content_item_id"
+    t.integer "current",         default: 0
+  end
+
+  add_index "content_store_payload_versions", ["content_item_id"], name: "index_content_store_payload_versions_on_content_item_id", unique: true, using: :btree
+
   create_table "draft_content_items", force: :cascade do |t|
     t.string   "content_id"
     t.string   "locale",               default: "en"
--- a/spec/models/content_store_payload_version_spec.rb
+++ b/spec/models/content_store_payload_version_spec.rb
@@ -1,0 +1,89 @@
+require "rails_helper"
+
+RSpec.describe ContentStorePayloadVersion do
+  let(:content_item_id) { 10 }
+  describe ".current_for" do
+    it "returns the current value for the supplied content_item_id" do
+      described_class.increment(content_item_id)
+      expect(described_class.current_for(content_item_id)).to eq(1)
+    end
+  end
+
+  describe ".increment" do
+    context "when there is no record" do
+      it "creates one" do
+        described_class.increment(content_item_id)
+        expect(described_class.count).to eq(1)
+      end
+
+      it "initializes current to 1" do
+        described_class.increment(content_item_id)
+        expect(described_class.last.current).to eq(1)
+      end
+
+      it "sets the content_item_id" do
+        described_class.increment(content_item_id)
+        expect(described_class.last.content_item_id).to eq(content_item_id)
+      end
+
+      it "returns the current value" do
+        expect(described_class.increment(content_item_id)).to eq(1)
+      end
+    end
+
+    context "when a record already exists" do
+      let!(:content_store_payload_version) do
+        described_class.create(
+          content_item_id: content_item_id,
+          current: 10
+        )
+      end
+
+      it "increments the current value before returning it" do
+        expect(described_class.increment(content_item_id))
+          .to eq(11)
+      end
+    end
+  end
+
+  describe ContentStorePayloadVersion::V1 do
+    describe ".current" do
+      it "returns the current value from the nil content_item_id record" do
+        described_class.increment
+        expect(described_class.current).to eq(1)
+      end
+    end
+
+    describe ".increment" do
+      context "when there is no record" do
+        it "creates one" do
+          described_class.increment
+          expect(ContentStorePayloadVersion.count).to eq(1)
+        end
+
+        it "sets content_item_id to null" do
+          described_class.increment
+          expect(ContentStorePayloadVersion.last.content_item_id).to be_nil
+        end
+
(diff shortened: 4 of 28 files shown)


app/models/content_store_payload_version.rb
```diff
@@ -0,0 +1,41 @@
+class ContentStorePayloadVersion < ActiveRecord::Base
+  def self.increment(content_item_id:)
```
benlovell: I think this could be a regular argument, defaulting to `nil`, instead of a keyword argument. This means callers wouldn't have to pass a named argument, given the method signature only has an (assumed, given they're keyword arguments) arity of one.

app/models/content_store_payload_version.rb
```diff
@@ -0,0 +1,41 @@
...
+        WHERE content_item_id #{ content_item_id_sql(content_item_id) }
+        RETURNING current;
+      SQL
+
+    ContentStorePayloadVersion.connection
+      .execute(sql)
+      .first["current"].to_i
+  end
+
+  def self.current_for(content_item_id:)
```
benlovell: Same here, I don't think the keyword argument makes much sense in this case.

app/models/content_store_payload_version.rb
```diff
@@ -0,0 +1,41 @@
...
+        WHERE content_item_id #{ content_item_id_sql(content_item_id) });
+
+        UPDATE content_store_payload_versions
+        SET current = COALESCE(current, 0) + 1
+        WHERE content_item_id #{ content_item_id_sql(content_item_id) }
+        RETURNING current;
+      SQL
+
+    ContentStorePayloadVersion.connection
+      .execute(sql)
```
benlovell: This isn't implicitly transacted. You probably want to ensure this is running in a transaction.

db/schema.rb
```diff
@@ -163,6 +170,9 @@
 
   add_index "path_reservations", ["base_path"], name: "index_path_reservations_on_base_path", unique: true, using: :btree
 
+  create_table "receipt_order", force: :cascade do |t|
```
benlovell: Should this be here?

app/models/content_store_payload_version.rb
gpeng: I had it as a regular argument at first but it didn't read all that well when it is called.
`ContentStorePayloadVersion.increment(content_item_id: 10)` looked better to me than `ContentStorePayloadVersion.increment(10)`.

benlovell: That's true enough in your tests, but in the real world usage you're passing `content_item.id` so reading the full signature in the caller, it's currently something like `ContentStorePayloadVersion.increment(content_item_id: content_item.id)` which feels odd when you could just `ContentStorePayloadVersion.increment(content_item.id)`. So the provided argument forms part of the signature.

boffbowsh:
The V1 implementation feels brittle. Is it possible to extract the content_id from the payload then fetch the content_item_id from that and increment? Or do things not exist in the DB at that point?

app/models/content_store_payload_version.rb
benlovell: Actually the more I think about this, `increment` may be a misnomer. Would `next` be more suitable? Given it's something resembling a sequence.

boffbowsh: Personally I'm happy with this as it is. I think @gpeng's argument for readability and clear intent makes sense.

benlovell: This feels odd to me and somewhat without precedent. Keyword arguments are intended to solve connascence of name. Sure, if it provides hints then great, but then aren't we saying the method isn't suitably named?

boffbowsh: I don't see how it's different to the `root to: "home#index"` pattern, or `has_many :x, through: :y`

app/models/content_store_payload_version.rb
boffbowsh: If it's not in a transaction, it can't be forced into a relevant transaction. It still needs to be run though.

app/models/content_store_payload_version.rb
benlovell: Both of those examples take multiple options. The method in question takes a single named argument.

app/models/content_store_payload_version.rb
benlovell: This is also racier than Lewis Hamilton, without running serially in a transaction.

boffbowsh: I'm not sure it matters. The counter will always be going up, which is what we want.

benlovell: In that case we should add uniqueness over `current` and `content_item_id` else under load we'll end up with duplicates.

boffbowsh: Yes, true :+1:

app/models/content_store_payload_version.rb
boffbowsh: So maybe `.next_for` or `.increment_for`, similar to `.current_for`. FWIW I'm fine with increment.

gpeng: I think the using regular args is fair enough here. You're right Ben that it causes 'duplication' when you're passing `content_item_id: content_item.id`. It's in the tests where it is more readable but that causes uglification™ in 'real life'. 

As for the `increment`/`next`... I actually had it called `next` originally but it felt a bit passive. Like a 'get' kind of thing. `increment` felt more descriptive in that that's what it actually does. The fact that it returns the value is almost a byproduct (and arguably unnecessary now the use of it has fleshed out a bit)

benlovell: Cool, that makes sense to me. :+1:

app/models/content_store_payload_version.rb
gpeng: Yeah.. I wrestled with this a bit. It's doing the creation in the same call that's the issue. I don't think wrapping it in a transaction in the sql helps with that by my understanding. The read bit 'not existing' could still happen more than once before the write takes place I think. I think it's the `content_item_id` that we need to enforce uniqueness on though. I don't think `current` needs to be involved.

db/schema.rb
gpeng: err... no.

gpeng:
@boffbowsh The 'V1' comment. Turns out that through a few 'paths' of the V1 commands, nothing actually gets saved to the database and so we never have a `ContentItem` to get the `id` from. It's unclear whether anything still takes these paths but they're still being supported so this was the best I could come up with. As a happy by product the `current` value on that record will be a count of the number of requests that actually take those paths so we can check along the way and see. Most V1 stuff now gets sent through the V2 commands anyway so predominantly all is well in the world.

boffbowsh:
V1 reasoning sounds good to me :+1:

gpeng:
I've addressed the above comments in their relevant commits and added statsd.

db/migrate/20160225113309_create_content_store_payload_versions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateContentStorePayloadVersions < ActiveRecord::Migration
+  def change
+    create_table :content_store_payload_versions do |t|
+      t.integer :content_item_id
+      t.integer :current, default: 0
+    end
+
+    add_index :content_store_payload_versions, :content_item_id
```
benlovell: This should be unique.

spec/models/content_store_payload_version_spec.rb
```diff
@@ -0,0 +1,89 @@
+require "rails_helper"
+
+RSpec.describe ContentStorePayloadVersion do
+  let(:content_item_id){ 10 }
```
benlovell: Missing a space.

spec/models/content_store_payload_version_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        expect(described_class.last.content_item_id).to eq(content_item_id)
+      end
+
+      it "returns the current value" do
+        expect(described_class.increment(content_item_id)).to eq(1)
+      end
+    end
+
+    context "when a record already exists" do
+      let!(:content_store_payload_version){
```
benlovell: Same here.

spec/models/content_store_payload_version_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+          expect(ContentStorePayloadVersion.last.content_item_id).to be_nil
+        end
+
+        it "returns current value of 1" do
+          expect(described_class.increment).to eq(1)
+        end
+      end
+
+      context "when a record already exists" do
+        let!(:content_store_payload_version){
```
benlovell: Same here. :cop:

db/migrate/20160225113309_create_content_store_payload_versions.rb
gpeng: Oh.. I added that. Must have gone missing in my rebase-o-thon.

spec/support/request_helpers/downstream_requests.rb
```diff
@@ -4,18 +4,16 @@ def sends_to_draft_content_store
...
+        expect(PublishingAPI.service(:draft_content_store)).to receive(:put_content_item)
+          .with(
+            base_path: base_path,
+            content_item: content_item_for_draft_content_store
+              .merge(payload_version: anything)
+          )
+
+        do_request
+
+        expect(response.status).to eq(200), response.body
```
benlovell: `expect(response.status).to be_ok, response.body` might read nicer but to be honest, this style of multiple expectations on a single line makes me have to stop and think every time. :speak_no_evil:

spec/support/request_helpers/downstream_requests.rb
```diff
@@ -23,18 +21,16 @@ def sends_to_live_content_store
...
+            content_item: content_item_for_live_content_store
+              .merge(payload_version: anything)
+          )
 
-          do_request
+        do_request
 
-          expect(response.status).to eq(200), response.body
-        end
+        expect(response.status).to eq(200), response.body
```
benlovell: Same here.

boffbowsh:
:+1: good work :)

PR #214, opened 2016-02-26
Add govuk-lint in diff mode

boffbowsh:
There's a few hundred violations, the vast majority of which can be auto corrected. I'll fix them in a separate PR and remove the diff mode switch.


jamiecobbett:
Would it be better to hook it to the default rake task as per https://github.com/alphagov/government-frontend/pull/107 ?

boffbowsh:
I don't particularly see the merit of that approach versus putting it in `jenkins.sh`. Especially while a repo is in diff mode.

PR #216, opened 2016-02-29
Clone content register

elliotcm:
This endpoint is to serve as a replacement for the current content register endpoint, namely for populating dropdowns in publishing applications.

Although we're making use of the same code as the index endpoint, they serve different purposes.

To my understanding the purpose of the index endpoint is to provide all the information required to render an index page in a publishing application.  This requires some fields that the dropdowns do not, and vice versa.  The index endpoint also has authentication, and soon ordering, filtering and pagination, none of which the dropdowns need.

I believe the eventual intent is for the publishing API to also provide an endpoint which gives all the information required to render an edit page for a given format, at which point this dropdowns endpoint will become redundant and can be removed.

I've named the endpoint `linkables` because the content register used "entries" which needs more context in an app with multiple endpoints.  It provides lists of things which can be linked to your content item, hence the name.

There are possibly further performance improvements that could be made to the underlying code and I'd love to hear them, but I don't feel that they should block this code going out as testing seems to show it to be Fast Enough™.

https://trello.com/c/jzI3JlLH/515-remove-content-register

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -1,19 +1,32 @@
 module V2
   class ContentItemsController < ApplicationController
     def index
-      content_format = params.fetch(:content_format)
-      fields = params.fetch(:fields)
-      locale = params[:locale]
+      format = query_params.fetch(:content_format) { query_params.fetch(:format) }
+
       render json: Queries::GetContentCollection.new(
-        content_format: content_format,
-        fields: fields,
+        content_format: format,
+        fields: query_params.fetch(:fields),
         publishing_app: publishing_app,
-        locale: locale,
+        locale: query_params[:locale],
+      ).call
+    end
+
+    def linkables
+      # Base path is returned to facilitate rummager indexing.
+      # This can be removed once link updates are picked up by rummager from the message bus.
+      render json: Queries::GetContentCollection.new(
+        content_format: query_params.fetch(:format),
+        fields: %w(
+          title
+          content_id
+          publication_state
+          base_path
+        ),
       ).call
     end
 
     def show
-      render json: Queries::GetContent.call(params[:content_id], params[:locale])
+      render json: Queries::GetContent.call(path_params[:content_id], query_params[:locale])
     end
 
     def put_content
@@ -34,7 +47,7 @@ module V2
   private
 
     def content_item
-      payload.merge(content_id: params[:content_id])
+      payload.merge(content_id: path_params[:content_id])
     end
 
     def publishing_app
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -21,6 +21,7 @@ module V2
           content_id
           publication_state
           base_path
+          internal_name
         ),
       ).call
     end
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -33,14 +33,19 @@ module Presenters
           most_recent_item = draft || live
           next unless most_recent_item
 
+          parse_json_fields!(most_recent_item)
+
+          if output_fields.include?("internal_name")
+            details = most_recent_item.fetch("details")
+            most_recent_item["internal_name"] = details["internal_name"] || most_recent_item.fetch("title")
+          end
+
           most_recent_item["publication_state"] = publication_state(draft, live)
           most_recent_item["lock_version"] = most_recent_item.fetch("lock_version").to_i
 
           if live
             most_recent_item["live_version"] = live.fetch("lock_version").to_i
           end
-
-          parse_json_fields!(most_recent_item)
 
           most_recent_item.slice(*output_fields)
         end
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -68,7 +68,7 @@ module Presenters
 
       def select_fields(scope)
         scope.select(
-          *(ContentItem::TOP_LEVEL_FIELDS - ["publication_state"]),
+          *ContentItem::TOP_LEVEL_FIELDS,
           "content_id",
           "states.name as state_name",
           "lock_versions.number as lock_version",
--- a/spec/presenters/queries/content_item_presenter_spec.rb
+++ b/spec/presenters/queries/content_item_presenter_spec.rb
@@ -171,5 +171,46 @@ RSpec.describe Presenters::Queries::ContentItemPresenter do
         expect(results.first.keys).to match_array(%w(title phase publication_state))
       end
     end
+
+    context "when internal_name is requested" do
+      let(:fields) { %w(internal_name) }
+
+      context "and an internal_name is present" do
+        before do
+          details = content_item.details
+          details[:internal_name] = "An internal name"
+
+          content_item.update_attributes(
+            details: details,
+          )
+        end
+
+        it "returns the internal_name" do
+          content_items = ContentItem.where(content_id: content_id)
+
+          results = described_class.present_many(content_items, fields: fields)
+          expect(results.first["internal_name"]).to eq("An internal name")
+        end
+      end
+
+      context "but an internal_name is not present" do
+        before do
+          details = content_item.details
+          details.delete(:internal_name)
+
+          content_item.update_attributes(
+            details: details,
+            title: "A normal title",
+          )
+        end
+
+        it "falls back to the title" do
+          content_items = ContentItem.where(content_id: content_id)
+
+          results = described_class.present_many(content_items, fields: fields)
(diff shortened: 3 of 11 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -46,10 +51,25 @@ def present
...
+      end
+
     private
 
       attr_accessor :content_item_scope, :fields
 
+      def link_sets_by_content_id
+        @link_sets ||= begin
+          content_ids = content_item_scope.pluck(:content_id)
+          link_sets = LinkSet.where(content_id: content_ids).includes(:links)
```
tuzz: I was thinking that it might be more efficient to inner join to link sets rather than evaluating the query and feeding the results into another.

elliotcm: Yeah probably.  Should we write a ticket for another performance pass?

tijmenb:
I'm not a 100% convinced we need a separate endpoint for this.

> To my understanding the purpose of the index endpoint is to provide all the information required to render an index page in a publishing application. This requires some fields that the dropdowns do not, and vice versa. 

The current index endpoint has support to request arbitrary fields. This is used by [policy-publisher](https://github.com/alphagov/policy-publisher/blob/63c3d30730464d96b5073c6ee8dd4fd00f898d10/lib/content_item_fetcher.rb#L3), [collections-publisher](https://github.com/alphagov/collections-publisher/blob/ec02dbb168898afa0d5be5ac7b32b46e8310601c/app/services/taxonomy/taxon_fetcher.rb#L7) (needs `base_path`) and [content-tagger](https://github.com/alphagov/content-tagger/blob/bd6483113e77e9a6dab8db2a49722a2e4c05c39b/app/services/tag_fetcher.rb#L15) (needs `details`). The fields exposed in this PR are not enough. I also hope that the dropdowns will never need the `links` of an item.

> The index endpoint also has authentication, and soon ordering, filtering and pagination, none of which the dropdowns need.

If a particular list of linkables (I like the name) becomes too large, I easily see a need for pagination, ordering and sorting.

> I believe the eventual intent is for the publishing API to also provide an endpoint which gives all the information required to render an edit page for a given format, at which point this dropdowns endpoint will become redundant and can be removed.

I don't know if this idea will work in practice. Currently, a edit page has attributes and links in the same edit page, but this will soon be split up in separate pages. For example, Finding Things has been working on splitting the content and tagging pages in Whitehall, because content & links have different workflows. Content has a draft state, while links are always directly set. Even if that weren't the case, content-tagger will still be using this endpoint, so it won't be removed.

In short, I think we don't need this particular endpoint, because the index endpoint already has the features needed for the dropdown (it was specifically created for that, see https://github.com/alphagov/publishing-api/pull/126). The only thing to change there is to remove the forced filtering introduced in https://github.com/alphagov/publishing-api/pull/188.

elliotcm:
@tijmenb Whitehall needs the links because it needs to know hierarchy of policies.  I believe collections publisher needs something similar.

Why would drop-downs ever need pagination?

The index endpoint may have been created for this purpose but it has gained a lot of functionality required to support phase 2 publishing applications.  As it is, this new endpoint is taking on the simpler role of the original endpoint you created.  You're right that the tagging part is moving into a separate flow, at which point a single request loading enough info for multiple dropdowns seems like a good idea anyway.  This could become that.

We have debated the need for separating these endpoints quite a bit.  @danielroseman could you chip in here?

tijmenb:
> Whitehall needs the links because it needs to know hierarchy of policies. I believe collections publisher needs something similar.

I was hoping to avoid that, because it is quite complex in the frontend. To make content-tagger display the hierarchy we've added `details.internal_name` in https://github.com/alphagov/govuk-content-schemas/pull/156 to achieve the same.

> Why would drop-downs ever need pagination?

The use case for sorting, filtering, pagination would be type-ahead dropdowns that send a search query to publishing-api (instead of downloading everything). A system like that could enable us to generate tagging interface for `related` content items (which can be of any type).

bradwright:
@tijmenb it sounds like you're talking theoretical use cases rather than existing ones. The intent of this is to replace content-register so we can delete the app, with an eye to making more complex pagination/queries available within the next iteration or two.

With that caveat in mind, it doesn't sound to me like the second of your objections is blocking, and I'm not sure about the first.

tijmenb:
Ok, I still think this is a needlessly duplicated endpoint but perhaps I'm missing something. I'll make some specific comments on the code so you can proceed.

app/controllers/v2/content_items_controller.rb
```diff
@@ -12,6 +14,18 @@ def index
...
     end
 
+    def linkables
+      render json: Queries::GetContentCollection.new(
+        content_format: request.query_parameters.fetch(:format),
+        fields: %w(
+          title
+          content_id
+          links
+          publication_state
```
tijmenb: If we want to use this endpoint for the tagging, we need at least `base_path` and `details` here.

app/controllers/v2/content_items_controller.rb
```diff
@@ -12,6 +14,18 @@ def index
       ).call
     end
 
+    def linkables
+      render json: Queries::GetContentCollection.new(
+        content_format: request.query_parameters.fetch(:format),
```
tijmenb: The index endpoint uses `content_format` instead of `format`. Would it be good to use that here as well?

elliotcm:
@tijmenb Let's have a chat when we're back in the office and we can look at consolidating them.

app/controllers/v2/content_items_controller.rb
```diff
@@ -41,5 +55,11 @@ def publishing_app
         current_user.app_name
       end
     end
+
+    def return_400_if_no_format
```
tijmenb: Using `params.fetch` like in the `index` method will raise a `ActionController::ParameterMissing` if the parameter is not provided. http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-fetch.

app/controllers/v2/content_items_controller.rb
elliotcm: Yeah not sure about this one.  The original was `content_format` but that was primarily to work around Rails nonsense which is no longer an issue.  There's other terms being used for the same thing so our conclusion was to defer this decision until this ticket: https://trello.com/c/1M8Njd0F/613-collating-the-names-for-everything

app/controllers/v2/content_items_controller.rb
elliotcm: Which apps need those?

app/controllers/v2/content_items_controller.rb
elliotcm: `params` includes magic rails params like `format` from the path extension, and clashes.

app/controllers/v2/content_items_controller.rb
```diff
@@ -12,6 +14,18 @@ def index
       ).call
     end
 
+    def linkables
```
tuzz: I'm still struggling with the term `linkables` – can anyone think of something more descriptive?

app/presenters/queries/content_item_presenter.rb
tuzz: Based on what we saw, I'm not sure this endpoint would be usable without tuning its performance?

Happy for that to go into a separate ticket.

tuzz:
We should add some client code and pacts in gds-api-adapters.

app/controllers/v2/content_items_controller.rb
tuzz: How many use cases do we think this endpoint will be used for? Will that list of fields cover those use cases or do we need to think about making it configurable?

spec/presenters/queries/content_item_presenter_spec.rb
```diff
@@ -130,27 +139,89 @@
...
+              ],
+            )
           end
         end
+
+        it "does not perform N+1 queries to retrieve link sets" do
+          expect {
+            described_class.present_many(ContentItem.all, fields: ["links"])
+          }.to make_database_queries(count: 0...content_item_count)
+        end
```
tuzz: :+1:

tuzz:
Mostly just questions from me. Code looks fine.

app/controllers/v2/content_items_controller.rb
elliotcm: I've been discussing this offline with @tijmenb

elliotcm:
@tuzz regarding the client code etc, I'm doing that as part of the client work.

app/controllers/v2/content_items_controller.rb
elliotcm: I can't, but I'm open.

elliotcm:
@tijmenb I've added the base path back and added the internal name as an explicit field.

elliotcm:
@tijmenb also removed links

PR #219, opened 2016-02-29
Ensure link set exists

tuzz:
This was the behaviour before the refactor and we don’t have
a good reason to change this.


app/commands/v2/put_content.rb
```diff
@@ -86,6 +87,14 @@ def create_supporting_objects(content_item)
         LockVersion.create!(target: content_item, number: lock_version_number_for_new_draft)
       end
 
+      def ensure_link_set_exists(content_item)
+        existing_link_set = LinkSet.find_by(content_id: content_item.content_id)
+        return if existing_link_set
```
elliotcm: Could collapse this into `return if LinkSet.exists?(content_id: content_item.content_id)`

PR #220, opened 2016-02-29
Guard against access limits going into the live content store...

tuzz:
We haven't got to the bottom of why some published content
items have access limits. However, we can at least guard
against them going downstream whilst we investigate.

**Note:** We haven't seen any new access limited items since we cleared them down in production – so it could have been residual data that wasn't migrated over correctly.


elliotcm:
It's probably not a bad idea to have a more permanent version of this in the future, but for now :+1:

PR #221, opened 2016-03-01
Deduplicate access limits...

steventux:
AccessLimit records have been unintentionally duplicated for some time in the
publishing-api, there should only be one access limit record per
draft content item and published content items should not have an
access limit associated with them.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160229133026) do
+ActiveRecord::Schema.define(version: 20160301120422) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/lib/data_hygiene/access_limits_cleaner.rb
+++ b/lib/data_hygiene/access_limits_cleaner.rb
@@ -1,0 +1,33 @@
+module DataHygiene
+  class AccessLimitsCleaner
+    def self.cleanup(log: STDOUT)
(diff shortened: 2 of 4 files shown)


db/schema.rb
```diff
@@ -54,6 +54,13 @@
   add_index "content_items", ["publishing_app"], name: "index_content_items_on_publishing_app", using: :btree
   add_index "content_items", ["rendering_app"], name: "index_content_items_on_rendering_app", using: :btree
 
+  create_table "content_store_payload_versions", force: :cascade do |t|
```
bradwright: Don't think this belongs here.

lib/data_hygiene/access_limits_cleaner.rb
```diff
@@ -0,0 +1,32 @@
...
+    def self.run(cleanup:, log:)
+      dupe_access_limits_by_content_item = AccessLimit
+        .joins("INNER JOIN content_items ON content_items.id = access_limits.content_item_id")
+        .group("content_items.id").having("COUNT(access_limits.id) > 1").count
+
+      dupe_access_limits_by_content_item.each do |ci|
+        dupe_access_limits = AccessLimit.where(content_item_id: ci.first).limit(ci.last - 1).order(updated_at: :asc)
+        log.puts "destroying AccessLimits: #{dupe_access_limits.map(&:id)} for content_item_id: #{ci.first}"
+        dupe_access_limits.destroy_all if cleanup
+      end
```
tuzz: Isn't the block variable `ci` an AccessLimit object?

PR #222, opened 2016-03-01
Log with request uuid

steventux:
This PR adds a basic request identifier to logging output which can be used in worker processes.
The `GdsApi::GovukHeaders.headers[:x_govuk_request_uuid]` header is populated with an `ActionDispatch::Request#uuid` and this is passed on to the `ContentStoreWorker` as an argument.

eg. in the application log:
screenshot from 2016-03-02 15 52 31

in the sidekiq log:
screenshot from 2016-03-02 15 57 12

I've assumed `GdsApi::GovukHeaders` is thread safe because https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/govuk_headers.rb#L15, this might not be the best mechanism for storing an internal identifier in the publishing-api.

I've also [logged the payload here](https://github.com/alphagov/publishing-api/commit/9471eacaf4436684ad98930c3faa0be1c8fc3a62#diff-e78d7903df6b6db2ce1780d8f618ca26R4), this might be too verbose, perhaps this could be a default to be overridden for noisier commands like `V2::PutContent`.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -167,9 +167,14 @@ module Commands
       def send_downstream(content_item)
         return unless downstream
 
+        message = "Enqueuing ContentStoreWorker job with "
+        message += "{ content_store: Adapters::DraftContentStore, content_item_id: #{content_item.id} }"
+        logger.info message
+
         PresentedContentStoreWorker.perform_async(
           content_store: Adapters::DraftContentStore,
           payload: Presenters::ContentStorePresenter.present(content_item, event),
+          request_uuid: GdsApi::GovukHeaders.headers[:x_govuk_request_uuid],
         )
       end
 
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -1,6 +1,8 @@
 module Commands
   class BaseCommand
     def self.call(payload, downstream: true)
+      logger.debug "#{self} called with payload:\n#{payload}"
+
       EventLogger.log_command(self, payload) do |event|
         new(payload, event: event, downstream: downstream).call
       end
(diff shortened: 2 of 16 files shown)


gpeng:
Is there a case for wrapping the `ContentStoreWorker` with a decorator to reduce the 'shotgun surgery' now that we've added a few common repetitive features to the enqueuing process? Within the decorator we could:

* set the `perform_in` delay
* increment the `content_item_payload_version`
* stamp the `request_guid`
* ...and enqueue the job

We're currently repeating these in all of the commands and then also testing in multiple places.

boffbowsh:
:+1: to @gpeng's comment.

tuzz:
Functionally, looks good. I also think that an abstraction that mixes in the request_uuid and adds the 1.second delay would be an improvement.

tuzz:
From Slack:

> If you are DRYing up the enqueuing of sidekiq jobs, please could you set the delay to 2.seconds instead of 1.second at the same time?
> There are a handful of requests that take longer than 1 second, but all fall under 2 seconds, so I think it would be more appropriate

steventux:
Thanks @gpeng @boffbowsh and @tuzz, doing that now.

app/commands/v2/put_content.rb
```diff
@@ -168,8 +168,12 @@ def send_downstream(content_item)
         return unless downstream
 
         ContentStorePayloadVersion.increment(content_item.id)
-        ContentStoreWorker.perform_in(
-          1.second,
+
+        message = "Enqueuing ContentStoreWorker job with "
+        message += "{ content_store: Adapters::DraftContentStore, content_item_id: #{content_item.id} }"
+        logger.info message
```
bradwright: Should this be `debug`?

steventux: It should, and I think this logging should be moved to `perform_with_defaults` to log anything going onto the queue.

PR #223, opened 2016-03-02
Add lint task with similar params to CI

steventux:
Adds a task to lint as per CI (omitting html output).
`bundle exec rake lint`
This could be added to our default task, but that's open for discussion.


benlovell:
Does this return the correct exit code upon failure?

boffbowsh:
You could follow the pattern here https://github.com/alphagov/government-frontend/pull/107

tuzz:
We're changing the workflow for linting in Jenkins, so this looks good to merge.

PR #224, opened 2016-03-03
Split format field into document_type and schema_name

danielroseman:
The format field is being replaced by two fields, see [RFC 41](https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+41%3A+Separate+document+type+from+format). document_type identifies the kind of the document for filtering and display purposes, and schema_name identifies which schema is used to validate the content. Often these will be the same, but some - in particular specialist documents - will have differing values.

The document_type/schema_name pair will eventually replace format everywhere. During the deprecation period we accept these as alternatives; on PUT, the one not supplied will be populated from the other. The index endpoint will filter by whichever is supplied.


app/models/content_item.rb
```diff
@@ -38,9 +40,24 @@ class ContentItem < ActiveRecord::Base
...
 
+  before_save :convert_format
+
 private
 
   def renderable_content?
     NON_RENDERABLE_FORMATS.exclude?(format)
   end
+
+  def format_required?
```
elliotcm: This doesn't seem to be used.

app/validators/schema_name_format_validator.rb
```diff
@@ -0,0 +1,9 @@
+class SchemaNameFormatValidator < ActiveModel::Validator
+  def validate(record)
+    if record.format.blank? && record.schema_name.blank?
+      record.errors.add(:format, 'must be supplied if schema_name is blank')
+    elsif record.schema_name.present? && record.document_type.blank?
+      record.errors.add(:document_type, 'must be supplied if schema_name is present')
+    end
```
elliotcm: The third branch of this is "what if we don't have a schema name at all" and the validation for that is currently undefined.

elliotcm:
Linkables spec is now failing.

app/validators/schema_name_format_validator.rb
elliotcm: After in-person discussion, this is fine.

PR #225, opened 2016-03-03
Update pact tests to no longer expect 409 responses

gpeng:
We are going to stop returning 409 for `payload_version`/`transmitted_at` conflicts in Content Store so this update is required to get the pacts passing.


gpeng:
Closing... Change of plan.

PR #226, opened 2016-03-03
Present in the commands and use the event.id as the content_store_payload_version

tuzz:
In order to preserve order in requests from publishing-api -> content-store we need an atomically incremented integer. An attempt was deployed yesterday using a per content item id (sent as `payload_version`) that was unsuccessful and caused updates to be rejected (with 409 conflict status).

This is another attempt to fix the problem. There are two main parts to the approach:

Firstly we've moved the presentation of the object back out of the worker so that we now queue the payload with Sidekiq rather than queuing the `ContentItem#id`. Due to the current transactional nature of the `EventLog` and the place in the workflow that the jobs are enqueued we can't seem to guarantee that the transaction is committed when the job runs. There is subsequent work to do to take the enqueuing out of the transaction but in the short term this should avoid that part of the issue.

The second part is to use the `Event#id` of the event that caused the payload to be generated and sent downstream as the `payload_version`. This is already atomically generated within the DB and as the `EventLog` is considered the 'source of truth' within the system it is a good candidate for use as the `payload_version`. This will also allow us to tie `ContentItem` records in content store to the event that generated them in Publishing API which will aid debugging.


benilovj:
Is it possible to add some context of this change to the PR description? Without it it's a bit cryptic.

spec/commands/v2/discard_draft_spec.rb
```diff
@@ -122,13 +126,16 @@
...
 
         it "sends the published content item to the draft content store" do
-          allow(ContentStoreWorker).to receive(:perform_in)
+          expect(Presenters::ContentStorePresenter).to receive(:present).with(
+            published_item,
+            instance_of(Event),
+          )
 
-          expect(ContentStoreWorker).to receive(:perform_in)
+          allow(PresentedContentStoreWorker).to receive(:perform_async)
```
benlovell: Doesn't the line below make this redundant?

tuzz: In this case, two requests are made downstream to the content store.

1) To delete the content item at the previous base path
2) To put the content item of the live content item in place of the draft

Because of this, we need to allow the `PresentedContentStoreWorker` to receive any request in addition to the specific one we're asserting against.

tuzz: Without that line:

```
expected: ({:content_store=>Adapters::DraftContentStore, :payload=>{:base_path=>"/vat-rates"}})
got: ({:content_store=>Adapters::DraftContentStore, :base_path=>"/vat-rates", :delete=>true})
```

benlovell: Ah, gotcha. That makes sense.

boffbowsh:
:+1: but will wait for a review from @elliotcm

PR #227, opened 2016-03-03
Paginate get content collection

steventux:
https://trello.com/c/OZF9rKeX/487-index-endpoint-should-support-pagination

Support pagination and add default ordering of content items index `/v2/content` endpoint.

Original PR here: https://github.com/alphagov/publishing-api/pull/211
It was easier, in light of the large refactor to the Publishing-API, to cherry-pick relevant commits to new branch. Paired with @Rosa-Fox on this story.

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -23,6 +23,8 @@ module V2
           base_path
           internal_name
         ),
+        locale: locale,
+        pagination: Pagination.new(params)
       ).call
     end
 
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -8,6 +8,7 @@ module V2
         fields: query_params.fetch(:fields),
         publishing_app: publishing_app,
         locale: query_params[:locale],
+        pagination: Pagination.new(query_params)
       ).call
     end
 
@@ -22,9 +23,7 @@ module V2
           publication_state
           base_path
           internal_name
-        ),
-        locale: locale,
-        pagination: Pagination.new(params)
+        )
       ).call
     end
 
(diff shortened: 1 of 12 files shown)


app/controllers/v2/content_items_controller.rb
```diff
@@ -23,6 +24,7 @@ def linkables
           base_path
           internal_name
         ),
+        locale: locale,
```
tuzz: Not sure how this relates?

steventux: I think this is the result of a bad rebase, will fix :+1:

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -107,6 +107,46 @@
...
+      end
+      it "is successful" do
+        expect(response.status).to eq(200)
+      end
+      it "responds with the content item as json" do
+        parsed_response_body = JSON.parse(response.body)
+        expect(parsed_response_body.first.fetch("content_id")).to eq("#{content_id}")
+      end
+    end
   end
```
tuzz: How does the controller pass order params through? (or doesn't it?)

tuzz:
Should we be concerned about the performance impact of ordering by some field? An index would help, but if the user can choose any field we might need indexes on lots of columns?

spec/controllers/v2/content_items_controller_spec.rb
steventux: https://github.com/alphagov/publishing-api/pull/227/files#diff-e23dffac69fdf54083615adffca39488R11 the controller passes a Pagination class initialized with query params to the command class.

steventux:
@tuzz: With regards to ordering, would it be acceptable to fix ordering as the current default : `public_updated_at` for the purposes of introducing pagination and then revisit custom ordering in another story or when the need arises?
I'll add an index to `public_updated_at` if it doesn't already exist.

tuzz:
Seems reasonable. :+1:

steventux:
@tuzz: I've updated with an index on `public_updated_at` and restricted ordering.

PR #229, opened 2016-03-08
Store published_at in content store for V2 content items

AndrewVos:
This is really just a hack to get the functionality we needed, and should probably
have some more tests. Just want to confirm that this is the correct approach.

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -85,9 +85,14 @@ module Commands
         end
 
         if live_content_item
+          set_last_published_at(live_content_item)
           send_to_content_store(live_content_item, Adapters::ContentStore)
           send_to_message_queue(live_content_item)
         end
+      end
+
+      def set_last_published_at(content_item)
+        content_item.update_attributes!(last_published_at: Time.zone.now)
       end
 
       def send_to_content_store(content_item, content_store)
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -41,6 +41,8 @@ module Commands
 
         clear_published_items_of_same_locale_and_base_path(content_item, translation, location)
 
+        set_last_published_at(content_item)
+
         set_public_updated_at(content_item, previously_published_item, update_type)
         State.publish(content_item)
 
@@ -87,6 +89,10 @@ module Commands
           new_item_content_id: content_item.content_id,
           state: "published", locale: translation.locale, base_path: location.base_path
         )
+      end
+
+      def set_last_published_at(content_item)
+        content_item.update_attributes!(last_published_at: Time.zone.now)
       end
 
       def set_public_updated_at(content_item, previously_published_item, update_type)
--- a/app/presenters/downstream_presenter.rb
+++ b/app/presenters/downstream_presenter.rb
@@ -15,6 +15,7 @@ module Presenters
       symbolized_attributes
         .slice(*content_item.class::TOP_LEVEL_FIELDS)
         .merge(public_updated_at)
+        .merge(last_published_at)
         .merge(links)
         .merge(access_limited)
         .merge(content_store_payload_version)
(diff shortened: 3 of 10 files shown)


elliotcm:
Yes this seems a reasonable approach, though you might want to use `last_published_at` to differentiate from the concept of "this is when it first went live".

gpeng:
Does it need another date field? Doesn't updated at already do this, assuming you touch the content item during the publish.

AndrewVos:
@gpeng No updated_at doesn't do anything because the content isn't really touched during publish. I'm not explaining this well, but long story short updated_at means the last time put_content was called. In our UI this amounts to updated_at being equal to the last time the user saved a document, not the last time they clicked Publish. We need the latter displayed in our frontend.

AndrewVos:
@elliotcm Renamed to `last_published_at`.

gpeng:
Ok... There are quite a few dates building up around `ContentItem` that I find a bit confusing. I see you need a date that represents the last time the content item as a whole was updated (I mean, including its `State` in this case). I think there's an argument for touching the `ContentItem` when any of the associated tables ('the pillars' etc) are updated. In words, publishing a content item changes its state. I think `updated_at` should reflect that despite the fact that the state isn't stored in the same record in the DB. Obviously, if your UI needs a date the `ContentItem` itself was last saved and the date that it was published then overloading that date is no good to you, I realise that. The `State` object has timestamps too doesn't it? Would that be any good?

AndrewVos:
The `State` object does, apparently, store these dates. This approach just seemed better than relying on some other object. A lot more obvious. I discuessed all this with @tuzz originally.

tuzz:
The `updated_at` on the state should be the last time the state was updated. I don't think we should touch it when some other arbitrary field changes on the content item. That would seem odd?

This approach looks long the right lines to me. :+1:

gpeng:
@tuzz I actually meant the other way around. Touching the `ContentItem` when the `State` was updated.

boffbowsh:
But from what @gpeng is saying, couldn't `last_published_at` in the `DownstreamPresenter` simply be something like:

```ruby
state.name == "published" ? state.updated_at : nil
``` 
 
Or am I confused (again) by the data model?

elliotcm:
@boffbowsh You could indeed do that.  I think I suggested it to someone yesterday but that might've been for something different.

elliotcm:
@boffbowsh however, you'd have to find the published content item if you're dealing with a draft at the time.

elliotcm:
@gpeng He doesn't need a date to represent the last time the content item was updated, he needs a date to represent the last time it was sent to the live content store.

tuzz:
> @tuzz I actually meant the other way around. Touching the ContentItem when the State was updated.

I'm also not sure that's the right thing to do. I'd expected the updated_at on ContentItem to be the last time the content of that ContentItem was changed. For example, when a ContentItem gets "superseded" because something else has published over the top of it, I wouldn't expect that to affect its timestamps because only the state has changed. Maybe I'm missing something, though? Happy to discuss.

elliotcm:
@AndrewVos Further to the chat above, this model might not be the best place to keep this timestamp (but maybe it is), could you pop over when you have time and we'll discuss again?

AndrewVos:
@elliotcm working from home today, I'll catch you on slack...

gpeng:
@tuzz Let's discuss that away from this PR as it's probably irrelevant here.

@elliotcm The `State#updated_at` would be the date it was sent wouldn't it, assuming `State#name` is 'published'?

elliotcm:
@gpeng it would, though it would require you finding the right object.  We'll have a chat with Andrew later and figure out the specifics.

gpeng:
@elliotcm Don't you already have the `ContentItem` passed in? ([here](https://github.com/alphagov/publishing-api/pull/229/files#diff-a8f5f38617f3679840b8740e6627ab92R47)).

elliotcm:
@gpeng yes, but we're talking about providing this information in a GET request.

gpeng:
🤔 Ok... not sure I understand but I'd like to if you'd fill me in (in the verbal rather than violent sense 🤓) when you've done the discussing. 👍

AndrewVos:
@thehenster I think that's there for when the item is in draft. Also, I've written this code to only set the value in the V2 api requests, so in pre-V2 I expect the value to not be set.

AndrewVos:
Closing this because after speaking to @elliotcm we decided this isn't needed: https://github.com/alphagov/government-frontend/pull/120

PR #230, opened 2016-03-08
Add /lookup-by-base-path endpoint

tijmenb:
This endpoint can be used by clients to fetch the `content_id` of one or more URLs on GOV.UK.

I've thought about making a `/content/?base_path=/foo` endpoint with the same result as `/content/:id`, but that seemed contrary to the spirit of referring to everything by `content_id`.

It will be used by:

- [content-tagger](https://github.com/alphagov/content-tagger/blob/2267c396e5c854aa8e398ce9be311e68b953fc63/app/forms/content_lookup_form.rb) to look up the content_id of a URL the user wants to tag.
- [content-tagger](https://github.com/alphagov/content-tagger/blob/4ccb14faf1bbcbfb1d7f3df1d227732b3b0169ac/lib/alpha_taxonomy/taxon_linker.rb#L76) to look up the `content_id` of items from a bulk-tagging spreadsheet.
- publisher during migration to fetch the `content_id` of tagged [topics](https://github.com/alphagov/publisher/blob/master/app/presenters/published_edition_presenter.rb#L24-L28) to send to the publishing-api.
- whitehall during migration to look up tags (see https://github.com/alphagov/whitehall/pull/2494/files#r55371472)

Also, I'm pretty sure this code is wrong, since it will return the `content_id` for all content-items that have existed at a certain `base_path`.

WDYT @danielroseman @elliotcm @tuzz?

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,0 +1,11 @@
+class LookupsController < ApplicationController
+  def by_base_path
+    base_paths_and_content_ids = ContentItemFilter
+      .filter(state: "published", base_path: params.fetch(:base_paths))
+      .pluck(:base_path, :content_id)
+      .uniq
+
+    response = Hash[base_paths_and_content_ids]
+    render json: response
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,8 @@ Rails.application.routes.draw do
 
     put "/paths(/*base_path)", to: "path_reservations#reserve_path"
 
+    post '/lookup-by-base-path', to: 'lookups#by_base_path'
+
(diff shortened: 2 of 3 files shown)


elliotcm:
Hullo @tijmenb.  You're on the right track IMO.  Wondering if we could think about the name, in a way that indicates we're looking up by location.

tijmenb:
I've scoped the lookup to published items, because I figured that there can be items with the same  `base_path`, but a different `content_id`.

@elliotcm `/lookup-by-base-path` or something like that?

elliotcm:
@tijmenb Sounds good to me.  You're going to want to also scope by locale.

tijmenb:
@elliotcm PR updated with new URL

PR #231, opened 2016-03-09
Remove old tables and columns

tuzz:
This cleans up old tables and columns that are left over since we moved the the new object model.

https://trello.com/c/5NOfjCPD/633-remove-old-content-item-tables-and-deprecated-columns

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,58 +11,21 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160307112043) do
+ActiveRecord::Schema.define(version: 20160309155913) do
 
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -11,42 +11,35 @@
...
   create_table "access_limits", force: :cascade do |t|
-    t.integer  "target_id"
-    t.string   "target_type"
     t.json     "users",           default: [], null: false
     t.datetime "created_at",                   null: false
     t.datetime "updated_at",                   null: false
     t.integer  "content_item_id"
   end
 
-  add_index "access_limits", ["target_type", "target_id"], name: "index_access_limits_on_target", using: :btree
```
elliotcm: This kinda highlights that we haven't replaced this index.

PR #233, opened 2016-03-10
Add provider state for lookup tests

tijmenb:
This will be used in gds-api-adapters to test the /lookup-by-base-path endpoint.

https://github.com/alphagov/publishing-api/pull/230


spec/service_consumers/pact_helper.rb
```diff
@@ -344,4 +344,20 @@
       stub_request(:put, Regexp.new('\A' + Regexp.escape(Plek.find("draft-content-store")) + "/content"))
     end
   end
+
+  provider_state "there are content items with base_paths /foo and /bar" do
```
danielroseman: Probably worth explicitly stating "there are live content items".

tijmenb:
@danielroseman done!

PR #234, opened 2016-03-11
Fix an order-dependent test

tuzz:
Both of these content item have the same `public_updated_at` and so it's not specified which order they'd come back in. We could set a `public_updated_at` to break the tie, but that's not really what this scenario is testing, so this seems more reasonable.


tuzz:
This is going to be taken care of as part of the pagination work. Closing.

PR #235, opened 2016-03-11
Fix pagination

dougdroper:
Moving the grouping into an SQL query and pushing the pagination down

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -12,7 +12,7 @@ module Presenters
         present_many(content_items).first
       end
 
-      def self.present_many(content_item_scope, fields: nil, order: { public_updated_at: :desc }, offset: 0, limit: 50)
+      def self.present_many(content_item_scope, fields: nil, order: { public_updated_at: :desc }, offset: 0, limit: nil)
         presenter = new(content_item_scope, fields)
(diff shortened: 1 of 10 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -24,20 +26,43 @@ def initialize(content_item_scope, fields = nil)
...
+        ActiveRecord::Base.connection.execute(aggregated_sql(scope.to_sql))
+      end
+
+      def aggregated_sql(sql)
+        <<-END.strip_heredoc
+          SELECT json_agg(json_rows) FROM (
+            SELECT row_to_json(item) json_item FROM (#{sql}) item
+          ) json_rows
+          GROUP BY json_item->>'content_id', json_item->>'locale'
+          ORDER BY max(json_item->>'public_updated_at') DESC
```
danielroseman: Since `order` is a parameter we should interpolate it here rather than hard-coding.

danielroseman:
This is also affecting the linkables endpoint - although that does not pass in the page or per_page params, it is still being limited to 50 items only. That endpoint always needs to return all items.

dougdroper:
@danielroseman, sure, will sort that out. :+1:

dougdroper:
@danielroseman the linkables endpoint is now no longer limited.

PR #238, opened 2016-03-15
Add an index on document_type

boffbowsh:
The `format` column had an index, but it hasn’t been created for the
new `document_type` index. This is being filtered on by the linkables
endpoint now used by Whitehall, but is slow to query for policies.

Explain before: http://explain.depesz.com/s/HVYa
Explain after: http://explain.depesz.com/s/9XY


boffbowsh:
This has been manually added in production, but the `schema_migration` row has not yet been added.

mattbostock:
:+1: for intent.

boffbowsh:
This migration has now been faked, or marked as run in production.

PR #239, opened 2016-03-16
Filter by the given link type

tuzz:
We were missing a restriction on `link_type`. It was being plumbed through but wasn't influencing the query.


mobaig:
LGTM :+1:

PR #240, opened 2016-03-16
Wrap results to GET /v2/content in pagination meta data

dougdroper:
Includes pagination and HATEOAS

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -12,43 +12,62 @@ module Presenters
         present_many(content_items).first
       end
 
-      def self.present_many(content_item_scope, fields: nil, order: { public_updated_at: :desc }, offset: 0, limit: nil)
-        presenter = new(content_item_scope, fields)
-        presenter.order = order
-        presenter.limit = limit
-        presenter.offset = offset
+      def self.present_many(content_item_scope, fields: nil, order: { public_updated_at: :desc }, offset: 0, limit: nil, locale: nil)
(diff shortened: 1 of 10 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -12,43 +12,62 @@ def self.present(content_item)
...
-      attr_accessor :content_item_scope, :fields
+      attr_accessor :content_item_scope, :fields, :locale
 
-      def groups
-        scope = join_supporting_objects(content_item_scope)
-        scope = select_fields(scope).order(order)
+      def total_query
+        sql = content_item_scope.select(:content_id)
+        sql = sql.group(:content_id) unless locale == "all"
+        "SELECT COUNT(*) FROM (#{sql.to_sql}) total"
```
danielroseman: Some local testing shows that this is pretty slow on document_types with a decent amount of data, eg cma_cases, but can be massively speeded up (70ms vs 1700ms) by adding an index to translations on (content_item_id, locale). Can you add a migration?

app/presenters/results_presenter.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def self_link
+      {
+        href: page_href(:+, 0),
+        rel: "self"
+      }
+    end
+
+    def page_href(operator, offset = 1)
+      request_url.gsub(/&page=(\d*)/, "") << "&page=#{current_page.send(operator, offset)}"
```
danielroseman: This is a bit magic. Can we not just use a negative offset to get the previous page?

dougdroper: Yes, good thinking :+1:

PR #241, opened 2016-03-17
Fix a race condition in PutContent

tuzz:
Someone had double clicked on the ‘Save & Publish’
button in travel advice publisher. This sent two
requests simultaneously to the Publishing API
which raced against each other.

This change uses row-level pessimistic locking,
which causes the second request to wait for a
mutex before it can proceed. This has the effect
that the second request is processed after the
first completes and we maintain data integrity.

The tests for this change use the truncation
strategy to clean the database instead of the
default transaction strategy. This is because the
threads use separate connections from the
connection pool which cannot be collectively
wrapped in a single transaction.

### DIFF
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -117,6 +117,26 @@ RSpec.describe Commands::V2::PutContent do
         expect(State.find_by!(content_item: content_item).name).to eq("draft")
         expect(UserFacingVersion.find_by!(content_item: content_item).number).to eq(6)
       end
+
+      describe "race condtitions", skip_cleaning: true do
+        after do
+          DatabaseCleaner.clean_with :truncation
+        end
+
+        it "copes with race conditions" do
+          described_class.call(payload)
+          Commands::V2::Publish.call(content_id: content_id, update_type: "minor")
+
+          expect {
+            thread1 = Thread.new { described_class.call(payload) }
+            thread2 = Thread.new { described_class.call(payload) }
+            thread1.join
+            thread2.join
+          }.to raise_error(CommandError, /conflicts with a duplicate/)
+
+          expect(State.all.pluck(:name)).to eq %w(superseded published draft)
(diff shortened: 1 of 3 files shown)


spec/commands/v2/put_content_spec.rb
```diff
@@ -117,6 +117,29 @@
         expect(State.find_by!(content_item: content_item).name).to eq("draft")
         expect(UserFacingVersion.find_by!(content_item: content_item).number).to eq(6)
       end
+
+      describe "race condtitions", skip_cleaning: true do
+        after do
+          DatabaseCleaner.clean_with :truncation
```
boffbowsh: Why is explicit cleaning needed?

spec/commands/v2/put_content_spec.rb
```diff
@@ -117,6 +117,29 @@
...
+        end
+
+        it "copes with race conditions" do
+          described_class.call(payload)
+          Commands::V2::Publish.call(content_id: content_id, update_type: "minor")
+
+          begin
```
boffbowsh: This could be:
```ruby
expect {
  thread1 = Thread.new { described_class.call(payload) }
  thread2 = Thread.new { described_class.call(payload) } 
  thread1.join
  thread2.join
}.to raise_error ExceptionClass
```

boffbowsh: In fact it should be, as the rescue doesn't assert that an error is raised.

spec/commands/v2/put_content_spec.rb
tuzz: I've amended the commit message to explain this:

The tests for this change use the truncation
strategy to clean the database instead of the
default transaction strategy. This is because the
threads use separate connections from the
connection pool which cannot be collectively
wrapped in a single transaction.

tuzz:
@boffbowsh thanks! I've amended the test to use an `expect { }` block.

PR #242, opened 2016-03-17
Filter GET index results by organisation

steventux:
https://trello.com/c/5n27561D/449-allow-the-get-index-endpoint-to-be-filtered-by-organisation

Makes the index endpoint for content items `/v2/content/` support an `organisation` parameter. This should be an organisation's content id. This will filter results to content items linked to the specified organisation.
The organisation is already associated with a content item via the `Link` and `LinkSet` models where `Link#link_type` is _organisations_.

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -7,8 +7,7 @@ module V2
       results = Queries::GetContentCollection.new(
         document_type: doc_type,
         fields: query_params.fetch(:fields),
-        publishing_app: publishing_app,
-        locale: query_params[:locale],
+        filters: filters,
         pagination: pagination
       )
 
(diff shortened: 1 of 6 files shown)


dougdroper:
LGTM :+1:

tuzz:
I think we should try and prevent knowledge of the different link_types leaking into Publishing API. There's nothing particularly special about `organisation` with regards to filtering, so I wonder if we could instead break this into two parameters:

`link_type=organisations&link_content_id=some-organisation-content-id`

What do you think?

rboulton:
I'd suggest copying the pattern used in the search API: you can filter on any field by using the parameter `filter_FIELDNAME=VALUE", where FIELDNAME is the field name, and VALUE is the value.

This would translate to something like `link_organisations=organisation-content-id`, or you could name it `has_link_organisations` or `is_link_organisations`.  Or even `filter_link_organisations=organisations-content-id`.

This has the advantage that it supports filtering on multiple fields, or multiple values in a field (you don't have to implement that now, but it would be easier to extend in future if that became needed)

steventux:
@tuzz, @rboulton thanks both for helpful comments, it's a good point that we should be able to filter on any link type, and extending this to a common convention as Richard describes seems a good option for future filtering and perhaps even moving `publishing_app` and `locale` filters into this pattern.

steventux:
@tuzz I've updated as per the discussion above, not 100% on the param naming. I've not gone as far as to use the convention on other filtering params yet, happy to do so if it makes sense in this PR.

PR #243, opened 2016-03-17
Write up the dependency resolution spike (part 2)

tuzz:
Here's my lengthy write up of the spike into the technical approach for dependency resolution.

Happy to go through it on a whiteboard, too.

### DIFF
--- a/doc/dependency-resolution.md
+++ b/doc/dependency-resolution.md
@@ -272,6 +272,36 @@ or similar and we can send it alongside the unexpanded links. As we migrate
 applications to use dependency resolution, we should make them use this field
 instead of "links" and look at retiring the expansion of links in content store.
 
+Here is a rough example of what this **might** look like:
+
+```javascript
+{
+  "expanded_links": {
+    "parent": [
+      {
+        "content_id": "111-111",
+        "title": "VAT Rates",
+        "base_path": "/tax/vat-rates",
+        "parent": [
+          {
+            "content_id": "222-222",
+            "title": "Taxation",
+            "base_path": "/tax",
+            "parent": [
+              {
+                "content_id": "333-333",
+                "title": "Home",
+                "base_path": "/"
+              }
+            ]
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
 **Bidirectional Dependencies**
 
 As mentioned, we have a requirement for bidirectional dependencies, in the case
--- a/doc/dependency-resolution.md
+++ b/doc/dependency-resolution.md
@@ -34,8 +34,8 @@ can change independently and its title should be updated on the collection page.
 
 Currently, publishing applications manually speak to Rummager which indexes
 their content for search. We'd like to centralise this behaviour so that the
-Publishing API forwards this content to Rummager in its expanded form for
-search indexing.
+Publishing API forwards this content in its expanded form onto a queue which is
+then consumed by Rummager.
 
 Note: In addition to these requirements, there are some non-functional
 requirements detailed below. This includes a architectural goal of moving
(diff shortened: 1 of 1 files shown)


doc/dependency-resolution.md
```diff
@@ -0,0 +1,98 @@
+## Dependency Resolution
+
+**Problem Statement**
```
danielroseman: There's a much more basic one too, which is that all the links in the frontend metadata block of an item - eg organisation, policies, countries etc - are calculated at request time, which is unnecessary; we'd like to do it at publish time, but we need the link text to update automatically when the item's title itself changes.

tuzz: I've mentioned that a bit in 'Tracking Dependencies' further down. I'll look at moving it up to make it more prominent.

tuzz: I've added a 'Note' at the bottom of this section. I hope that's ok.

doc/dependency-resolution.md
```diff
@@ -0,0 +1,319 @@
...
+updated as a result of the incoming request. It will do this using Postgres's
+WITH RECURSIVE functionality. This was explored in the following branch to
+demonstrate that it is feasible to detect dependent nodes in this way:
+
+[https://github.com/alphagov/publishing-api/tree/dependency-transitive-closure](https://github.com/alphagov/publishing-api/tree/dependency-transitive-closure)
+
+We should make it so that this query only follows edges in the dependency graph
+for groups which contain any of the fields that have changed on the content
+item. We can determine which fields have changed using Rails's
+[ActiveModel::Dirty](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html)
```
boffbowsh: This will only track individual ActiveModel fields, it won't be able to tell that e.g. `details.start_date` has changed but not `details.end_date`.

boffbowsh:
Worth clarifying that rather than
> Publishing API forwards this content to Rummager in its expanded form
the Publishing API should publish the resolved content item to some sort of fanout system for apps to subscribe to.

doc/dependency-resolution.md
```diff
@@ -0,0 +1,319 @@
...
+
+- Link expansion
+
+It is proposed that we move link expansion into the publishing api and send the
+fully-pressed content item downstream to the content store. For each of the link
+groups that appear on the content item, it will need to look up the rules for
+those groups from the schemas and expand those fields from the dependent content
+items.
+
+We also need to consider a new data structure to represent recursively expanded
```
boffbowsh: Some rough examples would be useful here to illustrate this.

doc/dependency-resolution.md
tuzz: That's a good point. I'll remove the reference to a specific technique and leave it open, for now.

doc/dependency-resolution.md
tuzz: Added a rough example.

We're going to design this structure relatively early in the development of dependency resolution, so will be able to provide a more detailed description of this once we start development.

boffbowsh:
:+1:

doc/dependency-resolution.md
```diff
@@ -0,0 +1,347 @@
...
+
+**Summary**
+
+In all of these cases, there is a dependency between content items. In the first
+case, the name of a page might change or the hierarchy might change. This would
+mean that we have to update the breadcrumbs for many pages to reflect this.
+
+In the second case, the title or path of a related link could change which would
+require an update to any content item that linked to the changed item. This is
+very similar in the third and fourht cases where embedded links would need to be
```
dougdroper: just a small type in fourht, but no biggie

tuzz: will fix

PR #252, opened 2016-03-24
PATCH /v2/links/:content_id for an absent content item returns a 404

thehenster:
Updating the links for a content item that doesn't exist is currently returning a 200.

Before I fix the large-ish quantity of tests this patch has broken I wanted to check that this is indeed a problem and that this is the right approach to fixing it?

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -2,6 +2,7 @@ module Commands
   module V2
     class PatchLinkSet < BaseCommand
       def call
+        raise_unless_content_item_exists
         raise_unless_links_hash_is_provided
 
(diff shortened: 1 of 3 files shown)


dougdroper:
Would this be in a case where a content_item has been deleted? Is that a thing that happens? 

But definitely agree that we should not be returning a 200 if that's that case, seems like a reasonable approach to me.

thehenster:
We discovered it by some good ole fat-fingering in our publisher. But yeah, if deleting drafts actually deletes the content item rather than changing state (_shrug_) then I guess this would be relevant for that too.

tuzz:
@thehenster this is intended behaviour. Jump into Slack or drop by and we can go through it as that might be easiest.

PR #253, opened 2016-03-24
Assert creating a link set without a content item returns success

thehenster:
Tuzz has explained why this works the way it does. This test is just supporting existing behaviour. I couldn't find it when trying to work out if this was indeed intended behaviour. It may/probably does already exist somewhere.

Please feel free to close if it already exists or isn't suitable!


tuzz:
Thanks :+1:

PR #254, opened 2016-03-30
Determine if there should be a dependency resolution lookup

dougdroper:
Based off a list of changed fields, determine if a content_item has
those fields changed which requires a content resolution lookup

https://trello.com/c/CjQNKShl/644-3-calculate-the-set-of-content-items-that-need-to-be-updated

@tuzz could you have a look please

### DIFF
--- a/app/queries/content_dependents.rb
+++ b/app/queries/content_dependents.rb
@@ -1,0 +1,42 @@
+module Queries
+  class ContentDependents
+    def initialize(content_id:, fields: [], link_sets: LinkSets.new, dependent_lookup: GetDependents.new)
+      @content_id = content_id
+      @fields = fields
+      @link_sets = link_sets
+      @dependent_lookup = dependent_lookup
+    end
+
+    def call
+      return [] unless dependents?
+      recursive_link_types, direct_link_types = link_types
+      dependent_lookup.call(
+        content_id: content_id,
+        recursive_link_types: Hash[recursive_link_types].keys,
+        direct_link_types: Hash[direct_link_types].keys
+      )
+    end
+
+  private
+
+    attr_reader :content_id, :fields, :dependent_lookup
+
+    def dependents?
+      (fields & fields_that_require_dependent_lookup).any?
+    end
+
+    def fields_that_require_dependent_lookup
+      link_sets.values.flat_map { |link| link[:expanded_links] }.uniq
+    end
+
+    def link_types
+      link_sets
+        .select { |_type, set| (fields & set[:expanded_links]).any? }
+        .partition { |_type, set| set[:recurse] }
+    end
+
+    def link_sets
+      @link_sets.call
+    end
+  end
+end
--- a/app/queries/get_dependents.rb
+++ b/app/queries/get_dependents.rb
@@ -1,0 +1,31 @@
+module Queries
+  module GetDependents
+    # Finds all content_ids that depend on the given content_id. This query is
+    # recursive and will also return content_ids that aren't direct neighbours
+    # to the given content_id. This is called a 'transitive closure'.
+    #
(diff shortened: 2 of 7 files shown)


app/queries/get_dependents.rb
```diff
@@ -0,0 +1,37 @@
...
+    # recursive and will also return content_ids that aren't direct neighbours
+    # to the given content_id. This is called a 'transitive closure'.
+    #
+    # If the optional link_type is given, only links that have that link_type
+    # will be followed when computing dependencies.
+    def call(content_id, link_type = nil)
+      results(content_id, link_type).map do |result|
+        result.fetch("content_id")
+      end
+    end
```
tuzz: I think we can achieve the same results by SELECTing multiple link types in the query itself. One query would be better than one for each link type if possible.

app/queries/content_dependents.rb
```diff
@@ -0,0 +1,50 @@
+module Queries
+  class ContentDependents
+    def initialize(content_id:, fields: [], link_sets: nil, dependent_lookup: Queries::GetDependents.new)
+      @content_id = content_id
+      @fields = fields
+      @dependent_lookup = dependent_lookup
```
tuzz: Do we need to inject the GetDependents query here?

app/queries/content_dependents.rb
```diff
@@ -0,0 +1,50 @@
...
+      link_sets.values.flatten.uniq
+    end
+
+    def link_types
+      link_sets.select(&method(:changed_fields_include?)).keys
+    end
+
+    def changed_fields_include?(_, link_set_field)
+      fields.any? { |field| link_set_field.include?(field) }
+    end
```
tuzz: `link_set_field` here is actually an array of fields, so it might help to name it as a plural.

I haven't seen many people using the method-proc syntax here (`&method`). I quite like it, but it might be unfamiliar to a lot of people.

I think you might be able to simplify this a bit by using the intersect operator `&`, e.g.

```ruby
def link_types
  link_sets.select { |_, expansion_fields| (fields & expansion_fields).any? }.keys
end
```

(This is untested.)

tuzz:
Looks good – minor comments.

One last thing that we need to account for is whether to recurse or not. Each link_type will have a boolean attribute that specifies the "recursive behaviour" for the query. The data structure will be something like:

```
{
  parent: {
    expansion_fields: [:title, :base_path],
    recurse: true,
  },
  attachment: {
    expansion_fields: [:title, :description, :base_path],
    recurse: false,
  }
}
```

This relates to what @elliotcm was looking at earlier.

That means we need to plumb this boolean into the query object and use it to change its behaviour – probably by running a much simpler query that just checks immediate neighbours – i.e. it joins to the link_sets just once.

Let me know if you have any questions or need help with this.

spec/queries/get_dependents_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    expect(subject.call(content_id: c, recursive_link_types: ["parent", "children"])).to match_array [a, b]
+  end
+
+  it "finds direct links only" do
+    create_edge(a, c, "parent")
+    create_edge(b, a, "parent")
+
+    expect(subject.call(content_id: c, direct_link_types: ["parent"])).to match_array [a]
+  end
```
tuzz: :+1: this confirms the direct query is directed correctly.

tuzz:
Looks ok. :+1: 

Do we need to be careful of SQL injection if we were to expose the query via a HTTP interface? i.e.

> Exposure via an Endpoint

> We might also consider exposing dependency resolution queries via an endpoint on the Publishing API. We currently support simple cases of this in the form of /links and /linked and it might make sense to collapse these into a single endpoint that leverages the dependency resolution queries to be implemented as part of this proposal.

PR #259, opened 2016-04-06
Expand the related links following the LinkSetSchema

dougdroper:
Links are currently expanded in the ContentStore. This moves the
functionality over to the PublishingApi

https://trello.com/c/Ahj9lNB0/645-4-update-content-items-whose-dependencies-have-changed


app/queries/link_set_schema.rb
```diff
@@ -0,0 +1,55 @@
...
+        :base_path,
+        :content_id,
+        :description,
+        :locale,
+        :title,
+        :web_url
+      ]
+    end
+  end
+end
```
tuzz: I'm a bit confused about this class. Doesn't [this](https://github.com/alphagov/publishing-api/blob/master/app/queries/link_sets.rb) serve the same purpose?

The link expansion configuration should be shared between the class that computes which nodes need to be updated and the the presenter that renders the expanded links.

If we need to account for default behaviour, we should keep all of that in the same place if possible.

app/presenters/queries/expanded_link_set.rb
```diff
@@ -0,0 +1,41 @@
...
+        @schema = schema
+      end
+
+      def links
+        link_types.each_with_object({}) do |type, hash|
+          hash[type] ||= {}
+          expanded_links(type).map do |field|
+            hash[type][field] ||= {}
+            hash[type][field] =
+              web_content_item.send(field)
```
tuzz: I'm a bit weary of `send` here. Is there a non-send way of doing this? If not, could we use `public_send` instead.

app/queries/link_set_schema.rb
dougdroper: yes this is the same thing. This is just a rename of that file, but both use this to compute and to update.
It may be a bit too far with the reducing the repetition since I am still not certain what the file will look like in the end.

Also this is arbitrarily split, I am not sure how we will decide which link_types will be recursive or not?

tuzz: Edit: Oh, I see. That class was superseded with this new one.

tuzz:
As discussed, the next steps are:

1) To get it to render the recursive link structure if that link type's configuration has `recurse: true`

2) To "reign in" the `LinkSetSchema` so that it can cope with link_types it's never seen before and use default expansion rules for this. We should only need to capture rules for things with bespoke rules.

:+1:

app/queries/link_set_schema.rb
```diff
@@ -0,0 +1,57 @@
+module Queries
+  class LinkSetSchema
+    def initialize(content_ids: [])
+      @content_ids = content_ids
+    end
+
+    def links
+      types.keys.each_with_object({}) do |type, hash|
+        hash[type] = {
+          expanded_links: expand_links(type),
```
dougdroper: fields_to_expand

app/queries/link_set_schema.rb
```diff
@@ -0,0 +1,57 @@
...
+          recurse: recursive_link_types.include?(type)
+        }
+      end
+    end
+
+    def types
```
dougdroper: types_with_ids

app/presenters/queries/expanded_link_set.rb
```diff
@@ -0,0 +1,91 @@
...
+        @schema_links ||= schema.links
+      end
+
+      def web_content_item(id)
+        @web_content_item ||= {}
+        @web_content_item[id] ||=
+          ::WebContentItem.new(content_item(id))
+      end
+
+      def web_item_valid?(web_item)
```
dougdroper: Double check that some fields can be expanded

app/commands/v2/discard_draft.rb
```diff
@@ -47,7 +47,7 @@ def delete_draft_from_draft_content_store(draft_path)
       def send_live_to_draft_content_store(live)
         PresentedContentStoreWorker.perform_async(
           content_store: Adapters::DraftContentStore,
-          payload: Presenters::ContentStorePresenter.present(live, event),
+          payload: Presenters::ContentStorePresenter.present(live, event, [:draft, :published]),
```
elliotcm: I'm not sure I follow this, does it send both the live and the published item?  Is this the fallback order?

dougdroper: Yeah this is the fallback order: there is a comment copied here:

We need to support an array of states to cater for the DiscardDraft
command which deletes the draft content item and sends the published item
to the draft content store. This means that we need to try to find a
draft, but fall back to the published item (if it exists).

elliotcm: OK, that makes sense.  Could we make it a named parameter so it's clearer at call time?

dougdroper: updated

PR #260, opened 2016-04-07
Remove redundant curriculum policy.

elliotcm:
Removes /government/policies/schools-and-college-qualifications-and-curriculum
content_id 3c04de88-9e4a-4ebb-bdc9-ef5946db17b9.

https://govuk.zendesk.com/agent/tickets/1299583

Background:
Back in the election times we seem to have created two of these with slightly different paths, but the "wrong" one is no longer in Policy Publisher.  It should be removed from the publishing API so it stops showing up in dropdowns.

The correct one is here: https://www.gov.uk/api/content/government/policies/school-and-college-qualifications-and-curriculum

### DIFF
--- a/db/migrate/20160407121458_remove_redundant_policy.rb
+++ b/db/migrate/20160407121458_remove_redundant_policy.rb
@@ -1,0 +1,16 @@
+class RemoveRedundantPolicy < ActiveRecord::Migration
+  # Remove /government/policies/schools-and-college-qualifications-and-curriculum
+  def up
+    content_items = ContentItem.where(content_id: "3c04de88-9e4a-4ebb-bdc9-ef5946db17b9")
+    Translation.where(content_item: content_items).destroy_all
+    Location.where(content_item: content_items).destroy_all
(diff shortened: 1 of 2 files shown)


db/migrate/20160407121458_remove_redundant_policy.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveRedundantPolicy < ActiveRecord::Migration
+  # Remove /government/policies/schools-and-college-qualifications-and-curriculum
+  def up
+    content_items = ContentItem.where(content_id: "3c04de88-9e4a-4ebb-bdc9-ef5946db17b9")
```
dougdroper: Might be a good idea to wrap this in a transaction in case something goes wrong half way through, or are migrations automatically wrapped? 

Also doesn't look like it removes the content_items?

elliotcm: Migrations are automatically wrapped in a transaction, but you're right about the content items.  I've updated the code.

PR #264, opened 2016-04-11
Fix two workflow problems with redirects

tuzz:
https://trello.com/c/MhxVMalt/678-process-withdrawals-before-supersedes-when-publishing-content

1) Draft redirects were not being created if there was only
    a published content item when the PutContent command was
    called. In this case, we should look up the previously
    published item and compare its base path to the item that
    has been put into the system.

Note: This should fix this class of errors: https://errbit.publishing.service.gov.uk/apps/552f95820da11577b4000032/problems/56f1cd6865786320a96c0600

2) Previously published items should be withdrawn rather
    than superseded if there is a choice between the two. This
    happens when an item changes its base path and the item is
    then published. Depending on which code runs first in the
    Publish command, the previously published item could either
    be superseded or withdrawn. This can cause conflicts later
    when the item is subsequently redrafted and published.

We'll probably need to fix the existing data, somehow.

I'll have a think about how to do this and create a ticket.

cc @elliotcm @steventux @MatMoore @davidbasalla


MatMoore:
:tada: Thanks for taking a look at this.

Would it be worth adding some examples to [publishing-application-examples.md](https://github.com/alphagov/publishing-api/blob/bd4debc81642522debe9ea5ec81a6667687ea3f6/doc/publishing-application-examples.md) or somewhere else to cover use cases where the content is renamed? It would be good to explicit about what the api is doing behind the scenes when we put content or publish, particularly in cases like this one where a request affects multiple content items.

tuzz:
Good idea @MatMoore – will do.

tuzz:
@MatMoore Added: https://github.com/alphagov/publishing-api/blob/fix-workflow-issue/doc/publishing-application-examples.md#automatic-redirects

MatMoore:
Awesome :+1: 

Very happy for this PR to be merged, but leaving it open in case anyone from the publishing platform wants to review first.

PR #267, opened 2016-04-12
[DISCUSS] Present content items with nil base_path downstream

steventux:
Following on from https://trello.com/c/DuueB1rh/686-prevent-substitution-of-content-for-put-content-requests-with-nil-base-path

It should be possible to put content with a nil base_path, these requests are persisted in the publishing-api so they should also be presentable to the content-store.
This change amends the `DownstreamPresenter` to allow this.

I'm not 100% on what the functionality should be for content with no base path going into the content store, so please comment away.
This work came from attempting to put content into the publishing api with a nil base_path, this is no longer destructive as of https://github.com/alphagov/publishing-api/pull/266, but still not fully working.

### DIFF
--- a/app/presenters/downstream_presenter.rb
+++ b/app/presenters/downstream_presenter.rb
@@ -59,7 +59,8 @@ module Presenters
     end
 
     def location
-      @location ||= Location.find_by!(content_item: content_item)
+      @location ||=
+        Location.find_by(content_item: content_item) || NullLocation.new
     end
 
     def translation
--- a/app/null_location.rb
+++ b/app/null_location.rb
@@ -1,0 +1,3 @@
+class NullLocation
+  attr_reader :base_path
+end
--- a/spec/requests/content_item_requests/downstream_requests_spec.rb
+++ b/spec/requests/content_item_requests/downstream_requests_spec.rb
@@ -111,6 +111,32 @@ RSpec.describe "Downstream requests", type: :request do
         expect(response).to be_ok, response.body
       end
     end
+
+    context "with a nil base_path" do
+      let(:content_item_payload) {
+        v2_content_item
+          .except(:format)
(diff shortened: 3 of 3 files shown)


app/presenters/downstream_presenter.rb
```diff
@@ -59,7 +59,7 @@ def access_limit
     end
 
     def location
-      @location ||= Location.find_by!(content_item: content_item)
+      @location ||= Location.find_by(content_item: content_item)
```
dougdroper: This may be overkill, but it seems like location without a base path is now part of the domain, so having something like:
`Location.find_by(content_item: content_item) || NoBasePathLocation.new` (bad name, maybe `NilLocation` or `GovernmentLocation` or whatever) and that guy has a method on it called `base_path` which is empty? It avoids that try down there :)

steventux: :+1: for less conditional snakey code

tuzz:
I think the plan is to prevent content with nil base paths going downstream to the content store. We can still refer to this content via other content's links. Once dependency resolution is done, we'll be able to access those content items' fields implicitly via the `expanded_links` of other content items that depend on it.

steventux:
OK thanks @tuzz that answers the question nicely, I'll close this then as we don't want this content to make it downstream. Be aware that currenty a request to PUT content with a nil base path will raise an error in the DownstreamPresenter, not ideal, perhaps we should revert [this commit](https://github.com/alphagov/publishing-api/pull/266/commits/4f09e6cfcdd69da4b8b1b4cb565cdcaa25c9ae8a) and leave the SubstitutionHelper to raise a `NilBasePathError`?

tuzz:
I think we should wait until DR is done then do another pass at this.

PR #268, opened 2016-04-12
Make the ContentItemPresenter much faster

tuzz:
https://trello.com/c/1wLclYR2/668-improve-efficiency-of-content-item-presenter

**Note:** Tests won't pass until this is merged: https://github.com/alphagov/gds-api-adapters/pull/488

Previously, the ContentItemPresenter could take a long time,
even with pagination and this was causing timeouts across a
number of projects and general instability. It was also
slowing down put content requests because the response goes
through the presenter.

This work effectively rewrites this presenter to work much
more quickly. It relies on `Queries::GetLatest` which is an
example of a "get rows with max value per group" query.
There are further improvements that could be made, perhaps
with how the count is computed, but these changes are
probably good enough for now.

The old presenter made it so that public_updated_at was
always output, regardless of whether this field was
requested or not. I have fixed this. It should now be
explicitly requested, just like any other field. In
addition to this, I made it so that the presenter defaults
to return all fields. The presenter is now quick enough to
do this and I think it serves as good documentation as to
which fields are available.

Finally, here are the results of some quick benchmarking
in development. Times are in milliseconds:

```
/v2/linkables?document_type=organisation
new: 95, 100, 96, 97, 95 (avg 97)
old: 553, 302, 535, 519, 563 (avg 494)
average: 509% as fast

/v2/linkables?document_type=policy
new: 39, 44, 37, 44, 38 (avg 40)
old: 109, 91, 92, 197, 210 (avg 140)
average: 350% as fast

/v2/content?document_type=organisation&fields[]=title&fields[]=base_path
new: 84, 87, 83, 90, 87 (avg 86)
old: 320, 361, 374, 351, 368 (avg 355)
average: 416% as fast

/v2/content?document_type=cma_case&fields[]=description&per_page=1000
new: 116, 111, 109, 114, 117 (avg 113)
old: 985, 678, 964, 953, 730 (avg 862)
average: 763% as fast

/v2/content?document_type=travel_advice&fields[]=details&per_page=3
new: 98, 99, 98, 105, 107 (avg 101)
old: 1363, 1048, 1020, 1025, 1030 (avg 1097)
average: 1086% as fast

PUT /v2/content/08d48cdd-6b50-43ff-a53b-beab47f4aab0 (travel advice index)
new: 904, 780, 804, 660, 883 (avg 806)
old: 3601, 2483, 2683, 2542, 2592 (avg 2780)
average: 345% as fast
```

Note: This last case was particularly problematic because the default
gds-api-adapters timeout is 4 seconds, which the old implementation
was close to hitting.

## Benchmarks graph

benchmark

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -1,182 +1,149 @@
+# This presenter has been carefully written to run quickly. Please be careful
+# if editing its behaviour and make sure to compare benchmarks.
 module Presenters
   module Queries
     class ContentItemPresenter
-      attr_accessor :order, :offset, :limit
+      attr_accessor :scope, :fields, :order, :limit, :offset
+
+      DEFAULT_FIELDS = [
+        *ContentItem::TOP_LEVEL_FIELDS,
+        :publication_state,
+        :user_facing_version,
+        :base_path,
+        :locale,
+        :lock_version,
+        :internal_name,
+      ]
+
+      def self.present_many(scope, params = {})
+        new(scope, params).present_many
+      end
 
       def self.present(content_item)
         translation = Translation.find_by!(content_item: content_item)
 
-        content_items = ContentItem.where(content_id: content_item.content_id)
-        content_items = Translation.filter(content_items, locale: translation.locale)
+        scope = ContentItem.where(content_id: content_item.content_id)
+        scope = Translation.filter(scope, locale: translation.locale)
 
-        present_many(content_items).first
+        present_many(scope).first
       end
 
-      def self.present_many(content_item_scope, fields: nil, order: { public_updated_at: :desc }, offset: 0, limit: nil, locale: nil)
-        presenter = new(content_item_scope, fields: fields, order: order, limit: limit, offset: offset, locale: locale)
-        presenter.present
+      def initialize(scope, params = {})
+        self.scope = scope
+        self.fields = (params[:fields] || DEFAULT_FIELDS).map(&:to_sym)
+        self.order = params[:order]
+        self.limit = params[:limit]
+        self.offset = params[:offset]
       end
 
-      def initialize(content_item_scope, fields: nil, order: {}, limit: nil, offset: 0, locale: nil)
-        self.content_item_scope = content_item_scope
-        self.fields = fields
-        self.order = order
-        self.limit = limit
-        self.offset = offset
-        self.locale = locale
-      end
+      def present_many
+        scope = ::Queries::GetLatest.call(self.scope)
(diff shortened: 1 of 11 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -1,182 +1,137 @@
...
+          when :user_facing_version
+            "user_facing_versions.number AS user_facing_version"
+          when :lock_version
+            "lock_versions.number AS lock_version"
+          when :description
+            "description->>'value' AS description"
+          when :internal_name
+            "COALESCE(details->>'internal_name', title) AS internal_name"
+          when :public_updated_at
+            "to_char(public_updated_at, 'YYYY-MM-DD\"T\"HH24:MI:SS\"Z\"') as public_updated_at"
```
elliotcm: Worth a comment to indicate this is iso8601?

app/presenters/queries/content_item_presenter.rb
```diff
@@ -1,182 +1,137 @@
...
+          when :publication_state
+            "#{publication_state_sql} AS publication_state"
+          when :user_facing_version
+            "user_facing_versions.number AS user_facing_version"
+          when :lock_version
+            "lock_versions.number AS lock_version"
+          when :description
+            "description->>'value' AS description"
+          when :internal_name
+            "COALESCE(details->>'internal_name', title) AS internal_name"
```
elliotcm: Maybe worth a comment to explain this.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -1,182 +1,137 @@
...
-        items.detect { |i| i.fetch("state_name") == "draft" }
+      def parse_int_column(result, column)
+        return unless result.key?(column)
+        result[column] = result[column].to_i
       end
 
-      def detect_live(items)
-        items.detect { |i| i.fetch("state_name") == "published" }
+      # It is substantially faster to evaluate directly against Postgres than
+      # to use: #as_json, #pluck or even ActiveRecord::Base.connection.execute
```
elliotcm: Has this been demonstrated in a production environment?  Maybe an artefact of development environment.

app/queries/get_latest.rb
```diff
@@ -0,0 +1,26 @@
+module Queries
+  module GetLatest
+    class << self
+      # Returns a new scope for the content items with the highest user-facing
+      # version number per (content_id, locale) group of the given scope.
```
elliotcm: Might be clearer as "per content_id and locale of the given scope".

elliotcm:
Some small comments otherwise 👍

PR #273, opened 2016-04-18
Add request id to Event...

steventux:
Part of this https://trello.com/c/Io2JYTFv/27-ensure-request-id-is-available-to-multipage-frontend-and-frontend-apps and this https://trello.com/c/mfOWQkE0/350-instrumenting-admin-apps

In order to trace publishing flow from the publishing application,
through the publishing-api and into the content store we should
also add the request id to the Event records for consistency and
completeness.

### DIFF
--- a/app/event_logger.rb
+++ b/app/event_logger.rb
@@ -8,7 +8,8 @@ module EventLogger
         event = Event.create!(
           action: action(command_class),
           payload: payload,
-          user_uid: GdsApi::GovukHeaders.headers[:x_govuk_authenticated_user]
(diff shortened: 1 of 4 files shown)


gpeng:
:+1:

boffbowsh:
There's no way to link an event to the content item it created or published. Part of the ticket is to expose the request id to downstream consumers, e.g. content store and frontend, so I think the request id should probably be copied into the content item too? Or we should add an event_id column to the content_item to show which event generated it.

gpeng:
Persisting it on the item sounds like a good idea to me, https://github.com/alphagov/govuk-content-schemas/pull/287#issuecomment-211770773 notwithstanding.

PR #275, opened 2016-04-20
Handle race condition in path_reservations

boffbowsh:
Path reservations might be attempted to be created concurrently due to
publish intents being added for Stats Announcements. Postgres raises an
error for this, we should swallow the error if the other reservation is
the same as the one this request is trying to create.

/cc @gpeng @tuzz @steventux

### DIFF
--- a/app/models/path_reservation.rb
+++ b/app/models/path_reservation.rb
@@ -7,5 +7,11 @@ class PathReservation < ActiveRecord::Base
     record = find_or_initialize_by(base_path: base_path)
     record.publishing_app = publishing_app
     record.save!
+  rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation
+    # If a path is already reserved by the same publishing app, ignore the error
+    record = find_by(base_path: base_path)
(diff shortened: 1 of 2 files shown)


app/models/path_reservation.rb
```diff
@@ -7,5 +7,11 @@ def self.reserve_base_path!(base_path, publishing_app)
     record = find_or_initialize_by(base_path: base_path)
     record.publishing_app = publishing_app
     record.save!
+  rescue PG::UniqueViolation
+    # If a path is path is already reserved by the same publishing app, ignore the error
```
gpeng: Bit of a typo in the comment 'path is path is'

gpeng:
:+1: to me apart from that little comment comment...

boffbowsh:
Don't merge. I don't think the test actually tests what I think it is.

boffbowsh:
Fixed.

steventux:
:+1: Looks good to me, will be glad to see this noise reduced in Errbit.

tuzz:
Thanks for this, @boffbowsh. 👍

PR #276, opened 2016-04-21
Restore links that were accidently erased

MatMoore:
## The problem
On March 14th a task was run from policy publisher that unintentionally cleared all organisations, lead_organisations, people and working groups from policies.

We don't see any updates since the event that cleared the links, and the event log seems to be mostly automated tasks, for example:

```
2015-12-02 10:42:58 UTC PutContentWithLinks

# Policy publish lead organisation/supporting organisation split
2016-01-29 14:10:29 UTC PatchLinkSet

# Policy publish lead organisation/supporting organisation split
2016-02-08 15:13:04 UTC PatchLinkSet

# Policy publisher republish rake task
2016-03-14 12:04:20 UTC PutContent
2016-03-14 12:04:20 UTC Publish
2016-03-14 12:04:20 UTC PatchLinkSet
```

## How we are correcting it
To correct this, we are going to replay the event log, ignoring the commands that caused the problem.

We need to replay more than one event because we now have PatchLinks, which doesn't require the whole entity.

All except one policy had links populated by a PutContentWithLinks command
in december 2015. We can ignore anything earlier than that, since this action replaces the entire links hash.

## Policy Links after running

219 policies:

```
     link_type      | count
--------------------+-------
 alpha_taxons       |     1
 policy_areas       |     6
 organisations      |   400
 lead_organisations |   219
 working_groups     |    90
 email_alert_signup |   219
 related            |     6
 people             |   340
```

## Review

The PR is currently a long rake task specifically for this fix. I'm not sure if it would be worth factoring out some of the functionality into something more reusable, and whether there is anything else we can do test this better, or any other constraints we should enforce when we run it.

Having general purpose functionality for replaying events does seem like something that will be useful in the future.

At a first glance the data looks correct on dev, but I want to check this more thoroughly before merging.

Part of https://trello.com/c/rQL3rtOD/599-fix-taggings-of-policies

### DIFF
--- a/lib/tasks/sanitize_data.rake
+++ b/lib/tasks/sanitize_data.rake
@@ -2,3 +2,66 @@ desc "Sanitize access limited data"
 task sanitize_data: :environment do
   Tasks::DataSanitizer.delete_access_limited(STDOUT)
 end
+
+task :restore_policy_links, [:action] => [:environment] do |_, args|
+  dry_run = args["action"] != 'apply'
+  puts "Dry run" if dry_run
+
+  policy_query = Queries::GetContentCollection.new(
+    document_type: 'policy',
+    fields: %w(content_id),
+    pagination: NullPagination.new
+  )
+  policy_content_ids = policy_query.call.map { |item| item["content_id"] }
+
+  # Identify the destructive events
+  policy_wipe_events = Event
+    .where("created_at::date = '2016-03-14' and action = 'PatchLinkSet'")
+    .where(content_id: policy_content_ids)
+    .order("content_id, created_at asc")
+
+  wipe_events_by_content_id = Hash.new { |h, k| h[k] = [] }
+
+  puts "Found #{policy_wipe_events.size} destructive events"
+
+  policy_wipe_events.each do |event|
+    puts "#{event.content_id} #{event.action} #{event.created_at} #{event.payload[:links]}"
+
+    wipe_events_by_content_id[event.content_id] << event.id
+  end
+
+  puts
+
+  # Restore the links
+  policy_wipe_events.each do |wipe_event|
+    puts "Restoring #{wipe_event.content_id}..."
+
+    policy_events = Event.where(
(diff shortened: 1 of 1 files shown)


boffbowsh:
This looks good to me, pending testing of the data. I don't think this needs to be extracted out into generic functionality yet. The next time someone needs to do something similar they can use this as prior art and then decide whether to extract it out.

Good sleuthing 👍

boffbowsh:
P.s. there's some lint failures that need fixing

tijmenb:
I think this looks good. Could you add a dry-run option so that we can inspect the expected changes?

fofr:
> On March 14th a task was run from policy publisher that unintentionally cleared all organisations, lead_organisations, people and working groups from policies.
Is there protection against this happening again?

MatMoore:
@fofr currently we've just removed the broken task, but I'm going to go back to policy publisher and clean up some of the technical debt I left behind when migrating it to the publishing api.

There are also gaps in our testing, which I'm going to look into once the data is restored.

I've added an `action` argument which is dry run by default.
` bundle exec rake restore_policy_links[apply]` to apply for real.

lib/tasks/sanitize_data.rake
```diff
@@ -2,3 +2,66 @@ desc "Sanitize access limited data"
...
+  policy_query = Queries::GetContentCollection.new(
+    document_type: 'policy',
+    fields: %w(content_id),
+    pagination: NullPagination.new
+  )
+  policy_content_ids = policy_query.call.map { |item| item["content_id"] }
+
+  # Identify the destructive events
+  policy_wipe_events = Event
+    .where(["created_at::date = '2016-03-14' and action = 'PatchLinkSet'"])
```
jackscotti: Why is this string within square brackets/

MatMoore: Don't think there's a reason, I'll remove it.

PR #277, opened 2016-04-22
Search content items by title and base_path

dougdroper:
If searching for a field that is not supported, that field is ignored
and it doesn't change the behaviour of the rest of the query

https://trello.com/c/9rpo1V2u/683-search-functionality-on-get-content-collection-endpoint-searching-documents-by-slug-and-title-in-sp

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -38,7 +38,7 @@ module Presenters
       end
 
       def present_many
-        parsing_enumerator(evaluate_query(ordered_extracted_fields))
+        parse_results(results)
       end
 
       def total
@@ -47,8 +47,12 @@ module Presenters
 
     private
 
-      def ordered_extracted_fields
-        extract_fields(order_and_paginate(full_scope))
+      def results
+        execute_query(ordered_fields)
+      end
+
+      def ordered_fields
+        select_fields(order_and_paginate)
       end
 
       def full_scope
@@ -68,14 +72,15 @@ module Presenters
         LockVersion.join_content_items(scope)
       end
 
-      def order_and_paginate(scope)
+      def order_and_paginate
+        scope = full_scope
         scope = scope.order(order.to_a.join(" ")) if order
         scope = scope.limit(limit) if limit
         scope = scope.offset(offset) if offset
         scope
       end
 
-      def extract_fields(scope)
+      def select_fields(scope)
         fields_to_select = fields.map do |field|
           case field
           when :publication_state
@@ -99,6 +104,7 @@ module Presenters
       end
 
       def search(scope)
+        return scope unless search_query.present?
         scope.where("title ilike ? OR base_path ilike ?", "%#{search_query}%", "%#{search_query}%")
       end
 
@@ -124,7 +130,7 @@ module Presenters
         "COALESCE(details->>'internal_name', title) "
       end
 
-      def parsing_enumerator(results)
+      def parse_results(results)
         json_columns = %w(details routes redirects need_ids)
         int_columns = %w(user_facing_version lock_version)
 
@@ -150,8 +156,8 @@ module Presenters
 
       # It is substantially faster to evaluate in this way rather than calling
       # the #pluck or #as_json methods.
-      def evaluate_query(scope)
-        ActiveRecord::Base.connection.execute(scope.to_sql)
+      def execute_query(query)
+        ActiveRecord::Base.connection.execute(query.to_sql)
       end
     end
   end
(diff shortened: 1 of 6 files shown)


danielroseman:
I don't think this quite supports the use case of specialist-publisher, which is to search for a single string and match against both title and slug.

dougdroper:
@danielroseman ah I see, I thought they needed some level of granularity. In the case for a single string, they could just call the API with q_title=SEARCH&q_base_path=SEARCH which will match SEARCH in title AND base_path, or do they need an interface that just does q=SEARCH and match against title OR base_path?

tuzz:
If an item previously had a title of `foo` and it changed to `bar` and was published, what would happen if you search for `q_title=foo`? Would it be an empty set of results or would it include the superseded item?

danielroseman:
@dougdroper the latter, it needs to be OR.

dougdroper:
@tuzz If the latest content_item has a title of `foo`, then searching for `bar` would return an empty set. This is just reducing the scope of the existing query, ie, it still gets passed to `GetLatest`

dougdroper:
@danielroseman updated, let me know if that supports the specialist-publishers needs.

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -45,6 +45,27 @@
       )
     end
 
+    context "searching a field" do
+      context "when there is a valid query" do
+        let!(:content_item) do
+          FactoryGirl.create(:live_content_item,
+                             base_path: "/foo",
+                             format: "topic",
+                             title: "foo bar")
```
danielroseman: Would be better remove "foo" from the title so that there is no overlap between the fields.

danielroseman:
Looks good, one minor comment and there is a linting issue.

tuzz:
That's not quite what I asked. I'm not sure what will happen if you search for a title that is no longer the title of the latest content item. Could we possibly write a test to check this?

I'm not sure what the behaviour of this should be, either. Whether we should return the superseded item or whether it should be scoped to only the latest (draft/published) versions.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -34,22 +34,31 @@ def initialize(scope, params = {})
...
       end
 
       def present_many
-        scope = ::Queries::GetLatest.call(self.scope)
-        scope = join_supporting_objects(scope)
-        scope = order_and_paginate(scope)
-
-        extract_fields(scope)
+        parsing_enumerator(evaluate_query(ordered_extracted_fields))
       end
```
tuzz: I find it quite difficult to read these nested method calls. I found the procedural style easier to read, but maybe that's just me? By breaking everything into one line methods, I personally struggle to build a global understanding of what this class does.

dougdroper: Yes I suppose this particular method is a little tricky to parse inside out. I was trying to expose the ordering of these methods, but have done a bit more refactoring in f997d4297f23157e18fcab20c13d6a1f827b1a4e to try and express the intent more clearly. 

I guess from a global understanding of the class I was trying to sum up what this does in the one public method, and then encapsulate the complexity lower down. When I read this class I want to understand what it does as fast as possible without looking below `private` so tend to prefer this style, but it certainly increases the amount you have to jump around, but hopefully with clearer method names it is clearer overall. Lets go through this class together and see if there is a way to make it clearer.

tuzz: Ok, sure. Let's go through this when you're at a natural breaking point. 😃

tuzz:
Docs for that endpoint are here:
https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#get-v2content

We might also want to consider adding a pact to gds-api-adapters:
https://github.com/alphagov/gds-api-adapters/blob/master/test/publishing_api_v2_test.rb#L906

app/presenters/queries/content_item_presenter.rb
```diff
@@ -86,10 +100,11 @@ def extract_fields(scope)
...
 
-        scope = scope.select(*fields_to_select)
-        results = evaluate_query(scope)
+        scope.select(*fields_to_select)
+      end
 
-        parsing_enumerator(results)
+      def search(scope)
+        scope.where("title ilike ? OR base_path ilike ?", "%#{search_query}%", "%#{search_query}%")
       end
```
tuzz: What happens if no search query is provided? Will it still attach this scope and does that impact performance?

dougdroper: Good catch 👍  my guard clause seems to have disappeared in the refactoring. Back now.

PR #278, opened 2016-04-26
Get content-store to know all the links of specialist-publisher owned documents

jackscotti:
specialist-publisher is being converted to a "standard" Rails app that uses the publishing-api V2 endpoints. We don't know when the conversion completed and we want all the links of specialist-publisher owned documents to be in the Content Store ASAP. We decided to introduce this quick-hack over having to migrate the "non-standard" Rails version of specialist-publisher.

### DIFF
--- a/app/commands/put_content_with_links.rb
+++ b/app/commands/put_content_with_links.rb
@@ -51,6 +51,8 @@ module Commands
     end
 
     def delete_existing_links
+      return if protected_apps.include?(payload[:publishing_app])
+
       link_set = LinkSet.find_by(content_id: payload[:content_id])
       return unless link_set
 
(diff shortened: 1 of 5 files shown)


tijmenb:
👍  this looks great!

dougdroper:
Not sure about the context here, but do you need something similar for the v2 endpoint? 
https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/patch_link_set.rb

jackscotti:
@dougdroper no because the V2 endpoint actually patches the links rather than PUTting them. This hack is to make the V1 endpoint behave like the V2 but only for Specialist Publisher. This is for not having to migrate the whole application.

dougdroper:
@jackscotti cool makes sense, just double checking.

app/commands/put_content_with_links.rb
```diff
@@ -61,5 +63,15 @@ def delete_existing_links
     def protected_link_types
       ["alpha_taxons"]
     end
+
+    def protected_apps
+      # specialist-publisher is being converted to a "standard" Rails app
+      # that uses the publishing-api V2 endpoints. We don't know when the
+      # conversion completed and we want all the links of specialist-publisher
```
elliotcm: Do you mean "when the conversion will be completed"?  Or has it completed already?

tuzz:
This looks fine to me. Code is cohesive with the `alpha_taxons` thing and it's tested. Once Elliot's question has been resolved, 👍  from me.

elliotcm:
@jackscotti 👍 merge when you're content

PR #280, opened 2016-04-29
Change withdrawn to unpublished

elliotcm:
- Renames "withdrawn" to "unpublished" (and other conjugations)
- Creates the `unpublishings` table
- Populates from existing data

https://trello.com/c/zhSso5gS/699-change-all-withdrawn-content-states-to-unpublished

### DIFF
--- a/app/models/unpublishing.rb
+++ b/app/models/unpublishing.rb
@@ -1,0 +1,18 @@
+class Unpublishing < ActiveRecord::Base
+  self.inheritance_column = nil
+
+  belongs_to :content_item
+
+  validates :content_item, presence: true, uniqueness: true
+  validates :type, presence: true
+  validates :explanation, presence: true, if: :withdrawal?
+  validates :alternative_url, presence: true, if: :redirect?
+
+  def withdrawal?
+    type == "withdrawal"
+  end
+
+  def redirect?
+    type == "redirect"
+  end
+end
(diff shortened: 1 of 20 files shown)


db/schema.rb
```diff
@@ -134,6 +134,15 @@
 
   add_index "translations", ["content_item_id", "locale"], name: "index_translations_on_content_item_id_and_locale", using: :btree
 
+  create_table "unpublishings", force: :cascade do |t|
+    t.integer  "content_item_id", null: false
```
tuzz: Should we add an index here? Perhaps a composite of `content_item_id, type` as well?

app/models/unpublishing.rb
```diff
@@ -0,0 +1,17 @@
+class Unpublishing < ActiveRecord::Base
+  self.inheritance_column = nil
+
+  belongs_to :content_item
+
+  validates :type, presence: true
+  validates :explanation, presence: true, if: :withdrawal?
+  validates :alternative_url, presence: true, if: :redirect?
```
tuzz: Are there any cases where we'd have more than one unpublishing for the same content_item? If not, could we add a uniqueness validator for belt and braces?

tuzz:
👍 LGTM. Minor comments.

elliotcm:
@tuzz Updated

app/substitution_helper.rb
```diff
@@ -10,7 +10,7 @@ def clear!(new_item_format:, new_item_content_id:, base_path:, locale:, state:)
         allowed_to_substitute = (substitute?(new_item_format) || substitute?(blocking_item.format))
 
         if mismatch && allowed_to_substitute
-          State.withdraw(blocking_item)
+          State.unpublish(blocking_item)
```
tuzz: We should also create an unpublishing here.

PR #281, opened 2016-04-29
First published at

elliotcm:
Adds `first_published_at` behaving in a similar way to `public_updated_at` and populates using some SQL from the GetLatest query object (but with the comparator flipped).  Returns as ISO8601 from the presenter.

https://trello.com/c/Si54omdL/698-return-the-first-published-at-time-to-publishing-apps-for-content-items


elliotcm:
Depends on https://github.com/alphagov/content-store/pull/205

elliotcm:
This has been tested on integration and the content store PR has been deployed to production.

app/commands/v2/publish.rb
```diff
@@ -39,6 +39,7 @@ def call
         clear_published_items_of_same_locale_and_base_path(content_item, translation, location)
 
         set_public_updated_at(content_item, previous_item, update_type)
+        set_first_published_at(content_item)
```
danielroseman: Not the fault of this line specifically, but this method is getting really long.

elliotcm: We do a lot of stuff when we publish.  Might be worth another refactor later, but I'm not sure how much it can be reduced down while still being clear.

PR #282, opened 2016-05-03
Handle nil error_details

boffbowsh:
`upstream_error.error_details` will be nil in the case of the response
body not being JSON encoded. We should handle this case by returning an
empty hash.

This is masking real errors in production: https://errbit.publishing.service.gov.uk/apps/552f95820da11577b4000032/problems/571fff906578631f8b522d00


elliotcm:
Normally we'd want to fix the source to return a proper error structure, but this code is on the chopping block so 👍

PR #285, opened 2016-05-05
Add rake task to requeue all published content items

mobaig:
The goal here is to allow us to requeue all published content
items and observe how the Rummager message queue workers perform under
this kind of load.

trello: https://trello.com/c/X3IbrAY7

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -101,7 +101,6 @@ module Commands
       def send_to_message_queue(content_item)
         payload = Presenters::MessageQueuePresenter.present(
           content_item,
-          event,
           fallback_order: [:published],
           update_type: "links",
         )
--- a/app/presenters/downstream_presenter.rb
+++ b/app/presenters/downstream_presenter.rb
@@ -1,14 +1,13 @@
 module Presenters
   class DownstreamPresenter
-    def self.present(content_item, event, fallback_order:)
+    def self.present(content_item, fallback_order:)
       link_set = LinkSet.find_by(content_id: content_item.content_id)
-      new(content_item, link_set, event, fallback_order: fallback_order).present
+      new(content_item, link_set, fallback_order: fallback_order).present
     end
 
-    def initialize(content_item, link_set, event, fallback_order:)
+    def initialize(content_item, link_set, fallback_order:)
       self.content_item = content_item
       self.link_set = link_set
-      self.event = event
       self.fallback_order = fallback_order
     end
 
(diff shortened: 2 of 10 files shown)


lib/requeue_content.rb
```diff
@@ -0,0 +1,38 @@
+class RequeueContent
+  def initialize(number_of_items: nil)
+    @number_of_items = number_of_items
+  end
+
+  attr_accessor :number_of_items
+
+  def call
+    if number_of_items.present?
+      ContentItemFilter.filter(state: :published).limit(number_of_items).each do |content_item|
```
jackscotti: `number_of_items` might be a big number can you use `find_each` here?

lib/requeue_content.rb
```diff
@@ -0,0 +1,38 @@
+class RequeueContent
+  def initialize(number_of_items: nil)
+    @number_of_items = number_of_items
+  end
+
+  attr_accessor :number_of_items
+
+  def call
```
jackscotti: could this be a class method?

app/commands/v2/patch_link_set.rb
```diff
@@ -101,7 +101,6 @@ def send_to_content_store(content_item, content_store)
       def send_to_message_queue(content_item)
         payload = Presenters::MessageQueuePresenter.present(
           content_item,
-          event,
```
boffbowsh: Why has this been removed?

lib/requeue_content.rb
dougdroper: I think `find_each` will ignore the limit and return all items

lib/requeue_content.rb
dougdroper: Out of interest why would a class method be better?

jackscotti: No need to instantiate an object. Nicer to read, no other reasons.

app/presenters/downstream_presenter.rb
```diff
@@ -1,14 +1,13 @@
 module Presenters
   class DownstreamPresenter
-    def self.present(content_item, event, fallback_order:)
```
dougdroper: Do we need to updated the ContentStorePresenter to remove the event too?

dougdroper: Actually never mind

lib/requeue_content.rb
mobaig: Yep, that's pretty much it. 

Alternative might be to write some SQL to handle batching through a limited subset but I think that's overkill here - the limit is really only used when testing the rake task at this stage.

boffbowsh:
We should remember to remove this task when email alerts are driven by the queues, or build some sort of deduplication into the email service.

jackscotti:
@mobaig can you add a comment mentions what @boffbowsh said?

elliotcm:
@boffbowsh Not sure I understand.  Email alerts are already driven by the queues.

elliotcm:
@boffbowsh also the email alert service already has deduplication

boffbowsh:
You're right, I'm being stupid.

elliotcm:
Not to keep laying it on but these updates are being sent as update type "links" whereas the email service is only subscribed to update type "major".

boffbowsh:
That was the bit where I realised I was being stupid

PR #286, opened 2016-05-06
Remove redundant Special Education Needs policy

brenetic:
- There are two copies of this policy. This removes one of them as per
  this [Zendesk ticket](https://govuk.zendesk.com/agent/tickets/1299583).

cc @elliotcm @tijmenb

### DIFF
--- a/db/migrate/20160506124002_remove_redundant_special_education_needs_policy.rb
+++ b/db/migrate/20160506124002_remove_redundant_special_education_needs_policy.rb
@@ -1,0 +1,16 @@
+class RemoveRedundantSpecialEducationNeedsPolicy < ActiveRecord::Migration
+  # Remove government/policies/special-education-needs-sen
+  def change
+    content_items = ContentItem.where(content_id: "f656d065-43aa-4ab0-91f7-a6809ce5b68b")
+    Translation.where(content_item: content_items).destroy_all
+    Location.where(content_item: content_items).destroy_all
+    State.where(content_item: content_items).destroy_all
+    UserFacingVersion.where(content_item: content_items).destroy_all
+    AccessLimit.where(content_item: content_items).destroy_all
+    LockVersion.where(target: content_items).destroy_all
+    content_items.destroy_all
+
+    link_sets = LinkSet.where(content_id: "f656d065-43aa-4ab0-91f7-a6809ce5b68b")
+    link_sets.destroy_all
+  end
+end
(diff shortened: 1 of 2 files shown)


brenetic:
This is not a url that is visible on gov.uk and it is also not in Rummager. @tijmenb is there somewhere else that we should also be looking just in case?

tijmenb:
I don't know. I do have the feeling it would be better if policy-publisher sent a redirect / gone or something. Doing these kind of things should be possible with the public API instead of doing complex data migrations. If we do this here we'll end up with a million data migrations in publishing-api. @danielroseman thoughts?

tijmenb:
I see now that https://github.com/alphagov/publishing-api/pull/260 is the precedent for this, so apparently this is fine.

elliotcm:
@tijmenb I wasn't intending to set a precedent for this being the right solution to this, I was expecting us to fix the underlying issue, though I can't remember whether I reported it as a bug.

MatMoore:
@elliotcm do we know how the policies got into this state? I thought about issuing a redirect as well but as far as I can tell there is no route set up for this base path?

@tijmenb I think I agree; it seems like it would be useful to have a general way to withdraw content from the publishing api for migrated apps, even if its not part of the normal publishing workflow for users, since this kind of thing will inevitably happen again. Do we have a rake task that does anything like this?

Also, not really relevant to the topic at hand, but it would be nice to have proper functionality for merging "linkables" in collections publisher as well at some point, so users can do this themselves.

elliotcm:
@MatMoore Not sure, I didn't get a chance to do a proper investigation.

As for the withdrawing, we have an endpoint incoming shortly which could be used.

brenetic:
@elliotcm Since this particular policy has no public facing base path it would seem reasonable in this instance to use the db migration to remove it?

Can we add this to a team's backlog to investigate the reason it happened in the first place?

elliotcm:
@brenetic Fine by me.  @tijmenb are you OK with that, and as your team owns policy publisher would that be you making a ticket?

tijmenb:
I've made a PR to do this in policy-publisher with the new unpublish endpoint: https://github.com/alphagov/policy-publisher/pull/133

tijmenb:
@brenetic closing this in favour of https://github.com/alphagov/policy-publisher/pull/133. I'll take it from here. Thanks!

PR #287, opened 2016-05-06
Move the presentation of the ContentItem to the worker

dougdroper:
Moves the finding and presentation of the ContentItem into the worker.
Moves the dependency of fallback_order into the ContentStore adapters.

### DIFF
--- a/app/adapters/content_store.rb
+++ b/app/adapters/content_store.rb
@@ -2,6 +2,7 @@
 # to
 module Adapters
   class ContentStore
+    DEPENDECY_FALLBACK_ORDER = [:published]
     def self.put_content_item(base_path, content_item)
       CommandError.with_error_handling do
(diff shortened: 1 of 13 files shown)


elliotcm:
We initially had this in the worker but moved it out for reasons I can't remember.  @danielroseman?

danielroseman:
See https://github.com/alphagov/publishing-api/pull/226 for the reasoning.

dougdroper:
Ah right ok, makes sense.. Might move the enqueuing out of the transaction then to guarantee the content item is committed to the DB first.

dougdroper:
Although this might be better since we were enqueuing a content_item that may not have been committed to the DB, if we raise in the worker to indicated that the content_item hasn't been committed then it wont get sent down to the content store.

PR #288, opened 2016-05-06
Resolve invalid versions

steventux:
https://trello.com/c/1domfbpK/690-fix-problems-flagged-up-by-rake-db-validate-versions


danielroseman:
When I run this locally, I get `NoMethodError: undefined method `number' for nil:NilClass` - I think some of the items have only two versions, might need some extra safeguards.

Also Jenkins is complaining of lint errors.

steventux:
Thanks @danielroseman there were a few edge cases with real data like version duplication in the middle of a workflow sequence eg `[1, 2, 2, 3, 4]`. Updated.

PR #289, opened 2016-05-09
Unpublishing endpoint

elliotcm:
Add an endpoint for unpublishing content, including "gone" items, "withdrawal"s, and redirects.

More details here https://trello.com/c/vXVnrgbB/701-add-v2-content-content-id-unpublish-endpoint and in the included docs.

### DIFF
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -46,21 +46,28 @@ module Commands
         "(#{previous_version_number}) is not the same as the current " +
         "lock version of the content item (#{current_version.number})."
 
-      conflict_error = CommandError.new(
-        code: 409,
-        message: "Conflict",
+      fields = {
+        fields: {
+          previous_version: ["does not match"],
+        },
+      }
+
+      if current_version.conflicts_with?(previous_version_number)
+        raise_command_error(409, "Conflict", fields, friendly_message: friendly_message)
+      end
+    end
+
+    def raise_command_error(code, message, fields, friendly_message: nil)
+      raise CommandError.new(
+        code: code,
+        message: message,
         error_details: {
           error: {
-            code: 409,
-            message: friendly_message,
-            fields: {
-              previous_version: ["does not match"],
-            }
-          }
+            code: code,
+            message: friendly_message || message,
+          }.merge(fields)
         }
       )
-
-      raise conflict_error if current_version.conflicts_with?(previous_version_number)
     end
   end
 end
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -129,8 +129,6 @@ module Commands
       end
 
       def send_downstream(content_item, update_type)
-        return unless downstream
-
         PresentedContentStoreWorker.perform_async(
           content_store: Adapters::ContentStore,
           payload: Presenters::ContentStorePresenter.present(
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -147,19 +147,6 @@ module Commands
 
         PublishingAPI.service(:queue_publisher).send_message(queue_payload)
       end
-
-      def raise_command_error(code, message, fields)
-        raise CommandError.new(
-          code: code,
-          message: message,
-          error_details: {
-            error: {
-              code: code,
-              message: message,
-            }.merge(fields)
-          }
-        )
-      end
     end
   end
 end
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -2,7 +2,8 @@ module Commands
   module V2
     class Unpublish < BaseCommand
       def call
-        content_item = find_live_content_item
+        content_id = payload.fetch(:content_id)
+        content_item = find_live_content_item(content_id)
 
         case payload.fetch(:type)
         when "withdrawal"
@@ -23,7 +24,7 @@ module Commands
           content_item: content_item,
         )
 
-        send_downstream(content_item, unpublishing) if downstream
+        send_content_item_downstream(content_item, unpublishing) if downstream
       end
 
       def redirect(content_item)
@@ -33,18 +34,24 @@ module Commands
           content_item: content_item,
         )
 
-        base_path = Location.find_by(content_item: content_item).base_path
+        send_redirect_downstream(
+          base_path: Location.find_by(content_item: content_item).base_path,
+          publishing_app: content_item.publishing_app,
+          destination: unpublishing.alternative_path
+        ) if downstream
+      end
 
+      def send_redirect_downstream(base_path:, publishing_app:, destination:)
         downstream_payload = {
           format: "redirect",
           base_path: base_path,
-          publishing_app: content_item.publishing_app,
+          publishing_app: publishing_app,
           public_updated_at: Time.zone.now.iso8601,
           redirects: [
             {
               path: base_path,
               type: "exact",
-              destination: unpublishing.alternative_path,
+              destination: destination,
             }
           ],
         }
@@ -56,20 +63,16 @@ module Commands
         )
       end
 
-      def content_id
-        payload[:content_id]
-      end
-
       def locale
         payload.fetch(:locale, ContentItem::DEFAULT_LOCALE)
       end
 
-      def find_live_content_item
+      def find_live_content_item(content_id)
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
         filter.filter(locale: locale, state: "published").first
       end
 
-      def send_downstream(content_item, unpublishing)
+      def send_content_item_downstream(content_item, unpublishing)
         downstream_payload = Presenters::ContentStorePresenter.present(
           content_item,
           event,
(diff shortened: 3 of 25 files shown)


app/commands/v2/unpublish.rb
```diff
@@ -0,0 +1,137 @@
...
+          event,
+          fallback_order: [:published]
+        )
+
+        downstream_payload.merge!(
+          withdrawn_notice: {
+            explanation: unpublishing.explanation,
+            withdrawn_at: unpublishing.created_at.iso8601,
+          }
+        )
```
tuzz: Would it make sense to move this logic into the downstream presenter itself? It could check if the content item being presented has an unpublishing and merge in the withdrawn notice.

app/commands/base_command.rb
```diff
@@ -46,21 +46,28 @@ def check_version_and_raise_if_conflicting(current_versioned_item, previous_vers
...
-        code: 409,
-        message: "Conflict",
+      fields = {
+        fields: {
+          previous_version: ["does not match"],
+        },
+      }
+
+      if current_version.conflicts_with?(previous_version_number)
+        raise_command_error(409, "Conflict", fields, friendly_message: friendly_message)
```
tuzz: 👍

app/commands/v2/publish.rb
```diff
@@ -129,8 +129,6 @@ def lookup_published_item(content_item)
       end
 
       def send_downstream(content_item, update_type)
-        return unless downstream
-
```
tuzz: To clarify, where is this checked?

app/commands/v2/unpublish.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+
+      def redirect(content_item)
+        unpublishing = State.unpublish(content_item,
+          type: "redirect",
+          alternative_path: payload.fetch(:alternative_path),
+        )
+
+        redirect = RedirectPresenter.present(
+          base_path: Location.find_by(content_item: content_item).base_path,
```
tuzz: We supposedly support content items that don't have locations and this feature will likely be used more in the future. What do we do if that's the case? I guess we don't need to send a redirect?

tuzz: Maybe we should raise a 422, actually, as it probably doesn't make sense to unpublish with a "redirect" type for something without a base_path.

app/commands/v2/unpublish.rb
```diff
@@ -0,0 +1,137 @@
...
+          publishing_app: content_item.publishing_app,
+          alternative_path: payload[:alternative_path],
+          explanation: payload[:explanation],
+        )
+
+        send_arbitrary_downstream(gone) if downstream
+      end
+
+      def send_arbitrary_downstream(downstream_payload)
```
tuzz: "arbitrary" is a bit confusing. I understand that intent of this word after seeing where it's used, but maybe "generic" is better or just drop it all together?

app/commands/v2/unpublish.rb
elliotcm: Moved.

app/commands/v2/publish.rb
elliotcm: It's a private method, so elsewhere in this file.

tuzz: Gotcha. Thanks.

app/models/unpublishing.rb
```diff
@@ -3,10 +3,17 @@ class Unpublishing < ActiveRecord::Base
 
   belongs_to :content_item
 
+  VALID_TYPES = %w(
+    gone
+    withdrawal
+    redirect
+    substitute
+  ).freeze
```
tuzz: In the docs:

> Can be "gone", "withdrawal", or "redirect".

I presume that "substitute" is an internal-only type? Do we expose it back to publishing apps if they make get requests to other endpoints?

elliotcm: We don't expose any of them back to the publishing app, currently.

app/commands/v2/unpublish.rb
elliotcm: It doesn't make sense to unpublish *any* content without a location, so I've added checks for that.

app/commands/v2/unpublish.rb
elliotcm: Removed.

app/models/state.rb
```diff
@@ -22,11 +22,29 @@ def self.publish(content_item)
...
+        alternative_path: alternative_path,
+      )
+    end
+  end
+
+  def self.substitute(content_item)
+    unpublish(content_item,
       type: "substitute",
       explanation: "Automatically unpublished to make way for another content item",
     )
```
tuzz: What do you think about maybe making this `#unpublish_as_substitute` (or something better named) to distinguish it from those methods that relate directly with the `state.name`? Maybe I'm just being a pedant, though.

elliotcm: I did think about it, but iunno.

doc/publishing-api-syntactic-usage.md
```diff
@@ -68,6 +68,35 @@ Requests to update an existing draft content item:
...
+  - `type` the type of unpublishing to create/perform.
+
+ ### Other request params:
+  - `explanation` (optional) Message to display on page for `gone`,
+                  (required) for `withdrawal`,
+                  (ignored) for `redirect`.
+  - `alternative_path` (optional) path to turn into a URL to display on page for `gone`,
+                       (required) path to redirect to if `redirect`,
+                       (ignored) if `withdrawal`.
+  - `discard_drafts` (optional) anything other than `true` is considered `false`,
```
tuzz: What is this parameter plural? Do we ever have more than one draft?

doc/publishing-api-syntactic-usage.md
```diff
@@ -68,6 +68,35 @@ Requests to update an existing draft content item:
...
+  - Will refuse to unpublish a lone draft.
+  - Will refuse to unpublish a redrafted document unless `discard_drafts` is `true`.
+  - Validates that unpublishing `type` is one of `withdrawal`, `gone` or `redirect` and raises a 422 otherwise.
+  - Retrieves the live content item with the matching content_id and locale and changes its state to `unpublished`.
+  - Creates an `Unpublishing` with the provided details.
+  - Will update the `Unpublishing` if the document is already unpublished.
+  - Sends the gone/redirect/withdrawal to the live content store.
+  - Does not send to the draft content store (unless a draft was discarded).
+  - Does not send to the message queue.
+  - Returns 200 along with the content_id of the unpublished item.
```
tuzz: Would it be worth re-presenting the whole content item rather than just returning the content_id? We've used this pattern for other things (e.g. PutContent) and the ContentItemPresenter is really fast now.

doc/publishing-api-syntactic-usage.md
elliotcm: Not yet.  **Yet**.

elliotcm: Yet.

app/commands/v2/unpublish.rb
tuzz: 👍

doc/publishing-api-syntactic-usage.md
elliotcm: We don't for `Publish` and this is the reverse of that.

app/commands/v2/unpublish.rb
```diff
@@ -0,0 +1,135 @@
...
+      end
+
+      def previous_version_number
+        payload[:previous_version].to_i if payload[:previous_version]
+      end
+
+      def find_unpublishable_content_item(content_id)
+        filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
+        filter.filter(locale: locale, state: %w(published unpublished)).first
+      end
```
tuzz: 👍  for remembering to scope by locale! We've made that mistake so many times.

db/migrate/20160505133601_rename_alternative_url.rb
```diff
@@ -0,0 +1,5 @@
+class RenameAlternativeUrl < ActiveRecord::Migration
+  def change
+    rename_column :unpublishings, :alternative_url, :alternative_path
```
tuzz: When you unpublish something, can you set an external url to redirect to? If so, would we need the full URL rather than just the path?

tuzz:
Lots of questions mostly. Code looks great. 👍

db/migrate/20160505133601_rename_alternative_url.rb
elliotcm: You cannot.  Whitehall's validations require it to be a fully qualified GOV.UK URL.

doc/publishing-api-syntactic-usage.md
tuzz: Ok, fair enough 👍

doc/publishing-api-syntactic-usage.md
tuzz: 😱

PR #290, opened 2016-05-09
Present in worker

dougdroper:
Move downstream presentation to outside the transaction
Ensure that the transaction is committed before sending the ContentItem
to the ContentStore to make sure we do not end up in a state where the
ContentStore has invalid data

Moves the finding and presentation of the ContentItem into the worker.
Moves the dependency of fallback_order into the ContentStore adapters.

### DIFF
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -1,16 +1,17 @@
 module Commands
   class BaseCommand
+    attr_reader :callbacks
     class <<self
-      attr_accessor :callbacks
-
       def call(payload, downstream: true)
         logger.debug "#{self} called with payload:\n#{payload}"
+        instance = nil
 
         response = EventLogger.log_command(self, payload) do |event|
-          new(payload, event: event, downstream: downstream).call
+          instance = new(payload, event: event, downstream: downstream)
+          instance.call
         end
 
-        Array(callbacks).compact.each(&:call)
+        instance.callbacks.each(&:call)
 
         response
       rescue ActiveRecord::RecordInvalid => e
@@ -22,17 +23,16 @@ module Commands
       @payload = payload
       @event = event
       @downstream = downstream
-    end
-
-  protected
-
-    def after_transaction_commit
-      self.class.callbacks = yield
+      @callbacks = []
     end
 
   private
 
     attr_reader :payload, :event, :downstream
+
+    def after_transaction_commit(&block)
+      callbacks << block
+    end
 
     def self.raise_validation_command_error(e)
       errors = e.record.errors
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -1,29 +1,27 @@
 module Commands
   class BaseCommand
     attr_reader :callbacks
-    class <<self
-      def call(payload, downstream: true)
-        logger.debug "#{self} called with payload:\n#{payload}"
-        instance = nil
 
-        response = EventLogger.log_command(self, payload) do |event|
-          instance = new(payload, event: event, downstream: downstream)
-          instance.call
-        end
+    def self.call(payload, downstream: true)
+      logger.debug "#{self} called with payload:\n#{payload}"
+      callbacks = Callback.new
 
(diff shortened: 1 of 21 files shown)


app/commands/v2/unpublish.rb
```diff
@@ -44,7 +44,7 @@ def call
           raise_command_error(422, message, fields: {})
         end
 
-        Success.new(content_id: content_id)
+        [Success.new(content_id: content_id), callback]
```
elliotcm: Is the receiving code going to be OK with `callback` being `nil` for redirects and gones?

app/commands/v2/put_content.rb
```diff
@@ -222,11 +220,13 @@ def send_downstream(content_item)
         message += "{ content_store: Adapters::DraftContentStore, content_item_id: #{content_item.id} }"
         logger.info message
```
elliotcm: Is it worth moving this log line into the lambda?  I presume they're going to execute pretty quickly but technically it's lying.

app/commands/v2/unpublish.rb
dougdroper: Yeah totally. It is normalised and striped of `nils`

elliotcm: kk

app/commands/v2/put_content.rb
dougdroper: done

elliotcm:
Looks good.  Could I get a little whitespace breathing room between blocks of code that do different things?  e.g. between guard clauses and main function code, and between function code and return values

dougdroper:
yes definitely, it makes it more consistent. 
I wonder if we can add that to the linter rules, anyway updated.

app/commands/v2/discard_draft.rb
```diff
@@ -10,14 +10,13 @@ def call
 
         delete_supporting_objects
         delete_draft_from_database
-        delete_draft_from_draft_content_store(draft_path) if downstream
+        delete_draft_from_draft_content_store(draft_path)
```
tuzz: Would we also want the deletion of the draft to happen after the transaction completes?

I'm imagining the case where we've deleted the draft item from the draft content store, then this transaction fails (for whatever reason). We'd be in a state where we've deleted a draft but haven't written the live content item in its place. What do you think?

app/event_logger.rb
```diff
@@ -13,9 +14,11 @@ def self.log_command(command_class, payload, &_block)
           request_id: GdsApi::GovukHeaders.headers[:govuk_request_id]
         )
 
-        response = yield(event) if block_given?
+        response, callbacks = yield(event) if block_given?
```
tuzz: I was thinking about an alternative interface where `BaseCommand` has a `call_after_transaction_commits` method (or similar) that all of the commands inherit. The `self.call` method on `BaseCommand` could then play through those callbacks after its sure the event has been logged (and the transaction committed).

As it stands, it's not immediately obvious why those calls happen afterwards. Maybe this alternative might make the code more intention revealing? What do you think?

app/commands/v2/discard_draft.rb
dougdroper: Yes cool, I had it in there originally, don't know why I moved it out again. 
It is back now

tuzz:
👍  looks good – one question about an alternative interface.

app/event_logger.rb
dougdroper: Definitely, anything to make it clearer is a big plus 👍  Updated, have a look

app/commands/base_command.rb
```diff
@@ -16,6 +24,12 @@ def initialize(payload, event:, downstream: true)
       @downstream = downstream
     end
 
+  protected
+
+    def after_transaction_commit
+      self.class.callbacks = yield
```
elliotcm: You could just capture the block here and call it later, right?  Like
```
def after_transaction_commit(&block)
  self.class.callbacks << block
end
```

and then iterate over the callbacks.

PR #295, opened 2016-05-11
Migrate format to dual fields specialist pub

tuzz:
> Specialist publisher has a separate document_type from schema
> name. Currently, these are set to the same thing. This
> migration sets the document_type from details.metadata.document_type


benilovj:
(I've rebased this against master)

👍  this should be mergable now, since specialist publisher was deployed yesterday.

benilovj:
What are the implications of such a migration on the event log? Won't this migration mean that the state of the publishing API isn't reproducible by re-running the event log?

PR #296, opened 2016-05-12
Call all callbacks after the top level transaction.

elliotcm:
There was a race condition previously where the
nested commands would trigger out-of-transaction
code to run via sidekiq workers.

A transaction was still running because the top level
command's transaction wrapped the nested commands,
meaning the nested transactions were turned into
savepoints.

This meant that the work done in the nested command
was not committed properly until after the top level
transaction had finished.

This makes sure that all callbacks happen after the
top level transaction is complete.

We have tested this branch on integration and have reasonable confidence that it works to fix the problem.


boffbowsh:
Good sleuthing 🔍. 👍  from me, but will let others with more context merge.

PR #298, opened 2016-05-12
Add an order parameter to /v2/content

tuzz:
This pull request adds an order parameter to allow publishing apps to order their results by a field other than `public_updated_at` which was previously hardcoded. We have used the same interface as documented [here](https://github.com/alphagov/rummager/blob/master/docs/unified-search-api.md) as suggested by @tijmenb.

This work was completed as part of this ticket:
https://trello.com/c/XJRTeF9f/130-spv2-orders-by-the-wrong-timestamp-on-the-search-index

We decided to limit the fields that you may order by to those that had indexes and left a comment indicating that we should be careful to add an index to any additional fields with which we wish to order. Otherwise, performance would be affected when running the query.

### DIFF
--- a/spec/controllers/v2/content_items_controller_spec.rb
+++ b/spec/controllers/v2/content_items_controller_spec.rb
@@ -177,6 +177,93 @@ RSpec.describe V2::ContentItemsController do
       end
     end
 
+    context "with an order param" do
+      before do
+        @en_draft_content.update!(updated_at: Date.new(2016, 1, 1))
+        @ar_draft_content.update!(updated_at: Date.new(2016, 2, 2))
+
+        get :index, document_type: "topic", locale: "all", order: order, fields: fields
+      end
+
+      context "when ordering by updated_at ascending" do
+        let(:order) { "updated_at" }
+        let(:fields) { ["updated_at"] }
+
+        it "returns the ordered results" do
+          results = JSON.parse(response.body)["results"]
+
+          expect(results).to eq([
+            { "updated_at" => "2016-01-01 00:00:00" },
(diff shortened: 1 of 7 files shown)


benilovj:
The order direction behaviour feels a bit magic and inflexible (what if you want to order descending on a custom field). Is there a good reason to disallow setting it through a parameter as well?

tuzz:
Please could you elaborate? What do you mean by custom field? Are you referring to a field such as `publication_state` which is an alias of `states.name` ?

benilovj:
Actually, ignore me - I misread the code. It's fine :+1:

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -177,6 +177,93 @@
...
+          results = JSON.parse(response.body)["results"]
+
+          expect(results).to eq([
+            { "base_path" => "/content.ar" },
+            { "base_path" => "/content.en" },
+          ])
+        end
+      end
+
+      context "when ordering by base_path ascending" do
```
benilovj: Should this say 'descending'?

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -177,6 +177,93 @@
...
+          results = JSON.parse(response.body)["results"]
+
+          expect(results).to eq([
+            { "updated_at" => "2016-01-01 00:00:00" },
+            { "updated_at" => "2016-02-02 00:00:00" },
+          ])
+        end
+      end
+
+      context "when ordering by updated_at ascending" do
```
benilovj: this should say "when ordering by updated_at descending" ?

PR #303, opened 2016-05-13
Deduplicate links

boffbowsh:
Ensure links aren’t added twice for the same link type. Adds a migration to deduplicate the links.

Finds and deletes all duplicated rows that have an id greater than
another id in the duplicated set. This leaves the duplicated record
with the lowest id intact.

From development VM:

```
== 20160513163056 DeduplicateLinks: migrating =================================
417 unique records of 971 total duplicated
554 records deleted
== 20160513163056 DeduplicateLinks: migrated (1.0488s) ========================
```

### DIFF
--- a/db/migrate/20160513163056_deduplicate_links.rb
+++ b/db/migrate/20160513163056_deduplicate_links.rb
@@ -1,0 +1,30 @@
+class DeduplicateLinks < ActiveRecord::Migration
+  def up
+    res = Link.connection.execute(<<-SQL
+      SELECT SUM(cnt), COUNT(cnt)
+      FROM (
+        SELECT
+          link_set_id,
+          target_content_id,
+          link_type,
+          COUNT(id) cnt
+        FROM links
+        WHERE target_content_id IS NOT NULL
+        GROUP BY 1,2,3
+        HAVING COUNT(id) > 1
+      ) sub
+      SQL
+    )
+    puts "#{res[0]["count"]} unique records of #{res[0]["sum"]} total duplicated"
+
+    res = Link.connection.execute(<<-SQL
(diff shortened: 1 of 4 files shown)


boffbowsh:
@dougdroper @tvararu

spec/commands/v2/patch_link_set_spec.rb
```diff
@@ -10,7 +10,7 @@
     {
       content_id: content_id,
       links: {
-        topics: topics,
+        topics: topics * 2, # test deduplication
```
tvararu: Pretty cool that the behaviour and updated test case for it all sum up to 9 bytes (`.uniq` and ` * 2`).

db/migrate/20160513163056_deduplicate_links.rb
```diff
@@ -0,0 +1,30 @@
...
+      SQL
+    )
+    puts "#{res[0]["count"]} unique records of #{res[0]["sum"]} total duplicated"
+
+    res = Link.connection.execute(<<-SQL
+      DELETE FROM links USING links l2
+        WHERE links.link_set_id = l2.link_set_id
+        AND links.target_content_id = l2.target_content_id
+        AND links.link_type = l2.link_type
+        AND links.id < l2.id
```
tvararu: Interpreting this out loud using my rusty SQL comprehension:

Deletes all the items from the `links` table that have those 3 columns duplicated in another item with a higher `id`. Because the primary key is sequential and auto-incremented (as is convention), this works out to leave at least one of each, the "newer" (higher `id`) one.

Seems sane to me. 👌

tvararu:
Looks good to me. 👍

PR #305, opened 2016-05-16
New batch read endpoint

MatMoore:
This is a first pass at https://trello.com/c/DYFgXNYr/621-automate-link-checker-tool
Paired with @Davidslv and @benhyland 

This new endpoint allows us to page through content ids and retrieve all the content items
and links associated with that content id.

a response item looks like this:
```ruby
{
                  "content_id" => content_id,
                  "content_items" => [
                    {
                      "locale" => "en",
                      "base_path" => "/such-base-path",
                      "publishing_app" => "whitehall",
                      "format" => "guide",
                      "user_facing_version" => "2",
                      "state" => "draft",
                    },
                    {
                      "locale" => "en",
                      "base_path" => "/such-base-path",
                      "publishing_app" => "whitehall",
                      "format" => "guide",
                      "user_facing_version" => "1",
                      "state" => "published",
                    },
                  ],
                  "links" => {
                    "topics" => [topic_content_id],
                  }
}
```

## Motivation
We are planning to use this in tooling that will check consistency between
rummager and publishing api:
https://github.com/alphagov/finding-things-migration-checker

Content is currently indexed to rummager from multiple sources, and we don't
have a reliable way of rebuilding the index from scratch at the moment. In the
future, content will only be indexed from publishing api.  We want to ensure
that data is kept in sync until then.

We are also in the middle of a data migration from the old tagging
infrastructure (panopticon/content API) to the new one (publishing api/content
store).  Rummager is currently the most complete source of tagging data for
historical content, so comparing publishing API to that gives us a view on any
remaining discrepencies.

## Data returned
For now, we don't apply any filtering (e.g. by publication state). We also
don't want to expand the links hash to include things like base paths, since
this requires more information to uniquely identify the target content item
(such as locale) and we don't actually need this for our usage.

The fields returned for content items are based on the things that influence
content item uniqueness (see doc/object-model-explanation.md), plus
publishing-app and format. We could add more details, but we don't need this
now either.

The data is retrieved using 3 queries. A content ID has many content items and
many links. We query these separately because otherwise we would get a
cartesian product of rows (# of links x # of content items for each content
id).

The pagination works differently to other requests because content id does not
have an inherent ordering. This means that if we define a page by a limit and
and an offset, somebody could publish new content that appears in the middle of
our page between successive queries.

To ensure that the queries links and content items match up to the same set of
content ids, we query all the content ids on the page first, and then pass
those to the other queries.

## Suggestions for code review

We'd like at least one person from the publishing platform team to check the approach we've taken. 

The main commit adds a controller, a query, and a presenter.

The query itself runs all three queries and returns the results grouped together into an array of hashes.

We chose to mixin ActiveModel validations here as we had some fairly standard input validation to do and this seemed like a simple way of managing it. If there is a better way to do this please let us know. We're aware that the format of `render json: { errors: query.errors.full_messages }, status: 422` isn't quite correct for the publishing api but we wanted some feedback before we fix this.

We're using raw SQL because we thought the queries might get complicated, but it seems like we could do the same queries using activerecord too. Do you have a preference one way or the other?

The presenter takes the array returned by the query and returns something that can be returned as JSON to the client. We didn't use ResultPresenter because we are doing pagination a little differently. Passing in the last seen content id for pagination guarantees that if the requester pages through everything, they will never see the same item twice.

## Tests
While testing this I noticed a small issue that breaks CI sometimes:
 https://ci.dev.publishing.service.gov.uk/view/Publishing/job/govuk_publishing-api/303/console

The 2nd commit makes the expectation less strict. I can split this off into a separate PR if you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 
 # Ignore the generated pact
 spec/pacts/publishing_api-content_store.json
+.DS_Store
--- a/app/controllers/v2/grouped_content_and_links_controller.rb
+++ b/app/controllers/v2/grouped_content_and_links_controller.rb
@@ -1,0 +1,26 @@
+module V2
+  #
+  # This controller provides endpoints for querying content and links together.
+  #
+  class GroupedContentAndLinksController < ApplicationController
+    #
+    # Query all content items and links. This can be used to extract data for
+    # batch processing tasks.
+    #
+    # Results are grouped by content id and are paginated using the
+    # last_seen_content_id parameter.
+    #
+    def index
+      query = Queries::GetGroupedContentAndLinks.new(
+        last_seen_content_id: params[:last_seen_content_id],
+        page_size: params[:page_size],
+      )
+
+      if query.valid?
+        render json: Presenters::Queries::GroupedContentAndLinks.new(query.call).present
+      else
+        render json: { errors: query.errors.full_messages }, status: 422
+      end
+    end
+  end
+end
--- a/app/presenters/queries/grouped_content_and_links.rb
+++ b/app/presenters/queries/grouped_content_and_links.rb
@@ -1,0 +1,65 @@
+module Presenters
+  module Queries
+    class GroupedContentAndLinks
+      def initialize(results)
+        self.results = results
+      end
+
+      def present
+        {
+          "last_seen_content_id" => last_seen_content_id,
+          "results" => present_results,
+        }
+      end
+
+    private
+
+      attr_accessor :results
+
+      def present_results
+        results.map { |result| present_result(result) }
+      end
+
+      def last_seen_content_id
+        return if results.empty?
+
+        results.last["content_id"]
+      end
+
+      def present_result(query_result)
+        {
+          "content_id" => query_result["content_id"],
+          "content_items" => present_content_items(query_result["content_items"]),
+          "links" => present_links(query_result["links"])
+        }
+      end
+
+      def present_content_items(content_items)
+        content_items.map { |content_item| present_content_item(content_item) }
+      end
+
+      def present_content_item(content_item)
+        content_item.slice(
+          "locale",
+          "base_path",
+          "publishing_app",
+          "format",
+          "user_facing_version",
+          "state"
+        )
+      end
+
+      # Links is an array of individual link rows
+      # So we need to group them by type
+      def present_links(links)
+        groups = links.group_by { |link| link.fetch("link_type") }
+
+        groups.each_with_object({}) do |(link_type, type_links), links_hash|
+          links_hash[link_type] = type_links.map do |link|
+            link.fetch("target_content_id")
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/queries/get_grouped_content_and_links.rb
+++ b/app/queries/get_grouped_content_and_links.rb
@@ -1,0 +1,128 @@
+module Queries
+  class GetGroupedContentAndLinks
+    PAGE_SIZE = 10
+    MAX_PAGE_SIZE = 1000
+
+    include ActiveModel::Validations
+
+    attr_accessor :last_seen_content_id, :page_size
+
+    validates_numericality_of :page_size, less_than_or_equal_to: MAX_PAGE_SIZE, greater_than: 0
+
+    def initialize(last_seen_content_id: nil, page_size: nil)
+      @last_seen_content_id = last_seen_content_id
+      @page_size = page_size || PAGE_SIZE
+    end
+
+    # Returns an array of hashes where each hash contains a content_id, an
+    # array of content item rows, and an array of link rows.
+    #
+    # The pagination works differently to other requests because content ids
+    # are not generated in any kind of order. When paging through the results
+    # we need to be sure of:
+    #
+    # - not missing a content_id that already existed when we started the
+    #   first page
+    #
+    # - not seeing any content_id twice
+    #
+    # - seeing complete and consistent data for each content_id returned
+    def call
+      content_ids = query_content_ids_for_page(
+        last_seen_content_id: last_seen_content_id,
+        page_size: page_size
+      )
+
+      group_results(content_results(content_ids), link_set_results(content_ids))
+    end
+
+  private
+
+    def query_content_ids_for_page(last_seen_content_id:, page_size:)
+      groups = ContentItem.group(:content_id)
+
+      if last_seen_content_id.nil?
+        page = groups
+          .limit(page_size)
+          .order(:content_id)
+      else
+        page = groups
+          .having("content_id > ?", last_seen_content_id)
+          .limit(page_size)
+          .order(:content_id)
+      end
+
+      page.pluck(:content_id)
+    end
+
+    def group_results(content_results, link_set_results)
+      grouped_content_items = group_by_content_id(content_results)
+      grouped_links = group_by_content_id(link_set_results)
+
+      grouped_content_items.map do |content_id, content_items|
+        {
+          "content_id" => content_id,
+          "content_items" => content_items,
+          "links" => grouped_links[content_id] || []
+        }
+      end
+    end
+
+    def group_by_content_id(results)
+      results.group_by { |item| item["content_id"] }
+    end
+
+    def content_results(content_ids)
+      return [] if content_ids.empty?
+
+      query = <<-SQL
+        SELECT
+          content_items.content_id,
+          content_items.id AS content_item_id,
+          translations.locale,
+          locations.base_path,
+          states.name AS state,
+          user_facing_versions.number AS user_facing_version,
+          content_items.publishing_app,
+          content_items.format
+        FROM content_items
+        JOIN translations
+          ON translations.content_item_id = content_items.id
+        JOIN locations
+          ON locations.content_item_id = content_items.id
+        JOIN states
+          ON states.content_item_id = content_items.id
+        JOIN user_facing_versions
+          ON user_facing_versions.content_item_id = content_items.id
+        WHERE
+          content_items.content_id IN (#{sql_value_placeholders(content_ids.size)})
+        ORDER BY
+          content_items.updated_at DESC
+      SQL
+
+      ActiveRecord::Base.connection.raw_connection.exec(query, content_ids)
+    end
+
+    def link_set_results(content_ids)
+      return [] if content_ids.empty?
+
+      query = <<-SQL
+        SELECT
+          link_sets.content_id,
+          links.link_type,
+          links.target_content_id
+        FROM link_sets
+        JOIN links
+          ON link_sets.id = links.link_set_id
+        WHERE
+          link_sets.content_id IN (#{sql_value_placeholders(content_ids.size)})
+      SQL
+
+      ActiveRecord::Base.connection.raw_connection.exec(query, content_ids)
+    end
+
+    def sql_value_placeholders(number)
+      (1).upto(number).map { |i| "$#{i}" }.join(',')
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,6 +27,8 @@ Rails.application.routes.draw do
       get "/linked/:content_id", to: "link_sets#get_linked"
 
       get "/linkables", to: "content_items#linkables"
+
+      get "/grouped-content-and-links", to: "grouped_content_and_links#index"
     end
   end
 
--- a/spec/presenters/queries/grouped_content_and_links_spec.rb
+++ b/spec/presenters/queries/grouped_content_and_links_spec.rb
@@ -1,0 +1,89 @@
+require "rails_helper"
+
+RSpec.describe Presenters::Queries::GroupedContentAndLinks do
+  describe "#present" do
+    context "when receives an empty array" do
+      it "returns a hash with empty results" do
+        presenter = Presenters::Queries::GroupedContentAndLinks.new([])
+
+        expect(presenter.present).to eq(
+          "last_seen_content_id" => nil,
+          "results" => []
+        )
+      end
+    end
+
+    context "when receives an array of results" do
+      let(:content_id) { SecureRandom.uuid }
+      let(:topic_content_id) { SecureRandom.uuid }
+
+      before do
+        FactoryGirl.create(
+          :content_item,
+          content_id: content_id,
+          base_path: "/vat-rates",
+          publishing_app: "whitehall",
+          locale: "en",
+          format: "guide",
+          user_facing_version: 1,
+          state: "published"
+        )
+
+        FactoryGirl.create(
+          :content_item,
+          content_id: content_id,
+          base_path: "/vat-rates",
+          publishing_app: "whitehall",
+          locale: "en",
+          format: "guide",
+          user_facing_version: 2,
+          state: "draft",
+        )
+
+        FactoryGirl.create(
+          :link_set,
+          content_id: content_id,
+          links_hash: {
+            "topics" => [topic_content_id]
+          }
+        )
+      end
+
+      it "returns a hash with grouped results" do
+        results = ::Queries::GetGroupedContentAndLinks.new.call
+        presenter = Presenters::Queries::GroupedContentAndLinks.new(results)
+
+        presented_results = presenter.present["results"]
+
+        expect(presented_results).to eq(
+          [
+            {
+              "content_id" => content_id,
+              "content_items" => [
+                {
+                  "locale" => "en",
+                  "base_path" => "/vat-rates",
+                  "publishing_app" => "whitehall",
+                  "format" => "guide",
+                  "user_facing_version" => "2",
+                  "state" => "draft",
+                },
+                {
+                  "locale" => "en",
+                  "base_path" => "/vat-rates",
+                  "publishing_app" => "whitehall",
+                  "format" => "guide",
+                  "user_facing_version" => "1",
+                  "state" => "published",
+                },
+              ],
+              "links" => {
+                "topics" => [topic_content_id],
+              }
+            }
+          ]
+        )
+      end
+    end
+  end
+end
--- a/spec/queries/get_grouped_content_and_links_spec.rb
+++ b/spec/queries/get_grouped_content_and_links_spec.rb
@@ -1,0 +1,217 @@
+require "rails_helper"
+
+RSpec.describe Queries::GetGroupedContentAndLinks do
+  let(:ordered_content_ids) do
+    %w(
+      5f612a9f-7d3f-4b4c-a35e-ebd0b1e79019
+      fd161210-3e12-41e7-8e5e-c8cef607a95f
+    )
+  end
+
+  def new_content_item
+    SecureRandom.uuid
+  end
+
+  describe "#call" do
+    context "when no results exist" do
+      it "returns an empty array" do
+        expect(subject.call).to be_empty
+      end
+    end
+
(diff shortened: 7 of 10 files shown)


MatMoore:
cc @alphagov/team-finding-things

app/controllers/v2/grouped_content_and_links_controller.rb
```diff
@@ -0,0 +1,26 @@
+module V2
+  #
+  # This controller provides endpoints for querying content and links together.
+  #
+  class GroupedContentAndLinksController < ApplicationController
```
tijmenb: Should we name this closer to what we call it? `BatchReadController`?

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
+module Queries
+  class GetGroupedContentAndLinks
```
tijmenb: Would it be interesting to explain the reason for using the `content_id` in pagination here?

config/routes.rb
```diff
@@ -27,6 +27,8 @@
       get "/linked/:content_id", to: "link_sets#get_linked"
 
       get "/linkables", to: "content_items#linkables"
+
+      get "/grouped_content_and_links", to: "grouped_content_and_links#index"
```
tijmenb: I think the pattern is to use dashes in paths.

spec/presenters/queries/grouped_content_and_links_spec.rb
```diff
@@ -0,0 +1,93 @@
+require "rails_helper"
+
+module Presenters
+  module Queries
```
tijmenb: You can do `RSpec.describe Presenters:: Queries::GroupedContentAndLinks` so it doesn't need the deep nesting.

spec/requests/grouped_content_and_links_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      expect(response.status).to eq(200)
+      expect(results(response)).to match_array(expected_result)
+    end
+  end
+
+private
+
+  def results(response)
+    JSON.parse(response.body, symbolize_names: true)[:results]
```
tijmenb: You can replace this with `parsed_respons["results"]` (https://github.com/alphagov/publishing-api/pull/299)

spec/queries/get_grouped_content_and_links_spec.rb
```diff
@@ -0,0 +1,211 @@
...
+        @content_item = FactoryGirl.create(
+          :content_item,
+          content_id: ordered_content_ids.first,
+          base_path: "/capital-gains-tax",
+          state: "published"
+        )
+      end
+
+      context "with no links" do
+        it "returns the content item with empty links" do
```
tijmenb: This may be testing more than the description - perhaps you could just assert `expect(result["links"]).to eq([])` here, since the presence of the content is already tested in the request tests? Or perhaps split it up somehow?

config/routes.rb
MatMoore: Yep. Actually we had trouble coming up with a descriptive name at all, so I'm interested if anyone has suggestions for something better to call this endpoint.

I don't think we have established language for talking about the thing that the content ID identifies - it's not something that's explicitly named in the data model at the moment. Other apps have a Document/Edition distinction but publishing api just has groups of Content Items that relate to each other.

spec/queries/get_grouped_content_and_links_spec.rb
MatMoore: I think splitting it up into two tests would work. Will take a look.

app/controllers/v2/grouped_content_and_links_controller.rb
MatMoore: I'm not sure - I don't think either name describes this very well. Does anyone have any other suggestions?

dougdroper:
I am curious about the pagination here, it seems to add a lot of incidental complexity to the implementation. 
I had a look at one of the content_items that has about 76 versions, and ran this query against it:
```sql
select content_items.id, content_items.format, content_items.content_id, translations.locale, locations.base_path, states.name, lock_versions.number from content_items
join locations on locations.content_item_id = content_items.id
join states on states.content_item_id = content_items.id
join lock_versions on lock_versions.target_id = content_items.id
join translations on translations.content_item_id = content_items.id
where content_items.content_id = '08d48cdd-6b50-43ff-a53b-beab47f4aab0'
group by content_items.id, translations.locale, locations.base_path, states.name, lock_versions.number
order by number DESC
```
and that takes about 0.2ms, so might be able to get away without pagination. 

Also this kinda feels like it is more like `Versions` rather then `GroupedContentAndLinks` so might need some thinking about how other people could use it too (specifically around the fields that would be displayed)

What do you think?

benhyland:
@dougdroper 

Hi Doug, I just want to explain a bit more about the motivation for the pagination here, since we might be understanding different things by the term.

The new endpoint is intended to be used for bulk reading, in batches of (say) 1000 content_ids per request.

We organise the response data under a hierarchy keyed by content_id because the endpoint is intended to deliver data which only makes sense per-content_id, such as links, in addition to the per-content_item_id data.

We therefore wish to paginate in pages of content_ids. In order to keep our pages stable in the face of edits that happen during an iteration across all pages, we must pass in the last_seen_content_id, rather than a numeric offset, since content_ids are not generated in sequence.

The result is that in a single iteration we can be sure of
- not missing a content_id that already existed when we started the first page
- not seeing any content_id twice
- seeing complete and consistent data for each content_id returned

If we iterate by some other key, such as content_item_id, or by pages using an offset rather than last_seen_content_id, these guarantees cannot be made.

Hope that makes sense - we can come and chat in person if you like?

benhyland:
@dougdroper regarding the naming and more general use cases, we tried to ensure that the endpoint could be useful to others as well, but we wanted to gather feedback before adding more fields or filtering.
Any suggestions (especially of specific use cases) are very welcome. Thanks!

dougdroper:
Ah right I see, so the response would something like:

```
{
  'total': 500000,
  'page': 1,
  ...
  'results': {
   [
     { 'content_id': '123', 'content_items': [{"base_path" => "/foo"}], 'links': {} },
     { 'content_id': '456', 'content_items': [{"base_path" => "/bar"}], 'links': {} }
   ]
  }
}
```

MatMoore:
@dougdroper Yep, that's right, except rather than `page` we return `last_seen_content_id`, for the reasons @benhyland  mentioned above.

We don't return `total` at the moment, but we can add that it in if it's going to be useful. We seem to be missing a couple of specs for the pagination response fields, so we'll need to take a look at that.

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+        SELECT
+          link_sets.content_id,
+          links.link_type,
+          links.target_content_id
+        FROM
+          link_sets
+        JOIN
+          links on link_sets.id = links.link_set_id
+
+        WHERE link_sets.content_id IN (#{sql_value_placeholders(content_ids)})
```
kaplan9000: `sql_value_placeholders` might never be anything dangerous right now but I think that adding an extra level of security around this query will not hurt and will prevent SQL injection in the case that in the future the code will change but the query will remain the same. 

Sadly, I appear not to find an easy way to escape SQL in multiline strings.

jscott06: sql_value_placeholders might never be anything dangerous right now but I think that adding an extra level of security around this query will not hurt and will prevent SQL injection in the case that in the future the code will change but the query will remain the same.

Sadly, I appear not to find an easy way to escape SQL in multiline strings.

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+          ci.format
+
+        FROM
+          content_items ci
+        JOIN translations t on t.content_item_id = ci.id
+        JOIN locations loc on loc.content_item_id = ci.id
+        JOIN states s on s.content_item_id = ci.id
+        JOIN user_facing_versions ufv on ufv.content_item_id = ci.id
+
+        WHERE ci.content_id IN (#{sql_value_placeholders(content_ids)})
```
jackscotti: sql_value_placeholders might never be anything dangerous right now but I think that adding an extra level of security around this query will not hurt and will prevent SQL injection in the case that in the future the code will change but the query will remain the same.

Sadly, I appear not to find an easy way to escape SQL in multiline strings.

app/queries/get_grouped_content_and_links.rb
jackscotti: same as above

.gitignore
```diff
@@ -16,3 +16,4 @@
 
 # Ignore the generated pact
 spec/pacts/publishing_api-content_store.json
+.DS_Store
```
jackscotti: I have added this to the `gitignore` in the past and I've been told that as because not every contributor uses a machine running Apple OS X, the `.DS_Store` should be ignored locally in the global `gitignore` file. 

http://technicalpickles.com/posts/configure-git-to-globally-ignore-some-files/

app/queries/get_grouped_content_and_links.rb
dougdroper: http://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_array

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    attr_accessor :last_seen_content_id, :page_size
+
+    validate :has_valid_uuid_for_pagination
+    validates_numericality_of :page_size, less_than_or_equal_to: MAX_PAGE_SIZE # FIXME: > 0
+
+    def initialize(last_seen_content_id: nil, page_size: nil)
+      @last_seen_content_id = last_seen_content_id || DEFAULT_CONTENT_ID
+      @page_size = page_size || PAGE_SIZE
+    end
```
jackscotti:
```ruby
def initialize(last_seen_content_id: DEFAULT_CONTENT_ID, page_size: DEFAULT_PAGE_SIZE)
   @last_seen_content_id = last_seen_content_id
   @page_size = page_size
end
```

Would this be more readable and work the same way?

benhyland: These are semantically different. With your proposed change, it is possible to pass and assign nil. In the original it is not.

dougdroper: I believe if you call with `last_seen_content_id: nil`, it would override `DEFAULT_CONTENT_ID` with `nil` sadly

jackscotti: oh. Learnt something new, thanks!

.gitignore
Davidslv: There's nothing wrong in adding it to the repo, it should be there to avoid the mistake of adding a file that shouldn't be in the repo in the first place. If someone doesn't use a OSX machine that will have no impact in their workflow whatsoever.

dougdroper:
What is the rational around returning all versions of the content_items?

benhyland:
@dougdroper Only that we haven't yet decided if or how we wish to filter them.

As an aside, at the moment we think it will be more flexible and maintainable for the future if we retrieve all versions from the db and filter in the application layer. The tradeoff is that the potentially unbounded versions in the result set might cause perf/memory issues. We don't expect this and we haven't looked into other potential mitigations such as streaming the response.

app/queries/get_grouped_content_and_links.rb
MatMoore: We're not trying to inject anything into the query here, only build up a dynamic list of placeholders like `IN ($1, $2, $3)`. In the `exec` call we pass the SQL and the values to the database separately, so this should be safer than trying to sanitise strings.

I think this query could be rewritten using `ContentItemFilter` though, which might make the code more readable. Any thoughts?

spec/presenters/queries/grouped_content_and_links_spec.rb
MatMoore: Thanks, I've fixed this now.

spec/requests/grouped_content_and_links_spec.rb
MatMoore: Fixed

spec/queries/get_grouped_content_and_links_spec.rb
MatMoore: I've split the other assertions out into different tests. There's a bit of overlap but I think it's useful to test presence of content here too as it helps narrow down the source of a bug if you can rule out the presentation layer.

app/presenters/queries/grouped_content_and_links.rb
```diff
@@ -0,0 +1,68 @@
...
+    private
+
+      attr_accessor :results
+
+      def present_results
+        results.map { |result| present_result(result) }
+      end
+
+      def last_seen_content_id
+        return nil if results.empty?
```
dougdroper: the use of nil is redundant here
`return if results.empty?`

app/presenters/queries/grouped_content_and_links.rb
```diff
@@ -0,0 +1,68 @@
...
+        )
+      end
+
+      # Links is an array of individual link rows
+      # So we need to group them by type
+      def present_links(links)
+        groups = links.group_by { |link| link.fetch("link_type") }
+        links_hash = {}
+
+        groups.each do |link_type, type_links|
```
dougdroper: can use `each_with_object({})` to remove the links_hash variable

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+      content_ids = query_content_ids_for_page(
+        last_seen_content_id: @last_seen_content_id,
+        page_size: @page_size
+      )
+
+      group_results(content_results(content_ids), link_set_results(content_ids))
+    end
+
+    def has_valid_uuid_for_pagination
+      unless @last_seen_content_id == DEFAULT_CONTENT_ID || UuidValidator.valid?(@last_seen_content_id)
```
dougdroper: Why do we need a UuidValidator? Maybe easier to see if a content_item exists with the content_id? Then we can drop the DEFAULT_CONTENT_ID

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+    def group_by_content_id(results)
+      results.group_by { |item| item["content_id"] }
+    end
+
+    def content_results(content_ids)
+      return [] if content_ids.empty?
+
+      query = <<-SQL
+        SELECT
+          ci.content_id as content_id,
```
dougdroper: Can we use the names of the tables rather then the shortcut aliases, I think it is tricky to keep doing the conversion and helps with reading

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+    PAGE_SIZE = 10
+    MAX_PAGE_SIZE = 1000
+    DEFAULT_CONTENT_ID = '00000000-0000-0000-0000-000000000000'.freeze
+
+    include ActiveModel::Validations
+
+    attr_accessor :last_seen_content_id, :page_size
+
+    validate :has_valid_uuid_for_pagination
+    validates_numericality_of :page_size, less_than_or_equal_to: MAX_PAGE_SIZE # FIXME: > 0
```
dougdroper: What does the comment # FIXME: > 0 mean?

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+  private
+
+    def query_content_ids_for_page(last_seen_content_id:, page_size:)
+      ContentItem
+        .select(:content_id)
+        .group(:content_id)
+        .having("content_id > ?", [last_seen_content_id])
+        .limit(page_size)
+        .order(:content_id)
+        .map(&:content_id)
```
dougdroper: can we change this to `pluck(:content_id)`

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+      end
+    end
+
+  private
+
+    def query_content_ids_for_page(last_seen_content_id:, page_size:)
+      ContentItem
+        .select(:content_id)
+        .group(:content_id)
+        .having("content_id > ?", [last_seen_content_id])
```
dougdroper: why is this an array?

app/queries/get_grouped_content_and_links.rb
```diff
@@ -0,0 +1,116 @@
...
+    def initialize(last_seen_content_id: nil, page_size: nil)
+      @last_seen_content_id = last_seen_content_id || DEFAULT_CONTENT_ID
+      @page_size = page_size || PAGE_SIZE
+    end
+
+    # Returns an array of hashes where each hash contains a content_id, an
+    # array of content item rows, and an array of link rows.
+    def call
+      content_ids = query_content_ids_for_page(
+        last_seen_content_id: @last_seen_content_id,
```
dougdroper: we have an attr_reader for last_seen_content_id and page_size

app/queries/get_grouped_content_and_links.rb
MatMoore: It just makes for a friendlier API.

If the request is made with an invalid UUID, then the client is doing something wrong, and we could tell them why.

In this case though it's inconsistent with the other requests so I'll take it out. This means we will return 404 in all cases.

MatMoore:
@dougdroper @tijmenb I've addressed your comments in a new commit.

If there are no other issues that need addressing now I'll rebase everything on top of the latest master (some integration tests are failing at the moment, but it seems unrelated to this work)

To elaborate on what we've discussed offline:

- We plan to run our checker tool only on integration for the time being
- @elliotcm is currently working on derived tables to improve read performance
- We don't actually need all versions of the content, although the state of the latest version is not enough for us to filter on. We think we should be able to rewrite the query to use Elliot's new tables when they're ready.
- We'd prefer to get this PR merged in the meantime, so we can start deploying the tool to integration.

I've raised this ticket in our backlog specifically to address the remaining work https://trello.com/c/3qIYFDBI/634-finish-publishing-api-endpoint-for-link-checker-tool

We haven't come up with a good name for the endpoint yet - I think we should rename it when we remove the multiple versions.

app/queries/get_grouped_content_and_links.rb
MatMoore: I've added a comment further down that explains why pagination works the way it does.

elliotcm:
Based on conversations offline (mentioned above) it doesn't like we're going to resolve everything in this PR.

I'm OK with us merging this based on the agreement that [this ticket](https://trello.com/c/3qIYFDBI/634-finish-publishing-api-endpoint-for-link-checker-tool) will be used to take a fresh look at the needs for this endpoint and the ways we can provide a robust endpoint via some careful denormalisation.  A member of the publishing platform team will be available to help with the discussion/planning of that, with the finding things team taking on the bulk of the actual typing.

app/queries/get_grouped_content_and_links.rb
dougdroper: I meant rather that we could do `ContentItem.exists?(last_seen_content_id)` as the validation. But am fine with it not validating.

elliotcm:
Get this green and I'll merge it.

MatMoore:
@elliotcm thanks, should be ready to merge now.

PR #308, opened 2016-05-16
Add dependees to expansion group.

elliotcm:
https://trello.com/c/2BMiIf0y/671-8-add-reverse-dependency-groups-to-the-expanded-links-field

This also changes the behaviour to not include keys
for link types if there are no valid content items
to expand for that link type.

### DIFF
--- a/app/presenters/queries/expand_link.rb
+++ b/app/presenters/queries/expand_link.rb
@@ -1,60 +1,23 @@
 module Presenters
   module Queries
     class ExpandLink
-      attr_reader :item, :type, :visited
+      attr_reader :item, :type
 
-      def initialize(item, type, visited, set:)
+      def initialize(item, type, rules)
         @item = item
         @type = type
-        @visited = visited
-        @set = set
+        @rules = rules
       end
 
       def expand_link
-        expanded = rules.expansion_fields(type).each_with_object({}) do |field, subhash|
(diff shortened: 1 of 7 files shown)


dougdroper:
I thought there was a requirement to not expand different link types after the first parent/child? @tuzz?

tuzz:
@dougdroper: Yeah, I don't think we should "cross the streams" and expand `parent` then expand anything other than `parent` in that hierarchy of expansion.

The reason not to do this is because it makes it very difficult to identify which content items are affected when content item X has changed. We can no longer look at just the expansion rules for its link types, we have to look at all link types and feed that knowledge into the recursive query – which would be mind-boggling-ly difficult I reckon.

So although I think this change would be great, I don't think the "detection of change" part of the system will line up with it. This would result in stale content, i.e. content that has been rendered in the expanded_links that has since changed.

elliotcm:
Boo, that makes this side much more complicated.

elliotcm:
@dougdroper @tuzz Updated

dougdroper:
Looks good to me, `#expand_links` feels a bit bloated to me and might be good to do some minor refactoring but otherwise 👍

elliotcm:
@dougdroper I couldn't find a way to tidy it up further without additional DB requests, any suggestions?

elliotcm:
@dougdroper Updated.

PR #314, opened 2016-05-18
Dependency queue

dougdroper:
When sending downstream, ensure that the dependencies are also checked
and sent downstream. 

https://trello.com/c/pXgLk6jB/646-5-enqueue-sidekiq-jobs-to-check-dependencies

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -52,7 +52,7 @@ module Commands
 
         PresentedContentStoreWorker.perform_async(
           content_store: Adapters::DraftContentStore,
-          payload: { content_item: live.id, payload_version: event.id },
+          payload: { content_item_id: live.id, payload_version: event.id },
           request_uuid: GdsApi::GovukHeaders.headers[:govuk_request_id],
         )
       end
--- a/app/workers/dependency_resolution_worker.rb
+++ b/app/workers/dependency_resolution_worker.rb
@@ -28,7 +28,7 @@ private
   end
 
   def present_content_store(dependent_content_id)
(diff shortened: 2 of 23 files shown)


app/commands/v2/discard_draft.rb
```diff
@@ -52,7 +52,7 @@ def send_live_to_draft_content_store(live)
 
         PresentedContentStoreWorker.perform_async(
           content_store: Adapters::DraftContentStore,
-          payload: { content_item: live.id, payload_version: event.id },
+          payload: { content_id: live.id, payload_version: event.id },
```
danielroseman: I got a little confused here; I don't think this helps reduce the ambiguity. To me "content_id" implies the content_id UUID, which is the opposite of what you intended. Perhaps "content_item_id"?

app/workers/dependency_resolution_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def content_item_dependees
+    Queries::ContentDependencies.new(content_id: content_id,
+                                     fields: fields,
+                                     dependent_lookup: Queries::GetDependees.new).call
+  end
+
+  def present_content_store(dependent_content_id)
+    latest_content_item ||= Queries::GetLatest.call(
```
danielroseman: Don't think we want `||=` here, it can't have any previous value.

app/commands/v2/discard_draft.rb
dougdroper: Yes sure, 👍

danielroseman:
Looks good although I really don't have enough context to comment on the GetDependents stuff.

PR #315, opened 2016-05-18
Improve performance of uniqueness validator

boffbowsh:
Uses Rails’ eager loading of relationships to reduce round trips to the
database, and allows Postgres to optimise queries better.

Also uses the database to filter out the current content item when
calculating uniqueness, instead of instantiating records and performing
set math.

Adds supporting object relationships and scope to `ContentItem`, which
will be useful for other performance improvements.

Before (over 100 random records):
```
$ bundle exec ruby bench/content_item_uniqueness_validator.rb
  0.590000   0.050000   0.640000 (  1.721408)
```

After:
```
$ bundle exec ruby bench/content_item_uniqueness_validator.rb
  0.550000   0.040000   0.590000 (  1.390422)
```

I ran the above benchmarks multiple times and they were consistent enough for me to present this with confidence.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -28,6 +28,15 @@ class ContentItem < ActiveRecord::Base
   NON_RENDERABLE_FORMATS = %w(redirect gone).freeze
   EMPTY_BASE_PATH_FORMATS = %w(government).freeze
 
+  has_one :state
+  has_one :location
+  has_one :translation
+  has_one :user_facing_version
+
+  scope :with_supporting_objects, -> {
+    eager_load(:state, :location, :translation, :user_facing_version)
+  }
+
   scope :renderable_content, -> { where.not(format: NON_RENDERABLE_FORMATS) }
 
   validates_with SchemaNameFormatValidator
--- a/app/validators/content_item_uniqueness_validator.rb
+++ b/app/validators/content_item_uniqueness_validator.rb
@@ -1,16 +1,17 @@
 class ContentItemUniquenessValidator < ActiveModel::Validator
   def validate(record)
-    content_item = record.content_item
+    content_item = ContentItem.with_supporting_objects.where(id: record.content_item_id).first
+    return unless content_item
 
     state = record if record.is_a?(State)
     translation = record if record.is_a?(Translation)
     location = record if record.is_a?(Location)
     user_facing_version = record if record.is_a?(UserFacingVersion)
 
-    state ||= State.find_by(content_item: content_item)
-    translation ||= Translation.find_by(content_item: content_item)
-    location ||= Location.find_by(content_item: content_item)
-    user_facing_version ||= UserFacingVersion.find_by(content_item: content_item)
+    state ||= content_item.state
+    translation ||= content_item.translation
+    location ||= content_item.location
+    user_facing_version ||= content_item.user_facing_version
 
     return unless state && translation && location && user_facing_version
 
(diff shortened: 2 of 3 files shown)


app/validators/content_item_uniqueness_validator.rb
```diff
@@ -29,16 +30,15 @@ def validate(record)
...
       state: state_name,
       locale: locale,
       base_path: base_path,
       user_version: user_version,
     )
 
-    additional_items = matching_items - [content_item]
-
-    if additional_items.any?
+    if matching_items.any?
```
h-lame: I wondered if there was another micro-optimisation if we used `.count > 0` here rather than `.any?`.  I thought that `.any?` would load all the records from the DB and use ruby stdlib `Enumerable#any?` which would be slow if there were many records, but it turns out that's not what it does as `ActiveRecord::Relation#any?` basically does `ActiveRecord::Relation#count` anyway (see: https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/activerecord/lib/active_record/relation.rb#L297-L300 and https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/activerecord/lib/active_record/relation.rb#L278-L288).

TIL &c.

h-lame:
:clap: on spotting this and making the benchmark to prove it!

app/validators/content_item_uniqueness_validator.rb
boffbowsh: I wouldn't call that a micro optimisation if it were the case, but yes this is a rare example of AR doing the right thing :)

app/models/content_item.rb
```diff
@@ -28,6 +28,15 @@ class ContentItem < ActiveRecord::Base
   NON_RENDERABLE_FORMATS = %w(redirect gone).freeze
   EMPTY_BASE_PATH_FORMATS = %w(government).freeze
 
+  has_one :state
+  has_one :location
+  has_one :translation
+  has_one :user_facing_version
+
```
danielroseman: We've resisted adding the relationship accessors to ContentItem for fear of turning it into a "god object" (on the advice of @tuzz ). If there's a significant improvement in performance and readability as a result of doing this, then we can certainly discuss revisiting that decision.

boffbowsh: There's quite a few places in publishing api that would benefit from eager loading, and the easiest way to do this is to let Rails manage it for us. I think while we have this multi-table model it makes sense to keep access between the objects as convenient as possible. As a new developer on publishing api I'd expect these associations to exist. Calling `Thing.find_by!(content_item: content_item)` in so many places seems wrong, and doesn't allow Rails to memoize objects correctly. This PR is one example, but if the calling code of this had also had its objects eager loaded, there would be just a single query at the end to find duplicates.

elliotcm: I don't want us to make this sort of decision in the context of a single change.  The current structure isn't there due to ignorance of ActiveRecord, it's there for a reason.

If we want to change approach it's something we should get together and discuss in a wider context.

tuzz:
Is this, on average, taking a query down from 17ms to 14ms or am I misreading the benchmark?

What's driving this. Is there a batch task that's taking too long?

elliotcm:
I'm not too familiar with ruby benchmarking but can I confirm that (1721ms - 1390ms) / 100 records = 3.1ms saved per write-request?

app/models/content_item.rb
boffbowsh: I agree, we need a wider discussion on this soon.

elliotcm: Book something in.

boffbowsh:
Creating each supporting object calls this validator, so its 4 calls per write request.

boffbowsh:
Closing this for now, to be discussed in meeting next week

PR #316, opened 2016-05-19
[Discuss] Don't present full content item to caller

boffbowsh:
Rendering the full content item on every PutContent call is expensive
and possibly unnecessary. Clients probably won’t need to have their
request rendered back to them. Whitehall certainly doesn’t, I haven’t
checked other apps.

Before:
```
$ bundle exec ruby bench/put_content.rb
  4.730000   0.220000   4.950000 (  7.536762)
4806 SQL queries
```

After:
```
$ bundle exec ruby bench/put_content.rb
  4.280000   0.210000   4.490000 (  6.338519)
4606 SQL queries
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,4 +46,5 @@ group :development, :test do
   gem "factory_girl_rails", "4.5.0"
   gem "pact_broker-client"
   gem "govuk-lint"
+  gem "faker", require: false
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,6 +92,8 @@ GEM
     factory_girl_rails (4.5.0)
       factory_girl (~> 4.5.0)
       railties (>= 3.0.0)
+    faker (1.6.3)
+      i18n (~> 0.5)
     faraday (0.9.2)
       multipart-post (>= 1.2, < 3)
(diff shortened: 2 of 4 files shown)


danielroseman:
It would be worth looking at whether specialist-publisher-rebuild, as the first of the phase 2 apps, uses the response.

tijmenb:
I like this. As for other apps, the "taxon manager" inside collections-publisher is also a phase 2 app, and it [doesn't use the return value](https://github.com/alphagov/collections-publisher/blob/master/app/forms/taxon_form.rb#L30-L33). 

Does it also apply to `publish` and `patch_links`?

boffbowsh:
Looking through SP v2 it looks like it also ignores the response of put_content. Paging @tuzz and @benilovj for people with more info.

tuzz:
It doesn't look like specialist-publisher-rebuild uses this.

This may be contentious, but as someone who has recently become a consumer of the Publishing API, it'd be handy if every response was standardised to include the full content item. It'd make things simpler, I think – although this would likely make request a bit slower.

boffbowsh:
Closing this for now, to be discussed in meeting next week

PR #317, opened 2016-05-19
Require explicit FactoryGirl calls

elliotcm:
We've been using these inconsistently but leaning
towards the explicit version (in line with reducing
sneaky RSpec monkeypatching in general).

This makes it a requirement.


danielroseman:
👍 , needs a rebase though.

elliotcm:
@danielroseman rebased

PR #321, opened 2016-05-23
Support new linkables endpoint

elliotcm:
https://trello.com/c/3PfssMJ2/716-more-robust-linkables-endpoint

Builds a slightly denormalised table to support quick reads for the `/v2/linkables` endpoint.  This is the sort of approach I'd like to recommend for many of our read endpoints where possible, since even a single table approach would require dealing with grouping and sub-sorting to find e.g. the latest of a given content item with state fallbacks.

There's another branch to switch the endpoint over to reading from this table once everything is deployed and ready: https://github.com/alphagov/publishing-api/tree/activate-new-linkables-endpoint

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -64,6 +64,7 @@ module Commands
         UserFacingVersion.find_by(content_item: draft).try(:destroy)
         LockVersion.find_by(target: draft).try(:destroy)
         AccessLimit.find_by(content_item: draft).try(:destroy)
+        Linkable.find_by(content_item: draft).try(:destroy)
       end
 
       def increment_live_lock_version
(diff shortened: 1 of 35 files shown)


boffbowsh:
Can you link the other branch please? I'd like to see how this is used.

elliotcm:
@boffbowsh https://github.com/alphagov/publishing-api/tree/activate-new-linkables-endpoint

danielroseman:
👍 

My only concern is that this adds yet another action in the long list of things that happen at publish time. And if we're doing a few of these denormalised tables, this will just get worse. Do we need to think about splitting this out somehow?

elliotcm:
I think we could do with having a team chat about how to handle all the things that happen during publish (and other actions) more cleanly rather than this one specifically.  As you say it's a very long list.

PR #322, opened 2016-05-23
Expand links according to locale

danielroseman:
When expanding links for an item in a non-default locale, return items in that same locale if they exist, otherwise fall back to the default locale.


dougdroper:
LGTM 👍

PR #323, opened 2016-05-23
Send all translations on PATCH links

boffbowsh:
When a PATCH links call is received, the links are updated for all
translations of the specified content_id. Therefore, we should
re-render and re-send all translations to the relevant content store.

/cc @tijmenb @danielroseman


tijmenb:
👍

PR #324, opened 2016-05-23
Add a spec showing first_published_at can be set

rgarner:
During a call to put_content, demonstrate that first_published_at can
be set by the caller rather than using a default value


tuzz:
Tested locally. Passes. Merging.

PR #330, opened 2016-05-27
Validate content_item and links against schema

dougdroper:
Schema's are made available to the application by syncing them to the
repo in production. This validates against them and sends any errors to
errbit for the time being so as to not block.

https://trello.com/c/O8qZQlLh/720-validate-schemas-in-publishing-api-for-error-messaging

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -114,15 +114,15 @@ module Commands
         PublishingAPI.service(:queue_publisher).send_message(payload)
       end
 
-      def format
+      def schema_name
         Queries::GetLatest.call(
           ContentItem.where(content_id: content_id)
-        ).last.try(:format)
+        ).limit(1).pluck(:schema_name).last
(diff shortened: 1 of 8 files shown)


app/commands/v2/patch_link_set.rb
```diff
@@ -112,6 +113,19 @@ def send_to_message_queue(content_item)
 
         PublishingAPI.service(:queue_publisher).send_message(payload)
       end
+
+      def format
+        Queries::GetLatest.call(
+          ContentItem.where(content_id: content_id)
+        ).last.try(:format)
```
boffbowsh: This is an expensive query. Can we get the format from elsewhere? At the least, can we do something like:
```ruby
Queries::GetLatest.call(
  ContentItem.where(content_id: content_id)
).limit(1).pluck(:format)
```
That way we don't need to instantiate ActiveRecord.

Also, I think this should be `schema_name` now, not `format`.

PR #331, opened 2016-05-31
Do not validate PatchLinkSet link schema when there is no ContentItem

dougdroper:
There is no way to determine which schema to validate against for link
sets without a content_item, unless that information was provided in the
payload.

Until the time where we can add that information to the payload, or
ensure there is a content_item before creating the link set. This will
remove errors that we can't determine which schema to validate against

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -115,12 +115,14 @@ module Commands
       end
 
       def schema_name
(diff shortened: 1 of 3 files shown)


elliotcm:
Code looks good.  Could you add an explanation in the commit message to explain why this is a problem for the sake of future devs?

dougdroper:
@elliotcm updated

PR #332, opened 2016-05-31
ADR for govspeak embedded content.

danielroseman:


doc/arch/embedded-content-in-govspeak.md
```diff
@@ -0,0 +1,99 @@
...
+The short ID continues to be used here as it is more friendly for editors,
+some of whom know their frequently-used IDs.
+
+When sending to publishing-api, Whitehall converts this to an `embed` element
+with the relevant content_id:
+
+```
+Some text
+
+[embed:2b4d92f3-f8cd-4284-aaaa-25b3a640d26c]
```
boffbowsh: I think this needs to be more than `embed` as Publishing API needs to know what markup to generate, e.g. `<!-- govspeak:contact -->`. Perhaps `[embed:contact:2b4d92f3-f8cd-4284-aaaa-25b3a640d26c]`?

doc/arch/embedded-content-in-govspeak.md
```diff
@@ -0,0 +1,99 @@
...
+
+More text
+```
+
+and additionally adds that content_id to the the relevant element of the
```
tijmenb: Is there a timing issue here? What if the `put_content` with this comes before the `patch_links`?

danielroseman: I don't think so. At the very least, this will only affect the draft stack, and only momentarily. Content is not pushed to the live content store until after publish, but put_links happens before that - if for some reason it doesn't, then the HTML will simply remain as a comment.

doc/arch/embedded-content-in-govspeak.md
danielroseman: Fair point, updated.

danielroseman:
@dsingleton

dsingleton:
> @dsingleton

Ta, taking a look. Is there a reason why this isn't in the usual ADR structure?

dsingleton:
We (@danielroseman, @elliotcm, @boffbowsh, @tvararu and @nickcolley)  had a discussion about this offline on Wednesday. The outcome was that we're not sure this approach (originally agreed in the the RFC last year) is definitely the right approach, and that other options may be viable.

The debate is over where in the our architecture the Govspeak to HTML conversion (including embedded content). This is a non-trivial part of the new architecture, so it's worth spending a bit of time re-evaluating the options vs 6 months ago.

The three options we discussed are:
1. Conversion happens entirely in publishing side - `content store` serves fully static HTML, with `publishing-api` re-generating HTML when embedded dependencies (this is close to what we're doing for migrated WH content right now)
2. Conversion is split across Publishing and FE - `content-store` serves HTML with placeholders that must be resolved by a frontend using dependant `links` (the solution proposed here)
3. Conversion happens entirely in frontend side - `content-store` serves govspeak only, frontends convert it to HTML and resolve embedded content from `links`.

The next step is to outline the options in more detail in a new RFC. I'm happy to do this, but won't have time until next week, so if someone else wanted to write it up in the meantime that'd be 💯

elliotcm:
FWIW I'm backing option 1.

tvararu:
FWIW I'm backing option 3. It's a simpler conceptual model (all HTML is generated in one place), separates concerns better, and fits the "view as a pure function over data" pattern.

My second preference is option 1. Worth mentioning that part of making it work is ensuring that republishing from the `publishing-api` to the `content-store` is quick and painless, which is a worthy goal in and of it itself, but this option makes it a hard requirement. Not necessarily a bad thing.

benhyland:
tijmenb suggested we have a look at this.
My preferences are the same as @tvararu for much the same reasons.

It is usually the right thing to do to delay rendering until the last possible moment. 
Working with the raw source data tends to allow greater flexibility for future design changes. It seems clear at the moment that we don't have a strong prediction of changes that might be needed as we migrate the architecture to use publishing-api.

For clients other than frontends which might need rendered content (e.g. Rummager), there's nothing stopping them from performing that rendering themselves, using the same mechanisms.

danielroseman:
Updated to reflect the decision to go with option 1.

gpeng:
Looks good to me...

tvararu:
👍

doc/arch/govspeak-and-embedded-content.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+The govspeak renderer will then convert this to HTML, replacing the embedded
+content with the data from the relevant entry in the expanded links.
+
+## Consequences
+
+As a consequence of rendering govspeak within the publishing platform, rather
+than in the frontends, we forego the ability to easily change the format of the
```
dsingleton: Format is a little vague (and overloaded). We mean the template logic / HTML output for an embedded right?

tvararu: Yeah, or `presentational HTML` etc.

boffbowsh:
👍

PR #334, opened 2016-05-31
Only pass links hash through to SchemaValidator when validating link_sets

dougdroper:
The other information in the payload is metadata that should not be
validated

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -124,8 +124,9 @@ module Commands
         return unless schema_name
 
         SchemaValidator.new(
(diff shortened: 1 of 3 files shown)


tijmenb:
@dougdroper @boffbowsh I was wondering about this after the `bulk_publishing` flag got added. Should we change the `links` schema to include those attributes? I'm not sure why we don't validate the entire payload - it seemed to have been the intention here: https://github.com/alphagov/govuk-content-schemas/pull/164

boffbowsh:
I feel like that's more an argument to the command than data. The `bulk_publishing` flag isn't stored anywhere.

PR #335, opened 2016-05-31
Don't instantiate AR objects for dep resolution

boffbowsh:
Some content items have thousands of links forwards and backwards.
Instantiating AR objects for each of these is very expensive.

This prefers using `pluck` to fetch an array of just the values needed
for dependency resolution.

There is some logic required where it needs to choose between a
`content_id` or a `passthrough_hash`, but the `passthrough_hash` workaround will be going
away soon so this logic won’t need to remain for long.

Benchmarks using largest sets of forward and reverse links (10 iterations):

Before:

```
$ bundle exec rails r bench/downstream_presenter.rb
Reverse dependencies
..........
 12.370000   0.610000  12.980000 ( 14.230377)
Forward dependencies
..........
 11.790000   0.450000  12.240000 ( 13.463747)
```

After:

```
$ bundle exec rails r bench/downstream_presenter.rb
Reverse dependencies
..........
  2.040000   0.130000   2.170000 (  4.054558)
Forward dependencies
..........
  1.870000   0.090000   1.960000 (  3.355650)
```

### DIFF
--- a/app/presenters/queries/expanded_link_set.rb
+++ b/app/presenters/queries/expanded_link_set.rb
@@ -64,8 +64,13 @@ module Presenters
       end
 
       def dependees
-        link_set.links.group_by(&:link_type).each_with_object({}) do |(type, links), hash|
-          links = links.map(&:target_content_id)
+        return {} unless link_set.is_a? LinkSet
+        grouped_links = link_set.links
+          .pluck(:link_type, :target_content_id, :passthrough_hash)
+          .group_by(&:first)
+
+        grouped_links.each_with_object({}) do |(type, links), hash|
+          links = links.map { |l| l[2] || l[1] }
           expansion_rules = ::Queries::DependeeExpansionRules
 
           expanded_links = expand_links(links, type.to_sym, expansion_rules)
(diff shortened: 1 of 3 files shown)


app/presenters/queries/expanded_link_set.rb
```diff
@@ -64,8 +64,12 @@ def expanding_this_type?(type)
...
 
       def dependees
-        link_set.links.group_by(&:link_type).each_with_object({}) do |(type, links), hash|
-          links = links.map(&:target_content_id)
+        return {} unless link_set.is_a? LinkSet
+        link_set.links
+          .pluck(:link_type, :target_content_id, :passthrough_hash)
+          .group_by(&:first)
+          .each_with_object({}) do |(type, links), hash|
+          links = links.map { |l| l[2] || l[1] }
```
danielroseman: Being picky but should the code inside the block be indented another level?

danielroseman:
Thanks for this. It seems fine.

I think there is also some optimisation to be done lower down in that presenter, in the `content_item` method, since that is doing a lot of queries; I'd like to see how that compares with doing a single query to get all possible items and sorting in Ruby, but that is probably a matter for another PR.

app/presenters/queries/expanded_link_set.rb
boffbowsh: I thought that too, but apparently not. You're right though, it doesn't look good. I've pushed a change that makes it look nicer.

boffbowsh:
Yeah, additional improvement sounds good. I'd like to get a suite of benchmark scripts that we can use to help measure performance. I realise these are micro optimisations but every little helps.

PR #339, opened 2016-06-01
Regenerate the workflow image

MatMoore:
This was already renamed in the dot file, but the png was still showing "withdrawn"


Davidslv:
👍 Well spotted!

PR #340, opened 2016-06-01
Migrate withdrawn_notice data

steventux:
https://trello.com/c/l3oyxKgq/723-add-withdrawn-notice-to-schema-metadata-and-use-in-frontends

We need to migrate the `withdrawn_notice` property from the details hash of existing withdrawal Unpublishings to the top level property so that they can render correctly on frontends.

### DIFF
--- a/db/migrate/20160601134855_move_existing_withdrawn_notices.rb
+++ b/db/migrate/20160601134855_move_existing_withdrawn_notices.rb
@@ -1,0 +1,22 @@
+class MoveExistingWithdrawnNotices < ActiveRecord::Migration
+  def change
+    Unpublishing.where(type: "withdrawal").find_each do |unpublishing|
+      content_item = unpublishing.content_item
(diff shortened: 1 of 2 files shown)


db/migrate/20160601134855_move_existing_withdrawn_notices.rb
```diff
@@ -0,0 +1,22 @@
+class MoveExistingWithdrawnNotices < ActiveRecord::Migration
+  def change
+    Unpublishing.where(type: "withdrawal").each do |unpublishing|
+      content_item = unpublishing.content_item
```
elliotcm: How many of these are there?  If lots should we use `find_each` and/or eager load the content item?

elliotcm:
Small performance question, otherwise 👍

steventux:
@elliotcm updated, I should have done this anyway :+1:

PR #341, opened 2016-06-02
Don't send draft content to live content store.

elliotcm:
This was happening during dependency resolution
because the GetLatest query will return drafts if
they exist.


boffbowsh:
LGTM

PR #342, opened 2016-06-02
Content store safety check

elliotcm:
Prevents draft content items from being sent to the
live content store at the worker level, reducing
the chance of doing so accidentally in the future.

### DIFF
--- a/app/workers/presented_content_store_worker.rb
+++ b/app/workers/presented_content_store_worker.rb
@@ -9,7 +9,7 @@ class PresentedContentStoreWorker
   sidekiq_options queue: HIGH_QUEUE
 
   def perform(args = {})
-    assing_attributes(args.deep_symbolize_keys)
+    assign_attributes(args.deep_symbolize_keys)
     set_headers
 
     if params[:delete]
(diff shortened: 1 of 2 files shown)


app/workers/presented_content_store_worker.rb
```diff
@@ -9,7 +9,7 @@ class PresentedContentStoreWorker
   sidekiq_options queue: HIGH_QUEUE
 
   def perform(args = {})
-    assing_attributes(args.deep_symbolize_keys)
+    assign_attributes(args.deep_symbolize_keys)
```
boffbowsh: THANK YOU

app/workers/presented_content_store_worker.rb
```diff
@@ -27,6 +27,16 @@ def perform(args = {})
 
   def send_to_content_store
     if content_item_id
+      content_item_is_draft = (State.find_by(content_item_id: content_item_id).name == "draft")
```
boffbowsh:
```ruby
State.where(content_item_id: content_item_id).pluck == ["draft"]
```

Every time you instantiate an ActiveRecord object, god kills a meg of RAM

boffbowsh:
Small comment otherwise yes this change is good 👍

elliotcm:
@boffbowsh ok updated

PR #345, opened 2016-06-03
Add a /debug endpoint to help debug content issues

dougdroper:
Graph of expanded link representation
Graph of states
Timeline of events which link to the specific event
Links to web representation and show other useful information

https://trello.com/c/INzLJqNi/726-add-a-debug-html-endpoint-to-help-developers-debug-content-issues

screen shot 2016-06-03 at 14 47 16

screen shot 2016-06-06 at 10 51 57

### DIFF
--- a/app/presenters/debug_presenter.rb
+++ b/app/presenters/debug_presenter.rb
@@ -1,0 +1,96 @@
+module Presenters
+  class DebugPresenter
+    attr_reader :content_id
+
+    def initialize(content_id)
+      @content_id = content_id
+    end
+
+    def content_items
+      @content_items ||= ContentItem.where(content_id: content_id)
+    end
+
+    def user_facing_versions
+      UserFacingVersion.join_content_items(content_items).pluck("user_facing_versions.number")
+    end
+
+    def locales
+      State.join_content_items(
+        Translation
+        .join_content_items(content_items))
+        .reorder('')
+        .select("locale, content_items.id as id, states.name as state")
+    end
+
+    def latest_content_items
+      ::Queries::GetLatest.call(content_items)
+    end
+
+    def latest_state_with_locale
+      Translation.join_content_items(State.join_content_items(latest_content_items)).pluck("translations.locale, states.name")
+    end
+
+    def link_set
+      @link_set ||= LinkSet.find_by(content_id: content_id)
+    end
+
+    def expanded_links
+      links = ::Queries::GetExpandedLinks.call(content_id, "en")
+      rows = []
+      rows << [links[:content_id], '', 'current']
+      links[:expanded_links].each do |type, sub_links|
+        rows << [type, links[:content_id], 'current']
+        sub_links.each do |link|
+          rows << [
+            {
+              v: (link[:content_id].to_s + type.to_s),
+              f: "<a href='debug?content_id=#{link[:content_id]}'>#{link[:content_id]}</a>"
+            },
+            type,
+            ''
+          ]
+        end
+      end
+
+      rows
+    end
+
+    def states
+      @states = []
+      locales.group_by(&:locale).each do |locale, local_states|
+        @states << [locale, '', '']
+        @states << [{ v: local_states.first.id.to_s, f: local_states.first.state }, locale, '']
+        local_states[1..-1].each_with_index do |state, index|
+          @states << [{ v: state.id.to_s, f: state.state }, local_states[(index + 1) - 1].try(:id).to_s, '']
+        end
+      end
+      @states
+    end
+
+    def event_timeline
+      events.map do |event|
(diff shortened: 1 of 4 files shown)


tuzz:
This is looking great!

app/views/debug/show.html.erb
```diff
@@ -0,0 +1,118 @@
...
+
+  <h3>User facing verisons</h3>
+  <%= @presenter.user_facing_versions %>
+
+  <h3>Events</h3>
+  <% @presenter.events.each do |event| %>
+    <div class='events' id="<%= event.id  %>">
+      <p><b><%= event.action %></b>, <%= event.created_at.strftime("%Y-%m-%d %H:%I:%S") %><br />
+      <span class='payload' >
+        <% event.payload.each do |k,v| %>
```
tijmenb: It would be possible to use the table helper from:

https://github.com/alphagov/search-admin/blob/master/app/helpers/table_helper.rb

That gives you nested tables for the data:

screen shot 2016-06-03 at 15 29 06

You could also try:

```
      <pre class='payload'>
        <%= JSON.pretty_generate(event.payload) %>
      <pre>
```

To make it look like:

screen shot 2016-06-03 at 15 26 00

tijmenb:
This is dope!

app/presenters/debug_presenter.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def user_facing_versions
+      UserFacingVersion.join_content_items(content_items).pluck("user_facing_versions.number")
+    end
+
+    def locales
+      State.join_content_items(
+        Translation
+        .join_content_items(content_items))
+        .reorder('')
```
elliotcm: ?

app/presenters/debug_presenter.rb
```diff
@@ -0,0 +1,96 @@
...
+        local_states[1..-1].each_with_index do |state, index|
+          @states << [{ v: state.id.to_s, f: state.state }, local_states[(index + 1) - 1].try(:id).to_s, '']
+        end
+      end
+      @states
+    end
+
+    def event_timeline
+      events.map do |event|
+        [event.action, '', event.id.to_s, event.created_at, event.created_at + 1.second]
```
elliotcm: Lots of magic empty strings in this file.  I saw you twiddling with the `states` method earlier but it's still pretty confusing.

config/routes.rb
```diff
@@ -34,4 +34,5 @@
   end
 
   get '/healthcheck', to: proc { [200, {}, ['OK']] }
+  get '/debug', to: "debug#show"
```
elliotcm: Should this be `/debug/:content_id` or similar?

elliotcm:
Good stuff, just some code tidy comments.

MatMoore:
Couple of thoughts:

1) Is this only going to be available on dev environments? It would be great if this was exposed on integration/staging too, as this would allow us to investigate bad data a lot more easily.

2) Can the event log be exposed through the API?

Currently, when I investigate the history of a content item I go direct to the database/console to query `payload->'links'` for the content id in question. This data can help us respond to zendesk tickets, and identify errors in the migration work we're doing.

@alphagov/team-finding-things are looking at tagging history more widely, since this is relevant to users too:
* https://trello.com/c/emr1o1Js/29-whitehall-tagging-history-what-do-users-need
* https://trello.com/c/reQdXotD/456-spike-how-might-we-handle-tag-history
* https://gov-uk.atlassian.net/wiki/pages/viewpage.action?pageId=69566472

It seems to me that event log data meets part of the need already and we could use this to prototype quickly.

We could also put up a similar /debug interface in content tagger that is more focused on links.

danielroseman:
I presume this will be accessed in integration/production via port forwarding, like we do with sidekiq-monitoring. If so we'll need to exempt this controller from signin permission, which is enforced in application_controller - going via signin will fail because the redirect URI won't be valid.

app/views/debug/show.html.erb
dougdroper: Feast your eyes on: 

screen shot 2016-06-06 at 10 51 57

app/presenters/debug_presenter.rb
dougdroper: Yeah, annoyingly the empty string are there to map to google charts table definitions. I have removed some of them that I am no longer using to clean it up a bit, but some are still required in that order.

config/routes.rb
dougdroper: 👍  done

dougdroper:
@danielroseman I have removed the need for signin on this controller: 
https://github.com/alphagov/publishing-api/pull/345/files#diff-00f3ef513f0aded7383455b25fd42be4R2

PR #346, opened 2016-06-03
Correct the order of grouped content results

benhyland:
The order of the results for the groupd_content_and_links
endpoint was incorrect.
This meant that any client code would have a hard time paging
because the last_seen_content_id which determines which page
comes next would be wrongly reported.

@matmoore and @benhyland

### DIFF
--- a/app/queries/get_grouped_content_and_links.rb
+++ b/app/queries/get_grouped_content_and_links.rb
@@ -33,7 +33,7 @@ module Queries
         page_size: page_size
       )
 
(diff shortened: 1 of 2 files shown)


tuzz:
This seems relevant:
https://gov-uk.atlassian.net/wiki/display/FS/RFC+36%3A+Stop+preserving+order+for+links

benhyland:
@tuzz Thanks for the heads up.
The ordering fix here is on the content_id level, not on the links level, but it is good to remember that we shouldn't guarantee link ordering in this endpoint, as per the RFC.

PR #347, opened 2016-06-06
Return schema_name & document_type for bulk read

tijmenb:
`schema_name` & `document_type` are replacing `format`.


dougdroper:
LGTM, the spec is failing though

PR #349, opened 2016-06-06
Delete HTML publications from publishing api

boffbowsh:
Due to the various stages of migration of HTML publications, it's best to clear out the partially migrated content items and start from scratch. Example reasons we need to do this:

- Some HTML publications had different content ids across their various editions
- Some HTML publications had their locale set to en when it wasn't, which causes `patch_link_set` calls to send the wrong locale to the Content Store

These are not currently used anywhere in production code. I've included a safety check to prevent this migration being run twice accidentally in production.

Q: This covers links and supporting objects. Are there other tables I should be deleting rows from?

/cc @elliotcm

### DIFF
--- a/db/migrate/20160606142809_delete_html_publications.rb
+++ b/db/migrate/20160606142809_delete_html_publications.rb
@@ -1,0 +1,26 @@
+class DeleteHtmlPublications < ActiveRecord::Migration
+  def up
+    # Safety check to ensure this doesn't get run more than once.
+    # At time of writing, the max content_item_id in production
+    # is 796460. Refuse to run this migration if there are no
+    # HTML publications lower than this ID, as this means that
+    # this migration has already been run.
+
+    return if ContentItem.where("id < 796460").where(document_type: "html_publication").empty?
+
+    ContentItem.transaction do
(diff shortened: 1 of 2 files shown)


gpeng:
Looks good to me. Left open for @elliotcm to review...

db/migrate/20160606142809_delete_html_publications.rb
```diff
@@ -0,0 +1,26 @@
...
+    # At time of writing, the max content_item_id in production
+    # is 796460. Refuse to run this migration if there are no
+    # HTML publications lower than this ID, as this means that
+    # this migration has already been run.
+
+    return if ContentItem.where("id < 796460").where(document_type: "html_publication").empty?
+
+    ContentItem.transaction do
+      results = ContentItem.where(document_type: "html_publication").pluck(:id, :content_id)
+      ids = results.map { |r| r[0] }.uniq
```
elliotcm: ids can't be anything other than unique.

elliotcm:
👍 Not sure if you can be bothered about the small comment I made, merge if you can't.

As a double safety maybe delete this once it's run.

PR #355, opened 2016-06-09
Create an async experiment for the linkables endpoint.

elliotcm:
We were going to use Github Scientist for this but
they avoid the async complexity by running everything
sequentially.  This means that our linkables
endpoint would take twice as long to respond while
we were running the test.

Scientist doesn't actually do much (you still have
to write your comparators and reporting) so I
wrote an async version.

Here's the flow:
fullsizerender

I'd love to get feedback from (at least):
- [x] @danielroseman 
- [ ] @tijmenb 
- [x] @dougdroper 
- [ ] @steventux


elliotcm:
After some feedback from @dougdroper I've restructured the way the workers report the data to avoid locking issues.

Now only the control branch will try to report and will retry if the candidate isn't ready.

app/candidates/base_candidate.rb
```diff
@@ -0,0 +1,26 @@
...
+    def experiment_candidate(experiment)
+      start_time = Time.now
+      result = yield
+      duration = (Time.now - start_time).to_f
+
+      ExperimentResultWorker.perform_async(experiment[:name], experiment[:id], result, duration, :candidate)
+
+      result
+    rescue StandardError => exception
+      if Rails.env.test?
```
boffbowsh: Maybe dev too?

boffbowsh:
This is good, I like the API. I think this will be really useful :) 👍

dougdroper:
I agree 👍

app/candidates/base_candidate.rb
elliotcm: I'll make it configurable in puppet, defaulting to re-raise.

elliotcm: Or maybe the other way round.

elliotcm:
@boffbowsh I've changed it to an environment variable and enable it for test.  People who need it can activate it or if we find we need it all the time we can add it to the dev VM in puppet.

elliotcm:
We've had enough top qual feedback I think, I'll get this merged and we can tweak it as we go if needed.

PR #356, opened 2016-06-09
Delete stale placeholder policies

benhyland:
There are a few placeholder policies which the policy-publisher
doesn't know about.

This is because those policies were renamed at some point before
the migration to publishing-api was completed.

We should delete these manually, since there is no api to do so.
We'll introduce another PR to deal with content-store.

See https://trello.com/c/xARBi6HF/661-policy-publisher-5-pages-missing-from-rummager

@Davidslv and @benhyland

### DIFF
--- a/db/migrate/20160609153413_remove_policy_placeholders.rb
+++ b/db/migrate/20160609153413_remove_policy_placeholders.rb
@@ -1,0 +1,22 @@
+class RemovePolicyPlaceholders < ActiveRecord::Migration
+  def change
+    ContentItem.transaction do
+      results = ContentItem.where(content_id: content_ids_to_delete).pluck(:id, :content_id)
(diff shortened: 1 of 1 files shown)


MatMoore:
I think these items should be changed to a redirect in the publishing API.

https://www.gov.uk/government/policies/carer-s-health is currently 404ing which is wrong if it ever existed.

benhyland:
@MatMoore Do we know it ever existed? We just know there is a placeholder for it in publishing-api.

Also, at some point we have to draw the line between perfecting content and fixing the things we need to fix to continue with the tagging work. I'm not experienced enough with GDS to say for sure where that line is though.

tijmenb:
Closed in favour of doing a `unpublish` API call to the publishing-api.

PR #359, opened 2016-06-13
Allow the lookup-by-base-path endpoint to work for 'unpublished' content

benhyland:
The endpoint was created when there were fewer states.
The main client is currently content-tagger, which is concerned
with tagging all user-visible content. This includes 'unpublished' items,
since items that have been withdrawn are still user-visible.

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,7 +1,12 @@
 class LookupsController < ApplicationController
   def by_base_path
+    # return content_ids for content that is visible on the live site
+    # withdrawn items are still visible
+    states = %w(published unpublished)
+    base_paths = params.fetch(:base_paths)
+
     base_paths_and_content_ids = ContentItemFilter
-      .filter(state: "published", base_path: params.fetch(:base_paths))
+      .filter(state: states, base_path: base_paths)
       .pluck(:base_path, :content_id)
       .uniq
 
--- a/spec/requests/lookups_spec.rb
+++ b/spec/requests/lookups_spec.rb
@@ -1,24 +1,40 @@
 require "rails_helper"
 
 RSpec.describe "POST /lookup-by-base-path", type: :request do
-  it "returns content_ids for base_paths" do
-    FactoryGirl.create(:content_item, state: "published", base_path: "/my-page", content_id: "b9b2da0a-ec50-4b0e-b29c-b7cbc8195307")
-    FactoryGirl.create(:content_item, state: "draft", base_path: "/my-page", content_id: "f7cdb359-c8ab-4d6d-b1f0-5c5640b24c09")
-    FactoryGirl.create(:content_item, state: "published", base_path: "/other-page", content_id: "b879bcdb-6160-4bfd-b758-f546bbb408c4")
+  context "validating" do
+    it "requires a base_paths param" do
+      expected_error_response = { "error" => { "code" => 422, "message" => "param is missing or the value is empty: base_paths" } }
 
-    post "/lookup-by-base-path", base_paths: ["/my-page", "/other-page", "/does-not-exist"]
+      post "/lookup-by-base-path"
 
-    expect(parsed_response).to eql(
-      "/my-page" => "b9b2da0a-ec50-4b0e-b29c-b7cbc8195307",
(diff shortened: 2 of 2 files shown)


app/controllers/lookups_controller.rb
```diff
@@ -1,7 +1,12 @@
 class LookupsController < ApplicationController
   def by_base_path
+    # return content_ids for content that is visible on the live site
+    # withdrawn items are still visible
+    states = %w(published unpublished)
```
gpeng: Not all 'unpublished' items items are 'withdrawn' so some will no longer be 'visible' having been redirected. I'm not sure how you're consuming this and whether that is an issue but thought it would be worth mentioning on account of the comment.

MatMoore:
I think this is fine but would prefer someone who knows the publishing api to review as well.

Is there ever a case where the ContentItemFilter query could return multiple rows for the same path now that we've included unpublished content items? I'm working from the assumption that you should never be able register a content item to a base path if that path has been used for another content item before.

app/controllers/lookups_controller.rb
MatMoore: Can we tell if an item is withdrawn? I think we care about this distinction in publishing apps and rummager.

benhyland: A useful query. We are aware of that point, although we (I) don't have a good grasp of the other possible reasons for withdrawal.
It shouldn't affect clients currently, and we wanted to avoid introducing a more complex dependency on the details of the publishing-api workflow model.

benhyland:
@MatMoore ContentItemFilter can return multiple rows with the same base path. The endpoint calls `uniq!` on the result.

app/controllers/lookups_controller.rb
MatMoore: Oh yeah, I was asking more out of general interest for things like the [tagging migration checker](https://github.com/alphagov/finding-things-migration-checker).

I think this request should be able to give you an answer as long as the thing has ever been published.

benhyland: @MatMoore AIUI clients of this endpoint are currently the checker tool and content-tagger. Neither of these needs to distinguish withdrawn content using the results of this endpoint (although they might need to know in other circumstances). In particular, the checker tool only uses this output to link up the iterative imports it has done from Rummager and Publishing API.

benhyland: @MatMoore I think this does return a base_path iff the item has ever been published. There are 4 possible states, draft, published, superseded, unpublished. Items in draft need not ever have been published. Items in superseded necessarily have a published successor, which may become unpublished, but is never deleted or changed to any other state. So the filter to include published and unpublished should be sufficient to find the base_path for all content that has ever been published.

gpeng: The `Unpublishing` has a [`type`](https://github.com/alphagov/publishing-api/blob/master/app/models/unpublishing.rb#L7-L10) if that helps...

benhyland: Thanks that's good to know :)

tijmenb: Not sure if I'm following everything here, but I'd like the behaviour of this endpoint to be the same as if you're looking up a content_id from the site by checking the _live_ content-store (https://www.gov.uk/passport-fees -> https://www.gov.uk/api/content/passport-fees -> `8b59b474-9775-4366-97ee-97a66740411c`). 

Is the `%w[published withdrawn]` filter enough for that?

benhyland: @tijmenb No. The `%w(published unpublished)` (note - not `withdrawn`) filter will return more content_ids than just those exposed by the live content store, because it will return ids for all types of unpublishings. See the link Graham provided for the other types: https://github.com/alphagov/publishing-api/blob/master/app/models/unpublishing.rb#L7-L10

tijmenb: So which type of things would be not in the content-store?

benhyland: @tijmenb Oh I see what you mean. I guess all types of unpublishing will have some kind of record in the live content store - @gpeng could confirm.

tijmenb:
👍  from me

PR #364, opened 2016-06-14
Fix Unpublishing update bug

benhyland:
When updating an existing Unpublshing record, we erroneously returned
a boolean from `update_attributes` instead of returning the updated object.
Add a test and return the value expected by client code.

@benhyland and @davidslv


elliotcm:
👍

PR #365, opened 2016-06-15
Add last_edited_at to content item

cbaines:
Trello card: https://trello.com/c/QJPHGeQf/738-add-last-edited-at-timestamp-to-content-items-small-ish

These commits add a new "last_edited_at" field to the content item model, along with the corresponding schema changes and add tests to cover this new behaviour. Existing tests are also changed to pass given this new functionality. 

@tuzz (as the RFC author)
@alphagov/team-publishing-platform (who need to be aware of this work)

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -62,6 +62,7 @@ module Commands
         check_version_and_raise_if_conflicting(content_item, payload[:previous_version])
 
         update_content_item(content_item)
+        update_last_edited_at_if_needed(content_item, payload[:last_edited_at])
         increment_lock_version(content_item)
 
         if path_has_changed?(previous_location)
@@ -219,6 +220,14 @@ module Commands
       def update_content_item(content_item)
         content_item.assign_attributes_with_defaults(content_item_attributes_from_payload)
         content_item.save!
+      end
+
+      def update_last_edited_at_if_needed(content_item, last_edited_at = nil)
+        if last_edited_at.nil? && ["major", "minor"].include?(payload[:update_type])
+          last_edited_at = Time.zone.now
+        end
+
+        content_item.update_attributes(last_edited_at: last_edited_at) if last_edited_at
       end
 
       def increment_lock_version(content_item)
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -223,7 +223,7 @@ module Commands
       end
 
       def update_last_edited_at_if_needed(content_item, last_edited_at = nil)
-        if last_edited_at.nil? && ["major", "minor"].include?(payload[:update_type])
+        if last_edited_at.nil? && %w(major minor).include?(payload[:update_type])
           last_edited_at = Time.zone.now
         end
 
--- a/app/presenters/downstream_presenter.rb
+++ b/app/presenters/downstream_presenter.rb
@@ -14,6 +14,7 @@ module Presenters
     def present
       symbolized_attributes
         .slice(*content_item.class::TOP_LEVEL_FIELDS)
+        .except(:last_edited_at) # only intended to be used by publishing applications
         .merge(first_published_at)
         .merge(public_updated_at)
         .merge(links)
--- a/db/migrate/20160614144815_add_last_edited_at_to_content_item.rb
+++ b/db/migrate/20160614144815_add_last_edited_at_to_content_item.rb
@@ -1,0 +1,6 @@
+class AddLastEditedAtToContentItem < ActiveRecord::Migration
+  def change
+    add_column :content_items, :last_edited_at, :datetime
+    add_index :content_items, :last_edited_at
+  end
+end
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -338,6 +338,26 @@ front-end application, when the content item is rendered.
 
 ---
 
+### last_edited_at
+
+Example: *2015-01-01T12:00:00Z*
+
+Required: No
+
+The `last_edited_at` records the last time that the content received a major or
+minor update. This can be used by the publishing applications as a ordering
+parameter and shown to editors.
+
+If the `last_updated_at` is specified when PUTing content, the given value will
+be stored. If `last_updated_at` is not specified, and the update type is major
+or minor, the `last_updated_at` will be set to the current time. For any other
+update type, the `last_updated_at` will be unchanged.
+
+The `last_edited_at` must use the [ISO 8601
+format](https://en.wikipedia.org/wiki/ISO_8601).
+
+---
+
 ### links
 
 Example: *{ “related”: [“8242a29f-8ad1-4fbe-9f71-f9e57ea5f1ea”] }*
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -849,6 +849,21 @@ RSpec.describe Commands::V2::PutContent do
 
           expect(content_item.last_edited_at.iso8601).to eq(last_edited_at.iso8601)
         end
+
+        context "with update_type not major or minor" do
+          it "stores the provided timestamp" do
+            last_edited_at = 1.year.ago
+
+            described_class.call(payload.merge(
+                                   update_type: "republish",
+                                   last_edited_at: last_edited_at
+                                 ))
+
+            content_item.reload
+
+            expect(content_item.last_edited_at.iso8601).to eq(last_edited_at.iso8601)
+          end
+        end
       end
     end
 
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -820,8 +820,8 @@ RSpec.describe Commands::V2::PutContent do
           last_edited_at = 1.year.ago
 
           described_class.call(payload.merge(
-            last_edited_at: last_edited_at
-          ))
+                                 last_edited_at: last_edited_at
+                              ))
 
           content_item = ContentItem.last
 
@@ -842,8 +842,8 @@ RSpec.describe Commands::V2::PutContent do
           last_edited_at = 1.year.ago
 
           described_class.call(payload.merge(
-            last_edited_at: last_edited_at
-          ))
+                                 last_edited_at: last_edited_at
+                               ))
 
           content_item.reload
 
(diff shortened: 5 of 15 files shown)


app/commands/v2/put_content.rb
```diff
@@ -221,6 +222,14 @@ def update_content_item(content_item)
         content_item.save!
       end
 
+      def update_last_edited_at_if_needed(content_item, last_edited_at = nil)
+        if last_edited_at.nil? && %w(major minor).include?(payload[:update_type])
+          last_edited_at = Time.zone.now
```
tuzz: Would it be tidier to use `||= Time.zone.now` here?

app/presenters/downstream_presenter.rb
```diff
@@ -14,6 +14,7 @@ def initialize(content_item, link_set, state_fallback_order:)
     def present
       symbolized_attributes
         .slice(*content_item.class::TOP_LEVEL_FIELDS)
+        .except(:last_edited_at) # only intended to be used by publishing applications
```
tuzz: 👍

db/migrate/20160614144815_add_last_edited_at_to_content_item.rb
```diff
@@ -0,0 +1,6 @@
+class AddLastEditedAtToContentItem < ActiveRecord::Migration
+  def change
+    add_column :content_items, :last_edited_at, :datetime
+    add_index :content_items, :last_edited_at
```
tuzz: 👍

doc/semantics-of-the-publishing-api.md
```diff
@@ -338,6 +338,26 @@ front-end application, when the content item is rendered.
...
+Required: No
+
+The `last_edited_at` records the last time that the content received a major or
+minor update. This can be used by the publishing applications as a ordering
+parameter and shown to editors.
+
+If the `last_updated_at` is specified when PUTing content, the given value will
+be stored. If `last_updated_at` is not specified, and the update type is major
+or minor, the `last_updated_at` will be set to the current time. For any other
+update type, the `last_updated_at` will be unchanged.
```
tuzz: It might be worth elaborating the behaviour when `update_type` is `nil`.

doc/semantics-of-the-publishing-api.md
```diff
@@ -338,6 +338,26 @@ front-end application, when the content item is rendered.
...
+
+The `last_edited_at` must use the [ISO 8601
+format](https://en.wikipedia.org/wiki/ISO_8601).
```
tuzz: Does this actually matter for the request payload? Won't Rails parse this from a wide-range of possible formats?

spec/commands/v2/put_content_spec.rb
```diff
@@ -812,5 +813,77 @@
...
+          described_class.call(payload.merge(
+                                 update_type: "republish"
+                               ))
+
+          content_item.reload
+
+          expect(content_item.last_edited_at).to eq(old_last_edited_at)
+        end
+      end
+    end
```
tuzz: Would it be worth adding a test for the case when `update_type` is `nil`? I was under the impression this would also set the `last_edited_at` because @elliotcm explained this is assumed to be a "major" update for historical reasons.

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -219,6 +225,38 @@
...
+
+          expect(response.status).to eq(200), response.body
+
+          expect(results).to eq([
+            { "last_edited_at" => "2016-02-02T00:00:00Z" },
+            { "last_edited_at" => "2016-01-01T00:00:00Z" },
+          ])
+        end
+      end
+
```
tuzz: 👍

tuzz:
Really good work. 🎉

A few minor questions / suggestions.

Looking forward to having this feature from the Specialist Publisher side.

Thanks

doc/semantics-of-the-publishing-api.md
cbaines: I don't know. In terms of this documentation, I added this to maintain consistency with `public_updated_at` field documentation, as `last_edited_at` is handled in a very similar way.

spec/commands/v2/put_content_spec.rb
cbaines: Hmm, there is possibly an issue with `update_type`. I would hope that `update_type` `nil` is considered invalid input, but as I understand, it is valid to not specify the `update_type` (which is not the same as `nil`?).

Currently, unless `update_type` is set to "major" when its not specified, this won't change `last_edited_at` when a put_content call is made. I'm not sure if this is correct or not, but it deserves more thought.

doc/semantics-of-the-publishing-api.md
cbaines: I think the best way to resolve this is either make sure that ISO8601 is the only accepted format, as its better to accept a smaller set of valid values. What do you think @tuzz ?

danielroseman: JSON Schema already enforces RFC 3339 for datetimes, which is actually more restrictive than ISO8601 (eg it mandates a timezone component). As long as this value is declared as a date-time there, there's probably not much point including this detail here.

spec/commands/v2/put_content_spec.rb
cbaines: @elliotcm we talked through the behavior for different `update_type`s, but I can't remember if we decided what the correct behavior is when its not specified?

elliotcm: Yeah I think we should distinguish between the client sending `nil` and not sending the param at all.

If it sends `nil` we shouldn't do anything because it's explicitly neither major nor minor, but if they don't send anything we should assume major or minor because the most common use case for that will be apps which choose to send the update type on publish, and thus part of the normal editor workflow (which only deals in minor or major updates).

Does that sound right?

cbaines: The publish endpoint is quite strict, which I like. Could something similar be done for this endpoint also? https://github.com/alphagov/publishing-api/blob/7a05330e2ffbb0b8ad2af101dba7d86414d32452/app/commands/v2/publish.rb#L17

app/commands/v2/put_content.rb
dougdroper: Could be 
```
def major_or_minor_update?
  %w(major minor).include?(payload[:update_type])
end
```

then 

```
def update_last_edited_at
   return unless major_or_minor_update?
   last_edited_at ||= Time.zone.now
   etc
end
```

cbaines: I still like just having one if, as I think it reflects the way in which I think about the condition. I understand the desired behavior as "set last_edited_at to the current time if it has not been specified in the request and the update_type is major or minor", and so having an if statement that is similar to this is quite helpful. While the alternatives above are tidier, I think they are all harder to read than just having a big if statement.

What does everyone think? I'm happy to defer to your better judgement, as this could just be an issue with my inexperience with reading Ruby?

dougdroper: I am happy to make the `last_edited_at.nil?` check explicit and make it read like the desired behaviour: 
`if last_edit_at_specified? && major_or_minor_update?` etc

cbaines: Something like this @dougdroper ?
```ruby
      def last_edited_at_unspecified?
        payload[:last_edited_at].nil?
      end

      def major_or_minor_update_type?
        %w(major minor).include?(payload[:update_type])
      end

      def update_last_edited_at_if_needed(content_item)
        last_edited_at = payload[:last_edited_at]

        if last_edited_at_unspecified? && major_or_minor_update_type?
          last_edited_at = Time.zone.now
        end

        content_item.update_attributes(last_edited_at: last_edited_at) if last_edited_at
      end
```

Pulling out last_edited_at from the payload in two methods does not seem great, but neither does passing it in just to do .nil?, so I'm not sure how to improve on this.

tuzz: Maybe it's fine as-is. 👍

dougdroper:
```ruby
      def update_last_edited_at_if_needed(content_item)
        return unless %w(major minor).include?(payload[:update_type])
        last_edited_at = payload[:last_edited_at] || Time.zone.now
        content_item.update_attributes(last_edited_at: last_edited_at)
      end
```

dougdroper: Ah, so sounds like there is a case where if you explicitly set the last_edited_at, then it should use that (regardless of the update_type) Is that correct @elliotcm? If so we need a test around that and probably use the original method.

elliotcm: Yes, that is correct.

cbaines: New test added in 1d6c22116e2521d6c6f15d006e8039d7cb6681e3

cbaines:
Does a modification need to be made to a schema in https://github.com/alphagov/govuk-content-schemas because of the changes being made here?

doc/semantics-of-the-publishing-api.md
cbaines: Currently `put_content` looks to process this fine. I don't think this should be the case, but the correct place to stop this is at the schema validation level, which is before any of the logic associated with processing `last_edited_at`?

elliotcm:
@cbaines Yes schemas will need updating (though in this case that can happen after) and we need a rake task to fill out the column.

elliotcm:
Doesn't have to be part of this PR unless it would prevent us shipping this code as is in the meantime.

cbaines:
@elliotcm For filling out the column, I've had a quick look, and just using the value in `public_updated_at` seems to be the best value to use, does this make sense?

elliotcm:
`public_updated_at` will only give you major instead of both, if you want to be accurate with it you can use the `events` table and get the max timestamp for all "Publish" or "PutContent" requests where the `update_type` in the `payload` was "major" or "minor".

Check with @tuzz as to whether `public_updated_at` would be good enough to fill things out.

cbaines:
@elliotcm ah, yeah, I've had a quick look in to this, and I'm not quite sure how to relate the events in the log to the entries in the content_items table, unless only the most relevant/recent entry for a given piece of content is updated?

@tuzz just let me know if you want to discuss this.

tuzz:
`public_updated_at` will be fine. We're going to republish all content to Publishing API from our side so it doesn't matter to us.

elliotcm:
@cbaines Yeah, it would just be the most recent for that content ID, but it looks like we don't need to go into that much detail.

tuzz:
`GetLatest` is probably what you need.

elliotcm:
@tuzz not in this case, no

PR #371, opened 2016-06-16
Don't perform safety check for every job

boffbowsh:
Only perform the state == draft check if we would send to the live content store.

### DIFF
--- a/app/workers/presented_content_store_worker.rb
+++ b/app/workers/presented_content_store_worker.rb
@@ -27,14 +27,15 @@ private
 
   def send_to_content_store
     if content_item_id
-      content_item_is_draft = (State.where(content_item_id: content_item_id).pluck(:name) == ["draft"])
       content_store_is_live = (content_store == Adapters::ContentStore)
 
-      if content_item_is_draft && content_store_is_live
-        raise CommandError.new(
-          code: 500,
-          message: "Will not send a draft content item to the live content store",
-        )
+      if content_store_is_live
+        if State.where(content_item_id: content_item_id).pluck(:name) == ["draft"]
+          raise CommandError.new(
+            code: 500,
+            message: "Will not send a draft content item to the live content store",
+          )
+        end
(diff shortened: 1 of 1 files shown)


app/workers/presented_content_store_worker.rb
```diff
@@ -27,14 +27,15 @@ def perform(args = {})
...
-      content_item_is_draft = (State.where(content_item_id: content_item_id).pluck(:name) == ["draft"])
       content_store_is_live = (content_store == Adapters::ContentStore)
 
-      if content_item_is_draft && content_store_is_live
-        raise CommandError.new(
-          code: 500,
-          message: "Will not send a draft content item to the live content store",
-        )
+      if content_store_is_live
+        if State.where(content_item_id: content_item_id).pluck(:name) == ["draft"]
```
Davidslv: It feels that it should be a `find_by` instead of using `where`, you will only receive one result with that query

```
State.find_by(content_item_id: content_item_id).name == "draft"
```

boffbowsh: Using `find_by` involves expensive ActiveRecord instantiation. Using `pluck` is many many times quicker.

cbaines:
It would be useful to have the reasoning behind this change in the commit message.

app/workers/presented_content_store_worker.rb
tuzz: Perhaps `State.exists?(content_item_id: content_item_id, name: "draft")`

boffbowsh:
Commit message updated

app/workers/presented_content_store_worker.rb
boffbowsh: I didn't change the code from the original, just moved it.

tuzz:
Merged to avoid bikeshedding.

cbaines:
> Commit message updated

:+1:

PR #373, opened 2016-06-17
Cache the signon token for applications

tijmenb:
This bumps the gds-sso gem and configures it to use a cache, which will prevent publishing-api from making many requests to Signon for the user data associated with the bearer token.

See https://github.com/alphagov/gds-sso/pull/97 for the change.

This speeds up most calls to publishing-api by ~120ms.

https://trello.com/c/Gpc22bUW/684-speed-up-publishing-api-by-caching-the-token

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ else
   gem "gds-api-adapters", "30.2.0"
 end
 
-gem "gds-sso", "11.2.1"
+gem "gds-sso", github: "alphagov/gds-sso"
 
 gem 'bunny', '2.0.0'
 gem 'whenever', '0.9.4', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ else
   gem "gds-api-adapters", "30.2.0"
 end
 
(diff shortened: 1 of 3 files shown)


boffbowsh:
Great catch! 👍 . I feel though this should go into `gds-sso` as an opt-in thing for apps that are basically API exclusive and request heavy such as Publishing API. I think having a default cache for user facing apps is not a good thing, and they wouldn't notice 300ms on their request. 

However, if this monkey patch works as-is for publishing API, I have no problems with merging this in lieu of a proper change to `gds-sso`. The patch can then get removed when the gem gets bumped in this app.

Good 🔍 ing.

benhyland:
@boffbowsh http://glinden.blogspot.co.uk/2006/11/marissa-mayer-at-web-20.html
There's quite a bit of evidence that users do notice and care about page render delays on the order of 100s of ms btw.

boffbowsh:
Yes of course, but we're talking about publishing users not frontend users. It's obviously not ideal but it's of less consequence than delayed API calls, and it's not affecting the frontend.

PR #374, opened 2016-06-19
Disable linkables experiment

boffbowsh:
The recording of results is running Redis out of memory

/cc @alexmuller @elliotcm


elliotcm:
There's a change going up at 10am which will reduce the amount of cruft left over in redis during this process, and we intend to complete the experiment this morning.

PR #377, opened 2016-06-20
Filter out redirects on `/lookup-by-base-path`

tvararu:
Some base paths have multiple redirects, which are returned as part of this lookup, but not in any particular order:

```ruby
> ContentItemFilter.filter(state: %w(published unpublished), base_path: ['/renewing-your-tax-credits-claim']).pluck(:base_path, :content_id, :document_typ    e, :updated_at)
[["/renewing-your-tax-credits-claim", "64d2120f-2396-4033-8e91-45f3e1f49ad8", "redirect", Tue, 19 Apr 2016 15:02:19 UTC +00:00],
 ["/renewing-your-tax-credits-claim", "6aecd670-0451-49d2-bce7-a4258be1adb3", "redirect", Tue, 19 Apr 2016 15:02:22 UTC +00:00],
 ["/renewing-your-tax-credits-claim", "bd88b94e-9fb0-44ea-84e8-6c3a4ef962a7", "answer", Tue, 07 Jun 2016 12:27:13 UTC +00:00],
 ["/renewing-your-tax-credits-claim", "fefde62e-a660-41eb-a222-4f711a3208da", "redirect", Wed, 04 May 2016 18:51:16 UTC +00:00]]
 ```

In the example above, the most recently published content item for this base path is `bd88b94e-9fb0-44ea-84e8-6c3a4ef962a7`. However, the existing implementation will return the bottom one which is
`fefde62e-a660-41eb-a222-4f711a3208da`, because it's the last value written to the hash for that base path key.

This commit attempts to address this by filtering out redirects.

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -7,7 +7,9 @@ class LookupsController < ApplicationController
 
     base_paths_and_content_ids = ContentItemFilter
       .filter(state: states, base_path: base_paths)
-      .pluck(:base_path, :content_id)
+      .pluck(:base_path, :content_id, :document_type)
+      .reject { |arr| arr[2] == 'redirect' }
+      .map { |arr| arr[0..1] }
       .uniq
 
     response = Hash[base_paths_and_content_ids]
--- a/spec/requests/lookups_spec.rb
+++ b/spec/requests/lookups_spec.rb
@@ -23,6 +23,15 @@ RSpec.describe "POST /lookup-by-base-path", type: :request do
         "/only-unpublished-page" => "cc6d416c-f369-4b7c-bac7-5e9401e79362",
       )
     end
+
+    it "ignored redirects" do
+      FactoryGirl.create(:redirect_content_item, state: "unpublished", base_path: "/published-page", content_id: "aa491126-77ed-4e81-91fa-8dc7f74e9657")
+      base_paths = ['/published-page']
+
+      post "/lookup-by-base-path", base_paths: base_paths
+
+      expect(parsed_response).to eql({})
+    end
   end
 
   def create_test_content

### CODE REVIEW

File: app/controllers/lookups_controller.rb [hunk 1]
```diff
@@ -7,7 +7,9 @@ def by_base_path
 
     base_paths_and_content_ids = ContentItemFilter
       .filter(state: states, base_path: base_paths)
-      .pluck(:base_path, :content_id)
```
dougdroper: Does this work? 
```ruby
.where("document_type != 'redirect'")
```

boffbowsh:
It seems to me more that this call should instead use `Queries::GetLatest`, because it's not just redirects that shouldn't be sent back, it's gones and older content items that shouldn't be sent.

I think that's a question for someone like @elliotcm or @danielroseman though

elliotcm:
Looks like a data and/or assumptions problem.  We shouldn't have more than one "visible" content item for a single base path; content store doesn't work that way.

If we're considering unpublishings to be "visible" then we need to make sure we supersede them when there's a newer unpublishing or publishing.

tvararu:
It does seem like a data problem, since it's only a one-off content item which has it.

What's the right way to go about fixing it? Deleting the multiple redirects? What's the proper way to do that?

elliotcm:
Hi @tvararu, sorry replying took a while.

We've got some stuff in our roadmap to review these data constraints, is this currently blocking your work?

tvararu:
@elliotcm there's only one affected document in the detailed guide migration: https://www.gov.uk/guidance/tax-credits-working-out-income

The document republishes fine, but it links to a `redirect` rather than the real content item for the `/renewing-your-tax-credits` association, and redirects have `null` titles and other attributes.

The net effect of this is that the frontend will go from this:

screen shot 2016-06-22 at 10 24 02

To this, which is the best the frontend can do given that the linked expanded content item will only have a URL:

screen shot 2016-06-22 at 10 24 40

Not a particularly egregious regression, but it seems like it can be cleanly solved in this situation by simply deleting the multiple redirects.

elliotcm:
It'd be better to set the older content items to `superseded`, so if that makes sense for those content items then go ahead, and we'll look at the root cause separately.

elliotcm:
That said, pop round to second line some time and we can double check that's the best approach.

tvararu:
Discussed offline with @elliotcm: because of the circular nature of these redirects and that it's a one-off case, we've decided that it's safe to delete the offending content items.

I'm going to close this and open a new PR.

elliotcm:
@tvararu instructions for deleting content items here: https://github.com/alphagov/publishing-api/pull/379

PR #379, opened 2016-06-22
Add README instructions for deleting content items.

elliotcm:
@alphagov/team-publishing-platform This needs to be kept updated when we add or remove
supporting objects.


steventux:
:+1: Should any `Unpublishing`s also be removed?

elliotcm:
@steventux yep, thanks.  Added.

PR #380, opened 2016-06-22
Document required fields for PUT content

thehenster:
`schema_name` and `document_type` are required fields and `rendering_app` is nearly always required for `PUT /v2/content/:content_id`.

It looks like we're gradually removing the `format` field. Should the pact test in gds-api-adapters be updated to send `schema_name` and `document_type`?

https://github.com/alphagov/gds-api-adapters/blob/master/test/publishing_api_v2_test.rb#L14

### DIFF
--- a/doc/publishing-api-syntactic-usage.md
+++ b/doc/publishing-api-syntactic-usage.md
@@ -21,7 +21,9 @@ PUT and POST endpoints take an optional integer field `previous_version` in the 
  - `content_id` the primary identifier for the content being created or updated.
 Requests to create a new draft content item:
  - `base_path` must be a valid path format
+ - `document_type`
  - `publishing_app`
+ - `schema_name`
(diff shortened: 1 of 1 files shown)


thehenster:
I noticed https://github.com/alphagov/publishing-api/blob/master/doc/semantics-of-the-publishing-api.md today. It'd be easier to maintain if the list of required/optional parameters was only in one place.

From a consumers point of view.. I like the extra detail provided in the semantics doc but I think knowing which API call they relate to is important, which the syntactic document does.

My preference would be to see the semantics doc merged into the syntactic doc.

boffbowsh:
I've invited you to a documentation workshop @thehenster, I think this sort of thing will end up being discussed there.

thehenster:
:+1:

On second thought, I'm going to close this. If it's worth doing this it should probably be done in one go.

PR #381, opened 2016-06-22
Remove redundant tax credits redirects

tvararu:
Previous discussion: https://github.com/alphagov/publishing-api/pull/377

These are 3 redirects which are currently confusing `/lookup-by-base-path` queries for `/renewing-your-tax-credits-claim`. They redirect `/renewing-your-tax-credits-claim` to `/renew-your-tax-credits-claim`, which then redirects back to the original path, which is where the real content item for this path exists anyway. (`bd88b94e-9fb0-44ea-84e8-6c3a4ef962a7`)

While queries to the content-store seem to resolve fine by returning the final content item, `/lookup-by-base-path` lookups on publishing-api don't work well when more than one content item exists for a particular route. It returns one of these redirects, which isn't useful for dependency resolution expansion.

This commit removes the offending redirects using a data migration.

### DIFF
--- a/db/migrate/20160622100129_remove_redundant_tax_credits_redirects.rb
+++ b/db/migrate/20160622100129_remove_redundant_tax_credits_redirects.rb
@@ -1,0 +1,32 @@
+class RemoveRedundantTaxCreditsRedirects < ActiveRecord::Migration
+  def up
+    redirect_content_ids = [
+      "64d2120f-2396-4033-8e91-45f3e1f49ad8",
+      "6aecd670-0451-49d2-bce7-a4258be1adb3",
+      "fefde62e-a660-41eb-a222-4f711a3208da"
+    ]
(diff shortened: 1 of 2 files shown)


db/migrate/20160622100129_remove_redundant_tax_credits_redirects.rb
```diff
@@ -0,0 +1,31 @@
...
+    ]
+    redirect_content_ids.each do |content_id|
+      content_items = ContentItem.where(content_id: content_id)
+
+      supporting_classes = [
+        AccessLimit,
+        Linkable,
+        Location,
+        State,
+        Translation,
```
dougdroper: We are missing Unbublishing

tvararu: Updated! Thanks.

PR #383, opened 2016-06-23
Start validating message queue payloads with the schema

cbaines:
Currently just log failures, and continue as normal.

This is dependant on https://github.com/alphagov/govuk-content-schemas/pull/333

### DIFF
--- a/lib/queue_publisher.rb
+++ b/lib/queue_publisher.rb
@@ -43,6 +43,8 @@ class QueuePublisher
 private
 
   def publish_message(routing_key, message_data, options = {})
+    validate_message(message_data)
+
     publish_options = options.merge(routing_key: routing_key)
 
     exchange.publish(message_data.to_json, publish_options)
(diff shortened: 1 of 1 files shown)


lib/queue_publisher.rb
```diff
@@ -62,6 +64,29 @@ def publish_message(routing_key, message_data, options = {})
     raise
   end
 
+  def validate_message(message_data)
+    schema = JSON.load(File.read("govuk-content-schemas/message_queue.json"))
```
dougdroper: Is this validation different to the SchemaValidator? 
https://github.com/alphagov/publishing-api/blob/master/app/validators/schema_validator.rb

cbaines: I did originally try using that, but ran in to difficulties as it tried to validate using a sub schema, which didn't work in this case.

I ended up just implementing the necessary behaviour outside of it, but it would be good to improve the SchemaValidator class to make it usable in this case by improving the validation output (which would make using the sub schema unnecessary).

thehenster:
Hi. Is the plan to do this validation in all environments? The integration environment, even though it represents the master branch, is a shared testing area so it's quite likely that a change from one team could stop another team testing their work.

Maybe that is a good thing of course..

cbaines:
@thehenster I'm still getting up to speed on the use of the schemas, so I'm not too sure.

This change should not break anything, as it just does reporting on the output of the publishing api.

I've gone round the wiki and tagged all the pages relevant to this repository [1], which I have not read in detail yet, but might describe the intent of others regarding the use of the schemas.

1: https://gov-uk.atlassian.net/wiki/dosearchsite.action?cql=siteSearch+~+%22labelText%3Agovuk-content-schemas%22&queryString=labelText%3Agovuk-content-schemas

thehenster:
Cool.

I'm not sure what the answer to the problem is.. The purpose of the environment integration is in the name isn't it. It's supposed to check that stuff integrates together.

An "integrates in isolation" environment would require the whole stack with just your changes in it. As we don't have QAs here then perhaps this is just our development environments..

cbaines:
Closed in favor of https://github.com/alphagov/publishing-api/pull/387

lib/queue_publisher.rb
cbaines: Done in https://github.com/alphagov/publishing-api/pull/387

PR #384, opened 2016-06-23
Add 'state_history' to response bodies

tuzz:
https://trello.com/c/kyRoUV0h/756-add-a-state-history-field-to-presented-items-in-publishing-api-medium

This adds a `state_history` field to the response from ContentItemPresenter:

```javascript
"state_history": {
  "1": "superseded",
  "2": "published",
  "3": "draft"
}
```

This approach uses a correlated subquery. There are probably better ways to retrieve this data, but this seemed the simplest. I have benchmarked this change and it has a slight impact on performance. To serve 64 items, it has slowed from ~75ms to ~85ms. I don't know whether that is a reasonable slow-down in order to meet the requirement. What do people think?

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -14,6 +14,7 @@ module Presenters
         :lock_version,
         :internal_name,
         :updated_at,
+        :state_history,
       ]
 
       def self.present_many(scope, params = {})
@@ -100,6 +101,8 @@ module Presenters
             "to_char(public_updated_at, '#{iso8601_sql}') as public_updated_at"
           when :first_published_at
             "to_char(first_published_at, '#{iso8601_sql}') as first_published_at"
+          when :state_history
+            "#{state_history_sql} AS state_history"
           else
             field
           end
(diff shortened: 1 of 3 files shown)


boffbowsh:
If you run `bin/rails r bench/downstream_presenter.rb`, that should exercise your code, as long as it's involved in the downstream presentation. Otherwise, you'll want to write a new benchmark script for `ContentItemPresenter` following the style of the other scripts.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -125,6 +128,19 @@ def publication_state_sql
         SQL
       end
 
+      def state_history_sql
+        <<-SQL
+          (
+            SELECT json_agg((number, name))
+            FROM content_items c
```
dougdroper: I am not sure about the aliasing of the table names, I find that I have to keep doing the conversion from `c` to `content_items` in my head when reading it.

dougdroper:
~75ms to ~85ms seems reasonable. 

What was the reason to add it to the default fields?

app/presenters/queries/content_item_presenter.rb
tuzz: In this case, an alias is needed to disambiguate the joined table in this restriction:
`WHERE c.content_id = content_items.content_id`

Happy to change the `s` and `u` aliases if you think, though?

tuzz:
> @dougdroper: What was the reason to add it to the default fields?

> From [this commit](https://github.com/alphagov/publishing-api/commit/90e0090b62420aea37d37870c912ff77ce75a976): I made it so that the presenter defaults to return all fields. The presenter is now quick enough to do this and I think it serves as good documentation as to which fields are available.

Happy to re-visit / discuss my (somewhat arbitrary) decision to expose all fields. I think the performance is still pretty good. Not sure.

app/presenters/queries/content_item_presenter.rb
tuzz: Also – if you know a better way to do this `SELECT`, I'm all ears. I try to avoid correlated subqueries if at all possible as (traditionally?) they've been terrible for performance.

dougdroper: on no don't worry about it, I didn't see that restriction. 👍

PR #385, opened 2016-06-27
Migrate links

dougdroper:
Rake task to Represent the content item downstream without state changes.

Use the expanded link set presenter for links that are ready to migrate. (currently service_manual)

Use `links` for and second level expansion instead of `expanded_links` so that there are no changes to frontend apps. 

https://trello.com/c/yNL1Lziv/727-prepare-formats-to-switch-to-expanded-links

### DIFF
--- a/app/commands/v2/represent_downstream.rb
+++ b/app/commands/v2/represent_downstream.rb
@@ -2,27 +2,39 @@ module Commands
   module V2
     class RepresentDownstream
       def self.name
-        "Commands::V2::RepresentDownstream"
+        self.to_s
       end
 
       def call(scope)
         filter = ContentItemFilter.new(scope: scope)
 
-        draft_content_items = Queries::GetLatest.call(
-          filter.filter(state: %w{draft published}))
-        live_content_items = filter.filter(state: "published")
-
-        draft_content_items.each do |draft_content_item|
-          send_to_content_store(draft_content_item, Adapters::DraftContentStore)
+        items_for_draft_store(filter).find_each do |draft_store_item|
+          send_to_content_store(draft_store_item, Adapters::DraftContentStore)
         end
 
-        live_content_items.each do |live_content_item|
-          send_to_content_store(live_content_item, Adapters::ContentStore)
+        items_for_live_store(filter).find_each do |live_store_item|
+          send_to_content_store(live_store_item, Adapters::ContentStore)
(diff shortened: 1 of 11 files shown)


lib/tasks/migrate_links.rake
```diff
@@ -0,0 +1,23 @@
...
+  Usage
+  rake 'migrate_links:format[:schema_name]'
+  """
+  task :format, [:schema_name] => :environment do |_t, args|
+    schema_name = args[:schema_name]
+
+    raise "Schema name needs to be in MigrateExpandedLinks" unless Presenters::MigrateExpandedLinks.document_types.include?(schema_name)
+
+    Commands::V2::RepresentDownstream.new.call(
+      ContentItem.where(document_type: schema_name)
```
dougdroper: change to schema_name

app/commands/v2/represent_downstream.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        draft_content_items = Queries::GetLatest.call(
+          filter.filter(state: %w{draft published}))
+        live_content_items = filter.filter(state: "published")
+
+        draft_content_items.each do |draft_content_item|
+          send_to_content_store(draft_content_item, Adapters::DraftContentStore)
+        end
+
+        live_content_items.each do |live_content_item|
```
dougdroper: find_each

app/commands/v2/represent_downstream.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      def call(scope)
+        filter = ContentItemFilter.new(scope: scope)
+
+        draft_content_items(filter).find_each do |draft_content_item|
+          send_to_content_store(draft_content_item, Adapters::DraftContentStore)
+        end
+
+        live_content_items(filter).find_each do |live_content_item|
+          send_to_content_store(live_content_item, Adapters::ContentStore)
```
danielroseman: Should we not also re-send live content items to the draft content-store? (If so this should probably happen before the draft_content_items loop.)

danielroseman: Ah ignore me, I've just seen that draft_content_items includes live ones. Maybe rename to something like `items_for_live_store`?

danielroseman:
Apart from the one name change, :+1:

PR #386, opened 2016-06-27
Schema validator improvements

cbaines:
This applies the same approach I used for schema validation here [1], to the SchemaValidator class, such that it can be used for validation of message queue payloads. These changes will improve the error report usefulness for all other uses also.

1: https://github.com/alphagov/publishing-api/pull/383

### DIFF
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -26,6 +26,8 @@ private
 
     return true if errors.empty?
 
+    errors = errors.map { |e| present_error e }
+
     Airbrake.notify_or_ignore(
       {
         error_class: "SchemaValidationError",
@@ -37,6 +39,20 @@ private
       }
     )
     false
+  end
+
+  def present_error(error_hash)
+    # The schema key just contains an addressable, which is not informative as
+    # the schema in use should be clear from the error class and message
+    error_hash = error_hash.reject { |k,v| k == :schema }
+
+    if error_hash.has_key? :errors
+      error_hash[:errors] = Hash[
+        error_hash[:errors].map {|k,errors| [k,errors.map { |e| present_error e }]}
+      ]
+    end
+
+    error_hash
   end
 
   def schema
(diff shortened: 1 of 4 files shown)


app/validators/schema_validator.rb
```diff
@@ -1,35 +1,56 @@
...
+
+    errors = JSON::Validator.fully_validate(
+      schema,
+      payload,
+      errors_as_objects: true,
+    )
+
+    return true if errors.empty?
+
+    errors = errors.map { |e| present_error e }
```
danielroseman: Generally we use parens when calling methods with arguments.

app/validators/schema_validator.rb
```diff
@@ -1,35 +1,56 @@
...
-      explanation: "#{payload} schema validation error"
-    })
-    false
-  end
+  def present_error(error_hash)
+    # The schema key just contains an addressable, which is not informative as
+    # the schema in use should be clear from the error class and message
+    error_hash = error_hash.reject { |k, _| k == :schema }
+
+    if error_hash.has_key? :errors
```
danielroseman: Again, parens needed here.

app/validators/schema_validator.rb
```diff
@@ -1,35 +1,56 @@
...
+      error_hash[:errors] = Hash[
+        error_hash[:errors].map { |k, errors| [k, errors.map { |e| present_error e }] }
```
danielroseman: This would probably be clearer with the outer block as a do...end:

    error_hash[:errors].map do |k, errors|
        [k, errors.map { |e| present_error(e) }]
    end

app/validators/schema_validator.rb
cbaines: Fixed in  	018373c

app/validators/schema_validator.rb
cbaines: Fixed in 018373c

app/validators/schema_validator.rb
cbaines: Agreed, changed in a6ffab1

PR #387, opened 2016-06-27
Message queue schema validation (take 2)

cbaines:
Currently just log failures, and continue as normal.

This is dependant on https://github.com/alphagov/govuk-content-schemas/pull/333


cbaines:
I believe the test failures from Jenkins are just because the govuk-content-schemas are missing the message queue schema.

cbaines:
Closing, as this is not being worked on currently.

PR #388, opened 2016-06-28
Update state of five inconsistent Whitehall documents to 'published'

mobaig:
This is a precursor to https://github.com/alphagov/whitehall/pull/2635. Read that PR for more detail about the context of this change.

The Finding Things migration checker has detected an inconsistency in
the state of 5 documents across Whitehall > Publishing API > Content
Store. These items do not exist in the content store, and have
mismatched content IDs in Whitehall/Publishing API.

To fix this, we set the state of these documents to 'published' so that
in Whitehall we can fetch their content IDs via the lookup_content_id
endpoint and write the correct value to Whitehall's database.

trello https://trello.com/c/W3TLYeaM/683-make-the-links-checker-green

### DIFF
--- a/db/migrate/20160627100338_update_state_on_whitehall_docs.rb
+++ b/db/migrate/20160627100338_update_state_on_whitehall_docs.rb
@@ -1,0 +1,34 @@
+class UpdateStateOnWhitehallDocs < ActiveRecord::Migration
+  def up
+    # These are all currently in a 'draft' state. This is at odds with the
+    # state of these documents in Whitehall, where the slugs below represent
+    # published content.
+    doc_fixes = [
+      {
+        base_path: "/guidance/how-to-appeal-your-rateable-value",
+        content_id: "e76c9e81-9fce-48e5-8ccb-100fe77ac14c",
+      },
+      {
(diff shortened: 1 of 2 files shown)


tijmenb:
Would it be possible to do this via the API or a command, so that we have the entries in the event log?

mobaig:
@tijmenb sure, wanted to keep side-effects minimal given I'm manually fudging the data here but on second thought that's probably fine.

db/migrate/20160627100338_update_state_on_whitehall_docs.rb
```diff
@@ -0,0 +1,34 @@
...
+      },
+      {
+        base_path: "/government/world-location-news/uk-visa-operations-in-south-india-are-impacted-by-the-floods-in-chennai",
+        content_id: "4be55cf7-316b-4961-a293-05ebdd786c78",
+      },
+    ]
+
+    for_updating = ContentItem.where(content_id: doc_fixes.map { |doc| doc[:content_id] })
+    for_updating.each do |content_item|
+      Commands::V2::Publish.call(content_id: content_item.content_id, update_type: 'minor')
```
tijmenb: Should this be `content_item[:content_id]`?

PR #390, opened 2016-06-29
Support unpublishing drafts

cbaines:
Trello: https://trello.com/c/uXUhd3oJ/757-allow-unpublishing-of-draft-items-medium

This is being done to assist in the Specialist Publisher migration @tuzz

### DIFF
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -2,6 +2,11 @@ module Commands
   module V2
     class Unpublish < BaseCommand
       def call
+        if payload[:allow_draft] && payload[:discard_drafts]
+          message = "allow_draft and discard_drafts cannot be used together"
+          raise_command_error(422, message, fields: {})
+        end
+
         content_id = payload.fetch(:content_id)
         content_item = find_unpublishable_content_item(content_id)
 
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -17,7 +17,7 @@ module Commands
 
         check_version_and_raise_if_conflicting(content_item, previous_version_number)
 
-        if draft_present?(content_id)
+        if draft_present?(content_id) && !payload[:allow_draft]
           if payload[:discard_drafts] == true
             DiscardDraft.call(
               {
@@ -121,6 +121,11 @@ module Commands
 
       def find_unpublishable_content_item(content_id)
         allowed_states = %w(published unpublished)
+
+        if payload[:allow_draft]
+          allowed_states << "draft"
+        end
+
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
         filter.filter(locale: locale, state: allowed_states).first
       end
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -128,7 +128,7 @@ module Commands
         allowed_states = %w(published unpublished)
 
         if payload[:allow_draft]
-          allowed_states << "draft"
+          allowed_states = %w(draft)
         end
 
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -120,8 +120,9 @@ module Commands
       end
 
       def find_unpublishable_content_item(content_id)
+        allowed_states = %w(published unpublished)
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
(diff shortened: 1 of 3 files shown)


app/commands/v2/unpublish.rb
```diff
@@ -120,8 +120,14 @@ def previous_version_number
       end
 
       def find_unpublishable_content_item(content_id)
+        allowed_states = %w(published unpublished)
+
+        if payload[:allow_draft]
+          allowed_states.push("draft")
```
danielroseman: quibble, but `allowed_states << "draft"` is probably more idiomatic

spec/commands/v2/unpublish_spec.rb
```diff
@@ -91,6 +91,57 @@
...
+        end
+
+        it "deletes the linkable" do
+          described_class.call(payload_with_allow_draft)
+
+          linkable = Linkable.find_by(base_path: base_path)
+          expect(linkable).to be_nil
+        end
+
+        it "sends an unpublishing to the live content store" do
```
danielroseman: I guess we do want it to send the unpublishing, but probably worth just checking with @tuzz

app/commands/v2/unpublish.rb
cbaines: fa04f55

spec/commands/v2/unpublish_spec.rb
tuzz: Yes. We'd want the unpublishing to be sent.

tuzz:
Questions:

1) What should happen if I set the `allow_draft` flag and there's no draft to unpublish, but there is a published item? Currently, it unpublishes the published item. Is this what we want?

2) What should happen if I set both the `allow_draft` and `discard_drafts` flags? Currently it silently skips the discarding of drafts. Is this what we want?

Neither of these questions are leading, i.e. I don't know what the behaviour should be. I think it would be good to specify this behaviour in the tests and document its behaviour for clients.

cbaines:
> 1) What should happen if I set the allow_draft flag and there's no draft to unpublish, but there is a published item? Currently, it unpublishes the published item. Is this what we want?

Probably best to just error (status 422), I'll implement this. 

> 2) What should happen if I set both the allow_draft and discard_drafts flags? Currently it silently skips the discarding of drafts. Is this what we want?

Again, probably best to error. I'll implement this also.

I've spotted another issue with the current implementation, where the behaviour will be non-deterministic and possibly broken if an attempt is made to unpublish a draft for which there is a published or unpublished content item in existence. I'll fix this also.

app/commands/v2/unpublish.rb
```diff
@@ -2,6 +2,11 @@ module Commands
   module V2
     class Unpublish < BaseCommand
       def call
+        if payload[:allow_draft] && payload[:discard_drafts]
+          message = "allow_draft and discard_drafts cannot be used together"
+          raise_command_error(422, message, fields: {})
+        end
+
```
tuzz: 👍

app/commands/v2/unpublish.rb
```diff
@@ -17,7 +22,7 @@ def call
 
         check_version_and_raise_if_conflicting(content_item, previous_version_number)
 
-        if draft_present?(content_id)
+        if draft_present?(content_id) && !payload[:allow_draft]
```
tuzz: I wonder if there's a cleaner way to express this. Perhaps extracting a method and adding `!payload[:allow_draft]` as a guard clause?

It's a shame that the `discard_drafts` behaviour is intertwined with `allow_draft`. I was hoping the "cannot be used together" error would flatten out some of this logic but I guess that's not the case?

tuzz:
Looks good. 👍 

Did this get resolved?
> I've spotted another issue with the current implementation, where the behaviour will be non-deterministic and possibly broken if an attempt is made to unpublish a draft for which there is a published or unpublished content item in existence.

app/commands/v2/unpublish.rb
cbaines: I've had a look, but I'm unsure how to make this better. I'm not sure extracting it out in to a method would help, as that is just moving the complexity.

cbaines:
> Did this get resolved?
> 
> >  I've spotted another issue with the current implementation, where the behaviour will be non-deterministic and possibly broken if an attempt is made to unpublish a draft for which there is a published or unpublished content item in existence.

Yep, that actually ended up being simple to guard against here [1], as the issue was when the query in `find_unpublishable_content_item` would return some content item, where there is a choice of say a published item, and a draft. Just making unpublish only unpublish drafts when `allow_draft` is used works around that issue.

1: https://github.com/alphagov/publishing-api/pull/390/commits/dc0e2b63fd3b8fbce4dab2186a2301384207165b

PR #393, opened 2016-07-04
Make Sidekiq Redis namespace match GOVUK_APP_NAME.

danielroseman:
This is part of moving Sidekiq configuration to the govuk_sidekiq gem.


config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,7 @@
 redis_config = {
   host: ENV["REDIS_HOST"] || "127.0.0.1",
   port: ENV["REDIS_PORT"] || 6379,
-  namespace: "publishing_api",
+  namespace: "publishing-api",
```
dougdroper: Should we use the environment variable?

PR #396, opened 2016-07-05
Use constants when interpolating SQL.

elliotcm:
There was a worry that some of our SQL allowed SQL
injection.  This turned out not to be true, but
there was still uncertainty about the unescaped
string interpolation that occurs in the content
item presenter.

This interpolation is of static SQL strings with
no user input so it's safe, but for the sake of
clarity I've moved these into frozen constants.

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -86,7 +86,7 @@ module Presenters
         fields_to_select = fields.map do |field|
           case field
           when :publication_state
-            "#{publication_state_sql} AS publication_state"
+            "#{PUBLICATION_STATE_SQL} AS publication_state"
           when :user_facing_version
             "user_facing_versions.number AS user_facing_version"
           when :lock_version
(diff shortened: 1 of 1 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -116,40 +116,33 @@ def search(scope)
...
+          SELECT json_agg((number, name))
+          FROM content_items c
+          JOIN states s ON s.content_item_id = c.id
+          JOIN user_facing_versions u ON u.content_item_id = c.id
+          WHERE c.content_id = content_items.content_id
+          GROUP BY content_id
+        )
+      SQL
+
+      ISO8601_SQL = "YYYY-MM-DD\"T\"HH24:MI:SS\"Z\"".freeze
```
boffbowsh: Could use `'` here to avoid having to escape the `"`s?

boffbowsh:
Tiny comment otherwise 👍

app/presenters/queries/content_item_presenter.rb
```diff
@@ -86,23 +86,23 @@ def select_fields(scope)
...
+            "#{PUBLICATION_STATE_SQL} AS publication_state"
           when :user_facing_version
             "user_facing_versions.number AS user_facing_version"
           when :lock_version
             "lock_versions.number AS lock_version"
           when :description
             "description->>'value' AS description"
           when :internal_name
-            "#{internal_name_sql} AS internal_name"
+            "#{INTERNAL_NAME_SQL} AS internal_name"
```
steventux: Not entirely related to this PR but would it make sense to constantize the whole statement _and_ alias? Seems like the interpolation is redundant.

app/presenters/queries/content_item_presenter.rb
elliotcm: Trying to reduce the amount of change.

app/presenters/queries/content_item_presenter.rb
elliotcm: Frankly the entire branch of the case statement is redundant since we changed the data source for the linkables endpoint, but I didn't want to get into that here.

steventux: Fair enough :+1: I'm happy as is.

PR #397, opened 2016-07-05
[do not merge] Supersede previous unpublished/published items when unpublishing/publishing

tuzz:
This fixes a bug whereby the unpublish command was
finding the first ‘published’ or ‘unpublished’
item rather than the one with the highest
user-facing version.

This fixes the Specialist Publisher bug here:
https://trello.com/c/Ur355YbL/174-bug-cannot-unpublish-something-more-than-once-small-mediumish


cbaines:
Looks good to me

app/commands/v2/unpublish.rb
```diff
@@ -132,7 +132,9 @@ def find_unpublishable_content_item(content_id)
         end
 
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
-        filter.filter(locale: locale, state: allowed_states).first
```
cbaines: When I have looked in to other issues related to the unpublishing change, I did look at whether `.first` could be changed to something which would error if there is more than one result. I didn't find anything simple, but it may be worth adding an explicit check here?

tuzz: I suppose we could have raised if `.count > 1`, but it is valid for there to be more than one item, in this case because it finds all unpublished and published items.

cbaines: From my limited understanding of the state in the Publishing API, there should only be a maximum of 1 unpublished/published content item at any one time.

I really need to learn some more about this, as its unclear to me, which is very bad...

tuzz: I'm fairly sure that you can have more than one, e.g this workflow:

1) create draft
2) publish
3) unpublish
4) redraft
5) publish

In this case, you'll have both an unpublished content item and a published content item (with a higher user-facing version).

tuzz: That would also make sense. I suppose we'd have to supersede on an "allow_draft" unpublishing too, in that case. Should I close this in favour of that ticket?

This is a blocker for the SP team currently, so we might need to chat about prioritising it.

tuzz: We might need some guidance from @danielroseman to decide.

elliotcm: I can't find that ticket in our planning board anymore so worth bringing it up with @danielroseman and our DM

danielroseman: I agree with @elliotcm ; when you publish a draft it would supersede the previous version, whether it's published or unpublished.

tuzz: 👍  following a discussion, I'm going to try to implement the superseding behaviour.

app/commands/v2/unpublish.rb
```diff
@@ -132,7 +133,23 @@ def find_unpublishable_content_item(content_id)
...
+          content_item,
+          state: "unpublished",
+          base_path: nil,
+          user_version: nil,
+        )
+
+        items.each do |item|
+          next if item == content_item
+          State.supersede(item)
+        end
```
cbaines: I would prefer if this just found the previous unpublished, or published item, and superseeded it, as this mimics the behaviour in the publish endpoint (apart from I think there is a bug there also... in that it does not superseed previously unpublished content items).

If existing content item states are migrated, this should have the same behaviour, but the code will be simpler.

db/migrate/20160706110723_supersede_previous_unpublished_items.rb
```diff
@@ -0,0 +1,25 @@
...
+      .joins("INNER JOIN translations t on t.content_item_id = content_items.id")
+      .joins("INNER JOIN user_facing_versions u on u.content_item_id = content_items.id")
+      .group(:content_id, :locale)
+      .having("count(content_id) > 1")
+      .pluck("json_agg((states.id, u.number))")
+      .map do |results|
+        results.map { |row| { state_id: row["f1"], version: row["f2"] } }
+      end
+
+    states_to_supersede = state_histories.map do |history|
```
boffbowsh: I think this should be `flat_map`? This will produce an array of arrays. The outer array represents content_ids, the inner array is the content_item_ids. `flat_map` will just give you a single flat array of content_ids. This probably doesn't matter in practice except your `.count` below will be more accurate.

app/commands/v2/unpublish.rb
cbaines: (have talked with @tuzz and he was already on top of the changes to the publish endpoint...)

tuzz:
Closed in favour of https://github.com/alphagov/publishing-api/pull/405.

PR #399, opened 2016-07-05
Remove incorrect placeholder `Publication`s

gpeng:
This PR is a migration to remove two `Publication` content items that are causing their `HtmlAttachment`
republishing sync checks to fail as their content ids don't match with the `HtmlAttachment`. 

There is a [corresponding PR on Whitehall](https://github.com/alphagov/whitehall/pull/2649)


boffbowsh:
Closing in favour of creating a `Command` to perform this

PR #400, opened 2016-07-05
Allow full deletion of content including URL.

elliotcm:
Content which should never have been on the site
and needs to be fully removed including existence
of its URL (for example, early release of a document
which references a company or stock in its URL)
can send an unpublishing of type "vanish" to
indicate that it was published in error.

This is similar to a "gone" unpublishing except
the URL will 404 instead of 410.

### DIFF
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -46,6 +46,8 @@ module Commands
           redirect(content_item)
         when "gone"
           gone(content_item)
+        when "vanish"
+          vanish(content_item)
         else
           message = "#{type} is not a valid unpublishing type"
           raise_command_error(422, message, fields: {})
@@ -102,6 +104,12 @@ module Commands
         send_downstream(gone)
       end
 
+      def vanish(content_item)
+        State.unpublish(content_item, type: "vanish")
+
+        delete_from_downstream(Location.find_by(content_item: content_item).base_path)
+      end
+
       def delete_linkable(content_item)
         Linkable.find_by(content_item: content_item).try(:destroy)
       end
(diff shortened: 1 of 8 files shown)


elliotcm:
Adding docs now..

app/commands/v2/unpublish.rb
```diff
@@ -46,6 +46,8 @@ def call
           redirect(content_item)
         when "gone"
           gone(content_item)
+        when "vanish"
+          vanish(content_item)
```
tuzz: Vanish

tuzz:
Makes sense to me. 👍

app/commands/v2/unpublish.rb
```diff
@@ -116,6 +124,16 @@ def send_downstream(downstream_payload)
...
+      def delete_from_downstream(base_path)
+        return unless downstream
+
+        PresentedContentStoreWorker.perform_async(
+          content_store: Adapters::ContentStore,
+          base_path: base_path,
+          delete: true,
+        )
+      end
+
```
tuzz: Should we also delete from the draft content store?

tuzz:
Should we also invalidate the fastly / varnish caches?

app/commands/v2/unpublish.rb
elliotcm: The "gone" behaviour is still appropriate here.

elliotcm:
@tuzz nah, that seems like it belongs as part of whatever task invokes this since it's expected to be a devops operation.

boffbowsh:
LGTM. I briefly agreed with Tuzz's point about the draft content store based on things ending up on non-SC developer laptops, but the same is true of it remaining in the Publishing API.

👍

PR #401, opened 2016-07-05
Add contacts to pathless formats

steventux:
Part of: https://trello.com/c/JxKRVn6c/731-export-whitehall-contacts-to-publishing-api-medium

Adds `contact` as a pathless content item format.
There was also some undesirable/missing behaviour for drafting and publishing content without a base path so this PR adds a way to draft and publish 'pathless' content without sending it to the content store(s).


boffbowsh:
LGTM 👍

PR #402, opened 2016-07-06
Revert "Remove unpublishing for previously published item..."

steventux:
This reverts commit 63feb0efa7e85ff522ec9201a63dfe612dd41a9f.

There's no need to remove an unpublishing when a piece of
content is unwithdrawn, the unpublished content item will
be superseded and the unpublishing will be related to this,
not the new (redrafted and published) item.


cbaines:
:+1:

PR #405, opened 2016-07-08
Supersede publish and unpublish

tuzz:
Re-opened as the other pull request was noisy (https://github.com/alphagov/publishing-api/pull/397).

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -28,13 +28,12 @@ module Commands
 
         check_version_and_raise_if_conflicting(content_item, previous_version_number)
 
-        previous_item = lookup_published_item(content_item)
+        previous_item = lookup_previous_item(content_item)
         previous_location = Location.find_by(content_item: previous_item)
 
-        publish_redirect_if_content_item_has_moved(location, previous_location, translation)
-
-        previous_item = lookup_published_item(content_item)
         State.supersede(previous_item) if previous_item
+
+        publish_redirect_if_content_item_has_moved(location, previous_location, translation)
 
         clear_published_items_of_same_locale_and_base_path(content_item, translation, location)
 
@@ -131,10 +130,10 @@ module Commands
         ) if draft_redirect
       end
 
-      def lookup_published_item(content_item)
+      def lookup_previous_item(content_item)
         ContentItemFilter.similar_to(
           content_item,
-          state: "published",
+          state: %w(published unpublished),
           base_path: nil,
           user_version: nil,
         ).first
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -131,12 +131,18 @@ module Commands
       end
 
       def lookup_previous_item(content_item)
-        ContentItemFilter.similar_to(
+        previous_items = ContentItemFilter.similar_to(
           content_item,
           state: %w(published unpublished),
           base_path: nil,
           user_version: nil,
-        ).first
+        ).to_a
+
+        if previous_items.size > 1
+          raise "There should only be one previous published or unpublished item"
+        end
+
+        previous_items.first
       end
 
       def send_downstream(content_item, update_type)
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -39,6 +39,9 @@ module Commands
           end
         end
 
+        previous_item = lookup_previous_item(content_item)
+        State.supersede(previous_item) if previous_item
+
         case type = payload.fetch(:type)
         when "withdrawal"
           withdraw(content_item)
@@ -54,7 +57,6 @@ module Commands
         end
 
         delete_linkable(content_item)
-        supersede_previous_items(content_item)
 
         Success.new(content_id: content_id)
       end
@@ -153,18 +155,13 @@ module Commands
         filter.filter(locale: locale, state: allowed_states).last
       end
 
-      def supersede_previous_items(content_item)
-        items = ContentItemFilter.similar_to(
+      def lookup_previous_item(content_item)
+        ContentItemFilter.similar_to(
           content_item,
           state: %w(published unpublished),
           base_path: nil,
           user_version: nil,
-        )
-
-        items.each do |item|
-          next if item == content_item
-          State.supersede(item)
-        end
+        ).first
       end
 
       def draft_present?(content_id)
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -54,6 +54,7 @@ module Commands
         end
 
         delete_linkable(content_item)
+        supersede_previous_items(content_item)
 
         Success.new(content_id: content_id)
       end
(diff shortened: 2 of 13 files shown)


config/application.rb
```diff
@@ -28,6 +28,7 @@ class Application < Rails::Application
     config.autoload_paths << "#{config.root}/app"
     config.autoload_paths << "#{config.root}/lib"
     config.autoload_paths += Dir["#{config.root}/lib/**/"]
+    config.autoload_paths << "#{config.root}/db/migrate"
```
boffbowsh: I think it'd be better to explicitly require that helper in the migration that needs it. They shouldn't ever be needed in production code outside of migrations.

db/migrate/20160708150030_delete_substitutions.rb
```diff
@@ -0,0 +1,38 @@
+class DeleteSubstitutions < ActiveRecord::Migration
```
elliotcm: Should we extract this into a rake task in another PR, and ship it first?

tuzz: Yes, let's do that. 👍

elliotcm:
Half of this has been extracted to https://github.com/alphagov/publishing-api/pull/407

elliotcm:
@boffbowsh Updated.

elliotcm:
Tests will fail until https://github.com/alphagov/publishing-api/pull/407 is merged and this is rebased onto master.

cbaines:
Its unclear to me how https://github.com/alphagov/publishing-api/pull/407/commits/7978bd7621f8dc5aa2e47373bb94b99e8ee712e1 is necessary as preparation for fixing the state history regarding published and unpublished items?

app/commands/v2/unpublish.rb
```diff
@@ -149,7 +150,21 @@ def find_unpublishable_content_item(content_id)
...
+        items = ContentItemFilter.similar_to(
+          content_item,
+          state: %w(published unpublished),
+          base_path: nil,
+          user_version: nil,
+        )
+
+        items.each do |item|
+          next if item == content_item
+          State.supersede(item)
```
cbaines: There should only be 0, or 1 item meeting these criteria, so I don't think this should be a loop.

elliotcm:
@cbaines It makes the other task simpler.  The other task supersedes all but the most recent published or unpublished version of the content item, but with this bad data floating around there'll be some where the most recent unpublished version is actually an old draft which should've been deleted.

If we get rid of them all from the beginning there's only 50+ pieces of content which still have both a published and unpublished version at the same time, rather than the 1100+ currently.

app/commands/v2/unpublish.rb
elliotcm: We don't have any validations which ensure this is the case (yet, there's a ticket in the backlog) so for the sake of robustness we agreed to take the more careful approach.

app/commands/v2/publish.rb
```diff
@@ -131,10 +130,10 @@ def publish_redirect_if_content_item_has_moved(new_location, previous_location,
...
 
-      def lookup_published_item(content_item)
+      def lookup_previous_item(content_item)
         ContentItemFilter.similar_to(
           content_item,
-          state: "published",
+          state: %w(published unpublished),
           base_path: nil,
           user_version: nil,
         ).first
```
cbaines: I think it would be best if an explicit error was raised here, if more than one content item is returned.

app/commands/v2/unpublish.rb
```diff
@@ -153,18 +155,13 @@ def find_unpublishable_content_item(content_id)
...
           state: %w(published unpublished),
           base_path: nil,
           user_version: nil,
-        )
-
-        items.each do |item|
-          next if item == content_item
-          State.supersede(item)
-        end
+        ).first
```
cbaines: Same as https://github.com/alphagov/publishing-api/pull/405/commits/ee5626ed7503e0615e6238de397dede2606a3463#r70237338 , it would be good to have an explicit error.

app/commands/v2/unpublish.rb
cbaines: On 11/07/16 11:48, Elliot Crosby-McCullough wrote:
>> @@ -149,7 +150,21 @@ def find_unpublishable_content_item(content_id)
>>          end
>>  
>>          filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
>> -        filter.filter(locale: locale, state: allowed_states).first
>> +        filter.filter(locale: locale, state: allowed_states).last
>> +      end
>> +
>> +      def supersede_previous_items(content_item)
>> +        items = ContentItemFilter.similar_to(
>> +          content_item,
>> +          state: %w(published unpublished),
>> +          base_path: nil,
>> +          user_version: nil,
>> +        )
>> +
>> +        items.each do |item|
>> +          next if item == content_item
>> +          State.supersede(item)
> 
> We don't have any validations which ensure this is the case (yet, there's a ticket in the backlog) so for the sake of robustness we agreed to take the more careful approach.

I would say that the most careful approach would be to check that
reality matches the assumptions (only one item to supersede), and error
if it does not.

As I see it, the lack of confidence that this assumption holds makes it
more important to do this, not less.

app/commands/v2/publish.rb
tuzz: I think it would be better to go even further and make this a hard constraint of the system, i.e. that you can only have on published/unpublished content item in this history of a content_id.

There's a ticket here to address this: https://trello.com/c/EZXtVEpz/247-only-allow-one-of-unpublished-or-published-content-items

I think this would be a better approach because you'd check in a single place that this constraint has been met rather than repeatedly checking your assumptions and potentially littering the code.

I can add it now, but I think it would better to do the ticket linked above. What do you think?

cbaines: On 11/07/16 12:49, Chris Patuzzo wrote:
>>          ContentItemFilter.similar_to(
>>            content_item,
>> -          state: "published",
>> +          state: %w(published unpublished),
>>            base_path: nil,
>>            user_version: nil,
>>          ).first
> 
> I think it would be better to go even further and make this a hard constraint of the system, i.e. that you can only have on published/unpublished content item in this history of a content_id.
> 
> There's a ticket here to address this: https://trello.com/c/EZXtVEpz/247-only-allow-one-of-unpublished-or-published-content-items
> 
> I think this would be a better approach because you'd check in a single place that this constraint has been met rather than repeatedly checking your assumptions and potentially littering the code.
> 
> I can add it now, but I think it would better to do the ticket linked above. What do you think?

I would like to see a check here.

I'm uncertain what form the implementation of that ticket will take, and
even if it works out well, I think a explicit check is still useful, as
there are other things that could go wrong and cause multiple results to
be returned (e.g. if the query is wrong). Even excluding correctness,
having a explicit check informs the reader that there should be one result.

tuzz: Ok. Will add.

elliotcm: I'm with @tuzz on this one, in that it's best to provide a consistent worldview in which to work and then work more simply within it, rather than constantly validating that worldview whenever taking any action.

Up to him whether he implements the check, though.

steventux: I wouldn't want to slow this PR up but I'd also prefer constraint over a noisy assertion here.

tuzz:
@cbaines: I have addressed your comments. Please could you take another look? Thanks.

cbaines:
> @cbaines: I have addressed your comments. Please could you take another look? Thanks.

Yep, looks good to me.

Thanks for going above and beyond with this as well :)

PR #406, opened 2016-07-11
Speed up represent downstream

cbaines:
Instead of pulling out batches of ContentItems, just select the two required
fields for all instances.


boffbowsh:
Yassssss 👍

PR #407, opened 2016-07-11
Supersede prep

elliotcm:
Split out from https://github.com/alphagov/publishing-api/pull/405

We need to remove all unpublishings of type
"substitution".  This is due to some inconsistent
data where some draft content items where unpublished
and thus impossible to tell apart from previously
live content.

Since we can't easily tell them apart, we're
taking a scorched earth approach and deleting all
of them.  Since no apps are yet phase 2 no data
will be permanently lost.


app/substitution_helper.rb
```diff
@@ -10,7 +10,12 @@ def clear!(new_item_document_type:, new_item_content_id:, base_path:, locale:, s
         allowed_to_substitute = (substitute?(new_item_document_type) || substitute?(blocking_item.document_type))
 
         if mismatch && allowed_to_substitute
-          State.substitute(blocking_item)
+          if state == "draft"
+            Commands::V2::DiscardDraft.call(content_id: blocking_item.content_id)
```
steventux: Is there a risk of this command nesting producing a race condition on something relying on persistence outside the outer transaction?

elliotcm: Hopefully not; the DiscardDraft command uses the same `after_transaction_commit` block that the other commands do for talking downstream.

PR #408, opened 2016-07-11
Update documentation for deprecated format field

kevindew:
I've attempted to update the documentation for the deprecation of the format field. I don't think this quite complete and I have a few questions to resolve to complete this:

1. How to define what a document_type should match? I understand it’s the document type that is selected when creating a document on whitehall or similar systems, but don’t know of a list.
2. Is document_type, schema_name or both used to determine which layout is used to render the item?
3. Whether key information should be on document_type field and/or schema_name field.
    1. Redirect or gone types - I’ve tied to document_type
    2. Placeholder I’ve put on the schema_name
4. The format we’d use to mark a field as deprecated - as is the case with the format field

### DIFF
--- a/doc/semantics-of-the-publishing-api.md
+++ b/doc/semantics-of-the-publishing-api.md
@@ -57,31 +57,6 @@ a tendency to change.
 
 ---
 
-### format
-
-Examples: *manual, policy, redirect*
-
-Required: Yes
-
-The `format` specifies the data format of the content item as per the
-[GOV.UK content schemas](https://github.com/alphagov/govuk-content-schemas).
-It is used downstream to render the content item according a specific layout for
-that `format`.
-
-If the `format` is one of either *redirect* or *gone*, the content item is
-considered non-renderable and this waives the requirement for some of the other
-fields in the content item to be present, namely `title`, `rendering_app` and
-`public_updated_at`.
-
-At present, not all content goes through the publishing pipeline, but there is
-still a need to link to content items on our legacy infrastructure. There are
-some special formats that can be used in these cases. The `format` should be
-prefixed with *placeholder_* or set to *placeholder*. See
-[here](https://github.com/alphagov/content-store/blob/master/doc/placeholder_item.md)
-for more information.
-
----
-
 ### publishing_app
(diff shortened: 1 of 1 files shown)


elliotcm:
1.  There's no single list (yet?), so the best validation we have for now is that it's an a-z string with underscores.
2.  The URL routing typically determines what layout to use, and the content store is looked up by the path from the URL, so the type doesn't really come into it at that point.  It's more often used during bulk requests, e.g. to the `linkables` endpoint in publishing API.  `schema_name` isn't used there.

Not sure about the others.

cbaines:
While `schema_name` seems to have replaced `format` in the value sense, as far as I understand the situation, they have very different semantics.

I would prefer if `schema_name` is used only for identifying which schema file to use, and nothing more (and I think it would be good to even get rid of the need to do this in the future).

So, instead of replacing `format` with `schema_name`, I would suggest using `document_type` (apart from if its being used to refer to the name of the schema).

doc/semantics-of-the-publishing-api.md
```diff
@@ -228,6 +241,27 @@ non-renderable (see [**format**](#format)).
 
 ---
 
+### schema_name
+
+Examples: *manual, policy, redirect*
+
+Required: Conditionally
+
+The `schema_name` specifies the data format of the content item as per the
```
cbaines: I would just say that `schema_name` specifies the name of the schema file in the govuk-content-schemas repository.

danielroseman:
See [the original RFC](https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+41%3A+Separate+document+type+from+format): the intention was definitely that `schema_name` is only used to identify the schema to validate against.

kevindew:
Thanks - I've updated the doc now based off your collective feedback and to be closer to the RFC. 

I wonder if we should give advice to users who have no idea what the `document_type` might be? In many cases it is the same as the `schema_name` or is it ok to assume that end users are familiar with the `document_type` terminology?

kevindew:
Superseded by the refactor documentation drive.

PR #410, opened 2016-07-11
Allow base_path lookups for a single given state

rgarner:
* Gives us a way to see what existing item is causing a collision in
  dfid-transition
* Don't allow multiple state query, as some combinations (e.g.
  live/draft) are inadvisable

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -2,7 +2,7 @@ class LookupsController < ApplicationController
   def by_base_path
     # return content_ids for content that is visible on the live site
     # withdrawn items are still visible
-    states = %w(published unpublished)
+    states = params[:state] || %w(published unpublished)
     base_paths = params.fetch(:base_paths)
 
     base_paths_and_content_ids = ContentItemFilter
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,8 +1,11 @@
 class LookupsController < ApplicationController
+  # return content_ids for content that is visible on the live site
+  # withdrawn items are still visible
+  DEFAULT_STATES_FILTER = %w(published unpublished).freeze
+
   def by_base_path
-    # return content_ids for content that is visible on the live site
-    # withdrawn items are still visible
-    states = params[:state] || %w(published unpublished)
+    states = params[:state] ? valid_state_params : DEFAULT_STATES_FILTER
+
     base_paths = params.fetch(:base_paths)
 
     base_paths_and_content_ids = ContentItemFilter
@@ -12,5 +15,17 @@ class LookupsController < ApplicationController
 
     response = Hash[base_paths_and_content_ids]
     render json: response
+  end
+
+private
+
+  def valid_state_params
+    unless State.allowed_values.include?(params[:state])
+      raise CommandError.new(
+        code: 422,
+        message: "state: '#{params[:state]}' not one of #{State.allowed_values}"
+      )
+    end
+    params[:state]
   end
 end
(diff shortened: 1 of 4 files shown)


tuzz:
Looks good. I wonder if it'd be worth validating that the provided state is in an enumerated set and raising a helpful error message if it is not?

rgarner:
Good idea. Where's the authoritative source for that set?

rgarner:
`State.select(:name).group(:name).to_a` gets me a list of four – that look ok?

```
  State Load (287.7ms)  SELECT "states"."name" FROM "states" GROUP BY "states"."name"
=> [#<State:0x007fc30b598e88 id: nil, name: "published">,
 #<State:0x007fc30b598d20 id: nil, name: "draft">,
 #<State:0x007fc30b598bb8 id: nil, name: "superseded">,
 #<State:0x007fc30b598a50 id: nil, name: "unpublished">]
```

kevindew:
@rgarner yup that's all of them

app/controllers/lookups_controller.rb
```diff
@@ -1,8 +1,11 @@
 class LookupsController < ApplicationController
+  DEFAULT_STATES_FILTER = %w(published unpublished)
```
kevindew: can we run freeze on this? consistent with other array constants in app

rgarner: good point

dougdroper:
@rgarner is this waiting for anything?

rgarner:
@dougdroper Don't think so. The *need* for it kind of lessened (or vanished) after #411 was merged. I kind of assumed this would be merged in the same pass, but it turned out not to be important if it was merged or not for the [republish-on-conflict use case](https://github.com/alphagov/dfid-transition/blob/master/lib/dfid-transition/load/output.rb#L51-L54) in dfid-transition. If it's considered generally useful for other use cases, though, go ahead and merge.

BTW, I'm still unclear on the modern etiquette for merging in alphagov; it used to (circa 2011) be the case that you seldom merged your own, but this seems to have been supplanted on some projects by self-merge in the presence of sufficient 👍 👍 👍 s. What's the consensus in this repo?

dougdroper:
@rgarner I think the self-merge is project specific. In this case we tend to get other people to merge, but we are fairly pragmatic. 

I am not sure if it is generally useful but it offers flexibility which is excellent, although the only thing I would like to see is the corresponding documentation change. https://github.com/alphagov/publishing-api/blob/5e7597f7d80d411c636091078cf0663d9f5004eb/doc/api.md#post-lookup-by-base-path

And annoyingly it has lost the build information, so don't know if it is green or not.

rgarner:
Ah, ok, well fixing the docs will queue another build, so let's go with it after I've done that.

rgarner:
(disclaimer: will be done in spare time, live date Monday ;) )

tijmenb:
Would someone object to closing this PR and https://github.com/alphagov/gds-api-adapters/pull/536? I'd rather not add parameters to this endpoint if it's not strictly necessary. If we do have a use day-to-day use case for this, I'd rather go for a parameter that allows you only to switch between draft and live, so that we don't need to deal with other states here.

rgarner:
Hi Tijmen, on holiday at the moment but as I recall we dealt with the use
case for this by getting the conflict exception to tell us with which draft
content ID it was conflicting (optimistic drafting on PUT rather than this
pessimistic pre-lookup). I left this open because there was a comment that
liked the flexibility, but it's not strictly necessary. Do close if that's
what you'd like - seems like it might be more trouble than it's worth.

tijmenb:
Cool, thanks for responding! When the need arises again we'll revisit this.

PR #411, opened 2016-07-12
Return the content_id of the duplicate item on error

rgarner:
Allows us to recover in batch import processes like dfid-transition
without having to issue a second lookup to the publishing API to find
out what content we're colliding with. Although `additional_items.any?`
and the concomitant `additional_items.first.content_id`
might imply that there could be more than one conflicting item,
discussions with the publishing team indicate that this situation
could be described as an exotic corner case.


tuzz:
Huge 👍  from me. This will help tremendously with debugging.

PR #414, opened 2016-07-13
Allow publishing apps to request a specific version.

elliotcm:
During the Specialist Publisher migration work, we found that we need to be able to retrieve the published version of content so that we can re-trigger email alerts if they fail to send (for whatever reason). Currently, you can only retrieve the latest piece of content, which is sometimes a draft if the content has been redrafted.

Additionally, existing applications provide features that allow users of those apps to view superseded content. Travel Advice Publisher, for example, allows users to view "archived" content which is not currently supported by the Publishing API.

https://trello.com/c/rPoiiBbz/762-allow-get-content-item-requests-to-specify-a-user-facing-version

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -24,7 +24,11 @@ module V2
     end
 
     def show
-      render json: Queries::GetContent.call(path_params[:content_id], query_params[:locale])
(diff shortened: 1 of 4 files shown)


elliotcm:
Implementation here: https://github.com/alphagov/gds-api-adapters/pull/538

PR #421, opened 2016-07-14
Return 422 on put-content requests without publishing_app

kevindew:
When a request is submitted to PUT /v2/content/:content-id without a
publishing_app a 500 response is returned. These changes instead return a 422
and a descriptive message.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -5,6 +5,12 @@ module Commands
         raise_if_links_is_provided
         validate_schema
 
+        if publishing_app.blank?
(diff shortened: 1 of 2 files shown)


app/commands/v2/put_content.rb
```diff
@@ -219,7 +225,7 @@ def locale
       end
 
       def publishing_app
-        payload.fetch(:publishing_app)
+        payload.fetch(:publishing_app, nil)
```
elliotcm: This is equivalent to `payload[:publishing_app]` which is a bit clearer for this use case.

PR #422, opened 2016-07-15
Add indexes for content_item foreign keys

kevindew:
This commit adds indexes on the foreign key of tables that have a
relationship with content_item.
This can speed up some of the queries we run accessing these tables via
content_item_id

### DIFF
--- a/db/migrate/20160715153009_add_content_item_indexes.rb
+++ b/db/migrate/20160715153009_add_content_item_indexes.rb
@@ -1,0 +1,8 @@
+class AddContentItemIndexes < ActiveRecord::Migration
+  def change
+    add_index :access_limits, :content_item_id
+    add_index :linkables, :content_item_id
+    add_index :states, :content_item_id
+    add_index :translations, :content_item_id
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160706110723) do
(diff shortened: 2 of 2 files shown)


dougdroper:
do we know what the performance impact of this will be? Are there some metrics we can see.

kevindew:
Sure discussion on this is on the trello card: https://trello.com/c/b0mrOatq/758-publishing-api-performance-improvements-publish

Putting an index on content_id for linkables has quite a significant reduction in time spent running this task: https://github.com/alphagov/publishing-api/pull/423 however the indexes on the other tables hasn't actually had too large an impact (which was a little surprising given how frequently we instantiate objects by querying via content_id)

dougdroper:
```
== 20160715153009 AddContentItemIndexes: migrating ============================
-- add_index(:access_limits, :content_item_id)
   -> 0.0386s
-- add_index(:linkables, :content_item_id)
   -> 0.3375s
-- add_index(:states, :content_item_id)
   -> 0.9427s
-- add_index(:translations, :content_item_id)
   -> 0.7850s
== 20160715153009 AddContentItemIndexes: migrated (2.1041s) ===================
```

I think that should be ok to run on production, @danielroseman is that ok? or do we need to run this out of hours? (this will only block writes)

danielroseman:
I am slightly nervous; 2 seconds is a noticeable amount of time. If someone hit publish in specialist-publisher for example during that time, they might well get a timeout error.

danielroseman:
As discussed, I'm going to merge this, we should make sure we deploy it at a quiet time.

PR #426, opened 2016-07-19
Specific ordering for links relation of LinkSet

kevindew:
This PR introduces ordering to the links relation of LinkSet. This introduces some additional consistency to the results and resolves an intermittent test failure: https://ci.dev.publishing.service.gov.uk/job/govuk_publishing-api_branches/2067/

Technically the ordering of the links isn't something we care about or something that can be defined by the request. Although by having no ordering specified on the relation we're leaving it up to Postgres to determine the ordering which means it may vary. Making this consistent will mean more uniformity in responses, which would be beneficial if anyone is diffing responses.

### DIFF
--- a/app/models/link_set.rb
+++ b/app/models/link_set.rb
@@ -1,3 +1,3 @@
 class LinkSet < ActiveRecord::Base
-  has_many :links, dependent: :destroy
+  has_many :links, -> { order(id: :asc) }, dependent: :destroy
 end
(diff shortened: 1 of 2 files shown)


cbaines:
:+1:

app/models/link_set.rb
```diff
@@ -1,3 +1,3 @@
 class LinkSet < ActiveRecord::Base
-  has_many :links, dependent: :destroy
+  has_many :links, -> { order(id: :asc) }, dependent: :destroy
```
cbaines: It might be better to order by `content_id`, rather than `id` as that means that the order can be communicated to users, as `content_id` is external, and `id` is internal?

PR #427, opened 2016-07-19
Make version check in put content clearer

kevindew:
I thought I had found a bug in the code where it allowed the state of the application to be mutated on a validation error. It turns out that this is actually not the case as all the database changes are wrapped in a transaction which the exception causes a rollback.

Despite this I thought these tests and the change in the ordering of the code could make it easier to understand the intentions of the code and could help someone else not come to the conclusion I came to.

Can understand though if people agree that this isn't necessary to merge in.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -62,14 +62,14 @@ module Commands
       end
 
       def update_existing_content_item(content_item)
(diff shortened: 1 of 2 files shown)


cbaines:
:+1: but I don't feel confident enough to just merge this...

cbaines:
Doug and Kevin have talked it though with me, and I'm probably being overly pessimistic, as this is a really safe change to make.

PR #429, opened 2016-07-20
Allow pathless formats to accept a base_path

steventux:
https://govuk.zendesk.com/agent/tickets/1371252

The `contact` format is ambiguous in accepting a `base_path` but not requiring it.
Contacts from whitehall do not have a path and are never sent to the content store whereas contacts-admin content does use this field and is presented to the content store as expected.
This change allows for content items of a pathless format to use the `base_path` if supplied.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -33,7 +33,7 @@ module Commands
 
         State.supersede(previous_item) if previous_item
 
-        unless content_item.pathless?
+        unless pathless?(content_item)
           publish_redirect_if_content_item_has_moved(location, previous_location, translation)
           clear_published_items_of_same_locale_and_base_path(content_item, translation, location)
         end
@@ -107,6 +107,11 @@ module Commands
         content_item.update_attributes!(first_published_at: Time.zone.now)
       end
 
+      def pathless?(content_item)
+        !Location.exists?(content_item: content_item) &&
+          ContentItem::EMPTY_BASE_PATH_FORMATS.include?(content_item.schema_name || content_item.format)
+      end
+
       def publish_redirect_if_content_item_has_moved(new_location, previous_location, translation)
         return unless previous_location
         return if previous_location.base_path == new_location.base_path
@@ -154,7 +159,7 @@ module Commands
 
         PublishingAPI.service(:queue_publisher).send_message(queue_payload)
 
-        return if content_item.pathless?
+        return if pathless?(content_item)
 
         queue = update_type == 'republish' ? PresentedContentStoreWorker::LOW_QUEUE : PresentedContentStoreWorker::HIGH_QUEUE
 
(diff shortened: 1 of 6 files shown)


app/commands/v2/publish.rb
```diff
@@ -107,6 +107,11 @@ def set_first_published_at(content_item)
         content_item.update_attributes!(first_published_at: Time.zone.now)
       end
 
+      def pathless?(content_item)
+        !Location.find_by(content_item: content_item) &&
```
kevindew: Could we use an exists query here to avoid the object instantiation?

app/commands/v2/publish.rb
```diff
@@ -107,6 +107,11 @@ def set_first_published_at(content_item)
...
+          ContentItem::EMPTY_BASE_PATH_FORMATS.include?(content_item.schema_name || content_item.format)
```
kevindew: Could be less verbose to have a `requires_path?` method on content_item or in this class?

tijmenb:
Can you elaborate on the backstory about this? It seems weird that a format has an _optional_ base path. This could perhaps lead to confusion. Are whitehall-contacts and contacts-admin-contacts actually different things?

app/commands/v2/put_content.rb
```diff
@@ -216,6 +217,10 @@ def base_path
         payload[:base_path]
       end
 
+      def content_with_base_path?
+        payload.has_key?(:base_path) || base_path_required?
+      end
+
       def base_path_required?
         !ContentItem::EMPTY_BASE_PATH_FORMATS.include?(payload[:format] || payload[:schema_name])
```
kevindew: Though not part of this diff but in this area it could be good to switch the ordering of this `||` to indicate format is the fallback rather than primary value

app/commands/v2/put_content.rb
```diff
@@ -32,7 +32,8 @@ def call
     private
 
       def fill_out_new_content_item(content_item)
-        clear_draft_items_of_same_locale_and_base_path(content_item, locale, base_path) if base_path_required?
+        clear_draft_items_of_same_locale_and_base_path(
+          content_item, locale, base_path) if content_with_base_path?
```
kevindew: Since this method call is no longer on one line it might be more readable to have an if block

steventux:
@tijmenb sure, we needed to be able to send whitehall contacts to the publishing-api as a precursor to upcoming govspeak work. Content schemas were amended to accommodate these pathless contacts as part of the existing `contact` format. The formats are very similar and it seemed reasonable to amend an existing format. `base_path` was no longer required for contacts but is still an defined property. Contacts admin still relies on sending contacts with a path.
TL;DR we either have two very similar formats for contacts or we accept base_path as an optional property for pathless formats. We opted for the latter, this PR is a bugfix to accept the optional `base_path`.

kevindew:
LGTM 👍

PR #430, opened 2016-07-21
Stop skipping uniqueness validation for "superseded" content items

cbaines:
This is unnecessary, as there can only be one superseded content item per user
facing version.


danielroseman:
:+1:, needs rebase though.

PR #431, opened 2016-07-21
Represent

dougdroper:
Sleep every 10000 items so that sidekiq can process the queues and
memory can be kept under control

Increase concurrency level to default (25 threads)

### DIFF
--- a/app/commands/v2/represent_downstream.rb
+++ b/app/commands/v2/represent_downstream.rb
@@ -5,14 +5,17 @@ module Commands
         self.to_s
       end
 
-      def call(scope)
+      def call(scope, draft = false)
         filter = ContentItemFilter.new(scope: scope)
 
-        items_for_draft_store(filter).pluck(:id, :content_id).each do |(content_item_id, content_id)|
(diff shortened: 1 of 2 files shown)


app/commands/v2/represent_downstream.rb
```diff
@@ -5,14 +5,17 @@ def self.name
...
-          send_to_content_store(content_item_id, content_id, Adapters::DraftContentStore)
+        if draft
+          items_for_draft_store(filter).pluck(:id, :content_id).each do |(content_item_id, content_id)|
+            send_to_content_store(content_item_id, content_id, Adapters::DraftContentStore)
+          end
         end
 
-        items_for_live_store(filter).pluck(:id, :content_id).each do |(content_item_id, content_id)|
+        items_for_live_store(filter).pluck(:id, :content_id).each_with_index do |(content_item_id, content_id), index|
+          sleep 60 if (index + 1) % 10000 == 0
```
kevindew: V. Minor but the number could be slightly easier to digest in 10_000 form. Of course could be slightly worse if we formatting large numbers like that is uncommon in our codebases?

PR #432, opened 2016-07-21
Refactor documentation of API

kevindew:
Continuing the work of @cbaines we have a refactored document that lists all the API endpoints with their purpose, their inputs and the state changes they produce. 

There's bound to be a few things in here that can be phrased better and/or may be slightly incorrect or inconsistent. Any pointers appreciated.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -3,7 +3,7 @@
 This is the primary interface from publishing apps to the publishing pipeline. Applications PUT items as JSON conforming to a schema specified in govuk-content-schemas.
 Content paths are arbitrated internally by the Publishing API, the content is then forwarded to the live and draft content stores, and placed on the message queue for other apps (eg email-alert-service) to consume.
 
-## Index
+## Endpoint index
 - [`PUT /v2/content/:content_id`](#put-v2contentcontent_id)
 - [`POST /v2/content/:content_id/publish`](#post-v2contentcontent_idpublish)
 - [`POST /v2/content/:content_id/unpublish`](#post-v2contentcontent_idunpublish)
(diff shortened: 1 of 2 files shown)


doc/api.md
```diff
@@ -27,332 +31,406 @@ question.
...
+### Path parameters
+- [`content_id`](model.md#content_id)
+  - Identifies the content item to unpublish.
+
+### JSON attributes
+- `allow_draft` *(optional)*
+  - Boolean value, cannot be `true` if `discard_drafts` is also true.
+  - Specifies that if a draft content item is present it will be transitioned to
+  "unpublished" rather than the published version of this content item, which
+  itself will transition to "superseded".
```
cbaines: I don't think this is right. I think the important thing to get across here is that passing `allow_draft` as `true` means that it will only unpublish drafts.

I'm also not sure what is meant by the bit about published and superseded.

doc/api.md
```diff
@@ -1,9 +1,13 @@
 # Publishing API's API
 
-This is the primary interface from publishing apps to the publishing pipeline. Applications PUT items as JSON conforming to a schema specified in govuk-content-schemas.
-Content paths are arbitrated internally by the Publishing API, the content is then forwarded to the live and draft content stores, and placed on the message queue for other apps (eg email-alert-service) to consume.
-
-## Index
+This is the primary interface from publishing apps to the publishing pipeline.
+Applications PUT items as JSON conforming to a schema specified in
+[govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas).
+Content paths are arbitrated internally by the Publishing API, the content is
```
cbaines: It may be better to use `base_path`, or use Location (as in content item location) here?

doc/api.md
kevindew: Ah yes I misread the code. Good catch. If there's a published content item that exists the state of that will change to superseded. I'll have a look at rephrasing it.

doc/api.md
kevindew: Hmm I feel like that might veer this off into needing a longer introduction whilst it's good to keep it succinct. I think the use of "paths" does throw this off a little. Might be good with just "Content is arbitrated..."

cbaines: I've made a change 30c1136

PR #433, opened 2016-07-21
Improve 404 messages for GET /v2/content/:content-id

kevindew:
When a user tries to access a content item which exists but the
requested version or locale is not available they will now receive a 404
message that states that the version and/or locale of that content item
is not found. Previously all 404s returned the same message that implied
there was not a content item for the content_id when it could be a
version or locale not found.

### DIFF
--- a/app/queries/get_content.rb
+++ b/app/queries/get_content.rb
@@ -1,10 +1,10 @@
 module Queries
   module GetContent
     def self.call(content_id, locale = nil, version: nil)
-      locale ||= ContentItem::DEFAULT_LOCALE
+      locale_to_use = locale || ContentItem::DEFAULT_LOCALE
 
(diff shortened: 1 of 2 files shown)


tijmenb:
Do we know what could cause a version or locale to not exist? This seems an error internal to the publishing-api. Even if the user now knows what caused the error, they won't be able to fix it via the API.

kevindew:
@tijmenb Sure, they're not found for
a locale: when a user hasn't created one for that locale
a version: when the user facing version hasn't reached the number requested

So both are items controlled by the user. Although to make it easier to fix for a user though it could be a future scope to return which locales/version ranges are available.

tijmenb:
Yeah you're right, I was thinking of something else completely. Never mind me!

PR #434, opened 2016-07-21
Content items don't need a location any more

MatMoore:
Some formats, like contacts, don't always have an associated page on GOV.UK.
Support this in grouped-content-and-links.


cbaines:
:+1: thanks Mat :)

PR #437, opened 2016-07-26
Refactor uniqueness validator

kevindew:
The ContentItemUniqueness validator has been identified as one of our performance bottlenecks. We've found we can improve performance modestly by refactoring this to use less queries and of those be specific ones that don't instantiate ActiveRecord models.

Against the `bench/publish.rb` script I get the following results on my machine with this branch:
```
vagrant publishing-api git:(refactor_uniqueness_validator) ruby bench/publish.rb
Creating drafts...
Publishing...
  3.110000   0.560000   3.670000 (  5.347455)
3003 SQL queries
Creating new drafts...
Superseding...
  3.140000   0.680000   3.820000 (  5.487163)
3300 SQL queries
```
On the master branch (@5e7597f) I get: 
```
vagrant publishing-api git:(master) ruby bench/publish.rb
Creating drafts...
Publishing...
  3.140000   0.690000   3.830000 (  5.404747)
3203 SQL queries
Creating new drafts...
Superseding...
  2.840000   1.440000   4.280000 (  5.911925)
3700 SQL queries
```

### DIFF
--- a/app/queries/content_item_uniqueness.rb
+++ b/app/queries/content_item_uniqueness.rb
@@ -1,0 +1,57 @@
+module Queries
+  module ContentItemUniqueness
+    extend ArelHelpers
+    extend self
+
+    def unique_fields_for_content_item(content_item)
+      scope = unique_fields_scope
+      scope
+        .where(table(:content_items)[:id].eq(content_item.id))
+        .take(1)
+      get_rows(scope).first.try(:symbolize_keys)
+    end
+
+    def first_non_unique_item(content_item, base_path:, locale:, state:, user_facing_version:)
+      scope = unique_fields_scope
+      scope
+        .where(table(:content_items)[:id].not_eq(content_item.id))
+        .where(table(:states)[:name].eq(state))
+        .where(table(:translations)[:locale].eq(locale))
+        .where(table(:locations)[:base_path].eq(base_path))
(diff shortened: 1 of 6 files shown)


app/queries/content_item_uniqueness.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def first_non_unique_item(content_item, base_path:, locale:, state:, user_facing_version:)
+      scope = unique_fields_scope
+      scope
+        .where(table(:content_items)[:id].not_eq(content_item.id))
+        .where(table(:states)[:name].eq(state))
+        .where(table(:translations)[:locale].eq(locale))
+        .where(table(:locations)[:base_path].eq(base_path))
+        .where(table(:user_facing_versions)[:number].eq(user_facing_version))
+        .take(1)
```
cbaines: Could the API be simplified if this used `exists?` rather than returning a unused content item?

kevindew: the results are used in the error message at: https://github.com/alphagov/publishing-api/blob/bb069e68f21d8b47b0010ea8312e4924e26ad5d8/app/validators/content_item_uniqueness_validator.rb#L33

cbaines: Ah, yes, I missed the `non_unique` at the end.

PR #438, opened 2016-07-27
Remove race condition when discarding a draft where a live item exists

cbaines:
There was a possibility of a race condition with the discard draft command, as
if the Sidekiq job to discard the draft gets run after the job to send the live
item, the content store could end up with nothing at that base path, rather
than the live content item as intended.

This can be avoided by not attempting to delete the draft from the draft
content store, if the live item is being sent.

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -13,8 +13,14 @@ module Commands
         increment_live_lock_version if live
 
         after_transaction_commit do
-          delete_draft_from_draft_content_store(draft_path)
-          send_live_to_draft_content_store(live)
+          if live
+            send_live_to_draft_content_store(live)
+
+            live_path = Location.find_by!(content_item: live).base_path
+            delete_draft_from_draft_content_store(draft_path) if live_path != draft_path
+          else
+            delete_draft_from_draft_content_store(draft_path)
+          end
         end
 
         Success.new(content_id: content_id)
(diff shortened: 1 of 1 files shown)


app/commands/v2/discard_draft.rb
```diff
@@ -13,8 +13,18 @@ def call
         increment_live_lock_version if live
 
         after_transaction_commit do
-          delete_draft_from_draft_content_store(draft_path)
-          send_live_to_draft_content_store(live)
+          if live
+            live_path = Location.find_by!(content_item: live).base_path
+
+            if live_path == draft_path
```
gpeng: Could we simplify this to make it a bit easier to follow?

Maybe

```
send_live_to_draft_content_store(live)
delete_draft_from_draft_content_store(draft_path) if live_path != draft_path
```
or even extract `live_path == draft_path` into a method named to describe what that actually means?

cbaines: I tried to optimise for making it clear what actions are taken in each case. I tried a few different variations, but the above it the best I could come up with.

In particular, the suggestion you give above looks like it always sends the "live" version to the draft content store, which is not always the case.

gpeng: Both branches of that `if` send the live version. I was suggesting...
```
if live
  live_path = Location.find_by!(content_item: live).base_path
  send_live_to_draft_content_store(live)
  delete_draft_from_draft_content_store(draft_path) if live_path != draft_path
else
  delete_draft_from_draft_content_store(draft_path)
end
```

I just find nested `if`s  make you have to pause and jump back a bit when you're reading the code so I try to avoid them.

cbaines: Ok, I've reworked the commit now b4cf549c69e40b23e2dd5987906324644e14b360

PR #439, opened 2016-07-27
[DO NOT MERGE] Downstream publish worker

kevindew:
This is the first of a selection of classes to replace the PresentedContentStoreWorker with workers that are more specialised and thus require less configuration and their use is more apparent from their name.

The impetus for this work was to increase the speed of publishing by pushing the message queue into an asynchronous task, this however meant repurposing the worker.

This has different queue names for sidekiq and I am not sure if we need to set them up in advance of a merge/deployment. We would also have to maintain the PresentedContentStoreWorker until the queue for that is clear.

The performance difference with `ruby bench/publish.rb` is as such:
**This branch:**
```
Creating drafts...
Publishing...
  2.610000   0.400000   3.010000 (  3.883270)
2301 SQL queries
Creating new drafts...
Superseding...
  3.050000   0.320000   3.370000 (  4.306301)
2800 SQL queries
````
**Master branch:**
```
Creating drafts...
Publishing...
  3.460000   0.700000   4.160000 (  6.069500)
3203 SQL queries
Creating new drafts...
Superseding...
  4.200000   0.560000   4.760000 (  6.731551)
3700 SQL queries
```

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -151,22 +151,14 @@ module Commands
       def send_downstream(content_item, update_type)
         return unless downstream
 
-        queue_payload = Presenters::MessageQueuePresenter.present(
-          content_item,
-          state_fallback_order: [:published],
-          update_type: update_type
-        )
+        queue = update_type == 'republish' ? DownstreamPublishWorker::LOW_QUEUE : DownstreamPublishWorker::HIGH_QUEUE
 
-        PublishingAPI.service(:queue_publisher).send_message(queue_payload)
-
-        return if pathless?(content_item)
-
-        queue = update_type == 'republish' ? PresentedContentStoreWorker::LOW_QUEUE : PresentedContentStoreWorker::HIGH_QUEUE
-
-        PresentedContentStoreWorker.perform_async_in_queue(
+        DownstreamPublishWorker.perform_async_in_queue(
           queue,
-          content_store: Adapters::ContentStore,
-          payload: { content_item_id: content_item.id, payload_version: event.id },
+          content_item_id: content_item.id,
+          send_to_content_store: !pathless?(content_item),
+          message_queue_update_type: update_type,
+          payload_version: event.id,
         )
       end
     end
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -156,7 +156,6 @@ module Commands
         DownstreamPublishWorker.perform_async_in_queue(
           queue,
           content_item_id: content_item.id,
-          send_to_content_store: !pathless?(content_item),
           message_queue_update_type: update_type,
           payload_version: event.id,
         )
--- a/app/workers/downstream_publish_worker.rb
+++ b/app/workers/downstream_publish_worker.rb
@@ -44,8 +44,6 @@ private
     payload = presented_content_store_payload
     base_path = payload.fetch(:base_path)
     live_content_store.put_content_item(base_path, payload)
-  rescue => e
-    handle_content_store_error(e)
   end
 
   def broadcast_to_message_queue
@@ -55,18 +53,6 @@ private
 
   def live_content_store
     Adapters::ContentStore
-  end
-
-  def handle_content_store_error(error)
-    if !error.is_a?(CommandError) || error.code >= 500
-      raise error
-    else
-      # @TODO
-      # I'm not sure this is the correct explanation - feels like it was written
-      # for 409s which are swallowed
-      explanation = "The message is a duplicate and does not need to be retried"
-      Airbrake.notify_or_ignore(error, parameters: { explanation: explanation })
-    end
   end
 
   def web_content_item
--- a/app/workers/downstream_publish_worker.rb
+++ b/app/workers/downstream_publish_worker.rb
@@ -1,5 +1,5 @@
 class DownstreamPublishWorker
-  attr_reader :content_item_id, :send_to_content_store, :payload_version, :message_queue_update_type
+  attr_reader :content_item_id, :send_to_content_store, :payload_version, :message_queue_update_type, :update_dependencies
   include Sidekiq::Worker
   include PerformAsyncInQueue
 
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: spec/commands/v2/publish_spec.rb [hunk 1]
```diff
@@ -28,8 +28,8 @@
       stub_request(:put, %r{.*content-store.*/content/.*})
 
       allow(DependencyResolutionWorker).to receive(:perform_async)
-      allow(Presenters::ContentStorePresenter).to receive(:present)
-        .and_return(expected_content_store_payload)
+      #allow(Presenters::ContentStorePresenter).to receive(:present)
+      #  .and_return(expected_content_store_payload)
```
cbaines: Just commenting to point this out

kevindew: Thanks! forgot to clear that

app/workers/downstream_publish_worker.rb
```diff
@@ -0,0 +1,93 @@
...
+  sidekiq_options queue: HIGH_QUEUE
+
+  def perform(args = {})
+    assign_attributes(args.symbolize_keys)
+
+    unless web_content_item
+      raise CommandError.new(
+        code: 404,
+        message: "The content item for id: #{content_item_id} was not found",
+      )
```
cbaines: I don't think CommandError is appropriate here, as this is not a command.

app/workers/downstream_publish_worker.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    unless content_item_state?(:published)
+      raise CommandError.new(
+        code: 500,
+        message: "Will not downstream publish a content item that isn't published",
+      )
```
cbaines: I think in this case, the worker should just exit immediately, without raising any error, as the state not being published is a sign that its no longer appropriate to continue, and does not definitely mean that something has gone wrong.

app/workers/downstream_publish_worker.rb
```diff
@@ -0,0 +1,93 @@
...
+  def should_send_to_content_store?
+    web_content_item.base_path != nil
+  end
+
+  def web_content_item
+    @web_content_item ||= Queries::GetWebContentItems.(content_item_id).first
+  end
+
+  def content_item_state?(allowed_state)
+    State.where(content_item_id: content_item_id).pluck(:name) == [allowed_state.to_s]
```
cbaines: I think this should work off of the WebContentItem, otherwise you have to deal with the reality that the value here, and the value given by web_content_item might differ.

app/workers/downstream_publish_worker.rb
```diff
@@ -0,0 +1,93 @@
...
+  def assign_attributes(attributes)
+    @content_item_id = attributes.fetch(:content_item_id)
+    @payload_version = attributes.fetch(:payload_version)
+    @message_queue_update_type = attributes.fetch(:message_queue_update_type)
+    @update_dependencies = attributes.fetch(:update_dependencies, true)
+  end
+
+  def send_to_live_content_store
+    payload = presented_content_store_payload
+    base_path = payload.fetch(:base_path)
```
cbaines: A really minor one, but I think the base_path here should be from the web_content_item, as this simplifies how the data is flowing through the worker, and protects against weird effects that would occur if the Presenter modified the base_path somehow.

app/commands/v2/publish.rb
```diff
@@ -151,22 +151,14 @@ def lookup_previous_item(content_item)
...
 
-        PresentedContentStoreWorker.perform_async_in_queue(
+        DownstreamPublishWorker.perform_async_in_queue(
           queue,
-          content_store: Adapters::ContentStore,
-          payload: { content_item_id: content_item.id, payload_version: event.id },
+          content_item_id: content_item.id,
+          send_to_content_store: !pathless?(content_item),
+          message_queue_update_type: update_type,
+          payload_version: event.id,
```
cbaines: I'm enjoying the effect this change has had here in terms of the simplicity of the code :)

config/sidekiq.yml
```diff
@@ -3,8 +3,10 @@
 :concurrency:  2
 :logfile: ./log/sidekiq.json.log
 :queues:
-  - content_store_high
-  - content_store_low
+  - content_store_high # Deprecated, to be removed once jobs completed
+  - content_store_low # Deprecated, to be removed once jobs completed
```
cbaines: Not all the uses of the PresentedContentStoreWorker have been removed, so I would hold off in marking the queues as deprecated.

cbaines:
I think this is really good and should be merged, modulo addressing the comments I have added.

I'm not going to merge it though, as I was too involved in the changes.

config/sidekiq.yml
kevindew: Good point, will be once I get them all in - but shouldn't be in this.

app/workers/downstream_publish_worker.rb
kevindew: sure, yeah I actually switched to that on one of my other workers. I'll modify this one accordingly.

app/workers/downstream_publish_worker.rb
kevindew: Cool - I've got a slight refactor going with WebContentItem that will make use of this easier and less verbose.

app/workers/downstream_publish_worker.rb
kevindew: Valid point - using it here is consistent with how existing workers use it: https://github.com/alphagov/publishing-api/blob/master/app/workers/presented_content_store_worker.rb#L33 so may need to get some further input on this on the next on whether or how this should error.

cbaines:
The new changes look good, although I can't see why this is necessary da1319238b8c811a83bd2bd85c55caf4f48b4ce0, I guess you are using it in classes in other branches?

kevindew:
@cbaines Yup - that's correct. I'm going to remark this pull as DNM while I get the other bits together.

kevindew:
Superseded by https://github.com/alphagov/publishing-api/pull/444

PR #440, opened 2016-07-28
Warn about the content index endpoint

tijmenb:
This endpoint will only return content that is "owned" by the app making the request:

https://github.com/alphagov/publishing-api/blob/bb069e68f21d8b47b0010ea8312e4924e26ad5d8/app/controllers/v2/content_items_controller.rb#L68

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -266,6 +266,10 @@ the draft content store with the published item, if one exists. Uses
 
  [Request/Response detail](https://pact-broker.dev.publishing.service.gov.uk/pacts/provider/Publishing%20API/consumer/GDS%20API%20Adapters/latest#a_get_entries_request_given_a_content_item_exists_in_multiple_locales_with_content_id:_bed722e6-db68-43e5-9079-063f623335a7)
 
+**This endpoint will only return content for which the value of `publishing_app`
+matches the name of the application making the request, unless that app's API
+key has `view_all` permissions.** 
+
 Retrieves a paginated list of content items for the provided query string
 parameters. If content items exists in both a published and a draft state, the
 draft is returned.


kevindew:
Looks good - Thanks 👍

I wonder if we need a matching restriction on show endpoint?

doc/api.md
```diff
@@ -266,6 +266,8 @@ the draft content store with the published item, if one exists. Uses
 
  [Request/Response detail](https://pact-broker.dev.publishing.service.gov.uk/pacts/provider/Publishing%20API/consumer/GDS%20API%20Adapters/latest#a_get_entries_request_given_a_content_item_exists_in_multiple_locales_with_content_id:_bed722e6-db68-43e5-9079-063f623335a7)
 
+⚠️ **This endpoint will only return content by the authenticating app, unless that app's API key has `view_all` permissions.**
```
cbaines: Rather than saying "by the", I think it would be better to use the same terminology as here [1], which would mean saying something like:
```This endpoint will only return content for which the value of `publishing_app` matches the name of the application making the request, unless that app's API key has `view_all` permissions```

1: https://github.com/alphagov/publishing-api/blob/master/doc/model.md#publishing_app

tijmenb: Cool, updated.

tijmenb:
> I wonder if we need a matching restriction on show endpoint?

I'm not really convinced of the need of this restriction. It's kind of weird that you can create a content item that you're not allowed to see. I'd rather have `publishing_app` be explicitly passed in so you can see from the code what is return values are expected (with this system you have to check `permissions` in publishing-api database to know what a certain request will do). But that's a story for another day.

kevindew:
ok cool. We'll keep it out of here :)

PR #442, opened 2016-08-01
Fix hmrc contacts with no Location

steventux:
https://govuk.zendesk.com/agent/tickets/1371252

The publishing-api previously had a bug which prevented contacts which have a base_path
from populating a corresponding Location record. This has now been updated so that
contacts and other formats can optionally specify a path and the relevant Location
object will be created. There are around 21 records without a Location object
which need updating.

### DIFF
--- a/db/migrate/20160721120915_clean_up_pathless_contacts.rb
+++ b/db/migrate/20160721120915_clean_up_pathless_contacts.rb
@@ -1,0 +1,14 @@
+class CleanUpPathlessContacts < ActiveRecord::Migration
+  def change
+    # There are approx 800 contact ContentItems.
+    ContentItem.where(schema_name: "contact", publishing_app: "contacts").find_each do |item|
+      path = item.routes.first[:path]
+
+      # Approx 25 records have no Location supporting object.
+      unless Location.find_by(content_item: item)
+        Location.create!(base_path: path, content_item: item)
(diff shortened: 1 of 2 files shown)


cbaines:
Given that it is a small and unchanging set of items that need fixing, I would prefer the migration to just contain the ids for the items which need fixing, rather than the query to select them (as its then possible to see what the migration changed). But, I'm fine if you want to stick with the current implementation.

db/migrate/20160721120915_clean_up_pathless_contacts.rb
```diff
@@ -0,0 +1,14 @@
+class CleanUpPathlessContacts < ActiveRecord::Migration
+  def change
+    # There are approx 800 contact ContentItems.
+    ContentItem.find_each(schema_name: "contact", publishing_app: "contacts") do |item|
```
kevindew: Did this work locally? this doesn't seem the correct arguments for [`find_each`](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) looks like it should be `ContentItem.where(schema_name: "contact", publishing_app: "contacts").find_each` ?

kevindew:
@cbaines a plus if it contained the ids is we could have a up and down implementation. Currently it'll perform same thing being rollbacked as migrated.

PR #443, opened 2016-08-01
Fix HMRC contacts with no Location

steventux:
The publishing-api previously had a bug which prevented contacts which have a base_path
from populating a corresponding Location record. This has now been updated so that
contacts and other formats can optionally specify a path and the relevant Location
object will be created. There are around 21 records without a Location object
which need updating.


kevindew:
Looks good 👍

PR #444, opened 2016-08-02
Downstream Workers

kevindew:
Continuing from the work done in: https://github.com/alphagov/publishing-api/pull/439 I have refactored all the uses of the `PresentedContentStoreWorker` to use either `DownstreamDraftWorker`, `DownstreamLiveWorker` or `DownstreamDiscardDraftWorker`. 

The advantages of having these classes is 
1. Simplified API to initialise workers
2. Workers more closely tied to their responsibility offering clearer opportunities to validate asynchronous actions
3.  Opportunities to re-use expensive downstream operations to improve performance.

In refactoring these I was able to address some bugs that we have in the Publishing API:

- We now send unpublishings to both the draft and live content stores.
- Replace a discarded draft with an unpublished item (if one exists)
- The content store action of discarding a draft is less vulnerable to race conditions by nesting all uses of it and checking for content conflicts.


app/workers/dependency_resolution_worker.rb
```diff
@@ -29,24 +29,43 @@ def content_item_dependees
...
+
+  def draft?
+    content_store == Adapters::DraftContentStore
+  end
+
+  def present_content_store(latest_content_item)
     PresentedContentStoreWorker.perform_async_in_queue(
       PresentedContentStoreWorker::LOW_QUEUE,
-      content_store: content_store,
+      content_store: Adapters::DraftContentStore,
```
boffbowsh: Why is this now only draft content store?

kevindew: It's because we have a different method to go to live content store. Though this has alerted me that I need to update the worker used here - so change incoming.

kevindew: Ah actually had already been done in this commit: https://github.com/alphagov/publishing-api/commit/d198f7eec3634141b9023f35564014864f831df0 

Reason for change was initially we replaced just publishing with a new worker, which then expanded into the various downstream workers.

config/sidekiq.yml
```diff
@@ -3,8 +3,8 @@
 :concurrency:  2
 :logfile: ./log/sidekiq.json.log
 :queues:
-  - content_store_high # Deprecated, to be removed once jobs completed
-  - content_store_low # Deprecated, to be removed once jobs completed
+  - content_store_high
```
boffbowsh: What's the reason for keeping separate queues for these now? I think we should move all content store/downstream jobs to `downstream_*`

kevindew: I think you're on an outdated diff there (from when this was closer to being merged gradually: https://github.com/alphagov/publishing-api/pull/439). This one here should be clearer: https://github.com/alphagov/publishing-api/pull/444/commits/f03cba17d4f360ef3f90875a8c4d105225277581

It would be keeping the content_store_* ones until the sidekiq queue for those is empty.

boffbowsh:
I like this a lot, great refactoring 💯 . I'm unsure about "mediator" as a term here, but I also can't quickly think of an alternative. I won't merge as this is a large change so I think it's worth some additional eyes.

boffbowsh:
Also double 💯  for clear commit messages

kevindew:
Thanks @boffbowsh 👍 - I struggled a bit about the mediator class, seemed the closest design pattern to employ. 

I'm working with @cbaines and @dougdroper to try reduce complexity in that class so might well get renamed/repurposed.

spec/downstream_payload_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      it "returns true" do
+        expect(downstream_payload.unpublished?).to be true
+      end
+    end
+
+    context "published content item" do
+      let(:web_content_item) { create_web_content_item(:live_content_item) }
+
+      it "returns true" do
```
dougdroper: does it?

kevindew: Doh! Good catch

dougdroper:
Lots of things going on here, but superficially looks good, have you pushed it to integration to double check that everything works as expected?

kevindew:
@dougdroper all looks good on my tests

PR #446, opened 2016-08-04
Simplify state name presentation (live to published)

dougdroper:
ContentItemPresenter was renaming published to live for no apparent reason, so now it is
using a consistent ubiquitous language


tuzz:
Don't merge yet please, we'll need to integrate against this in Specialist Publisher Rebuild. I'll add a ticket.

PR #447, opened 2016-08-08
Add the reverse names for documents and working_groups

dougdroper:
Document collections have links to documents. (Whitehall) Documents
need to know which collections they belong to. See 7a5b6e9 in
content_store.

To render a working_group, you need the list of policies which link to that
working_group. See 5292b6e in content_store


kevindew:
👍

PR #448, opened 2016-08-08
Don't validate uniqueness of pathless content

steventux:
https://trello.com/c/b6ZrXFTX/778-contacts-validation-produces-false-positives

Content items without a `base_path` will match anything with a similar locale, version and state.
This will lead to false positives as one of the four uniqueness identifiers is missing.
Instead, we don't attempt to validate pathless items, the existing version checking in the `V2::PutContent` command protects duplicating pathless content.

### DIFF
--- a/app/validators/content_item_uniqueness_validator.rb
+++ b/app/validators/content_item_uniqueness_validator.rb
@@ -12,8 +12,7 @@ class ContentItemUniquenessValidator < ActiveModel::Validator
     state = record.is_a?(State) ? record.name : unique_fields[:state]
     user_facing_version = record.is_a?(UserFacingVersion) ? record.number : unique_fields[:user_facing_version]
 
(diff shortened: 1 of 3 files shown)


app/validators/content_item_uniqueness_validator.rb
```diff
@@ -6,14 +6,14 @@ def validate(record)
     unique_fields = Queries::ContentItemUniqueness.unique_fields_for_content_item(content_item)
 
     return unless unique_fields
+    return unless content_item.requires_base_path?
```
kevindew: I don't think we want this line here, as I expect this will mean we don't catch instances where a contact has a base_path and it collides with a different contact content_item.

steventux:
@kevindew thanks for review, updated.

kevindew:
Looks good 👍

PR #450, opened 2016-08-09
[WIP][Feedback Needed] - Add endpoint to allow changing the ownership of content items

carvil:
We have recently migrated some taxonomy-related tools from `collections-publisher` into `content-tagger`. Since some of the Taxons have been created on `collections-publisher`, we now need to update their ownership in order to be able to update those content items from `content-tagger`.

As things stand, we cannot update the `publishing_app` attribute of content items via the API. There is a [validator](https://github.com/alphagov/publishing-api/blob/2481abe74fd0aa3b7fb948eade84bb60c74c9bd9/app/validators/publishing_app_validator.rb) that prevents us from doing that.

This could be done via a rake task (I wrote that in a separate branch [here](https://github.com/alphagov/publishing-api/compare/change-ownership-of-taxons?expand=1) ).

However, my suggestion is to add an endpoint that allows us to change the ownership of content items without adding more logic to the validator, leaving the `put_content` flow intact.

Would this be useful and/or does it make sense?

If that's not the case, I will close this PR and open the one with the rake task.

Note: this was a very rough version to see if it would work, it's lacking tests and robustness.

### DIFF
--- a/app/commands/v2/change_ownership.rb
+++ b/app/commands/v2/change_ownership.rb
@@ -1,0 +1,67 @@
+module Commands
+  module V2
+    class ChangeOwnership < BaseCommand
+      def call
+        if publishing_app.blank?
+          raise_command_error(422, "publishing_app is required", fields: {
+            publishing_app: ["is required"]
+          })
+        end
+
+        content_item = find_content_item
+
+        raise_command_error(
+          404,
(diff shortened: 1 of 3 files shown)


carvil:
Closing in favour of the rake task as discussed on slack.

PR #451, opened 2016-08-09
Rake task to change ownership of Taxon content items

carvil:
Currently, most taxons have been created via `collections-publisher`. We now
moved the taxonomy tools to `content-tagger`, therefore we should change the
ownership of the Taxon content items so we can edit them on `content-tagger`.

We can delete this rake task after it runs in production.


Davidslv:
👍 Looks good to me

rubenarakelyan:
👍

PR #455, opened 2016-08-11
Periodically check for the presence of duplicate content items

dougdroper:
Verify frequency of duplicate content items that are created and log to
errbit if any exist.

Duplicate being defined as a content_item that has the same state name,
locale, user facing version number and base path, and state name not in
superseded or unpublished

### DIFF
--- a/lib/data_hygiene/duplicate_content_item.rb
+++ b/lib/data_hygiene/duplicate_content_item.rb
@@ -1,0 +1,47 @@
+module DataHygiene
+  class DuplicateContentItem
+    def check
+      log if duplicates.any?
+    end
+
+  private
+
+    def log
+      Airbrake.notify_or_ignore(
+        DuplicateContentItemError.new("Duplicate content items"),
+        parameters: parameters
+      )
+    end
+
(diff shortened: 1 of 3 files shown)


lib/data_hygiene/duplicate_content_item.rb
```diff
@@ -0,0 +1,50 @@
...
+        INNER
+        JOIN translations
+        ON content_items.id = translations.content_item_id
+        INNER
+        JOIN locations
+        ON content_items.id = locations.content_item_id
+        INNER
+        JOIN user_facing_versions
+        ON content_items.id = user_facing_versions.content_item_id
+        GROUP BY content_items.content_id, states.name, translations.locale,
```
cbaines: Having `content_id` in the group by will mean that this will not pick up issues where all of the 4 "uniqueness" criteria are violated, but where the `content_id` is different. Is this intentional?

dougdroper: not particularly, blindly lifting from the original query.

PR #458, opened 2016-08-16
Refactor pathless

cbaines:
I started looking at fudging side effects in Publish in to Unpublish to make `allow_draft` less broken, but luckily that is no longer necessary :)

However, some of the refactoring I did on the way is worth finishing.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -48,6 +48,10 @@ class ContentItem < ActiveRecord::Base
     EMPTY_BASE_PATH_FORMATS.exclude?(document_type)
   end
 
+  def pathless?
+    !self.requires_base_path? && !Location.exists?(content_item: self)
+  end
+
 private
 
   def renderable_content?
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -32,7 +32,7 @@ module Commands
 
         State.supersede(previous_item) if previous_item
 
-        unless pathless?(content_item)
+        unless content_item.pathless?
           if previous_item
             previous_location = Location.find_by(content_item: previous_item)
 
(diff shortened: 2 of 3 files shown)


app/models/content_item.rb
```diff
@@ -45,7 +45,11 @@ class ContentItem < ActiveRecord::Base
   validates :details, well_formed_content_types: { must_include: "text/html" }
 
   def requires_base_path?
-    EMPTY_BASE_PATH_FORMATS.exclude?(document_type)
+    EMPTY_BASE_PATH_FORMATS.exclude?(document_type || format)
+  end
+
+  def pathless?
+    !self.requires_base_path? && !Location.exists?(content_item: self)
```
dougdroper: I am a bit concerned with slipping knowledge of paths/locations into content_item here. it seems more convenient, but it goes against the general theme that this was originally trying to achieve: https://github.com/alphagov/publishing-api/blob/master/doc/model.md#high-level-diagram

I think this would be a good place to be moved into the WebContentItem, or something similar. 

What do you think?

cbaines: One alternative that I see is that requires_base_path? and pathless? could be moved to Location (and take a content_item as an argument).

I really don't mind either way, both are internally consistent.

WebContentItem does not seem the right place, as I don't see any reason why that should be treated differently to ContentItem when it comes to a separation of concerns (then again, I don't really agree with this "theme" off limited visibility of the content item, I'm hoping this will sort itself out when the database structure improves as a whole).

app/models/content_item.rb
```diff
@@ -45,7 +45,7 @@ class ContentItem < ActiveRecord::Base
   validates :details, well_formed_content_types: { must_include: "text/html" }
 
   def requires_base_path?
-    EMPTY_BASE_PATH_FORMATS.exclude?(document_type)
+    EMPTY_BASE_PATH_FORMATS.exclude?(document_type || format)
```
boffbowsh: Every content item has a document type:
```
publishing_api_production=# select count(*) from content_items where document_type = '' or document_type is null;
 count
-------
     0
(1 row)
```

cbaines: In my head, I was not sure what the interaction with this model, and the data coming in to the system is. I've checked now, and the document_type is coped with in PutContent, so I've removed the commit adding this.

PR #459, opened 2016-08-16
Do not expand links that are non-renderable 

dougdroper:
Document types of redirect/gone should not be expanded since they are not
displayable on the frontend so no need to expand them here.

Prevents any page from potentially breaking for asking the link for one
of its attributes that are not there.

### DIFF
--- a/app/presenters/queries/expanded_link_set.rb
+++ b/app/presenters/queries/expanded_link_set.rb
@@ -39,9 +39,9 @@ module Presenters
             next_level = {}
           end
 
-          expanded_links.merge(
-            links: next_level,
-          )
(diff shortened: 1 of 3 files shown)


tijmenb:
I think the original behaviour was not to expand anything that wasn't renderable:

https://github.com/alphagov/content-store/blob/9303c9c14fa876ee8b478ae4a57b602fb861addb/app/queries/linked_items_query.rb
https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L5

Should we add `gone`?

dougdroper:
Yeah, was going to add in a separate PR, but makes sense to do it here too. 👍

PR #462, opened 2016-08-16
Update link to API documentation

davidbasalla:
The previous link was linking to a non-existent page. I assume the 'doc/api.md' page is the appropriate one to link to?


kevindew:
Yup. Thanks for that 👍

PR #463, opened 2016-08-16
Make 20160816141116 migration robust in absence of data

h-lame:
It's possible that the find_by call in this migration could fail so to
avoid breaking the migration in this situation we use

    where(...).destroy_all

instead which won't break if the data is already missing.

This is a follow up to #461 - I was having a discussion about this with @dougdroper but it was merged before I could push up this change.

### DIFF
--- a/db/migrate/20160816141116_remove_path_reservation_for_find_local_council.rb
+++ b/db/migrate/20160816141116_remove_path_reservation_for_find_local_council.rb
@@ -1,0 +1,11 @@
+class RemovePathReservationForFindLocalCouncil < ActiveRecord::Migration
+  def change
+    # Actually fully delete this path reservation.
+    # It was created to reserve a path for an artefact that was created in
+    # error on launch day and immediately archived without publishing (or
+    # even completing the creation of the associated edition in publisher).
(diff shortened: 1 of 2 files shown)


kevindew:
No worries, I'll revert the previous merge. It does make more sense this way since it's less fragile. @dougdroper do you want to handle merging this one just in case there's any more discussion?

kevindew:
Any chance you could rebase this?

db/migrate/20160816141116_remove_path_reservation_for_find_local_council.rb
```diff
@@ -6,6 +6,6 @@ def change
     # even completing the creation of the associated edition in publisher).
     # We want to reuse the route without leaving something lying around
     # that could accidently be used to alter that route from panopticon.
-    PathReservation.find_by(base_path: '/find-local-council').destroy
+    PathReservation.where(base_path: '/find-local-council').destroy_all
```
jennyd: Do we also want to check the owning app for this, so that we don't accidentally delete the new route somehow in the future?

h-lame:
Rebased and includes publishing_app filter on the where to avoid deleting something we don't expect.

PR #465, opened 2016-08-16
Do not add duplicate jobs to Sidekiq queues

dougdroper:
This uses Sidekiq unique jobs `until executing` to enforce uniqueness,
which means the job will only be scheduled into redis once per content
item based off its content_item_id and which store it is
going to (live/draft).

Any jobs added until the first one of the same arguments has been
unlocked will just be dropped.

For a content item that has 751 links, these are the Sidekiq::Stats

Before:
{:processed=>5269, :failed=>12,
:scheduled_size=>0, :retry_size=>0, :dead_size=>0, :processes_size=>1,
:default_queue_latency=>0, :workers_size=>0, :enqueued=>0}

After:
{:processed=>751, :failed=>0,
:scheduled_size=>0, :retry_size=>0, :dead_size=>0, :processes_size=>1,
:default_queue_latency=>0, :workers_size=>0, :enqueued=>0}

### DIFF
--- a/app/workers/downstream_live_worker.rb
+++ b/app/workers/downstream_live_worker.rb
@@ -1,3 +1,5 @@
+require 'sidekiq-unique-jobs'
+
 class DownstreamLiveWorker
   attr_reader :content_item_id, :web_content_item, :payload_version, :message_queue_update_type, :update_dependencies
 
@@ -5,7 +7,16 @@ class DownstreamLiveWorker
   include Sidekiq::Worker
   include PerformAsyncInQueue
 
-  sidekiq_options queue: HIGH_QUEUE
+  sidekiq_options queue: HIGH_QUEUE,
+                  unique: :until_executing,
+                  unique_args: :uniq_args
+
+  def self.uniq_args(args)
+    [
+      args.first.fetch("content_item_id"),
+      'live'
+    ]
+  end
 
   def perform(args = {})
     assign_attributes(args.symbolize_keys)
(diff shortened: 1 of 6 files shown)


kevindew:
Will we still get duplicates if two content items with similar links are edited in quick succession?

dougdroper:
I dropped the need for the payload_version since it was unnecessary

boffbowsh:
The order of the queues also need switching to take full advantage of this. `dependency_resolution` clears through quickly and generates `downstream_low` jobs. If the `downstream_low` jobs are always worked first, there is less opportunity to find duplicate jobs because they reach the `executing` stage too quickly. The order should be:
- downstream_high
- dependency_resolution
- downstream_low

boffbowsh:
Otherwise 🎉

dougdroper:
@boffbowsh great suggestion! that will help a lot, thanks. done

app/workers/downstream_live_worker.rb
```diff
@@ -1,11 +1,22 @@
...
+  sidekiq_options queue: HIGH_QUEUE,
+                  unique: :until_executing,
+                  unique_args: :uniq_args
+
+  def self.uniq_args(args)
+    [
+      args.first.fetch("content_item_id"),
+      'live'
+    ]
+  end
```
kevindew: Sorry I'm a bit late on this.

I assume this is the unique criteria for the worker so if there's another item with this content_item_id it doesn't get queued - ignore me if I'm wrong.

We should also have message_queue_update_type and update_dependencies in there. For example if we had an item queued as the result of a link update it wouldn't update dependencies and just send a update_type of links, however if it had a major update while sat in the queue this message wouldn't get sent out or dependencies updated.

boffbowsh: Excellent catch :+1:

PR #466, opened 2016-08-17
Replace uniqueness validator

kevindew:
This replaces the previous system for validating the uniqueness of a ContentItem with 3 different validations that check for different conflicts.

### Previous uniqueness validation and it's problems

Before a ContentItem was deemed unique if it's state, locale, base_path and user version were distinct. Which is conceptually nice as it's simple to explain but has a number or problems, so of the most significant of these are:

1. Multiple ContentItems can occupy the same base_path, either with the same or separate content_id's
2. There can be multiple draft or published instances of a ContentItem existing at the same time if the user version is different

###  New uniqueness constraints

This PR introduces 3 different invariants and associated validations. To explain these I will paste in the updated documentation - to hopefully catch any errors in explaining them:

> The ContentItem model is a simple model that houses content. It is surrounded by
> objects that determine:
> 
> - The locale of the content (Translation)
> - The base_path of the content (Location)
> - The workflow state of the content (State)
> - The user-facing version of the content (UserFacingVersion)
> 
> The uniqueness of a ContentItem is ensured by invariants based on these objects.
> 
> Only one item in either a draft or a live state can be registered at a base_path.
> Thus the invariants for a base_path are:
> 
> 1. A base_path for a draft ContentItem must be distinct from the base_path of
> any other draft ContentItems.
> 2. A base_path for a live ContentItem must be distinct from the base_path for
> any other live ContentItems
> 
> Nb
> - A draft ContentItem is a ContentItem with a state of "draft"
> - A live ContentItem is a ContentItem with a state of "published" or
> "unpublished", however it does not include those with an unpublishing type of
> "substitute".
> 
> There can be only one instance of a ContentItem for a content_id and a
> particular locale with a state of draft.
> 
> There can be only one instance of a ContentItem for a content_id and a
> particular locale with a state of published/unpublished.
> 
> There can be multiple instance of a ContentItem for a content_id, a particular
> locale and a state of superseded.
> 
> Only one instance of a ContentItem for a content_id and a particular locale can
> be at a user-facing version. Thus an invariant exists that no two ContentItems
> can have the same content id, the same user-facing version and the same locale.

These are done through: 
- `BasePathStateValidator` - Which validates there are no draft items occupying a base_path for a draft or any live items occupying a base path for a live item. (live being published or unpublished - excluding unpublished type substitute)
- `StateForLocaleValidator` - Validates that for ContentItems sharing a content_id and locale that there is at maximum one item in a draft state and one item in a published/unpublished state
- `VersionForLocaleValidator` - Validates that for ContentItems sharing a content_id and locale that there are not multiple items with the same user_facing_version number

### Effect on performance
Running the `ruby bench/publish.rb` on my machine on this branch yields:

```
Creating drafts...
Publishing...
  1.520000   0.270000   1.790000 (  2.489066)
2102 SQL queries
Creating new drafts...
Superseding...
  1.760000   0.350000   2.110000 (  2.842613)
2600 SQL queries
```

Whereas on master branch:

```
Creating drafts...
Publishing...
  1.500000   0.200000   1.700000 (  2.449326)
1901 SQL queries
Creating new drafts...
Superseding...
  1.680000   0.280000   1.960000 (  3.008492)
2400 SQL queries
```

### Number of items that fail validation

Based on my current database which is approximately one week old.

We have 1784 😢 ContentItems that will fail the BasePathStateValidator
```sql
SELECT SUM(collisions) 
FROM (
    SELECT locations.base_path, 
        CASE states.name WHEN 'draft'THEN 'draft' ELSE 'live' END AS content_store,
        COUNT(*) AS collisions
    FROM content_items
    INNER JOIN locations
        ON locations.content_item_id = content_items.id
    INNER JOIN states
        ON states.content_item_id = content_items.id
    LEFT JOIN unpublishings
        ON states.name = 'unpublished'
            AND unpublishings.content_item_id = content_items.id
    WHERE locations.base_path IS NOT NULL
        AND states.name IN ('draft', 'published', 'unpublished')
        AND (unpublishings.type IS NULL OR unpublishings.type != 'substitute')
    GROUP BY locations.base_path, content_store
    HAVING COUNT(*) > 1
) AS results
```

We have 8 🎉 ContentItems that will fail the StateForLocaleValidator
```sql
SELECT SUM(collisions) 
FROM (
    SELECT content_items.content_id,
        translations.locale,
        CASE states.name WHEN 'draft'THEN 'draft' ELSE 'live' END AS content_store,
        COUNT(*) AS collisions
    FROM content_items
    INNER JOIN translations
        ON translations.content_item_id = content_items.id
    INNER JOIN states
        ON states.content_item_id = content_items.id
    WHERE states.name IN ('draft', 'published', 'unpublished')
    GROUP BY content_items.content_id, translations.locale, content_store
    HAVING COUNT(*) > 1
) AS results
```

We have 134 ContentItems that would fail the VersionForLocaleValidator
```sql
SELECT SUM(collisions) 
FROM (
    SELECT content_items.content_id,
        translations.locale,
        user_facing_versions.number,
        COUNT(*) AS collisions
    FROM content_items
    INNER JOIN translations
        ON translations.content_item_id = content_items.id
    INNER JOIN user_facing_versions
        ON user_facing_versions.content_item_id = content_items.id
    GROUP BY content_items.content_id, translations.locale, user_facing_versions.number
    HAVING COUNT(*) > 1
) AS results
```

### Tests
The biggest casualty in the change of this is that many of the tests relied upon the existing uniqueness constraints to run, so many of them have had to have been fixed in a sometimes inelegant style. 

At the time of writing there's still a few to fix.


boffbowsh:
Can we use a different term than "invariant" for this please? It's not a common term that I'm aware of outside of degree-level Computer Science or Mathematics. Not all our developers have a degree (I don't), and I had to look up the definition:

> _noun_
> a function, quantity, or property which remains unchanged when a specified transformation is applied.

It feels like we're using it here as a synonym for "condition" or "rule". 

From the description:

> Thus the invariants for a base_path are:
> 1. A base_path for a draft ContentItem must be distinct from the base_path of any other draft ContentItems.
> 2. A base_path for a live ContentItem must be distinct from the base_path for any other live ContentItems.

These don't feel like functions, quantities or properties to me. If you replace "invariant" with "rule", I think the meaning is a lot clearer.

We should keep the terminology we use accessible to people of all backgrounds.

boffbowsh:
(I should note I'm still reviewing the rest of the PR, and so far it seems the intention is great and is very well-described otherwise 👍 )

app/queries/base_path_state.rb
```diff
@@ -0,0 +1,71 @@
...
+        )
+        .where(content_items_table[:id].not_eq(content_item_id))
+        .where(states_table[:name].eq("draft"))
+        .where(locations_table[:base_path].eq(base_path))
+        .order(content_items_table[:created_at].desc)
+        .take(1)
+      get_rows(scope).first.try(:symbolize_keys)
+    end
+
+    def self.conflict_with_live_content(content_item_id, base_path)
```
boffbowsh: This doesn't seem to match what's in the commit message:

> - A live content item fails validation if there is another live content
  item with the same content id

app/queries/base_path_state.rb
```diff
@@ -0,0 +1,71 @@
+module Queries
+  module BasePathState
+    extend ArelHelpers
+
+    def self.conflict_with_draft_content(content_item_id, base_path)
```
boffbowsh: Can we extract the common query parts from these two methods into some helper method or something?

kevindew: Yeah - I think they it can probably be just a single method that handles both cases. There was a more complex condition before with draft (checking if it clashed with live) but that didn't work out so I think I can simplify them.

Thanks for the review by the way 👍

app/validators/base_path_state_validator.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def check_live_conflict(record, content_item, base_path)
+    conflicts = Queries::BasePathState.conflict_with_live_content(
+      content_item.id, base_path
+    )
+    record.errors.add(:content_item, error_message(conflicts, "live")) if conflicts
+  end
+
+  def error_message(conflict, prefix = nil)
```
boffbowsh: You're always providing a prefix above, so you can remove the optional-ness and logic below

app/validators/base_path_state_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def content_item_base_path(record)
+    return record.base_path if record.is_a?(Location)
+    Location.where(content_item: record.content_item).pluck(:base_path).first
+  end
+
+  def content_item_live?(content_item, state)
+    return true if state == "published"
+    state == "unpublished" && Unpublishing.is_substitute?(content_item)
```
cbaines: I think this should be `state == "unpublished" && !Unpublishing.is_substitute?(content_item)` (notice the `!`), as you want to say that all types of unpublishing are "live", except "substitute".

cbaines:
I've had a quick glance at this, and its looking good, and I'll give it a deeper read when we are closer to merging it.

app/validators/base_path_state_validator.rb
kevindew: Good catch - Thanks

kevindew:
@cbaines and @boffbowsh all comments addressed now.

Now we've got the blocking items off I've removed the DNM, so good for final reviewing now

PR #470, opened 2016-08-18
Update taxonomy rake task to update path reservations

carvil:
We're moving a load of taxons over to the content-tagger app. This adds
to the existing (temporary) rake task to update publishing_app on the
path reservations as well as the content items.

### DIFF
--- a/lib/tasks/taxonomy.rake
+++ b/lib/tasks/taxonomy.rake
@@ -2,12 +2,24 @@
 namespace :taxonomy do
   desc 'Change ownership of taxons from collections-publisher to content-tagger'
   task change_ownership: :environment do
-    ContentItem.transaction do
-      taxons = ContentItem.where(document_type: 'taxon')
+    ActiveRecord::Base.transaction do
+      taxons = ContentItem.where(document_type: 'taxon', publishing_app: "collections-publisher")
       puts "Found #{taxons.count} taxons"
+
       taxons.each do |taxon|
         puts "Changing #{taxon.title} from #{taxon.publishing_app} to content-tagger"
         taxon.update!(publishing_app: 'content-tagger')
+      end
+
+      puts ""
+
+      path_reservations = PathReservation.where("base_path LIKE '/alpha-taxonomy%'").where(publishing_app: "collections-publisher")
+      if path_reservations.present?
+        puts "Found #{path_reservations.count} path reservations belonging to collections publisher"
+        path_reservations.each do |path_reservation|
+          puts "Updating #{path_reservation.base_path} to content-tagger"
+          path_reservation.update_column(publishing_app: "content-tagger")
+        end
       end
     end
   end


lib/tasks/taxonomy.rake
```diff
@@ -2,13 +2,25 @@
...
+    ActiveRecord::Base.transaction do
+      taxons = ContentItem.where(document_type: 'taxon', publishing_app: "collections-publisher")
       puts "Found #{taxons.count} taxons"
+
       taxons.each do |taxon|
         puts "Changing #{taxon.title} from #{taxon.publishing_app} to content-tagger"
         taxon.update!(publishing_app: 'content-tagger')
       end
+
+      puts
```
kevindew: I think this would be nicer with an empty string just to indicate that you do intend to send an empty string as it looks a little like it is missing the argument.

lib/tasks/taxonomy.rake
```diff
@@ -2,13 +2,25 @@
...
+
+      path_reservations = PathReservation.where("base_path LIKE '%alpha-taxonomy%'").where(publishing_app: "collections-publisher")
```
kevindew: it looks like  prefixes of /alpha-taxonomy so might be better to use '/alpha-taxonomy%'

lib/tasks/taxonomy.rake
```diff
@@ -2,13 +2,25 @@
...
+      if path_reservations.present?
+        puts "Found #{path_reservations.count} path reservations belonging to collections publisher"
+        path_reservations.each do |path_reservation|
+          puts "Updating #{path_reservation.base_path} to content-tagger"
+          path_reservation.update_column(publishing_app: "content-tagger")
```
kevindew: any reason why you're using `update_column` rather than `update!` ?

kevindew:
aside my nitpicking this looks good 👍 changes make it safer

lib/tasks/taxonomy.rake
mobaig: So as not to trip the `PublishingAppValidator`, which might have a very sensible purpose ordinarily but in this case just prevents us from moving these reservations to `content-tagger`.

mobaig:
Thanks for the review @kevindew . Updated.

lib/tasks/taxonomy.rake
kevindew: Ah yes - that makes sense! `PublishingAppValidator` not so much though 😕 I'll make a card on it

kevindew:
Cool - I'll just see whether anyone has any input on the card I've created before I look at merging. Ideally I'd prefer it to go through validations as it seems a bad president to skip them.

Pragmatically though I know this is a one time thing so I'll try not to let this delay you getting it out.

kevindew:
So it would be a bit complex to sort the validation out before this so I'll merge this now. Do shout if you come up against this validation constraint again

PR #471, opened 2016-08-19
Improve uniqueness constraints

dougdroper:
Adding the other arguments to the uniqueness to ensure that major
updates are not rejected, and that other dependencies are also updated
correctly.

See https://github.com/alphagov/publishing-api/pull/465 for more info

### DIFF
--- a/app/workers/downstream_live_worker.rb
+++ b/app/workers/downstream_live_worker.rb
@@ -14,7 +14,9 @@ class DownstreamLiveWorker
   def self.uniq_args(args)
     [
       args.first.fetch("content_item_id"),
-      'live'
+      args.first.fetch("message_queue_update_type", nil),
+      args.first.fetch("update_dependencies", true),
+      name,
     ]
   end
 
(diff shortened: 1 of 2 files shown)


kevindew:
Looks good 👍 some of those attributes are optional - could be an idea to move their default to constants so we don't repeat them.

kevindew:
Hey cool, you fixed it before I finished my lunch 🎉

app/workers/downstream_live_worker.rb
```diff
@@ -14,8 +14,10 @@ class DownstreamLiveWorker
   def self.uniq_args(args)
     [
       args.first.fetch("content_item_id"),
+      args.first.fetch("message_queue_update_type", nil),
+      args.first.fetch("update_dependencies", true),
       'live'
-    ]
+    ].compact
```
kevindew: niggle but I think this would be better without the compact here since they'd always have arguments rather than a varying number

app/workers/downstream_live_worker.rb
```diff
@@ -14,8 +14,10 @@ class DownstreamLiveWorker
   def self.uniq_args(args)
     [
       args.first.fetch("content_item_id"),
+      args.first.fetch("message_queue_update_type", nil),
+      args.first.fetch("update_dependencies", true),
       'live'
```
kevindew: Probably being over zealous but would using self.name make this name more distinct (oh and should have a trailing comma)

kevindew:
Looks good 👍

PR #473, opened 2016-08-22
Add parent_taxons / child_taxons fields to dependent expansion rules

mobaig:
Finding Things are building tools for editing and navigating taxonomies.
To support this, we need to be able to retrieve parts of the taxonomical
hierarchy from the publishing api.

Adding these additional fields to the expansion rules allows us to do
this efficiently, without making multiple network requests to construct
the hierarchy from the client application.

### DIFF
--- a/app/queries/dependent_expansion_rules.rb
+++ b/app/queries/dependent_expansion_rules.rb
@@ -14,12 +14,13 @@ module Queries
       {
         parent: "children",
         documents: "document_collections",
-        working_groups: 'policies'
+        working_groups: 'policies',
+        parent_taxons: "child_taxons",
       }[link_type.to_sym]
     end
 
     def recursive_link_types
-      [:parent]
+      [:parent, :parent_taxons]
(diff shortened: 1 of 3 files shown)


MatMoore:
👍 from me but might be worth somebody more familiar with dependency resolution functionality reviewing as well?

cbaines:
I'm not particularly familiar with the dependency resolution functionality, but, it looks like this change might also require changing the documentation?

mobaig:
What documentation did you have in mind @cbaines ? I've added a commit which includes a pact-broker link that was missing from the `api.md` doc. Other than that, I'm not sure I've seen anything else that refers to the return values of this endpoint explicitly.

dougdroper:
This part of dependency resolution is meant to be application specific, so no need to document in the publishing API

MatMoore:
It might be worth documenting this in collections publisher then, since that's the publishing app for taxons.

PR #476, opened 2016-08-23
Duplicate content items rake tasks

kevindew:
Continuing from the validator work in https://github.com/alphagov/publishing-api/pull/466 this PR provides rake tasks to check the 3 uniqueness constraints of:

1. There can only be one "draft" item at a `base_path` and there can only be one live ("published" or "unpublished" excluding unpublishing type "substitute") item at a `base_path`
2. For content items sharing a `content_id` and a `locale` there can be at maximum one item that is a "draft" state and one item that is in a "published" or "unpublished" 
3. For content items sharing a `content_id` and a `locale` there can not be items that share the same `user_facing_version`.

These tasks can be run via:
1. `rake duplicate_content_items:base_path_for_state`
2. `rake duplicate_content_items:state_for_locale`
3. `rake duplicate_content_items:version_for_state`

or all 3 can be ran with `rake duplicate_content_items:all`

These replace the previous checks that could be done with `rake duplicate_content_items:check` which I've just removed rather than aliasing since we're not aware of anything using it.

At the time of writing we get the following output (database about 1 week old):
```
vm  publishing-api git:(duplicate-content-items) rake duplicate_content_items:all

There are 67 version for locale conflicts 😭
There are 4 state for locale conflicts 😞
There are 892 base path for state conflicts 😭
```


lib/data_hygiene/duplicate_content_item/base_path_for_state.rb
```diff
@@ -0,0 +1,79 @@
...
+              ROW(content_items.id, content_items.updated_at)
+              ORDER BY content_items.updated_at DESC
+            ) as content_items
+          FROM content_items
+          INNER JOIN locations
+            ON locations.content_item_id = content_items.id
+          INNER JOIN states
+            ON states.content_item_id = content_items.id
+          LEFT JOIN unpublishings
+            ON states.name = 'unpublished'
```
cbaines: You don't need this join condition.

kevindew: It's just there to stop any risk of there being a non "unpublished" item having a matching entry in the `unpublishings` table causing a false positive here. Pretty edge case though.

cbaines: That does make sense, and I do like edge cases being handled :)

lib/data_hygiene/duplicate_content_item/version_for_locale.rb
```diff
@@ -0,0 +1,68 @@
...
+              ROW(content_items.id, content_items.updated_at)
+              ORDER BY content_items.updated_at DESC
+            ) as content_items
+          FROM content_items
+          INNER JOIN translations
+            ON translations.content_item_id = content_items.id
+          INNER JOIN user_facing_versions
+            ON user_facing_versions.content_item_id = content_items.id
+          GROUP BY content_id, locale, user_facing_version
+          HAVING COUNT(*) > 1
```
cbaines: I was having trouble visualising what this query was picking up in the database, so I wrote another query which pulls out the (user_facing_version, state) history grouped by content_id where there is a (user_facing_version, locale) pair which is in the history twice.

Also, I tweaked the above query to group by content_id, and it gave me the same number of rows locally :)
```
SELECT content_items.content_id, ARRAY_AGG(DISTINCT content_items.publishing_app) AS publishing_app,
       ARRAY_AGG((user_facing_versions.number, states.name) ORDER BY user_facing_versions.number) AS version_and_state_details
  FROM content_items
 INNER
  JOIN user_facing_versions
    ON content_items.id                                                   = user_facing_versions.content_item_id
 INNER
  JOIN states
    ON content_items.id                                                   = states.content_item_id
 INNER
  JOIN translations
    ON content_items.id                                                   = translations.content_item_id
 GROUP BY content_items.content_id
HAVING COUNT(DISTINCT (user_facing_versions.number, translations.locale)) < COUNT((user_facing_versions.number, translations.locale));
```

lib/tasks/duplicate_content_items.rake
```diff
@@ -1,6 +1,43 @@
...
+  end
+
+  def conflict_message(number, name)
+    plural = number != 1
+    emoji = case number
+            when 0 then "😀"
+            when 1...50 then "😞"
+            when 50...1_000 then "😭"
+            when 1_000...Float::INFINITY then "😱"
+            else "😕"
```
cbaines: Amazing :+1: after reading the description, I was wondering if I would find this in the diff. I'm sure it will make Doug happy also :P

kevindew: I think he cried a little 👍

lib/data_hygiene/duplicate_content_item/version_for_locale.rb
kevindew: Neat - that's a cool query. I like that it pulls out the version lists. I think that could be useful in a different context were we to validate that state and version are flowing the correct way.

PR #477, opened 2016-08-23
Remove PresentedContentStoreWorker and associated queues

kevindew:
Now that we have deployed the Downstream* workers and the sidekiq queues
are clear on production we can now remove the previous worker class, all
the code that references it and the queues that used to use it.


db/migrate/20160601134855_move_existing_withdrawn_notices.rb
```diff
@@ -1,22 +1,27 @@
...
+    # This has been commented out so that if anyone (crazily) runs all the
+    # migrations¬this won't error¬
 
-      PresentedContentStoreWorker.perform_async(
-        content_store: Adapters::ContentStore,
-        payload: { content_item_id: unpublishing.content_item_id, payload_version: event.id },
-        request_uuid: GdsApi::GovukHeaders.headers[:govuk_request_id],
-      )
-    end
+    # Unpublishing.where(type: "withdrawal").find_each do |unpublishing|
```
danielroseman: I think it would be better to delete the code and just leave the comment.

PR #481, opened 2016-08-24
Restrict unpublishing collisions

kevindew:
This adds a condition (and unfortunately an extra query) when we
unpublish an item to check whether it is unpublished with a type of
"substitute".

It's not really ideal as it could fail if a rogue published document had
a state of published and an unpublishing entry but that seems unlikely.

Rather than getting too bogged down in another substitute exception
clause I have suggested this as an alternative:
https://trello.com/c/EMIigaJ1/306-introduce-a-state-of-substituted

### DIFF
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -126,7 +126,8 @@ module Commands
         end
 
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
-        filter.filter(locale: locale, state: allowed_states).last
+        content_item = filter.filter(locale: locale, state: allowed_states).last
+        content_item if content_item && (payload[:allow_draft] || !Unpublishing.is_substitute?(content_item))
       end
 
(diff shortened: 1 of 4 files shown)


app/commands/v2/unpublish.rb
```diff
@@ -126,7 +126,8 @@ def find_unpublishable_content_item(content_id)
         end
 
         filter = ContentItemFilter.new(scope: ContentItem.where(content_id: content_id))
-        filter.filter(locale: locale, state: allowed_states).last
+        content_item = filter.filter(locale: locale, state: allowed_states).last
+        content_item if content_item && (payload[:allow_draft] || !Unpublishing.is_substitute?(content_item))
```
cbaines: I'm not sure what `allow_draft` is doing in here?

kevindew: It is a bit ugly - it's just to shortcut the need to run that query if we have a draft (as we'll only have drafts if `payload[:allow_draft]` is true

PR #482, opened 2016-08-24
Downstream draft race condition

kevindew:
This corrects a potential race condition problem which allowed a published/unpublished item to be sent to the draft content store when a draft should occupy the item.

This was the previous scenario:

1. Unpublished item (a) enqueued to be sent to the draft content store
2. A new item (b) which has the same base path as a is created as a draft and enqueued to the content store
3. Sidekiq job for b completes putting the draft item in the draft content store
4. Sidekiq job for a completes replacing the draft item with the now out of date unpublished item


app/services/downstream_service.rb
```diff
@@ -18,6 +18,11 @@ def self.update_draft_content_store(downstream_payload)
       message = "Can only send draft, published and unpublished items to draft content store"
       raise DownstreamInvalidStateError.new(message)
     end
+    if downstream_payload.state != "draft" && draft_at_base_path?(downstream_payload.base_path)
+      message = "Can't send #{downstream_payload.state} item to draft content store, as there is a draft occupying the same base path"
+      raise DownstreamInvalidStateError.new(message)
```
cbaines: If this is just an concurrency issue, I think this should be silent.

kevindew: Sure, it's not strictly a concurrency issue as it could happen any time you queued a published/unpublished item. 

I think it makes sense for an error to be raised here as it is an error situation, however it's in the concern of the workers on how to deal with it (currently log for indirect publishings and raise for direct ones).

cbaines: What do you mean by "error situation"? Disregarding some bug in submitting the job, this particular condition just means that the world state has changed between the job being submitted, and being given to a worker, which is perfectly normal, and not something that I consider an error.

kevindew: By "error situation" I meant that:
- It is an error to add a published item to the draft content store when there is a draft item occupying that base path

As for whether the worker should treat it as an error or an indiscretion, that's a reasonable argument but outside the context of this work. I'd be fine with them just alerting the logger.

cbaines: So Kevin and I discussed this and a few other things for a while.

I think this is ok, but it would be best just to log these errors, and not send them to Errbit.

PR #483, opened 2016-08-25
Remove one-off rake task used to change the publishing_app for taxons

carvil:
This task was added in order to change the ownership of existing taxons to
another publishing app. We ran this task in production successfully, therefore
it is no longer needed and we should delete it.

Related PRs:
- https://github.com/alphagov/publishing-api/pull/451
- https://github.com/alphagov/publishing-api/pull/470

Trello: https://trello.com/c/0LzcgY9n/68-change-publishing-app-for-old-taxons-owned-by-collections-publisher


mobaig:
👍

PR #484, opened 2016-08-25
Capture Unpublishing record invalid errors

dougdroper:
Previously was returning a 500, now returns 422 with the validation message 

https://trello.com/c/EXyXItHs/784-provide-necessary-validations-for-required-fields-small

also includes some refactoring work

### DIFF
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -4,7 +4,7 @@ module Commands
       def call
         validate
         State.supersede(previous_item) if previous_item
-        transision_state
+        transition_state
         delete_linkable
 
         after_transaction_commit do
@@ -16,32 +16,21 @@ module Commands
 
     private
 
-      def transision_state
-        raise_invalid_type unless valid_type?
-        method = {
-          "withdrawal" => method(:withdraw),
-          "redirect" => method(:redirect),
-          "vanish" => method(:vanish),
-          "gone" => method(:gone),
-        }[type]
-
-        begin
-          method.call
-        rescue ActiveRecord::RecordInvalid => e
-          raise_command_error(422, e.message, fields: {})
-        end
+      def transition_state
+        raise_invalid_unpublishing_type unless valid_unpublishing_type?
+        unpublish
       end
 
-      def valid_type?
-        %w(withdrawal redirect gone vanish).include?(type)
+      def valid_unpublishing_type?
+        %w(withdrawal redirect gone vanish).include?(unpublishing_type)
       end
 
-      def type
+      def unpublishing_type
         payload.fetch(:type)
       end
 
-      def raise_invalid_type
-        message = "#{type} is not a valid unpublishing type"
+      def raise_invalid_unpublishing_type
+        message = "#{unpublishing_type} is not a valid unpublishing type"
         raise_command_error(422, message, fields: {})
       end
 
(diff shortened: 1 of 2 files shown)


app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,73 @@ module Commands
...
+        end
+
+        Success.new(content_id: content_id)
+      end
+
+    private
+
+      def transision_state
+        raise_invalid_type unless valid_type?
+        method = {
```
dougdroper: Actually, I don't think this helps much... refactoring again

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        after_transaction_commit do
+          send_downstream
+        end
+
+        Success.new(content_id: content_id)
+      end
+
+    private
+
+      def transision_state
```
cbaines: s/transision/transition ?

dougdroper: yip thanks!

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        raise_invalid_type unless valid_type?
+
+        begin
+          unpublish
+        rescue ActiveRecord::RecordInvalid => e
+          raise_command_error(422, e.message, fields: {})
+        end
+      end
+
+      def valid_type?
```
cbaines: Maybe specify that this is an unpublishing type?

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        %w(withdrawal redirect gone vanish).include?(type)
+      end
+
+      def type
+        payload.fetch(:type)
+      end
+
+      def raise_invalid_type
```
cbaines: and again specify that this is about unpublishing types here.

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        message = "#{type} is not a valid unpublishing type"
+        raise_command_error(422, message, fields: {})
+      end
+
+      def content_item
+        @content_item ||= find_unpublishable_content_item
```
cbaines: I think it would be better remove a layer of indirection, and this method just by moving this test and assignment inside find_unpublishable_content_item.

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        rescue ActiveRecord::RecordInvalid => e
+          raise_command_error(422, e.message, fields: {})
+        end
+      end
+
+      def valid_type?
+        %w(withdrawal redirect gone vanish).include?(type)
+      end
+
+      def type
```
cbaines: I think it would be useful to call this unpublishing_type.

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+
+      def content_item
+        @content_item ||= find_unpublishable_content_item
+      end
+
+      def content_id
+        @content_id ||= payload.fetch(:content_id)
+      end
+
+      def validate_allow_discard_draft!
```
cbaines: What does the ! indicate here? In Scheme/Guile, I'm used to this meaning that it has side effects, but I don't normally consider errors side effects?

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,67 @@ module Commands
...
+        Success.new(content_id: content_id)
+      end
+
+    private
+
+      def transition_state
+        raise_invalid_type unless valid_type?
+
+        begin
+          unpublish
```
cbaines: I think this would be better inlined.

app/commands/v2/unpublish.rb
dougdroper: I think I am ok with this, moving it into the `find_unpublishable_content_item` would make the memorisation quite tricky.

app/commands/v2/unpublish.rb
dougdroper: I was using it to indicate that it sends that error, but it is not consistent, and perhaps there should be a version without the bang too, so dropping.

app/commands/v2/unpublish.rb
```diff
@@ -2,22 +2,62 @@ module Commands
...
+      end
+
+    private
+
+      def transition_state
+        raise_invalid_unpublishing_type unless valid_unpublishing_typ?
+        unpublish
+      end
+
+      def valid_unpublishing_typ?
```
cbaines: Why `typ` instead of `type`?

dougdroper: has a nicer ring to it :) fixed

PR #485, opened 2016-08-25
Add a migration to delete two duplicate drafts

cbaines:
When these content items where put in to the publishing api, there was a
matching duplicate.

### DIFF
--- a/db/migrate/20160825122452_remove_duplicate_drafts.rb
+++ b/db/migrate/20160825122452_remove_duplicate_drafts.rb
@@ -1,0 +1,34 @@
+class RemoveDuplicateDrafts < ActiveRecord::Migration
+  def up
+    content_items = ContentItem.where(id: [815434, 818621])
+
+    if content_items.any? { |x| x.updated_at != x.created_at }
+      raise "content item #{x.id} has been changed"
+    end
+
+    states = State.where(content_item: content_items)
+
+    if states.any? { |x| x.name != "draft" }
+      raise "content item #{x.id} not in draft state"
+    end
+
+    states.destroy_all
+
+    other_supporting_classes = [
+      AccessLimit,
+      Linkable,
+      Location,
+      Translation,
+      Unpublishing,
+      UserFacingVersion
+    ]
+
+    other_supporting_classes.each do |klass|
+      klass.where(content_item: content_items).destroy_all
+    end
+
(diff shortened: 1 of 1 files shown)


db/migrate/20160825122452_remove_duplicate_drafts.rb
```diff
@@ -0,0 +1,34 @@
+class RemoveDuplicateDrafts < ActiveRecord::Migration
+  def up
+    content_items = ContentItem.where(id: [815434, 818621])
+
+    if content_items.any? { |x| x.updated_at != x.created_at }
+      raise "content item #{x.id} has been changed"
```
kevindew: I think this should fail silently here as if you ran through all migrations you wouldn't want this to die here.

kevindew:
You'll also want the schema.rb class which is updated after the migration has run.

Aside this looks cool though 👍

db/migrate/20160825122452_remove_duplicate_drafts.rb
cbaines: As this is deleting the content items, if you rerun the migration, it does nothing as no content items are found with those ids.

These errors will only happen if the migration should not be run when run in production because the assumptions under which I wrote it do not match the database.

I can still change it just to puts and return, but having it fail and abort the transaction seems better.

kevindew: So in the scenario that the data has changed and your assumptions are incorrect
- Raising an error here would fail a production deploy and block us until the migration was altered for the new data
- Just logging that problem occurred (could use errbit or logger.warn rather than just puts) wouldn't block anything

We still need to monitor for this error so even if it failed completely silently we'd still catch it as we need to monitor for similar.

kevindew:
I've just noticed the items here will be removed by: https://github.com/alphagov/publishing-api/pull/488
see: https://github.com/alphagov/publishing-api/blob/30d8c06b1e33ff9380e14261228197af407ae84e/db/migrate/20160825141159_resolve_version_for_locale_conflicts.rb#L287 and https://github.com/alphagov/publishing-api/blob/30d8c06b1e33ff9380e14261228197af407ae84e/db/migrate/20160825141159_resolve_version_for_locale_conflicts.rb#L387 so might be able to forgo a migration here.

cbaines:
Yep, I'll close this.

PR #487, opened 2016-08-25
Optimise put_content endpoint.

danielroseman:
A series of optimisations to put_content.

There can be various different paths through this code, so this PR also adds command-line parameters to the benchmark script (`bench/put_content.rb`) to test those paths.  Results from each run:

```
New item each time: master
  7.140000   0.780000   7.920000 ( 10.570383)
4324 SQL queries

now
  5.660000   0.460000   6.120000 (  8.595015)
2928 SQL queries

Update same item each time: master
  4.640000   0.450000   5.090000 (  6.907528)
2249 SQL queries

now
  3.330000   0.590000   3.920000 (  5.665207)
1543 SQL queries

Redrafting: master
  4.090000   0.320000   4.410000 (  6.650513)
2218 SQL queries

now
  3.120000   0.260000   3.380000 (  5.184676)
1513 SQL queries
```

The main win is in doing the draft vs live validation for LockVersion and UserFacingVersion in a single query, and there are a few other minor changes to reduce the number of unnecessary queries more generally. There's still a lot to gain in the uniqueness validations, but that would conflict with other work going on around those so it's best to leave that for a separate PR.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -66,7 +66,7 @@ module Commands
       end
 
       def update_existing_content_item(content_item)
-        check_version_and_raise_if_conflicting(content_item, payload[:previous_version])
+        version = check_version_and_raise_if_conflicting(content_item, payload[:previous_version])
 
         if content_with_base_path?
           clear_draft_items_of_same_locale_and_base_path(content_item, locale, base_path)
@@ -77,7 +77,8 @@ module Commands
 
         update_content_item(content_item)
         update_last_edited_at_if_needed(content_item, payload[:last_edited_at])
-        increment_lock_version(content_item)
+
+        increment_lock_version(version)
 
         if path_has_changed?(previous_location)
           from_path = previous_location.base_path
@@ -252,8 +253,7 @@ module Commands
         content_item.update_attributes(last_edited_at: last_edited_at) if last_edited_at
       end
 
-      def increment_lock_version(content_item)
-        lock_version = LockVersion.find_by!(target: content_item)
+      def increment_lock_version(lock_version)
         lock_version.increment
         lock_version.save!
       end
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -1,6 +1,7 @@
 module Commands
   module V2
     class PutContent < BaseCommand
+      ITEM_NOT_FOUND = Class.new
       def call
         raise_if_links_is_provided
         validate_schema
@@ -42,7 +43,7 @@ module Commands
 
         update_last_edited_at_if_needed(content_item, payload[:last_edited_at])
 
-        if previously_published_item
+        if previously_published_item != ITEM_NOT_FOUND
           set_first_published_at(content_item, previously_published_item)
 
           previous_location = Location.find_by(content_item: previously_published_item)
@@ -168,7 +169,7 @@ module Commands
       end
 
       def lock_version_number_for_new_draft
-        if previously_published_item
+        if previously_published_item != ITEM_NOT_FOUND
           lock_version = LockVersion.find_by!(target: previously_published_item)
           lock_version.number + 1
         else
@@ -177,7 +178,7 @@ module Commands
       end
 
       def user_facing_version_number_for_new_draft
-        if previously_published_item
+        if previously_published_item != ITEM_NOT_FOUND
           user_facing_version = UserFacingVersion.find_by!(content_item: previously_published_item)
           user_facing_version.number + 1
         else
@@ -190,7 +191,7 @@ module Commands
           filter = ContentItemFilter.new(scope: pessimistic_content_item_scope)
           content_items = filter.filter(state: %w(published unpublished), locale: locale)
           UserFacingVersion.latest(content_items)
-        )
+        ) || ITEM_NOT_FOUND
       end
 
       def set_first_published_at(content_item, previously_published_item)
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -21,11 +21,11 @@ private
   end
 
   def draft_and_live_versions
-    result = Queries::DraftAndLiveVersions.call(target, self.class.name.tableize)
+    result = Queries::DraftAndLiveVersions.call(target, self.class.table_name)
     if result["draft"] == self.number
       draft_version = self.number
-      live_version = result["published"]
-    elsif result["published"] == self.number
+      live_version = result["live"]
+    elsif result["live"] == self.number
       draft_version = result["draft"]
       live_version = self.number
     end
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -20,13 +20,26 @@ private
     errors.add(:number, message)
   end
 
+  def draft_and_live_versions
+    result = Queries::DraftAndLiveVersions.call(target, self.class.name.tableize)
+    if result["draft"] == self.number
+      draft_version = self.number
+      live_version = result["published"]
+    elsif result["published"] == self.number
+      draft_version = result["draft"]
+      live_version = self.number
+    end
+
+    [draft_version, live_version]
+  end
+
   def draft_cannot_be_behind_live
     draft_version, live_version = draft_and_live_versions
 
     return unless draft_version && live_version
 
-    if draft_version.number < live_version.number
-      mismatch = "(#{draft_version.number} < #{live_version.number})"
+    if draft_version < live_version
+      mismatch = "(#{draft_version} < #{live_version})"
       message = "draft #{self.class.name} cannot be behind the live #{self.class.name} #{mismatch}"
       errors.add(:number, message)
     end
--- a/app/queries/draft_and_live_versions.rb
+++ b/app/queries/draft_and_live_versions.rb
@@ -1,0 +1,56 @@
+module Queries
+  module DraftAndLiveVersions
+    extend ArelHelpers
+    def self.call(content_item, target_class, locale = nil, base_path = nil)
+      content_items_table = table(:content_items)
+      states_table = table(:states)
+      translations_table = table(:translations)
+      locations_table = table(:locations)
+      versions_table = table(target_class)
+      versions_fk = target_class == "lock_versions" ? :target_id : :content_item_id
+
+      if base_path.nil?
+        base_path = locations_table.project(locations_table[:base_path])
+          .where(locations_table[:content_item_id].eq(content_item.id))
+      end
+      if locale.nil?
+        locale = translations_table.project(translations_table[:locale])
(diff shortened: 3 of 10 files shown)


app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,56 @@
...
+      states_table = table(:states)
+      translations_table = table(:translations)
+      locations_table = table(:locations)
+      versions_table = table(target_class)
+      versions_fk = target_class == "lock_versions" ? :target_id : :content_item_id
+
+      if base_path.nil?
+        base_path = locations_table.project(locations_table[:base_path])
+          .where(locations_table[:content_item_id].eq(content_item.id))
+      end
```
kevindew: This condition is removable as the base_path doesn't impact the user_facing_version of an item.

app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,56 @@
...
+        .outer_join(locations_table).on(
+          content_items_table[:id].eq(locations_table[:content_item_id])
+        )
+        .outer_join(versions_table).on(
+          content_items_table[:id].eq(versions_table[versions_fk])
+        )
+        .where(content_items_table[:content_id].eq(content_item.content_id))
+        .where(locations_table[:base_path].eq(base_path))
+        .where(translations_table[:locale].eq(locale))
+        .where(states_table[:name].in(%w(draft published)))
```
kevindew: The live version could also be unpublished here so we'd want to consider that value - this is probably a bug we have in our current version.

app/queries/draft_and_live_versions.rb
danielroseman: Thanks. I see what you're saying but the original version had a call to `ContentItemFilter.similar_to` which does do a query for the base_path - was that wrong or just unnecessary?

kevindew: I believe that was actually wrong
e.g. 
```ruby
 content_id = SecureRandom.uuid
FactoryGirl.create(:live_content_item, content_id: content_id, user_facing_version: 3, base_path: "/old-base-path")
draft = FactoryGirl.create(:draft_content_item, content_id: content_id, user_facing_version: 2, base_path: "/new-base-path")
Queries::DraftAndLiveVersions.call(draft, :user_facing_versions, "en")
```
`=> {"draft"=>2}`

Though I don't think it'll have come into view at all as you'd need to have a incorrect version at the same time as a new content item changes it's base_path

app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,56 @@
+module Queries
+  module DraftAndLiveVersions
+    extend ArelHelpers
+    def self.call(content_item, target_class, locale = nil, base_path = nil)
```
kevindew: I think the variable naming of target_class is a little misleading and would be better represented by either:
- Change variable name to target_table_name which more closely represents how it's used
- pass in the model class itself and then run `table_name` on it to get the table name

app/models/version.rb
```diff
@@ -20,13 +20,26 @@ def numbers_must_increase
     errors.add(:number, message)
   end
 
+  def draft_and_live_versions
+    result = Queries::DraftAndLiveVersions.call(target, self.class.name.tableize)
```
kevindew: wouldn't really affect us unless we started changing table names, but this would be less brittle if `self.class.name.tableize` was `self.class.table_name`

spec/queries/draft_and_live_versions_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  before do
+    3.times do |i|
+      FactoryGirl.create(:content_item,
+        state: 'superseded',
+        content_id: content_id,
+        user_facing_version: i,
+        base_path: base_path
+      )
+    end
+    @item = FactoryGirl.create(:live_content_item, :with_draft,
```
kevindew: Minor - but it'd be a bit more rspec if this was defined via a `let` rather than an instance variable.

app/commands/v2/put_content.rb
```diff
@@ -1,6 +1,7 @@
 module Commands
   module V2
     class PutContent < BaseCommand
+      ITEM_NOT_FOUND = Class.new
```
kevindew: This caught me as a bit of WTF at first, but I see what you did here and it's elegant 👍

app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,56 @@
...
+        .where(locations_table[:base_path].eq(base_path))
+        .where(translations_table[:locale].eq(locale))
+        .where(states_table[:name].in(%w(draft published)))
+
+      partitioned = cte(scope, as: :partitioned)
+      query = partitioned.table.project(Arel.star)
+        .with(partitioned.compiled_scope)
+        .where(partitioned.table[:r].eq(1))
+
+      Hash[get_rows(query).map { |i| [i["state"], i["version_number"].to_i] }]
```
cbaines: I would suggest using two unioned queries, instead of a window function, as I think that is better suited to what you are trying to do.

I've tried comparing this to the query generated by the above code, and it is faster, but more importantly should be much easier to understand.

```
(
    SELECT 'live' AS content_store, lock_versions.number AS lock_version
    FROM content_items
    INNER JOIN states ON content_items.id = states.content_item_id
    INNER JOIN lock_versions ON content_items.id = lock_versions.target_id
        AND lock_versions.target_type = 'ContentItem'
    WHERE content_items.content_id = '...'
        AND states.name IN ('published', 'unpublished')
    ORDER BY lock_version DESC
    LIMIT 1
)
UNION
(
    SELECT 'draft' AS content_store, lock_versions.number AS lock_version
    FROM content_items
    INNER JOIN states ON content_items.id = states.content_item_id
    INNER JOIN lock_versions ON content_items.id = lock_versions.target_id
        AND lock_versions.target_type = 'ContentItem'
    WHERE content_items.content_id = '...'
        AND states.name = 'draft'
    ORDER BY lock_version DESC
    LIMIT 1
);

```

cbaines: Ah, I'm missing filtering on base_path, and locale.

app/commands/v2/put_content.rb
cbaines: I don't know why this is Class.new here, can someone explain?

danielroseman: It's a sentinel object; the explanation is in the commit message but maybe it would be clearer if there was a comment here as well.

app/queries/draft_and_live_versions.rb
cbaines: And using a CTE makes this a little neater, so now I have something that looks like:

```
EXPLAIN ANALYZE WITH content_store_and_lock_versions AS (
    SELECT
        CASE WHEN states.name IN ('published', 'unpublished') THEN 'live'
             WHEN states.name = 'draft' THEN 'draft' END
        AS content_store,
        lock_versions.number AS lock_version
    FROM content_items
    INNER JOIN states ON content_items.id = states.content_item_id
    INNER JOIN locations ON content_items.id = locations.content_item_id
    INNER JOIN translations ON content_items.id = translations.content_item_id
    INNER JOIN lock_versions ON content_items.id = lock_versions.target_id
        AND lock_versions.target_type = 'ContentItem'
    WHERE content_items.content_id = '...'
        AND states.name IN ('published', 'unpublished', 'draft')
)
(
    SELECT content_store, lock_version
    FROM content_store_and_lock_versions
    WHERE content_store = 'live'
    ORDER BY lock_version DESC
    LIMIT 1
)
UNION
(
    SELECT content_store, lock_version
    FROM content_store_and_lock_versions
    WHERE content_store = 'draft'
    ORDER BY lock_version DESC
    LIMIT 1
);
```

Not sure exactly what the conditions on the content items need to be, but its all in one place now.

danielroseman:
@kevindew @cbaines All comments addressed; now using a UNION query, removed unnecessary base_path join, and cleared up param names.

boffbowsh:
Nice 👏. Will let one of the other reviewers merge though

app/models/version.rb
```diff
@@ -20,13 +20,26 @@ def numbers_must_increase
...
 
+  def draft_and_live_versions
+    result = Queries::DraftAndLiveVersions.call(target, self.class.table_name)
+    if result["draft"] == self.number
+      draft_version = self.number
+      live_version = result["live"]
+    elsif result["live"] == self.number
+      draft_version = result["draft"]
+      live_version = self.number
+    end
```
kevindew: So I got a little caught on what happens here when neither the `if` or `elsif` are called and as to how this returns `[nil, nil]` when neither of the variables seem to be defined - which is a quirk of ruby scoping rules: http://stackoverflow.com/questions/12614805/why-can-i-refer-to-a-variable-outside-of-an-if-unless-case-statement-that-never

I think if we defined them before the conditional or had an else it'd be clearer for a future reader. I know this is just code that was moved, so could resolve this outside of this PR.

kevindew:
Just ⬆️  from me (I should have commented on that first time around) otherwise looks good to merge 👍

app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,65 @@
...
+          content_items_table[:id].eq(translations_table[:content_item_id])
+        )
+        .join(versions_table).on(
+          content_items_table[:id].eq(versions_table[versions_fk])
+        )
+        .where(content_items_table[:content_id].eq(content_item.content_id))
+        .where(translations_table[:locale].eq(locale))
+        .where(states_table[:name].in(%w(draft published unpublished)))
+      if target_table_name == "lock_versions"
+        scope.where(versions_table[:target_type].eq("ContentItem"))
```
cbaines: I'm not sure how this works, as scope is not being assigned (it must be a mutation, as I tested it, and it works...).

app/queries/draft_and_live_versions.rb
```diff
@@ -0,0 +1,65 @@
...
+          cte[:state],
+          cte[:version_number]
+        )
+        .where(cte[:state].eq("live"))
+        .order(cte[:version_number].desc)
+        .take(1)
+
+      # Arel doesn't correctly add parens around complex UNION clauses, which
+      # Postgres requires, so we need to build up the SQL query manually.
+      query = "WITH cte AS (#{scope.to_sql}) (#{left.to_sql}) UNION (#{right.to_sql})"
```
cbaines: :+1:

cbaines:
This is ready to merge as far as I am concerned.

app/queries/draft_and_live_versions.rb
danielroseman: Yes, Arel objects mutate themselves when methods are called.

PR #488, opened 2016-08-26
Resolve known version for locale conflicts

kevindew:
This is a migration which removes items which have version conflicts. Inside the migration are details about all the items which would be deleted from the database.

Outputs from running locally
```
vm  publishing-api git:(nuke-invalid-data) ✗ rake duplicate_content_items:version_for_locale

There are 133 version for locale conflicts 😭
vm  publishing-api git:(nuke-invalid-data) ✗ rake db:migrate

== 20160825141159 ResolveVersionForLocaleConflicts: migrating =================
== 20160825141159 ResolveVersionForLocaleConflicts: migrated (1.4831s) ========

vm  publishing-api git:(nuke-invalid-data) ✗ rake duplicate_content_items:version_for_locale

There are 0 version for locale conflicts 😀
```

It can be recovered, I have a sql file with all of the items to be deleted. Although it could be dangerous to run that in full as it may introduce items that have been deleted elsewhere.

```
vm  publishing-api git:(nuke-invalid-data) ✗ psql -d publishing_api_development -f version-for-locale-recovery.sql

vm  publishing-api git:(nuke-invalid-data) ✗ rake db:rollback

== 20160825141159 ResolveVersionForLocaleConflicts: reverting =================
== 20160825141159 ResolveVersionForLocaleConflicts: reverted (0.0000s) ========

vm  publishing-api git:(nuke-invalid-data) ✗ rake duplicate_content_items:version_for_locale

There are 133 version for locale conflicts 😭
```

### DIFF
--- a/db/migrate/20160825141159_resolve_version_for_locale_conflicts.rb
+++ b/db/migrate/20160825141159_resolve_version_for_locale_conflicts.rb
@@ -1,0 +1,695 @@
+require_relative "helpers/delete_content_item"
+
+class ResolveVersionForLocaleConflicts < ActiveRecord::Migration
+  def up
+    updated_time_range = -> (hash) do
+      parsed = Time.zone.parse(hash[:updated_at])
+      hash.merge(updated_at: (parsed..(parsed + 1.second)))
+    end
+
+    to_delete.map(&updated_time_range).each do |criteria|
+      # we can use a datetime range in a where but not a find_by hence odd choice
+      content_item = ContentItem.where(criteria).first
+      delete(content_item) if content_item
+    end
+  end
+
+  def delete(content_item)
+    Helpers::DeleteContentItem.destroy_supporting_objects(content_item)
+    content_item.destroy
+  end
+
+  def to_delete
+    # This data was pulled using DataHygiene::DuplicateContentItem::VersionForLocale
+    # from a 2016-08-25 database
+    # The items to be deleted are chosen by first their state (supserseded items
+    # are removed preferably, then draft) and the timestamp where items added
+    # later are removed
+    [
+      # Content item: 0227770e-0be5-46db-87c4-e1eac21e2a63, locale: en, version: 1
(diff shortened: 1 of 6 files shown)


db/migrate/helpers/delete-content-item.rb
```diff
@@ -0,0 +1,21 @@
+module Helpers
+  module DeleteContentItem
+    def self.destroy_supporting_objects(content_items)
```
dougdroper: Nice, can we update the readme to point to this? 
https://github.com/alphagov/publishing-api#deleting-content-items

dougdroper: And can we `Array` the content_items?

kevindew: Good point about the readme

hmm the `Array` would be unnecessary as ActiveRecord handles it being a model or an array. Do you mean to add that so it's explicit that it accepts a single item or a collection?

dougdroper: explicit that it accepts a single item or a collection

db/migrate/20160825141159_resolve_version_for_locale_conflicts.rb
```diff
@@ -0,0 +1,690 @@
...
+      delete(content_item) if content_item
+    end
+  end
+
+  def delete(content_item)
+    Helpers::DeleteContentItem.destroy_supporting_objects(content_item)
+    content_item.destroy
+  end
+
+  def to_delete
```
cbaines: It would be good to comment here saying where the data below came from.

danielroseman:
:+1:, merge once conflicts are resolved

kevindew:
Thanks @danielroseman - Since I want these commits to sit before the base_pase_for_state_conflicts (so validations are in chronological order and latest validation is in db/schema.rb) I've got with a merge commit. Do shout if you think I should rebase it instead.

PR #492, opened 2016-08-31
Delete Duplicate About Pages

mgrassotti:
Remove one of the two duplicate “about” pages found

### DIFF
--- a/db/migrate/20160831103251_delete_duplicate_about_pages.rb
+++ b/db/migrate/20160831103251_delete_duplicate_about_pages.rb
@@ -1,0 +1,13 @@
+class DeleteDuplicateAboutPages < ActiveRecord::Migration
+  # Remove /government/organisations/animal-and-plant-health-agency/corporate_information_pages/623676
+  def up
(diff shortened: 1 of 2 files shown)


db/migrate/20160831103251_delete_duplicate_about_pages.rb
```diff
@@ -0,0 +1,13 @@
+class DeleteDuplicateAboutPages < ActiveRecord::Migration
+  # Remove /government/organisations/animal-and-plant-health-agency/corporate_information_pages/623676
+  def up
+    content_id = "5ff02f7a-7631-11e4-a3cb-005056011aef"
+    content_items = ContentItem.where(content_id: content_id).limit(1)
```
boffbowsh: What's the `limit(1)` for? I think it's best to remove all content items for that content id

mgrassotti: It was to limit the riskiness of the deletion, but yes it doesn't make much sense as we want to remove all content items for that content id.

PR #493, opened 2016-08-31
Update linkable on publish.

danielroseman:
Currently we only update the Linkable in put_content when a new draft
is created with a different base path. This means that if the base
path does not change, the Linkable still refers to the old content
item, so if the title changes the get_linkable endpoint will return
the old data.

This change ensures that on publish any linkable pointing at the old
content_item is deleted and the new one updated. There are three cases:

* if the base_path changed, we have a published linkable pointing at
the old content_item and a draft linkable pointing at the new item;
delete the old one and update the draft one to published.
* if the base_path did not change, we have one linkable but it is
pointing at the old content_item; the delete is a no-op but we update
the linkable to point to the new content_item.
* if there is no previously published item, just update the linkable
state to published.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -36,16 +36,19 @@ module Commands
           if previous_item
             previous_location = Location.find_by(content_item: previous_item)
 
-            publish_redirect_if_content_item_has_moved(location, previous_location, translation)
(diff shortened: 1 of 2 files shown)


kevindew:
looks good 👍

PR #494, opened 2016-09-01
Redirect validation reject self redirect

cbaines:
I've found one case of this in the Publishing API (content_item_id: 645143), and this does cause a redirect loop on the live site.

I'll put up another Pull Request to remove the offending content_item.


kevindew:
Looks cool to me 👍 

As there are no tests (I see there's none for that class at all) have you been able to test it ok?

cbaines:
> As there are no tests (I see there's none for that class at all) have you been able to test it ok?

I confirmed that it worked as intended for content item 645143.

PR #495, opened 2016-09-01
Add migration for removing a content item which redirects to itself

cbaines:
This causes a redirect loop on the live site. In a different branch, I have
proposed adding validation to reject data like this.


kevindew:
Good catch 👍

PR #497, opened 2016-09-01
Filter by publishing app - development issue

carvil:
Hi publishing-api team,

This is a follow-up discussion regarding the PR #202 opened a few months back.

The endpoint `/v2/content` requires the user to have permissions to access the content. The [filters](https://github.com/alphagov/publishing-api/blob/7aa9c18dd61b2c9cee53a04e9fa0e224725d9f77/app/controllers/v2/content_items_controller.rb#L10) passed into `Queries::GetContentCollection` have a `publishing_app` [parameter](https://github.com/alphagov/publishing-api/blob/7aa9c18dd61b2c9cee53a04e9fa0e224725d9f77/app/controllers/v2/content_items_controller.rb#L58-L62) which is defined based on the user permissions and the user `app_name` attribute, which in turn is set after authentication [here](https://github.com/alphagov/publishing-api/blob/d739517d73a409767838a930e48d558c41dfbe41/app/controllers/application_controller.rb#L16).

For example, an API user `content-tagger@alphagov.co.uk` would have an app name of `content-tagger`, which matches a known `publishing_app`.

This is all fine, however in development mode `gds-sso` uses a [dummy](https://github.com/alphagov/gds-sso/blob/a1a0d92564618fec10ff9855c12ab373fc8f6ba5/lib/gds-sso/bearer_token.rb#L55-L60) user instead. When we set the application name as mentioned above, the `app_name` of the user will be `dummyapiuser`, which doesn't match any of the `publishing_app` valid values (e.g. `content-tagger`).

The consequence is that the API call to `/v2/content` returns no results in development mode, which is not ideal.

We would like to replace some `get_linkables` calls to `get_content_items` in order to get more attributes for our content items (e.g. `public_updated_at`), but it wouldn't be practical to do that as things stand.

@tijmenb made some suggestions in the PR #202 .

Any more thoughts on those?


kevindew:
Sure, I can see how this would be a problem for you. I've passed it on to Product Manager and we can look into getting a discussion and/or story for this. 

Our tech lead is on leave today but he'll probably be able to chime in based off knowledge from the previous PR

tuzz:
As discussed in Slack, this work is not required because the signon user can set the `view_all` permission which bypasses the restriction by publishing app:

```ruby
unless current_user.has_permission?('view_all')
  current_user.app_name
end
```

On the assumption that works, I'm going to close this. Thanks.

PR #498, opened 2016-09-05
Fix bug with basepath reservation

carlosmartinez:
Resolves [a bug](https://trello.com/c/CrtqBpmu/793-fix-bug-with-basepath-reservation-small#) and introduces some refactoring.

### DIFF
--- a/app/models/path_reservation.rb
+++ b/app/models/path_reservation.rb
@@ -5,8 +5,9 @@ class PathReservation < ActiveRecord::Base
 
   def self.reserve_base_path!(base_path, publishing_app)
     record = find_or_initialize_by(base_path: base_path)
-    record.publishing_app = publishing_app
-    record.save!
+    ActiveRecord::Base.transaction do
+      record.update!(publishing_app: publishing_app)
+    end
   rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation
     # If a path is already reserved by the same publishing app, ignore the error
     record = find_by(base_path: base_path)
--- a/app/models/path_reservation.rb
+++ b/app/models/path_reservation.rb
@@ -1,18 +1,40 @@
 class PathReservation < ActiveRecord::Base
   validates :base_path, absolute_path: true
   validates :publishing_app, presence: true
-  validates_with PublishingAppValidator
 
   def self.reserve_base_path!(base_path, publishing_app)
-    record = find_or_initialize_by(base_path: base_path)
+    existing = find_by(base_path: base_path)
+    if existing.nil?
+      create_path_reservation(base_path, publishing_app)
+    else
+      existing.ensure_unique(publishing_app)
+    end
+  end
+
+  def self.create_path_reservation(base_path, publishing_app)
     ActiveRecord::Base.transaction do
-      record.update!(publishing_app: publishing_app)
+      create!(base_path: base_path, publishing_app: publishing_app)
     end
   rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation
     # If a path is already reserved by the same publishing app, ignore the error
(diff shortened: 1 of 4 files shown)


kevindew:
Woo 👏  a PR on day one 🎉

app/models/path_reservation.rb
```diff
@@ -1,18 +1,43 @@
 class PathReservation < ActiveRecord::Base
   validates :base_path, absolute_path: true
   validates :publishing_app, presence: true
-  validates_with PublishingAppValidator
 
-  def self.reserve_base_path!(base_path, publishing_app)
-    record = find_or_initialize_by(base_path: base_path)
-    ActiveRecord::Base.transaction do
-      record.update!(publishing_app: publishing_app)
+  class << self
```
danielroseman: The styleguide recommends avoiding `class << self`, see third point under https://github.com/alphagov/styleguides/blob/master/ruby.md#classes

app/models/path_reservation.rb
```diff
@@ -1,17 +1,43 @@
...
 
-  def self.reserve_base_path!(base_path, publishing_app)
-    record = find_or_initialize_by(base_path: base_path)
-    record.publishing_app = publishing_app
-    record.save!
-  rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation
-    # If a path is already reserved by the same publishing app, ignore the error
-    record = find_by(base_path: base_path)
-    raise unless record.publishing_app == publishing_app
+  class << self
```
danielroseman: The styleguide recommends avoiding `class << self`, see third point under https://github.com/alphagov/styleguides/blob/master/ruby.md#classes

danielroseman:
One style comment, and the build has also failed on lint error. Otherwise :+1:

app/models/path_reservation.rb
```diff
@@ -1,17 +1,43 @@
...
+        raise existing.already_reserved_error
+      end
+    end
 
-    record
+    private
+
+    def create_path_reservation(base_path, publishing_app)
+      ActiveRecord::Base.transaction do
+        create!(base_path: base_path, publishing_app: publishing_app)
```
dougdroper: Is it necessary to wrap this in a transaction?

app/models/path_reservation.rb
```diff
@@ -1,17 +1,41 @@
...
+    ActiveRecord::Base.transaction do
+      create!(base_path: base_path, publishing_app: publishing_app)
+    end
   rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation
     # If a path is already reserved by the same publishing app, ignore the error
     record = find_by(base_path: base_path)
-    raise unless record.publishing_app == publishing_app
-
-    record
+    if record.already_associated_with?(publishing_app)
```
dougdroper: is there a way to remove this duplication with `reserve_base_path!`

cbaines:
For future reference, here is the [Git styleguide][1]. In particular, on this branch you don't always use the present tense, and don't always wrap to 72 characters.

[1]: https://github.com/alphagov/styleguides/blob/master/git.md

cbaines:
(I've only just realised that the styleguide includes Git, otherwise I would have pointed this out at the time)

app/models/path_reservation.rb
carlosmartinez: good point. resolved.

app/models/path_reservation.rb
carlosmartinez: The transaction is kind of the substance of the bug fix!

app/models/path_reservation.rb
carlosmartinez: Good spot. Done.

PR #500, opened 2016-09-07
Small performance improvement

carlosmartinez:
[Trello](https://trello.com/c/WMvRKsAt/792-publishing-api-performance-improvements-patchlinkset-2-days)

As part of a wider task to improve performance of Publishing API, do the PatchLinkSet deletes all in one go rather than one by one. It's a very small performance improvement, but has the advantage of also being a refactoring in this case.


boffbowsh:
👏

PR #502, opened 2016-09-08
Add requires_new to ActiveRecord to use savepoints

cbaines:
https://trello.com/c/CrtqBpmu/793-fix-bug-with-basepath-reservation-small


carlosmartinez:
approve21

(This is an interesting quirk; perhaps it's worth explaining in the commit message?)

cbaines:
I don't understand exactly why even without this change, the tests that cover this use savepoints. Saying that, I think that means that there is room for improvement in those tests.

PR #503, opened 2016-09-08
Don't filter content items by publishing app

carlosmartinez:
As per [Trello](https://trello.com/c/op0k4PYl/800-remove-permission-check-on-get-content-items-small)

This restriction is no longer required, and content-tagger in particular
needs to be able to see content items from all apps (in order to be able
to tag them to taxons).

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -56,9 +56,7 @@ module V2
     end
 
     def publishing_app
-      unless current_user.has_permission?('view_all')
-        current_user.app_name
-      end
+      query_params[:publishing_app]
     end
 
     def filters
--- a/spec/controllers/v2/content_items_controller_spec.rb
+++ b/spec/controllers/v2/content_items_controller_spec.rb
@@ -462,20 +462,20 @@ RSpec.describe V2::ContentItemsController do
       FactoryGirl.create(:draft_content_item, publishing_app: 'specialist_publisher', base_path: '/item3')
     end
 
-    it "displays items filtered by the user's app_name" do
-      request.env['warden'].user.app_name = 'whitehall'
-      get :index, document_type: 'guide', fields: %w(base_path publishing_app)
+    it "displays items filtered by publishing_app parameter" do
+      get :index,
+        document_type: "guide",
+        fields: %w(base_path publishing_app),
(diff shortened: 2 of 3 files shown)


kevindew:
This looks good to me 👍 I think it's best to check with the second part of the card before merging. I'll grab someone from specialist publisher to work out the best way to check.

carlosmartinez:
Thanks, @kevindew

kevindew:
This doesn't cause any problems with specialist publisher so it's looking good. Though we're going to look at making publishing_app one of the query string parameters so it can be set explicitly rather than implicitly.

carlosmartinez:
Updated as discussed.

kevindew:
Aces thanks - sorry you had to delete a lot a bunch of your changes.

Would you be able to add a line to the documentation for this method? https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content

app/controllers/v2/content_items_controller.rb
```diff
@@ -56,9 +56,7 @@ def content_item
     end
 
     def publishing_app
-      unless current_user.has_permission?('view_all')
-        current_user.app_name
-      end
+      query_params.fetch(:publishing_app, nil)
```
kevindew: I think this would be more consistent as `query_params[:publishing_app]`

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -462,20 +462,20 @@
...
-      get :index, document_type: 'guide', fields: %w(base_path publishing_app)
+    it "displays items filtered by publishing_app parameter" do
+      get :index,
+        document_type: "guide",
+        fields: %w(base_path publishing_app),
+        publishing_app: "whitehall"
       items = parsed_response["results"]
       expect(items.length).to eq(2)
-      expect(items.all? { |i| i["publishing_app"] == 'whitehall' }).to be true
+      expect(items.all? { |i| i["publishing_app"] == "whitehall" }).to be true
```
kevindew: 👍

carlosmartinez:
Aaaand documentation updated.

app/controllers/v2/content_items_controller.rb
carlosmartinez: your wish, my command

PR #504, opened 2016-09-12
Remove passthrough

dougdroper:
Passthrough hashes are not required any more since dependency resolution
does it now. The only thing that was using them was travel advice
documents types, and they have now been fixed in multipage frontend.

also there is no need to send links to content-store
Content-store uses expanded-links to display its links, and no longer
expands them. This removes the links from the payload.


cbaines:
Great :)

PR #505, opened 2016-09-12
Return error when payload schema validation fails

carlosmartinez:
Rather than allowing payloads that fail schema validation to be saved
and sent downstream, return a ValidationFailure instance to the
publishing app, containing details of the validation errors. This
will prompt the apps to fix any remaining validation problems.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -4,7 +4,7 @@ module Commands
       ITEM_NOT_FOUND = Class.new
       def call
         raise_if_links_is_provided
-        validate_schema
+        raise_if_payload_fails_schema_validation
 
         if publishing_app.blank?
           raise_command_error(422, "publishing_app is required", fields: {
@@ -304,8 +304,23 @@ module Commands
         )
       end
 
-      def validate_schema
-        SchemaValidator.new(type: :schema).validate(payload.except(:content_id))
+      def raise_if_payload_fails_schema_validation
+        return unless schema_validation_errors.any?
+        message = "The payload did not conform to the schema"
+        raise CommandError.new(
+          code: 422,
+          message: message,
+          error_details: schema_validation_errors,
+        )
+      end
+
+      def schema_validation_errors
+        @_validation_errors ||=
+          begin
+            validator = SchemaValidator.new(type: :schema)
+            validator.validate(payload.except(:content_id))
+            validator.errors
+          end
       end
     end
   end
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -29,7 +29,7 @@ private
   def schema
     @schema || JSON.load(File.read(schema_filepath))
   rescue Errno::ENOENT => error
-    msg = "#{payload} is missing schema_name #{schema_name} or type #{type}"
+    msg = "Unable to find schema for schema_name #{schema_name} and type #{type}"
     Airbrake.notify_or_ignore(error, parameters: { explanation: msg })
     {}
   end
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -19,53 +19,29 @@ class SchemaValidator
       payload,
       errors_as_objects: true,
     )
-
-    return true if errors.empty?
-
-    Airbrake.notify_or_ignore(
-      {
-        error_class: "SchemaValidationError",
-        error_message: "Error validating payload against schema '#{schema_name}'"
-      },
-      parameters: {
-        errors: errors_for_airbrake,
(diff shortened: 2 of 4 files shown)


kevindew:
👏 So this is shaping pretty good.

I think the `ValidationFailure` approach is going to cause some problems though. We currently catch instances of `CommandError` and convert that into a HTTP response: https://github.com/alphagov/publishing-api/blob/bb5284337c9cc04ff68d61cea60085c3d72951d0/app/controllers/application_controller.rb#L6. Therefore I think it'd make sense to use the command error approach (this is also used in a bunch of other validations across, so I think it'd be easier to make this consistent with those as if we introduce the new `ValidationFailure` class we'd probably want them to use that too)

Once we're rejecting items I think it'll be fine to not log them to errbit anymore as we'll be able to see 422's in the log and check against the event log. That was there more to know in advance what would error.

I don't know if it's realistic or not, but maybe we could have the `SchemaValidator` run within the model rather than as part of the input. This has the plus that all models we validate would have their schemas checked, only downside is these could fall invalid if the schema changes - but I think that's cool. In general we seem to have an approach of: validate in model if possible, otherwise validate outside it. As you can probably tell it's a bit of a mess though so it's something I'd like to see us improve in the not too distant future - but outside scope of here.

I know you'll be disappointed if I don't include a gif here so:
great job

app/commands/v2/put_content.rb
```diff
@@ -304,8 +307,13 @@ def raise_if_links_is_provided
         )
       end
 
-      def validate_schema
-        SchemaValidator.new(type: :schema).validate(payload.except(:content_id))
+      def schema_validation_errors
+        @_validation_errors ||=
+          begin
```
kevindew: does this need to be in a begin block?

app/commands/v2/put_content.rb
```diff
@@ -304,8 +307,13 @@ def raise_if_links_is_provided
         )
       end
 
-      def validate_schema
-        SchemaValidator.new(type: :schema).validate(payload.except(:content_id))
+      def schema_validation_errors
+        @_validation_errors ||=
```
kevindew: Is the underscore because you get a linting error otherwise or just to provide a clue that this ivar isn't used elsewhere? just flagging it as we've not got any other `@_` vars in the codebase

app/commands/v2/put_content.rb
carlosmartinez: i believe so, yes. otherwise it would assign the output of only the first of the three lines in the begin/end block.

app/commands/v2/put_content.rb
carlosmartinez: ah right. it's a reasonably widely-used convention that says "this is an instance variable that is only used to avoid repeating a service/db call, rather than representing some information that's relevant to the domain". but if it's not used elsewhere in the codebase, that might indicate that i should avoid it!

carlosmartinez:
Thanks @kevindew - will look into all that.

app/commands/v2/put_content.rb
kevindew: Ah I see, I thinking it was to swallow exceptions but that makes sense and as there is no `rescure` my idea does not.

carlosmartinez:
@kevindew - as discussed, have made this conform to the One True Path, and have removed the airbrake notification on schema validation failure.

app/validators/schema_validator.rb
```diff
@@ -12,58 +14,36 @@ def validate(payload)
...
-  end
-
   def schema
-    @schema || JSON.load(File.read("govuk-content-schemas/formats/#{schema_name}/publisher_v2/#{type}.json"))
+    @schema || JSON.load(File.read(schema_filepath))
   rescue Errno::ENOENT => error
-    Airbrake.notify_or_ignore(error, parameters: {
-      explanation: "#{payload} is missing schema_name #{schema_name} or type #{type}"
-    })
+    msg = "#{payload} is missing schema_name #{schema_name} or type #{type}"
```
kevindew: Sorry I know you've only moved the code here so pre-existing:

Pay

kevindew:
@carlosmartinez Aces! looks like we're just about there 😸

app/validators/schema_validator.rb
carlosmartinez: Duly updated. Because why not!

kevindew:
Great stuff 👍  - do we know if this is ok to merge or do we need to communicate with other teams first?

carlosmartinez:
Better not merge yet. Have raised a question on the Trello card as to how to proceed with the other teams.

PR #507, opened 2016-09-13
Fixes links migration

dougdroper:
this was throwing a `undefined method `assert_valid_keys' for "passthrough_hash IS NOT NULL":String`

there are only 1700 items so will be fast.


cbaines:
Please remember to use present tense, and include information (like you did in the Pull Request) in the commit message.

PR #508, opened 2016-09-13
remove migration

dougdroper:
this still isn't working, so removing for now until the app has been deployed, and will create a new one.


kevindew:
has db/schema.rb updated?

dougdroper:
there are a few migration after this one, so db/schema.rb is still pointing to the latest.

PR #510, opened 2016-09-13
Basics of govspeak implementation

kevindew:
This is the basics for introducing govspeak into the Publishing API. 

`a_content_item.details[:body]` is checked, if it has a content_type of "text/govspeak" and not one of "text/html" the govspeak will be rendered to HTML and be included in `details[:body]`.

As no apps are (or should be) sending us govspeak at the moment this won't have any actual affect until they do. 

This does not handle attachments, links, or contacts. Attachments will be looked at when we look at changing the rendering of specialist publisher.

Only area I was a little unsure about was how safe it was to just check `body` as the only place where "text/govspeak" could be available, as I noticed later on that this wasn't consistent with the `WellFormedContentTypesValidator`.


app/presenters/details_presenter.rb
```diff
@@ -0,0 +1,41 @@
+require "govspeak"
+
+module Presenters
+  class DetailsPresenter
+    attr_reader :content_item_details, :body
+
+    def initialize(content_item_details)
+      @content_item_details = SymbolizeJSON.symbolize(content_item_details)
+      @body = content_item_details[:body]
+      @body = [@body] if @body.is_a?(Hash)
```
carlosmartinez: does [Array::wrap](http://apidock.com/rails/Array/wrap/class) help here?

kevindew: Yes! I didn't know that existed, I've been getting caught out running `Array()` on hashes, it looks like that resolves just that. 

@dougdroper you might be stoked if you didn't know about this before - or maybe you know a downside I don't know?

dougdroper: No I didn't know about it either, probably because it is a rails method. good to know 👍

app/presenters/details_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+  private
+
+    def presented_details
+      return content_item_details unless can_render_govspeak?
+      govspeak = { content_type: "text/html", content: rendered_govspeak }
+      content_item_details.merge(body: body + [govspeak])
+    end
+
+    def can_render_govspeak?
+      return false unless body.respond_to?(:any?)
```
dougdroper: body respond to any? what's the reasoning behind that?

app/presenters/details_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+      has_html = body.any? { |format| format[:content_type] == "text/html" }
+      raw_govspeak.present? && !has_html
+    end
+
+    def raw_govspeak
+      return nil unless body.respond_to?(:find)
```
dougdroper: nil is redundant here

dougdroper: also respond to find?

app/presenters/details_presenter.rb
dougdroper: actually I think you can drop that line?

kevindew: We have instances of body as a string (definitely in tests not sure if in database), though this will become redundant with Array.wrap. Although in it's place we'll get a "no implicit conversion of Symbol into Integer"

app/presenters/details_presenter.rb
kevindew: whoops! thanks 👍

app/presenters/details_presenter.rb
dougdroper: maybe we can drop line 32 if we do that on this line, kill two birds with one stone type thing?

carlosmartinez:
@kevindew @dougdroper sorry guys, this needs another review - kevin pointed out yesterday that any field (not just `body`) could potentially contain govspeak. thankfully the changed requirement gave me a chance to simplify things a bit.

app/presenters/details_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+    attr_reader :content_item_details
+
+    def initialize(content_item_details)
+      @content_item_details = SymbolizeJSON.symbolize(content_item_details)
+    end
+
+    def details
+      @_details ||=
+        begin
+          content_item_details.inject({}) do |seed, hsh|
```
dougdroper: you could use `each_with_object` to avoid having to reset "seed" at the end. Also you can split key, val in the argument. `content_item_details.each_with_object({}) {|(key,val), seed| ... }`

app/presenters/details_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+      return value unless requires_govspeak_html_transform?(value)
+      govspeak = {
+        content_type: "text/html",
+        content: rendered_govspeak(value),
+      }
+      value + [govspeak]
+    end
+
+    def raw_govspeak(value)
+      value.find { |format| format[:content_type] == "text/govspeak" }[:content]
```
dougdroper: if find doesn't return anything will this still work?

carlosmartinez: no, it will blow up. but this method only gets called if `requires_govspeak_html_transform?` returns `true`, so `.find` should always return something.

PR #511, opened 2016-09-14
Make rendering_app optional for contacts

carlosmartinez:
Whitehall contacts are currently failing model validation because they
don't specify a rendering_app. Resolved so that the rendering_app
property is optional for ContentItems with a document_type of 'contact'

[Trello](https://trello.com/c/5itDAuZ5/801-fix-publishing-of-contacts-from-whitehall-medium)


kevindew:
Looks good 👍

PR #513, opened 2016-09-16
Add RabbitMQ production config

alexmuller:


boffbowsh:
I assume the env var is set in the environments? If so, 👍

alexmuller:
Not yet, was planning to set it if nobody was too offended by this pull request.

kevindew:
Does this mean we'll lose the magic file that's inserted as part of deployment? if so 🎉

alexmuller:
Once https://github.com/alphagov/govuk-puppet/pull/4876 is in production this can be merged.

alexmuller:
This is ready, you can verify it with:

```
fab staging class:backend do:'cat /etc/govuk/publishing-api/env.d/RABBITMQ_PASSWORD'
fab production class:backend do:'cat /etc/govuk/publishing-api/env.d/RABBITMQ_PASSWORD'
```

PR #517, opened 2016-09-20
Add GOVUK-Publishing-Source-Dependent-Content-Id header

cbaines:
This is for https://trello.com/c/04568g2y/796-view-request-path-more-easily-medium

Actually effecting the logging in the content store, and nginx will require changes there also.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem "gds-sso", "12.1.0"
 gem 'bunny', '2.5.1'
 gem 'whenever', '0.9.4', require: false
 gem "govuk_sidekiq", "~> 0.0"
+gem 'sidekiq', git: "https://github.com/alphagov/sidekiq.git", branch: "fix-testing-with-overridden-queue"
 gem "json-schema", require: false
 gem "hashdiff"
 gem "sidekiq-unique-jobs", require: false
--- a/app/workers/downstream_live_worker.rb
+++ b/app/workers/downstream_live_worker.rb
@@ -3,7 +3,7 @@ require 'sidekiq-unique-jobs'
 class DownstreamLiveWorker
   attr_reader :content_item_id, :web_content_item, :payload_version,
     :message_queue_update_type, :update_dependencies,
-    :alert_on_invalid_state_error
+    :alert_on_invalid_state_error, :dependency_resolution_source_content_id
 
   include DownstreamQueue
   include Sidekiq::Worker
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -20,6 +20,7 @@ gem "gds-sso", "12.1.0"
 gem 'bunny', '2.5.1'
 gem 'whenever', '0.9.4', require: false
 gem "govuk_sidekiq", "~> 0.0"
+gem 'sidekiq', git: "https://github.com/alphagov/sidekiq.git", branch: "fix-testing-with-overridden-queue"
```
dougdroper: Why is this pointing to a branch?

cbaines: I'm not sure what you mean?

I could have used a commit hash instead, which would have been more fixed, but I went with a more readable branch name instead.

app/workers/downstream_live_worker.rb
```diff
@@ -29,6 +29,12 @@ def perform(args = {})
       raise AbortWorkerError.new("The content item for id: #{content_item_id} was not found")
     end
 
+    unless source_dependent_content_id.nil?
+      DownstreamService.set_govuk_publishing_source_dependent_content_id_header(
+        @source_dependent_content_id
```
dougdroper: you have an attr_reader for this

Gemfile
dougdroper: No sorry, I mean why is the branch not merged into master? It also looks like it is behind? 
`This branch is 1 commit ahead, 81 commits behind mperham:master.`
Also what was the reason for not fixing this in mperham/sidekiq?

cbaines: I think having a specific branch is clearer than using master. The fix-testing-with-overridden-queue branch is based on the v4.1.4 tag, which is the version we are currently using in the Publishing API.This is why the branch is behind master, I just wanted to add this specific patch.

I created an issue in the mperham repository https://github.com/mperham/sidekiq/issues/3144

dougdroper: Ok, then should this go into govuk_sidekiq? Perhaps even upgrade to the latest release?

cbaines: It could go in govuk_sidekiq, but that would still require changes here, so for simplicity, I just made the change here.

As for upgrading sidekiq, personally, I prefer dependency changes to be kept separate from feature changes, unless they are actually related.

dougdroper: Except then no one else gets the benefit of the change :) 

Anyway lets take this offline.

cbaines: This change is now in Sidekiq master (https://github.com/mperham/sidekiq/pull/3152), and I'll keep an eye out for releases.

dougdroper: 👍

kevindew:
After a bunch of discussion and mixed opinions the general consensus seems to be that "GOVUK-Dependency-Resolution-Source-Content-Id" is the header that name that seems to tick a few of everyones boxes. 👍

cbaines:
The header name has now been changed to GOVUK-Dependency-Resolution-Source-Content-Id.

Had to rebase on master, as changes in govuk-content-schemas broke the tests for this branch.

PR #521, opened 2016-09-20
Update the publishing_app for Specialist Publisher manuals

steventux:
https://trello.com/c/2jXtF2jl/285-task-migrate-existing-sp-manuals-v1-in-the-publishing-api-small

Reviewers welcome!

Manuals will have their own publishing application so update
the relevant content items so that `publishing_app: "manuals-publisher"`.
Also update any relevant PathReservation records to the correct `publishing_app`.

### DIFF
--- a/db/migrate/20160920133337_update_publishing_app_of_manuals.rb
+++ b/db/migrate/20160920133337_update_publishing_app_of_manuals.rb
@@ -1,0 +1,19 @@
+class UpdatePublishingAppOfManuals < ActiveRecord::Migration
+  def change
+    specialist_publisher_manuals = ContentItem.where(
+      publishing_app: "specialist-publisher", document_type: ["manual", "manual_section"])
+
+    specialist_publisher_manuals.each do |manual|
+      # Find the Location of each specialist publisher content item
+      location = Location.find_by(content_item: manual)
+      # Find the PathReservation for that Location
+      path_reservation = PathReservation.find_by(publishing_app: "specialist-publisher",
+                                                 base_path: location.base_path)
+      # Update the PathReservation if it exists
+      path_reservation.update_attribute(:publishing_app, "manuals-publisher") if path_reservation
+    end
+
+    # Update the publishing_app for all specialist publisher manuals
+    specialist_publisher_manuals.update_all(publishing_app: "manuals-publisher")
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160913130622) do
+ActiveRecord::Schema.define(version: 20160920133337) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


cbaines:
I don't understand the why from the commit message?

This is putting false data in to the Publishing API, as there will be content that states that it was published by an application that did not exist at the time, which makes me uncertain about doing this.

steventux:
Having discussed this with @danielroseman we agreed that because the publishing application name is changing in for the _manuals publishing_ side of Specialist Publisher, the data within the publishing-api should reflect this as the origin. Given the new Manuals Publisher application will have a different `publishing_api_bearer_token`, it would filter out content with a `publishing_app` of `specialist-publisher` and so the manuals content would be invisible to Manuals Publisher, so sub-optimal.
Not sure what's meant by _false data_ but if you mean this should pass through an endpoint and the event log, I can't see how this would be achieved given `PathReservation` records also need to have their `publishing_app` updated in order to transfer ownership.
I'm also not certain there _is_ an endpoint which would allow us to change the `publishing_app` value of content.
There are ~2000 content items requiring this update (when Manuals Publisher is launched) so doing it quickly near the time of the application going into production is also desirable.

cbaines:
> Having discussed this with @danielroseman we agreed that because the publishing application name is changing in for the manuals publishing side of Specialist Publisher, the data within the publishing-api should reflect this as the origin.

In general, I prefer this kind of historical data to be unchanging, I'll try to elaborate more on the why below.

> Given the new Manuals Publisher application will have a different publishing_api_bearer_token, it would filter out content with a publishing_app of specialist-publisher and so the manuals content would be invisible to Manuals Publisher, so sub-optimal.

I don't think this is the case, as this was changed quite recently in d3e952d99773e4621b97ebeff47bb085a30c0a3f . It looks from the code like the publishing app is still used as a filter, but this doesn't have to be the case.

> Not sure what's meant by false data but if you mean this should pass through an endpoint and the event log, I can't see how this would be achieved given PathReservation records also need to have their publishing_app updated in order to transfer ownership.
> I'm also not certain there is an endpoint which would allow us to change the publishing_app value of content.
> There are ~2000 content items requiring this update (when Manuals Publisher is launched) so doing it quickly near the time of the application going into production is also desirable.

By false data, I mean that the `publishing_app` field in the database means something like the name of the application that made the request. As applications can change names, I view this as the name of the application at the time the request was made. So, under this interpretation, you get a contradiction if you change history in the Publishing API to say that some requests were made by an application that did not exist at the time.

As the core competency of the Publishing API is storing data, I think its important to think through things like this. My initial reaction is that the manuals publisher shouldn't care what Publishing App the content is related to, but instead just use the document_type, which I see as a more elegant solution than changing the history of ~2000 content items. Would doing something like this be possible?

cbaines:
Just talked to @steventux about this.

He made the point that as the name `specialist-publisher` is not uniquely identifying anyway, so even if the content items with a document_type of manual which were published by the original specialist publisher application are unchanged, once the original specialist publisher becomes the manuals publisher, keeping these as `publishing_app:specialist-publisher` will be contradictory, as the only Publishing App in use with the name specialist-publisher does not publish manuals.

I'm suitably convinced that the `publishing_app` field in the Publishing API is pretty vague in some cases (e.g. with the rebuild of specialist publisher), and that given this, its not worth trying to avoid changing it in history in this case.

cbaines:
The build failure might be fixed by just rebasing on master (as some schema things have changed).

PR #524, opened 2016-09-26
Include warnings when presenting content

cbaines:
This branch is for most of https://trello.com/c/CePxu8NA/809-include-warnings-in-payload-retrieved-from-publishing-api except the pact tests which I haven't looked in to yet.

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -3,7 +3,8 @@
 module Presenters
   module Queries
     class ContentItemPresenter
-      attr_accessor :scope, :fields, :order, :limit, :offset, :search_query
+      attr_accessor :scope, :fields, :order, :limit, :offset, :search_query,
+                    :include_warnings
 
       DEFAULT_FIELDS = [
         *ContentItem::TOP_LEVEL_FIELDS,
@@ -34,6 +35,7 @@ module Presenters
         self.limit = params[:limit]
         self.offset = params[:offset]
         self.search_query = params[:search_query]
+        self.include_warnings = params[:include_warnings] || false
       end
 
       def present_many
@@ -140,6 +142,8 @@ module Presenters
             int_columns.each { |c| parse_int_column(result, c) }
             parse_state_history(result)
 
+            result["warnings"] = get_warnings(result) if include_warnings
+
             yielder.yield result
           end
         end
@@ -159,6 +163,29 @@ module Presenters
         column = "state_history"
         return unless result.key?(column)
         result[column] = result[column].map(&:values).to_h
+      end
+
+      def get_warnings(result)
+        required_fields = %i{
+          content_id
+          state_history
+          user_facing_version
+          base_path
+          document_type
+        }
+
+        missing_fields = required_fields - fields
+
+        unless missing_fields.empty?
+          raise "#{missing_fields} must be included if the include_warnings parameter is true"
+        end
+
+        Presenters::Queries::ContentItemWarnings.call(
+          result["content_id"],
+          result["state_history"][result["user_facing_version"]],
+          result["base_path"],
+          result["document_type"],
+        )
       end
 
       # It is substantially faster to evaluate in this way rather than calling
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -22,10 +22,10 @@ module Presenters
         new(scope, params).present_many
       end
 
-      def self.present(content_item)
+      def self.present(content_item, params = {})
         scope = ContentItem.where(id: content_item.id)
 
-        present_many(scope).first
+        present_many(scope, params).first
       end
 
       def initialize(scope, params = {})
--- a/app/presenters/queries/content_item_warnings.rb
+++ b/app/presenters/queries/content_item_warnings.rb
@@ -1,0 +1,23 @@
+module Presenters
+  module Queries
+    module ContentItemWarnings
+      def self.call(content_id, state, base_path, document_type)
+        return unless state == "draft"
+
+        blocking_content_item_id = ::Queries::CheckForContentItemPreventingDraftFromBeingPublished.call(
+          content_id,
+          base_path,
+          document_type,
+        )
+
+        warnings = {}
+
+        if blocking_content_item_id
+          warnings["content_item_blocking_publish"] = "There is an item of content that prevents this draft from being published"
+        end
+
+        warnings
+      end
+    end
+  end
+end
--- a/app/queries/check_for_content_item_preventing_draft_from_being_published.rb
+++ b/app/queries/check_for_content_item_preventing_draft_from_being_published.rb
@@ -1,0 +1,51 @@
+module Queries
+  module CheckForContentItemPreventingDraftFromBeingPublished
+    extend ArelHelpers
+
+    # Checks for any content item which would prevent a content item with the
+    # specified content_id and base_path from being published (from the draft
+    # state).
+    def self.call(content_id, base_path, document_type)
+      if SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES.include?(document_type)
+        return # The SubstitionHelper will unpublish any item that is in the way
+      end
+
+      content_items_table = table(:content_items)
+      states_table = table(:states)
+      locations_table = table(:locations)
+      unpublishings_table = table(:unpublishings)
+
+      scope = content_items_table
+        .project(
+          content_items_table[:id]
+        )
+        .join(states_table).on(
+          content_items_table[:id].eq(states_table[:content_item_id])
+        )
+        .join(locations_table).on(
+          content_items_table[:id].eq(locations_table[:content_item_id])
+        )
+        .outer_join(unpublishings_table).on( # LEFT OUTER JOIN
+          content_items_table[:id].eq(unpublishings_table[:content_item_id])
+        )
+        .where(content_items_table[:content_id].not_eq(content_id))
+        .where(states_table[:name].in(%w(published unpublished)))
+        .where(content_items_table[:document_type].not_in(
+                 SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES
+        ))
+        .where(locations_table[:base_path].eq(base_path))
+        .where(
+          unpublishings_table[:type].not_eq("substitute")
+          .or(unpublishings_table[:type].eq(nil))
+        )
+
+      rows = get_rows(scope)
+
+      if rows.length > 1
+        raise "Multiple rows returned in CheckForContentItemPreventingDraftFromBeingPublished"
+      end
+
+      rows.first["id"].to_i if rows.first
+    end
+  end
+end
--- a/spec/queries/check_for_content_item_preventing_draft_from_being_published_spec.rb
+++ b/spec/queries/check_for_content_item_preventing_draft_from_being_published_spec.rb
@@ -1,0 +1,173 @@
+require "rails_helper"
+
+RSpec.describe Queries::CheckForContentItemPreventingDraftFromBeingPublished do
+  let(:content_id) { SecureRandom.uuid }
+  let(:base_path) { "/vat-rates" }
+  let(:document_type) { "guide" }
+
+  describe ".call" do
+    subject { described_class.call(content_id, base_path, document_type) }
+
+    shared_examples "check succeeds" do
+      it "returns nil" do
+        expect(subject).to eq(nil)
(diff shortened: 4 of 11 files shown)


app/presenters/queries/content_item_warnings.rb
```diff
@@ -0,0 +1,23 @@
+module Presenters
+  module Queries
+    module ContentItemWarnings
+      def self.call(content_id, state, base_path, document_type)
+        if state == "draft"
```
carlosmartinez: perhaps a case for getting this condition out the way (and thereby reducing reducing the number of clauses one has to hold in memory whilst reading the code):

`return unless state == "draft"`

app/presenters/queries/content_item_warnings.rb
```diff
@@ -0,0 +1,23 @@
...
+            document_type,
+          )
+
+          if blocking_content_item_id.nil?
+            nil
+          else
+            {
+              "content_item_blocking_publish" => "There is an item of content that prevents this draft from being published"
+            }
+          end
```
carlosmartinez: could you replace the if/else with an `unless blocking_content_item_id.nil?` - (as the method will return `nil` by default)

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
```diff
@@ -0,0 +1,53 @@
...
+        .where(content_items_table[:content_id].not_eq(content_id))
+        .where(states_table[:name].in(%w(published unpublished)))
+        .where(content_items_table[:document_type].not_in(
+                 SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES
+        ))
+        .where(locations_table[:base_path].eq(base_path))
+        .where(
+          unpublishings_table[:type].not_eq("substitute")
+          .or(unpublishings_table[:type].eq(nil))
+        )
```
carlosmartinez: impressive!

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
```diff
@@ -0,0 +1,53 @@
+module Queries
+  module CheckForContentItemPreventingDraftFromBeingPublished
+    extend ArelHelpers
+
+    # Checks for any content item which would prevent a content item with the
+    # specified content_id and base_path from being published (from the draft
+    # state).
+    def self.call(content_id, base_path, document_type)
+      if SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES.include?(document_type)
+        nil # The SubstitionHelper will unpublish any item that is in the way
```
dougdroper: I'm not sure I follow what this is doing?

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
```diff
@@ -0,0 +1,53 @@
...
+        .join(locations_table).on(
+          content_items_table[:id].eq(locations_table[:content_item_id])
+        )
+        .outer_join(unpublishings_table).on( # LEFT OUTER JOIN
+          content_items_table[:id].eq(unpublishings_table[:content_item_id])
+        )
+        .where(content_items_table[:content_id].not_eq(content_id))
+        .where(states_table[:name].in(%w(published unpublished)))
+        .where(content_items_table[:document_type].not_in(
+                 SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES
```
kevindew: I think this needs nudging a few spaces to the left.

spec/queries/check_for_content_item_preventing_draft_from_being_published_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      end
+
+      it "fails, returning the id of the content item" do
+        expect(subject).to eq(@blocking_content_item.id)
+      end
+    end
+
+    context "with a published item, and a draft at the same base path" do
+      before do
+        @blocking_content_item = FactoryGirl.create(:content_item,
```
kevindew: This looks a good candidate for using let and let!: http://betterspecs.org/#let

kevindew:
Haven't given this a thorough review but it's looking good to me 👍

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
cbaines: Do you mean what, or why?

Assuming you mean why, and not what, the publish command invokes the substitution helper, which will unpublish a "blocking" item if either the item being published, or the blocking item have a substitutable document type. This query is trying to work out if something can be published by detecting blocking items. To deal with the case where the substitution helper would be involved due to the blocking items document type, the query excludes content items with a substitutable document type. To deal with the case where the item being published has a substitutable document type, if the document type specified is substitutable, there is no point looking for blocking items, as if one does exist, it will be handled by the substitution helper. This if is that last check.

If this makes sense, I could add more of this information in to the code?

cbaines: Talked offline, the "what" is definitely wrong, as this should return nil, and currently it does nothing. I'll fix this, and add a test to cover this case.

Good spot Doug :)

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
cbaines: That is the only place on the line where the linter does not fail. I'm not really sure why.

spec/queries/check_for_content_item_preventing_draft_from_being_published_spec.rb
cbaines: I can't see any explanation on that page about why using "before do" is bad?

Using let! looks more complex to me, and as the setup for these examples are all about the database, it looks cleaner to me to just use "before do".

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
kevindew: That's a bit weird, bad linter 👎

spec/queries/check_for_content_item_preventing_draft_from_being_published_spec.rb
kevindew: `before` makes sense when you are performing an action for it's side effects, whereas let(!) is the predominant approach used across the test suite for setting variables that are used in tests.

app/presenters/queries/content_item_warnings.rb
cbaines: Done :)

app/presenters/queries/content_item_warnings.rb
cbaines: Done

kevindew:
Are you going to include the warnings as part of the response for `put_content`? From an API perspective it feels like that's a place they should belong - though it does feel that `show` is the place our apps would more likely care about them.

Also - I'm sure it's debatable - but it feels like it might be a more predictable API design if it always returned warnings field and it was just empty if there was nothing. Just feels like it'd be a more consistent API design if the field is always there rather than based on a conditional.

cbaines:
> Are you going to include the warnings as part of the response for put_content? From an API perspective it feels like that's a place they should belong - though it does feel that show is the place our apps would more likely care about them.

Yep, that's a different card [1] though, so I'll make that change (all several characters) in a separate branch once this one is merged.

1: https://trello.com/c/HW8dr2QM/810-warn-when-putting-a-draft-that-will-conflict-with-a-published-unpublished-item-occupying-a-base-path-medium

> Also - I'm sure it's debatable - but it feels like it might be a more predictable API design if it always returned warnings field and it was just empty if there was nothing. Just feels like it'd be a more consistent API design if the field is always there rather than based on a conditional.

Hmm, yeah, I'll put it in as an empty object. It might help people know that it is there.

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
```diff
@@ -0,0 +1,53 @@
+module Queries
+  module CheckForContentItemPreventingDraftFromBeingPublished
+    extend ArelHelpers
+
+    # Checks for any content item which would prevent a content item with the
+    # specified content_id and base_path from being published (from the draft
+    # state).
+    def self.call(content_id, base_path, document_type)
+      if SubstitutionHelper::SUBSTITUTABLE_DOCUMENT_TYPES.include?(document_type)
+        return nil # The SubstitionHelper will unpublish any item that is in the way
```
dougdroper: Not too important but nil is redundant here.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -143,6 +145,11 @@ def parse_results(results)
             int_columns.each { |c| parse_int_column(result, c) }
             parse_state_history(result)
 
+            if self.include_warnings
+              warnings = get_warnings(result)
+              result["warnings"] = warnings unless warnings.nil?
```
dougdroper: you could do `result["warnings"] = warnings if warnings`

app/presenters/queries/content_item_presenter.rb
```diff
@@ -164,6 +171,27 @@ def parse_state_history(result)
...
+      def get_warnings(result)
+        %i{
+          content_id
+          state_history
+          user_facing_version
+          base_path
+          document_type
+        }.each do |field|
+          unless self.fields.include?(field)
+            raise "The #{field} must be included if the include_warnings parameter is true"
```
dougdroper: Would this make sense to raise all the fields that must be included?

app/presenters/queries/content_item_warnings.rb
```diff
@@ -0,0 +1,21 @@
...
+      def self.call(content_id, state, base_path, document_type)
+        return unless state == "draft"
+
+        blocking_content_item_id = ::Queries::CheckForContentItemPreventingDraftFromBeingPublished.call(
+          content_id,
+          base_path,
+          document_type,
+        )
+
+        unless blocking_content_item_id.nil?
```
dougdroper: Again, I think would be easy to parse as `if blocking_content_item_id`

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+      rows = get_rows(scope)
+
+      if rows.length > 1
+        raise "Multiple rows returned in CheckForContentItemPreventingDraftFromBeingPublished"
+      end
+
+      if rows.first
+        rows.first["id"].to_i
```
dougdroper: `rows.first["id"].to_i if rows.first`

cbaines: Ok, changed.

app/presenters/queries/content_item_warnings.rb
cbaines: Yep, changed.

app/presenters/queries/content_item_presenter.rb
cbaines: Yep, that makes the code  a bit nicer also. Changed.

app/presenters/queries/content_item_presenter.rb
cbaines: I've changed this around now as a result of adding warnings in regardless of there being any.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -143,6 +145,11 @@ def parse_results(results)
             int_columns.each { |c| parse_int_column(result, c) }
             parse_state_history(result)
 
+            if self.include_warnings
```
cbaines: I've changed it to using the attr_accessor. Have a look at the code and tell me what you think? I'm unconvinced that there should be a method.

app/queries/check_for_content_item_preventing_draft_from_being_published.rb
cbaines: I've removed nil, but I've left it in the current form, as I quite like having the comment in the if.

cbaines:
@kevindew  After a lot of git faff, the warnings object is now always included if warnings are enabled.

@dougdroper Thanks for your comments, I've made some changes, both related and unrelated, so let me know what you think?

app/presenters/queries/content_item_presenter.rb
dougdroper: Yeah, the wrapper method is really just a convenience to indicate that it is a boolean, and is a convention, it just makes me pause when reading it. 
Although saying that, the style of having attr_accessors and assigning these instances to self is not very common so not really a big deal here.

cbaines:
Rebased to solve merge conflicts.

PR #530, opened 2016-09-29
Switch back to released Sidekiq.

danielroseman:
4.2.2 includes the fix we made for testing with overridden queue.

Also upgrade gds-api-adapters to fix some deprecation warnings from
RestClient.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,15 +12,14 @@ gem "pg"
 if ENV["API_DEV"]
   gem "gds-api-adapters", path: "../gds-api-adapters"
 else
-  gem "gds-api-adapters", "33.0.0"
+  gem "gds-api-adapters", "36.3.0"
 end
 
 gem "gds-sso", "12.1.0"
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -12,15 +12,15 @@ gem "pg"
...
+  gem "gds-api-adapters", "36.3.0"
 end
 
 gem "gds-sso", "12.1.0"
 
 gem 'bunny', '2.5.1'
 gem 'whenever', '0.9.4', require: false
 gem "govuk_sidekiq", "~> 0.0"
-gem 'sidekiq', git: "https://github.com/alphagov/sidekiq.git", branch: "fix-testing-with-overridden-queue"
+gem 'sidekiq', "4.2.2"
```
dougdroper: Do we know if we can update sidekiq in govuk_sidekiq?

danielroseman: Good point, I've done this instead - updated.

PR #531, opened 2016-09-29
[WIP] Enable use of gov.uk external campaign urls

ikennaokpala:
Related to changes in this PR https://github.com/alphagov/short-url-manager/pull/62
[Trello card](https://trello.com/c/YNWyi4dd/342-enable-campaign-redirects-in-publishing-api)

# Description 

Prior to this PR, external urls were not supported and validation
were put in place to negate them.

This PR enables the use of only gov.uk external campaign urls[1].

Within, external gov.uk campaign urls[1] have been validated with regex:

* possible campaign.service.gov.uk sub domain.
* scheme must either http or https.
* cannot start a hyphen.
* can be alphanumeric (or hyphen).
* number of characters range from 1-63. Refer to [2]

In theory [2], it is believed that DNS itself can have up to 127 levels
of label, each label can be up to 63 characters and the maximum length of
the whole record is limited to approximately 255 characters.

[1] Example: https://new-vat-rates.campaign.service.gov.uk/path/to/your/new/vat-rates
[2] https://en.wikipedia.org/wiki/Subdomain

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -147,11 +147,48 @@ private
 
   private
 
+    def acceptable_destination?(target)
+      target.starts_with?("/") || valid_govuk_campaign_url?(target)
+    end
+
+    def valid_govuk_campaign_url?(target)
+      uri = URI.parse(target)
+      host = uri.host
+      if host =~ /\A.+\.campaign\.gov\.uk\z/i &&
+          %w(http https).include?(uri.scheme)
+        label = host.split(".").first
+        label.present? && valid_subdomain?(label)
+      end
+    rescue
+      false
+    end
+
+    def valid_subdomain?(label)
+      valid_dns_label_range?(label) &&
+        starts_without_hyphen?(label) &&
+        contains_alphnumeric_or_hyphen?(label)
+    end
+
+    def valid_dns_label_range?(label)
+      (1..63) === label.length
+    end
+
+    def starts_without_hyphen?(label)
+      label =~ /\A[^-].*[^-]\z/i
+    end
+
+    def contains_alphnumeric_or_hyphen?(label)
+      label =~ /\A[a-z0-9\-]*\z/i
+    end
+
     def valid_exact_redirect_target?(target)
-      return false unless target.present? && target.starts_with?("/")
+      return false unless target.present? && acceptable_destination?(target)
 
       uri = URI.parse(target)
-      expected = uri.path
+      expected = ""
+      expected << "#{uri.scheme}://" if uri.scheme.present?
+      expected << uri.host if uri.host.present?
+      expected << uri.path if uri.path.present?
       expected << "?#{uri.query}" if uri.query.present?
       expected << "##{uri.fragment}" if uri.fragment.present?
       expected == target
--- a/spec/support/routes_and_redirects_validator.rb
+++ b/spec/support/routes_and_redirects_validator.rb
@@ -149,6 +149,35 @@ RSpec.shared_examples_for RoutesAndRedirectsValidator do
       expect { subject.valid? }.to_not raise_error
     end
 
+    context "when destination is external url" do
+      it "is invalid if it is not an external gov.uk campaign url" do
+        ["https://www.example.com/foo/bar", "https://www.gov.uk/foo/bar"].each do |destination|
+          content_item.redirects = [{ path: "#{subject.base_path}/foo", type: "exact", destination: destination }]
+
+          expect(subject).to be_invalid
+          expect(subject.errors[:redirects]).to eq(["is not a valid redirect destination"])
+        end
+      end
+
+      it "is invalid if the url is a malformed gov.uk campaign external url" do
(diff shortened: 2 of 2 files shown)


cbaines:
I can't see the Trello card?

ikennaokpala:
@cbaines I have added you to our trello board

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,37 @@ def validate(location, redirect)
 
   private
 
+    def acceptable_destination?(target)
+      target.starts_with?("/") || is_valid_govuk_campaign_url?(target)
```
dougdroper: I think I would drop the `is_` part of this method

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,37 @@ def validate(location, redirect)
...
+    end
+
+    def is_valid_govuk_campaign_url?(target)
+      uri = URI.parse(target)
+      host = uri.host
+      if (host =~ /\A.+\.campaign\.service\.gov\.uk\z/i) != nil &&
```
dougdroper: Do you need the != nil?

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,37 @@ def validate(location, redirect)
...
+          %w(http https).include?(uri.scheme)
+        label = host&.split(".")&.first
```
dougdroper: I am not sure in what scenario you wouldn't have a host at this point? Seems like the safe navigation is redundant? (and also possible a code smell :))

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,37 @@ def validate(location, redirect)
...
+        label.present? && is_subdomain?(label)
+      end
+    rescue
+      false
+    end
+
+    def is_subdomain?(label)
+      return false unless (1..63) === label.length
```
dougdroper: What is the significance of 63?

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,37 @@ def validate(location, redirect)
...
+      if (host =~ /\A.+\.campaign\.service\.gov\.uk\z/i) != nil &&
+          %w(http https).include?(uri.scheme)
+        label = host&.split(".")&.first
+        label.present? && is_subdomain?(label)
+      end
+    rescue
+      false
+    end
+
+    def is_subdomain?(label)
```
dougdroper: maybe `valid_subdomain?` rather then `is_`

app/validators/routes_and_redirects_validator.rb
ikennaokpala: 63 characters and the maximum length for a subdomain label.. Kindly read the description above

app/validators/routes_and_redirects_validator.rb
dougdroper: or you could use === if you want the boolean. 
https://ruby-doc.org/core-2.1.1/Regexp.html#method-i-3D-3D-3D

app/validators/routes_and_redirects_validator.rb
ikennaokpala: Yeah, sure will make this consistent with the other predicate methods

app/validators/routes_and_redirects_validator.rb
ikennaokpala: Well in the tests you will notice a set of malformed urls.. In this instance say `http://.campaign.gov.uk/path` The intention here is to split by dot which will return `["", "campaign", "gov", "uk"]`, so the first would not in this case suffice..

app/validators/routes_and_redirects_validator.rb
dougdroper: I mean can we make it more explicit. Without the context of the PR, I wouldn't know what the magic number 63 means. So you could pull the `(1..63)` into a method that explains that it is the DNS label range, or similar.

app/validators/routes_and_redirects_validator.rb
ikennaokpala: I have made amends to this.

app/validators/routes_and_redirects_validator.rb
ikennaokpala: ahh! ok.. will look into it

dougdroper:
Cool, looks good. 👍  Are you waiting for anything or can this be merged?

app/validators/routes_and_redirects_validator.rb
```diff
@@ -147,11 +147,48 @@ def validate(location, redirect)
...
 
+    def acceptable_destination?(target)
+      target.starts_with?("/") || valid_govuk_campaign_url?(target)
+    end
+
+    def valid_govuk_campaign_url?(target)
+      uri = URI.parse(target)
+      host = uri.host
+      if host =~ /\A.+\.campaign\.gov\.uk\z/i &&
+          %w(http https).include?(uri.scheme)
```
jennyd: Surely the campaigns platform will use HTTPS? We shouldn't be sending users to anything over unsecured HTTP if we can possibly avoid it.

ikennaokpala: Based on our inquiries.. it appears there might be a few http exceptions..

PR #533, opened 2016-09-30
Upgrade to Rails 5

danielroseman:
Mostly configuration and deprecation changes.

The biggest code change is because Arel now requires all columns to be typed. For the most part this means changing from `Arel::Table.new(:content_items)` to `ContentItem.arel_table`, but that also means all the arel queries will now be returning integers and bools rather than strings.

Rebuilding the gemfile also brought in a newer version of govuk-lint, which caused a whole load of lint errors which are fixed here.


app/presenters/queries/expanded_link_set.rb
```diff
@@ -52,7 +52,7 @@ def expand_links(links)
         links.map do |link|
           item = all_web_content_items[link["target_content_id"]]
           next unless item
-          next_level = expand_level(link["target_content_id"]) if link['cycle'] == 'f'
+          next_level = expand_level(link["target_content_id"]) unless link['cycle']
```
dougdroper: I am surprised this works, since link['cycle'] returns 't' or 'f'. Or is this a change in Rails 5 where it would now return a boolean?

danielroseman: See the comments on df00cee - all Arel columns are now typed.

PR #536, opened 2016-09-30
Use attachments in Govspeak

kevindew:
This adds attachments into our Govspeak rendering and ups our version of Govspeak to 5.0.

With this we should be set up for Specialist Publisher Govspeak rendering.


Gemfile.lock
```diff
@@ -418,4 +422,4 @@ DEPENDENCIES
   whenever (= 0.9.4)
 
 BUNDLED WITH
-   1.11.2
+   1.12.5
```
dougdroper: Does this match the bundler version on ci?

h-lame: It doesn't match the version defined in puppet: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_rbenv/manifests/all.pp#L64-L66 so it'll be a pain for those using the dev VM.

kevindew: I'd expect it's still 1.11.2 actually - that's what it is on integration. I'll fix it. 

Does this cause more pain than just a warning? I reckon I just updated my bundler at some point for a different repo.

h-lame: It might mean you end up with a permanently dirty tree if bundler decides to overwrite the BUNDLED_WITH and that means the update_git / update_all scripts in the development repo don't do their thing.  I mean, it's a minor annoyance for sure, but still annoying.

kevindew: Cool, thanks for explaining. I'll keep a closer eye on it in future 👍

kevindew: @dougdroper this is sorted now.

PR #538, opened 2016-10-04
Change handling of links to preserve ordering

rubenarakelyan:
**NOTE** To be reviewed by the Publishing API team.

Currently, the `PATCH link_sets` command works out the difference between the provided links payload and the links currently in the database for the link set concerned. Links are then created and/or deleted in the database to match the payload. This process does not preserve ordering, so new links are appended at the end of the link set.

This commit changes the command to delete all existing links in the database for the link set, and re-add all the links from the payload, thereby preserving the natural ordering in the payload. In addition, the ordering is also recorded in a new `position` field in the `links` table, to allow the ordering algorithm to be changed in future without affecting natural ordering.

All existing links in the database are given a `position` of zero to reflect the fact that they have no inherent order.

Trello: https://trello.com/c/LlO4fTIS/153-make-publishing-api-order-the-links-size-3-mazz

cc @kevindew

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -18,17 +18,14 @@ module Commands
         lock_version.save!
 
         grouped_links.each do |group, payload_content_ids|
-          links = link_set.links.where(link_type: group)
-          existing_content_ids = links.pluck(:target_content_id)
+          # For each set of links in a LinkSet scoped by link_type, this iterator
+          # deletes the entire existing set and then imports all the links in the
+          # payload, preserving their ordering.
+          link_set.links.where(link_type: group).delete_all
 
-          content_ids_to_create = payload_content_ids - existing_content_ids
-          content_ids_to_delete = existing_content_ids - payload_content_ids
-
-          content_ids_to_create.uniq.each do |content_id|
-            links.create!(target_content_id: content_id)
+          payload_content_ids.uniq.each_with_index do |content_id, i|
+            link_set.links.create!(target_content_id: content_id, link_type: group, position: i)
           end
-
-          links.where(target_content_id: content_ids_to_delete).delete_all
         end
 
         after_transaction_commit do
--- a/db/migrate/20161003145646_add_position_to_links.rb
+++ b/db/migrate/20161003145646_add_position_to_links.rb
@@ -1,0 +1,5 @@
+class AddPositionToLinks < ActiveRecord::Migration
+  def change
+    add_column :links, :position, :integer
+  end
+end
--- a/db/migrate/20161006115515_add_position_to_existing_links.rb
+++ b/db/migrate/20161006115515_add_position_to_existing_links.rb
@@ -1,0 +1,5 @@
+class AddPositionToExistingLinks < ActiveRecord::Migration
+  def change
+    Link.where(position: nil).update_all(position: 0)
+  end
+end
--- a/app/models/link_set.rb
+++ b/app/models/link_set.rb
@@ -1,3 +1,3 @@
 class LinkSet < ActiveRecord::Base
-  has_many :links, -> { order(id: :asc) }, dependent: :destroy
+  has_many :links, -> { order(link_type: :asc, position: :asc) }, dependent: :destroy
 end
--- a/db/migrate/20161006115850_make_position_required.rb
+++ b/db/migrate/20161006115850_make_position_required.rb
@@ -1,0 +1,5 @@
+class MakePositionRequired < ActiveRecord::Migration
+  def change
+    change_column :links, :position, :integer, null: false, default: 0
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160913130622) do
+ActiveRecord::Schema.define(version: 20161006115850) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -91,9 +91,10 @@ ActiveRecord::Schema.define(version: 20160913130622) do
   create_table "links", force: :cascade do |t|
     t.integer  "link_set_id"
     t.string   "target_content_id"
-    t.string   "link_type",         null: false
-    t.datetime "created_at",        null: false
-    t.datetime "updated_at",        null: false
+    t.string   "link_type",                     null: false
+    t.datetime "created_at",                    null: false
+    t.datetime "updated_at",                    null: false
+    t.integer  "position",          default: 0, null: false
   end
 
   add_index "links", ["link_set_id", "target_content_id"], name: "index_links_on_link_set_id_and_target_content_id", using: :btree
--- a/doc/api.md
+++ b/doc/api.md
@@ -27,7 +27,7 @@ for other apps (eg email-alert-service) to consume.
 ### Optimistic locking (`previous_version`)
 
 All PUT and POST endpoints take an optional JSON attribute `previous_version`
-in the request. If given, the coresponding value should be a integer. This
+in the request. If given, the corresponding value should be a integer. This
 allows the Publishing API to check that the publishing app sending the request
 intends to update the latest lock version of the model in question.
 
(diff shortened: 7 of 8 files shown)


db/migrate/20161003145646_add_position_to_links.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToLinks < ActiveRecord::Migration
+  def change
+    add_column :links, :position, :integer
```
kevindew: If we were to set this as non-null with a default of 0 could we drop the other migration?

kevindew:
Looks good, nicely done 👍 

I'll have a chat with our team about the idea of using activerecord-import and where we collectively stand on less queries vs validations/callbacks running.

db/migrate/20161003145646_add_position_to_links.rb
rubenarakelyan: Too right, done

rubenarakelyan:
@kevindew I think both of those can be done to some extent:

* Validations can be run by amending the import code (https://github.com/zdennis/activerecord-import/wiki/Examples)

* Callbacks can be run *if* we get an in-memory collection of ActiveRecord objects (https://github.com/zdennis/activerecord-import/wiki/Callbacks)

app/commands/v2/patch_link_set.rb
```diff
@@ -18,17 +18,25 @@ def call
...
+          # For each set of links in a LinkSet scoped by link_type, this iterator
+          # deletes the entire existing set and then imports all the links in the
+          # payload, preserving their ordering. The activerecord-import gem is
+          # used to create a single SQL statement for importing all links.
           links = link_set.links.where(link_type: group)
           existing_content_ids = links.pluck(:target_content_id)
 
-          content_ids_to_create = payload_content_ids - existing_content_ids
-          content_ids_to_delete = existing_content_ids - payload_content_ids
+          links.where(target_content_id: existing_content_ids).delete_all
```
rubenarakelyan: Mainly for two reasons:

1. Simpler code - the alternative would include working out which links to add/delete (as happens currently), then iterating over the links to update their positions in the database. That makes the code more complicated and adds more iteration.

2. Less database I/O - deleting and batch importing guarantees two SQL statements for every operation (one `DELETE` and one `INSERT`), whereas the alternative would execute more statements in many cases (one `DELETE` or `INSERT` per change plus `UPDATE`s and `SELECT`s for updating link positions).

dougdroper: I agree with the simpler code, although this way we lose the created_at timestamp (not sure that is an issue?)

rubenarakelyan: Yup, that's one downside to this approach. I checked this with @tijmenb and the view is this shouldn't be a problem since we don't rely on it for anything on our side. Not sure if you use that field elsewhere.

app/commands/v2/patch_link_set.rb
```diff
@@ -18,17 +18,25 @@ def call
...
+            links_to_create << Link.new(
+              link_set_id: link_set.id,
+              target_content_id: content_id,
+              link_type: group,
+              position: i
+            )
           end
-
-          links.where(target_content_id: content_ids_to_delete).delete_all
+          Link.import links_to_create
```
dougdroper: I am not sure about using activerecord-import. Our average count of links per group is only 1.23, and in worst case (at the moment) there are 813 links, which I am sure the DB can handle quite quickly. Seems like adding another dependency a bit early?

dougdroper: Also to @kevindew point about missing out on the Link validations.

rubenarakelyan: I don't have any problem with reverting to the standard "one insert per row" if you don't think it will case any noticeable performance issues. Maybe we should benchmark before/after?

dougdroper: Yes I agree, lets leave it out for now, then do some performance bench marks to see if this is an issue.

dougdroper:
@rubenarakelyan looks good to me now, can we get one or two tests just to cover the position logic. Also can we update the docs?

rubenarakelyan:
@dougdroper I've updated the `api.md` file and also added a specific test for making sure the order is correct. I also updated an existing order to use `eq` instead of `match_array` so that the ordering is also checked.

kevindew:
Can you update the ordering in the linkset model too: https://github.com/alphagov/publishing-api/blob/e4a5ad165a7917441b598e0817d155154e4a14b8/app/models/link_set.rb#L2 ?

dougdroper:
Is there any more work to do with the rendering to the content store? The expanded link set doesn't take into account any of the ordering (although it may be fine). Just wanted to know if that is part of this work.

rubenarakelyan:
@kevindew I've updated the `link_set` model to order by `link_type` and `position`. This is to keep links of the same type grouped together since the `position` is scoped to `link_set_id` then `link_type`.

@dougdroper Yes from discussing with @tijmenb I believe there is some work to do on the ordering of links when they are sent from the API to downstream consumers such as the Content Store. I think I'll need some more guidance on this though to work out what I need to change.

app/commands/v2/patch_link_set.rb
MatMoore: I agree with this approach but just wanted to note that there is still a separate insert for each link at the moment. So I'd expect this to perform roughly the same in the case where you're inserting a set of links for the first time, but if the user makes minor changes (adding a new tag to a long list) you lose the optimisation of just inserting the one you changed. 

That could maybe be remedied by inserting all the rows in one sql statement but I'm not sure you can do it with plain activerecord.

It might be worth checking for noops before doing anything, because publishing apps tend to send all their links when publishing a draft.

Have you run any of the benchmarks yet?

kevindew: @MatMoore We (Publishing API) persuaded Ruben to drop the single insert to multiple ones so that's our fault 😉  We felt it was best to wait until there was performance issues.

@rubenarakelyan I've had a look at the above and I think you can save some a query on each group iteration by dropping the need to get `existing_content_ids`:

```ruby
          link_set.where(link_type: group).delete_all
          payload_content_ids.uniq.each_with_index do |content_id, i|
            links.create!(link_type: group, target_content_id: content_id, position: i)
          end
```

I also dug in a little to see if it would be difficult to go down the update and no-op route if nothing has changed and think this isn't too complex:

```ruby
          links = link_set.links.where(link_type: group)
          existing_content_ids = links.pluck(:target_content_id)
          payload_content_ids.uniq.each_with_index do |content_id, position|
            next unless content_id == existing_content_ids[postion]
            links.find_or_create_by!(target_content_id: content_id).update!(position: position)
          end
          to_delete = existing_content_ids - payload_content_ids
          links.where(target_content_id: to_delete).delete_all unless to_delete.empty?
```

kevindew:
@rubenarakelyan great thanks for that.

Yeah I think you might have to dig into the [fun recursive query](https://github.com/alphagov/publishing-api/blob/master/app/presenters/queries/expanded_link_set.rb#L117) for the downstream ordering which does sound well suited to a separate PR.

kevindew:
Oh another thing, not really a concern for you @rubenarakelyan but more not for Publishing Platform team. 

I ran this migration locally and it took about 44seconds. Is this something we'll want to try deploy outside of a busy period? I think ALTER TABLE causes a write lock but I'm not certain.

db/migrate/20161003145646_add_position_to_links.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToLinks < ActiveRecord::Migration
+  def change
+    add_column :links, :position, :integer, null: false, default: 0
```
boffbowsh: The reason this takes so long is the default setting. It'd be better to do the following:

1. Add the column, allowing nulls with no default
2. `UPDATE "links" SET "position" = 0 WHERE "position" IS NULL;`
3. Alter the column and add the default

That way, PostgreSQL won't block at all.

kevindew: ha oh dear - that was how it was wrote originally before I intervened.

Thanks @boffbowsh 👍  - Sorry @rubenarakelyan

rubenarakelyan: No worries, I've changed this to do the migration in three steps.

db/migrate/20161006115515_add_position_to_existing_links.rb
```diff
@@ -0,0 +1,7 @@
+class AddPositionToExistingLinks < ActiveRecord::Migration
+  def change
+    Link.where("position IS NULL").each do |link|
```
boffbowsh:
```ruby
Link.where(position: nil).update_all(position: 0)
```

rubenarakelyan: And it was changed

PR #548, opened 2016-10-10
Use normal ruby recursion for link traversal

dougdroper:
The speed gains of using the recursive SQL doesn't out way the cost of
understanding the query, there are a few more queries to the DB, but it
is easier to reason about in the code, and is still calculated in roughly
the same time.

We also need to split the rules for going down the chain, and going up
the chain. For instance documents we want document_collections, but we
don't want to recurse on those document documents.

before:
..........  3.920000   0.150000   4.070000 (  7.515828)
queries: 165

..........  2.200000   0.060000   2.260000 (  3.754221)
queries: 120

..........  0.080000   0.010000   0.090000 (  0.113228)
queries: 100

..........  0.160000   0.010000   0.170000 (  0.302323)
queries: 160

after:
..........  3.920000   0.160000   4.080000 (  8.682721)
queries: 175

..........  2.730000   0.110000   2.840000 (  3.641704)
queries: 160

..........  0.070000   0.000000   0.070000 (  0.106539)
queries: 100

..........  0.140000   0.020000   0.160000 (  0.274945)
queries: 160

### DIFF
--- a/app/presenters/queries/expanded_link_set.rb
+++ b/app/presenters/queries/expanded_link_set.rb
@@ -5,6 +5,7 @@ module Presenters
         @content_id = content_id
         @state_fallback_order = Array(state_fallback_order)
         @locale_fallback_order = Array(locale_fallback_order)
+        @visited = []
       end
 
       def links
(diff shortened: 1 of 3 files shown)


boffbowsh:
I think this looks ok, but would appreciate another review from someone else due to the complexity

app/presenters/queries/expanded_link_set.rb
```diff
@@ -13,54 +14,61 @@ def links
...
-      def parents
-        @parents ||= all_parents.group_by { |e| e["parent_id"] }
-      end
-
-      def all_parents
-        @all_parents ||= recursive_query(recursive_link_types.map { |t| "'#{t}'" }.join(','))
+      def children(content_id, type = nil)
+        visited << content_id
+        links = all_links(content_id, type)
+        all_web_content_items = all_web_content_items(links)
```
dougdroper: updated

PR #550, opened 2016-10-10
Replace linkables with query

dougdroper:
Linkables have proven to be difficult to keep in sync with changes to
state in the system. This removes the code to create and update them and
uses a query to retrieve the same information.

Although not as fast as linkables, this removes the complication in
attempting to keep the Linkables table up to date. Since it is not as
fast as linkables, we cache the response.

old
topic
..........  0.060000   0.010000   0.070000 (  0.169602)
queries: 12

organisation
..........  0.090000   0.010000   0.100000 (  0.149797)
queries: 10

mainstream_browse_page
..........  0.030000   0.000000   0.030000 (  0.062847)
queries: 10

policy
..........  0.040000   0.010000   0.050000 (  0.159913)
queries: 10

taxon
..........  0.030000   0.000000   0.030000 (  0.034028)
queries: 10

new (warm cache)
topic
..........  0.040000   0.000000   0.040000 (  0.066904)
queries: 10

organisation
..........  0.060000   0.010000   0.070000 (  0.082757)
queries: 10

mainstream_browse_page
..........  0.020000   0.010000   0.030000 (  0.256291)
queries: 10

policy
..........  0.020000   0.000000   0.020000 (  0.058285)
queries: 10

taxon
..........  0.020000   0.010000   0.030000 (  0.180724)
queries: 10


app/queries/get_linkables.rb
```diff
@@ -1,17 +1,37 @@
...
-      Linkable
+      documents = ContentItem
         .where(document_type: [document_type, "placeholder_#{document_type}"])
-        .includes(:content_item)
-        .pluck(:content_id, :state, :title, :base_path, "content_items.details->>'internal_name' as internal_name")
-        .map { |linkable_data|
-          Presenters::Queries::LinkablePresenter.present(*linkable_data)
-        }
+        .order('updated_at ASC')
+        .pluck(:updated_at, :content_id)
```
boffbowsh: It surprises me that this is quicker than `SELECT MAX(updated_at) ...` but it really is. Even with a new index.

```psql
publishing_api_development=> CREATE INDEX idx_temp_dt_ua ON content_items (updated_at, document_type);
CREATE INDEX
Time: 6899.011 ms
publishing_api_development=> EXPLAIN SELECT MAX(updated_at) FROM content_items WHERE document_type = 'placeholder_policy_area';
                                                      QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------
 Result  (cost=20.85..20.86 rows=1 width=0)
   InitPlan 1 (returns $0)
     ->  Limit  (cost=0.42..20.85 rows=1 width=8)
           ->  Index Only Scan Backward using idx_temp_dt_ua on content_items  (cost=0.42..38387.41 rows=1879 width=8)
                 Index Cond: ((updated_at IS NOT NULL) AND (document_type = 'placeholder_policy_area'::text))
(5 rows)
publishing_api_development=> SELECT updated_at, content_id FROM content_items WHERE document_type = 'placeholder_policy_area' ORDER BY updated_at ASC;
Time: 0.801 ms
publishing_api_development=> SELECT MAX(updated_at) FROM content_items WHERE document_type = 'placeholder_policy_area';
            max
----------------------------
 2016-05-03 08:04:58.012402
(1 row)

Time: 14.633 ms
```

boffbowsh:
Were those benchmarks taken with caching enabled or disabled?

app/queries/get_web_content_items.rb
```diff
@@ -2,12 +2,12 @@ module Queries
...
     extend ArelHelpers
 
-    def self.call(content_item_ids)
+    def self.call(content_item_ids, presenter = WebContentItem)
       content_items = table(:content_items)
       filtered = scope
         .where(content_items[:id].in(content_item_ids))
       get_rows(filtered).map do |row|
-        WebContentItem.from_hash(row)
+        presenter.from_hash(row)
```
boffbowsh: 👏

dougdroper:
those benchmarks were with caching enabled, and warm, which seems to be fair since most of these calls to this endpoint will be cached.

boffbowsh:
Ok. That's a fair bit slower than before so we'll want to check it in the real world too. Can you make sure we have a kibana/grafana dashboard ready to track this during deployment please?

dougdroper:
Yes, will measure its performance

.rubocop.yml
```diff
@@ -1,3 +1,6 @@
+AllCops:
+  TargetRubyVersion: 2.3
```
kevindew: Yay! I think this will fix some nags I've been getting 👍

app/queries/get_linkables.rb
```diff
@@ -1,17 +1,37 @@
...
+
+      Rails.cache.fetch ["linkables", document_type, documents.last.first] do
+        Queries::GetWebContentItems.(
```
tijmenb: TIL `Class.(args)` sorcery. Is this something we do elsewhere?

config/environments/test.rb
```diff
@@ -28,6 +28,9 @@
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
+  # Disable cache in test
+  config.cache_store = :null_store
```
tijmenb: Perhaps it's better to enable caching in test & development. There's a risk that there are cache-invalidation bugs that would only surface on production otherwise.

app/queries/get_linkables.rb
dougdroper: I am not sure we do it? but it is the benefit of `#call`

boffbowsh: I always write `#call` like this. It's a Ruby 2.2 thing IIRC?

config/environments/test.rb
boffbowsh: Could just use [`:memory_store`](http://api.rubyonrails.org/classes/ActiveSupport/Cache/MemoryStore.html) ?

tijmenb: Yup, that'll be best I think.

app/queries/get_linkables.rb
dougdroper: looking into making this faster

app/presenters/message_queue_presenter.rb
```diff
@@ -4,11 +4,11 @@ def self.present(downstream_presenter, update_type:)
       attributes = downstream_presenter.present
       # FIXME: This only happens in tests. We need to fix the tests so that
       # `link_set` is always available.
-      if downstream_presenter.link_set
-        links = Presenters::Queries::LinkSetPresenter.new(downstream_presenter.link_set).links
-      else
-        links = {}
-      end
+      links = if downstream_presenter.link_set
```
danielroseman: There's been quite a lot of discussion over removing the cop that does this (Style/ConditionalAssignment). I think it hurts readability and looks ugly; we should probably disable it ourselves in rubocop.yml and revert the ones that got "fixed" here.

dougdroper: Agree, done

PR #552, opened 2016-10-11
Add ChangeNote to manage ContentItem change history

carlosmartinez:
[Trello](https://trello.com/c/PUldwy8A/823-public-content-history-medium)

We don't want the publishing apps to have to include change history as
part of the content item PUT request any more. Therefore we are moving
to a model whereby we save a ChangeNote for every such request.
Ultimately this will be a top level property in the schemas, but for the
moment we will continue to support `change_note` as part of the details
hash.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -22,6 +22,8 @@ module Commands
           fill_out_new_content_item(content_item)
         end
 
+        ChangeNote.create_from_content_item(payload, content_item)
+
         after_transaction_commit do
           send_downstream(content_item.content_id, locale)
         end
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -1,0 +1,57 @@
+class ChangeNote < ActiveRecord::Base
+  belongs_to :content_item
+
+  def self.create_from_content_item(payload, content_item)
+    ChangeNoteFactory.new(payload, content_item).build
+  end
+end
+
+class ChangeNoteFactory
+  def initialize(payload, content_item)
+    @payload = payload
+    @content_item = content_item
+  end
+
+  def build
+    return unless content_item.update_type == "major"
+    create_from_top_level_change_note ||
+      create_from_details_hash_change_note ||
+      create_from_details_hash_change_history
+  end
+
+private
+
+  attr_reader :payload, :content_item
+
+  def create_from_top_level_change_note
+    return unless change_note
+    ChangeNote.create!(change_note.merge(content_item: content_item))
+  end
+
+  def create_from_details_hash_change_note
+    return unless note
+    ChangeNote.create!(
+      content_item: content_item,
+      public_timestamp: content_item.updated_at,
+      note: note,
+    )
+  end
+
+  def create_from_details_hash_change_history
+    return unless history
+    history_element = history.sort_by { |h| h[:public_timestamp] }.last
+    ChangeNote.create!(history_element.merge(content_item: content_item))
+  end
+
+  def change_note
+    payload[:change_note]
+  end
+
+  def note
+    content_item.details[:change_note]
+  end
+
+  def history
+    content_item.details[:change_history]
+  end
+end
--- a/app/presenters/change_history_presenter.rb
+++ b/app/presenters/change_history_presenter.rb
@@ -1,0 +1,23 @@
+module Presenters
+  class ChangeHistoryPresenter
+    attr_reader :content_item
+
+    def initialize(content_item)
+      @content_item = content_item
+    end
+
+    def change_history
+      content_item.details[:change_history] || change_notes_for_content_item
+    end
+
+  private
+
+    def change_notes_for_content_item
+      ChangeNote
+        .joins(:content_item)
+        .where("content_items.content_id" => content_item.content_id)
+        .order(public_timestamp: :desc)
+        .map { |cn| cn.slice(:note, :public_timestamp) }
+    end
+  end
+end
--- a/app/presenters/change_history_presenter.rb
+++ b/app/presenters/change_history_presenter.rb
@@ -16,8 +16,17 @@ module Presenters
       ChangeNote
         .joins(:content_item)
         .where("content_items.content_id" => content_item.content_id)
+        .joins("INNER JOIN user_facing_versions ON user_facing_versions.content_item_id = content_items.id")
+        .where("user_facing_versions.number <= ?", version_number)
         .order(public_timestamp: :desc)
-        .map { |cn| cn.slice(:note, :public_timestamp) }
+        .pluck(:note, :public_timestamp)
+        .map do |note, timestamp|
+          { note: note, public_timestamp: timestamp }.stringify_keys
+        end
+    end
+
+    def version_number
+      UserFacingVersion.where(content_item_id: content_item.id).last.number
     end
   end
 end
--- a/app/presenters/details_presenter.rb
+++ b/app/presenters/details_presenter.rb
@@ -2,22 +2,29 @@ require "govspeak"
 
 module Presenters
   class DetailsPresenter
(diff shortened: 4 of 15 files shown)


app/models/content_item.rb
```diff
@@ -28,6 +28,8 @@ class ContentItem < ActiveRecord::Base
   NON_RENDERABLE_FORMATS = %w(redirect gone).freeze
   EMPTY_BASE_PATH_FORMATS = %w(contact government).freeze
 
+  has_many :change_notes
```
dougdroper: There has been quite a bit of work to avoid content item "knowing" about other objects. see https://github.com/alphagov/publishing-api/blob/master/doc/model.md#general-themes

It does mean that you would have to do a bit of a dance to work this in the same way, so might be something we want to discuss?

db/migrate/20161007080213_create_change_notes.rb
```diff
@@ -0,0 +1,9 @@
+class CreateChangeNotes < ActiveRecord::Migration
+  def change
+    create_table :change_notes do |t|
+      t.string :note
+      t.datetime :public_timestamp
```
dougdroper: Do we care about updated_at/created_at timestamps?

app/commands/v2/put_content.rb
```diff
@@ -22,6 +22,9 @@ def call
           fill_out_new_content_item(content_item)
         end
 
+        change_note = payload.fetch(:change_note, nil)
```
dougdroper: no need for a fetch here? `payload[:change_note]` does the same?

dougdroper:
Do we need to create one for publish?

kevindew:
Good work 👍 

I'd have thought that it needs expanding/coupling more between version numbers and `update_type` for the change history. As I understand we only want the change history to reflect update_types of "major" (at least the one that is public, there could be a need for a public / non public change history) and each entry in this list would be associated with a version number (thus you could then access that particular version via the Publishing API)

app/models/content_item.rb
carlosmartinez: thanks, i'll look into that

app/commands/v2/put_content.rb
carlosmartinez: yes, will change

app/models/change_note.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def self.create_from_content_item(payload, content_item)
+    return unless content_item.update_type == "major"
+    create_from_top_level_change_note(payload, content_item) ||
+      create_from_details_hash_change_note(content_item) ||
+      create_from_details_hash_change_history(content_item)
+  end
+
+  def self.create_from_top_level_change_note(payload, content_item)
+    return nil unless (change_note = payload[:change_note])
```
dougdroper: Just a redundant nil in this method and the ones below.

db/migrate/20161007080213_create_change_notes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateChangeNotes < ActiveRecord::Migration
+  def change
+    create_table :change_notes do |t|
+      t.string :note
```
dougdroper: Would it make sense to default this to an empty string?

app/models/change_note.rb
```diff
@@ -0,0 +1,30 @@
...
+  belongs_to :content_item
+
+  def self.create_from_content_item(payload, content_item)
+    return unless content_item.update_type == "major"
+    create_from_top_level_change_note(payload, content_item) ||
+      create_from_details_hash_change_note(content_item) ||
+      create_from_details_hash_change_history(content_item)
+  end
+
+  def self.create_from_top_level_change_note(payload, content_item)
```
dougdroper: Should these class methods be private? Which kinda makes me think that `self.create_from_content_item` shouldn't be a class method, maybe some other object, which would help with passing the payload and content_item around. What do you think?

app/presenters/details_presenter.rb
```diff
@@ -2,23 +2,30 @@
...
     end
 
     def details
       @_details ||=
         begin
-          content_item_details.each_with_object({}) do |(key, val), seed|
+          updated = content_item_details.each_with_object({}) do |(key, val), seed|
             seed[key] = append_transformed_govspeak(val)
           end
+          updated[:change_history] = change_history unless change_history.blank?
```
cbaines: This is making two method calls, which in some case will mean two database queries for the same data.

I think it would be preferable to inline the `change_history` method, and use a variable, which gets rid of a entire method, and means improves readability (as well as removing the duplicate calls/database queries) e.g.:
```
change_history = change_history_presenter.change_history

updated[:change_history] = change_history unless change_history.blank?
```

db/schema.rb
```diff
@@ -70,6 +78,16 @@
...
 
+  create_table "linkables", force: :cascade do |t|
+    t.integer  "content_item_id", null: false
+    t.string   "state",           null: false
+    t.string   "base_path",       null: false
+    t.string   "document_type",   null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
+    t.index ["content_id"], name: "index_link_sets_on_content_id", unique: true, using: :btree
+  end
```
carlosmartinez: Ah, you're going through commit by commit. This has been removed in a later commit. I'll interactive-rebase them all into a single commit before merging.

db/schema.rb
```diff
@@ -78,16 +79,6 @@
...
-  create_table "linkables", force: :cascade do |t|
-    t.integer  "content_item_id", null: false
-    t.string   "state",           null: false
-    t.string   "base_path",       null: false
-    t.string   "document_type",   null: false
-    t.datetime "created_at"
-    t.datetime "updated_at"
-    t.index ["content_id"], name: "index_link_sets_on_content_id", unique: true, using: :btree
-  end
-
```
cbaines: You could squash this in to the previous commit, for neatness (but I'll leave it up to you).

app/presenters/details_presenter.rb
dougdroper: or memorize the method

carlosmartinez: duly memoized

app/presenters/change_history_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def change_history
+      content_item.details[:change_history] || change_notes_for_content_item
+    end
+
+  private
+
+    def change_notes_for_content_item
+      ChangeNote.where(content_item_id: content_item.id).map do |cn|
```
cbaines: I'm surprised this query has no ordering, as the change notes should be ordered?

carlosmartinez: Thanks. Updated.

carlosmartinez:
@dougdroper @danielroseman - regarding whether we should create a ChangeNote on publish, I believe the answer is "no", since the change note is part of the put-content payload. The edge case is that the publish command sometimes changes the update_type. There's a fix in this PR to delete a content item's change note if the publish command changes the update_type from major to minor (because we shouldn't have change notes for minor updates). Beyond that, we need to think about whether publishing *should* be able to set update_type at all (I think @cbaines is writing a card for this).

app/presenters/change_history_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+  private
+
+    def change_notes_for_content_item
+      ChangeNote
+        .joins(:content_item)
+        .where("content_items.content_id" => content_item.content_id)
+        .order(public_timestamp: :desc)
+        .map { |cn| cn.slice(:note, :public_timestamp) }
```
dougdroper: `.pluck(:note, :public_timestamp).map {|n,t| {note: n, public_timestamp: t}}` might be faster

PR #556, opened 2016-10-12
[DO NOT MERGE] Return a 409 rather than 500 for concurrent content items

kevindew:
Bad things happen when we receive >1 requests concurrently to put an item with the same content_id. One of the problems that can occur is that we 500 due to trying to create the same link set twice.

This changes the behaviour in this scenario to return a 409 rather than a 500. 

I'm not 100% sure this is the best way to play it as:

- -ve it's not a consistent response to the same request
- +ve it's a better error response than a 500 for the scenario

Additional opinions on this would be great. Options:

A. 409 on this scenario as per this PR
B. 500 on this scenario as per current behaviour
C. Refactor so we can accept the content item - which will most likely result in us having conflicting data

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -161,6 +161,9 @@ module Commands
 
         link_set = LinkSet.create!(content_id: content_item.content_id)
         LockVersion.create!(target: link_set, number: 1)
+      rescue ActiveRecord::RecordNotUnique => e
+        Airbrake.notify(e)
+        raise_command_error(409, "Concurrent PUT Content requests for same content_id", {})
       end
 
       def lock_version_number_for_new_draft
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -584,6 +584,28 @@ RSpec.describe Commands::V2::PutContent do
         expect(lock_version).to be_present
         expect(lock_version.number).to eq(1)
       end
+
+      context "when a link set was created in between us checking and creating one" do
+        before do
+          allow(LinkSet).to receive(:create!).and_raise(ActiveRecord::RecordNotUnique)
+        end
+
+        it "returns a 409 Command Error" do
+          expect {
+            described_class.call(payload)
+          }.to raise_error(CommandError) { |e| expect(e.code).to eq 409 }
+        end
+
+        it "notifies airbrake" do
+          # this is required as - at the time of writing (2016-10-12) - notify
+          # gets called twice, once for a schema issue and then for this test.
+          allow(Airbrake).to receive(:notify)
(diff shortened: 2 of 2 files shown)


cbaines:
I think this would need to rescue `PG::UniqueValidation` as that is the error occurring. I'm also still hopeful that this can be resolved once and for all sometime soon, so I'd prefer just to do that, and not to change the behaviour in the mean time as it seems unnecessary.

kevindew:
Thanks for your input 👍 

`ActiveRecord::RecordNotUnique` wraps the `PG::UniqueViolation`

cbaines:
> ActiveRecord::RecordNotUnique wraps the PG::UniqueViolation

Ok... could you explain a bit more about what this means? I mainly brought this up as both a used explicitly elsewhere in the Publishing API https://github.com/alphagov/publishing-api/blob/master/app/models/path_reservation.rb#L20

kevindew:
Sure,

When ActiveRecord encounters database specific exceptions it will [convert them](https://github.com/rails/rails/blob/bc9dc418259f1db0f14ed895e8469b7839ef474b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L412-L429) to a vendor agnostic variant.

E.g. 
```
[4] pry(main)> exception = begin
[4] pry(main)*   2.times { LinkSet.create!(content_id: "123") }
[4] pry(main)* rescue Exception => e
[4] pry(main)*   e
[4] pry(main)* end
[5] pry(main)> exception.class
=> ActiveRecord::RecordNotUnique
[6] pry(main)> exception.message
=> "PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint \"index_link_sets_on_content_id\"\nDETAIL:  Key (content_id)=(123) already exists.\n: INSERT INTO \"link_sets\" (\"content_id\", \"created_at\", \"updated_at\") VALUES ($1, $2, $3) RETURNING \"id\""
[7] pry(main)> exception.cause
=> #<PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_link_sets_on_content_id"
DETAIL:  Key (content_id)=(123) already exists.
>
[8] pry(main)> exception.cause.class
=> PG::UniqueViolation
```

I doubt the `PG::UniqueViolation` exception has ever been caught there if my understanding is correct.

cbaines:
Ah, ok, thanks for explaining. So I guess that Airbrake/Errbit is using the cause, and not the actual exception (but I haven't dug in to airbrake to find out yet...).

kevindew:
After a couple of chats I think it's best to close this. Main reasons: 

- 409 implies we have a handle on what's going on but we don't really
- You wouldn't get the same result running this request again
- If we were handling concurrent requests ok we'd be hoping to 200 this request rather than 409.

PR #557, opened 2016-10-12
Prevent concurrent access to the same link set

kevindew:
We can use the uniqueness constraint on LinkSet as a means to prevent two link sets being changed at the same time.

Despite no successes with PUT Content we can at least make PATCH Link Set resilient to concurrent requests

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -3,16 +3,10 @@ module Commands
     class PatchLinkSet < BaseCommand
       def call
         raise_unless_links_hash_is_provided
+        link_set = find_or_create_link_set(content_id)
 
-        link_set = LinkSet.find_by(content_id: content_id)
-
-        if link_set
-          check_version_and_raise_if_conflicting(link_set, previous_version_number)
-          lock_version = LockVersion.find_by!(target: link_set)
-        else
-          link_set = LinkSet.create!(content_id: content_id)
-          lock_version = LockVersion.new(target: link_set)
-        end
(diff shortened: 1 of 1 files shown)


app/commands/v2/patch_link_set.rb
```diff
@@ -38,6 +32,14 @@ def call
 
     private
 
+      def find_or_create_link_set(content_id)
+        begin
+          LinkSet.find_or_create_by!(content_id: content_id).lock!
+        rescue ActiveRecord::RecordNotUnique
+          retry
```
danielroseman: It seems a bit dangerous to always retry here; a deadlock might mean this continues indefinitely. Instead it would be better to keep a count and retry a maximum of, say, 3 times.

kevindew: @danielroseman It's updated

app/commands/v2/patch_link_set.rb
```diff
@@ -38,6 +32,14 @@ def call
 
     private
 
+      def find_or_create_link_set(content_id)
+        begin
+          LinkSet.find_or_create_by!(content_id: content_id).lock!
+        rescue ActiveRecord::RecordNotUnique
```
kevindew: @cbaines that's weird - I seem to have just got comments through from you that no doubt you wrote a number of days ago. Maybe they were caught in the github review process?

cbaines: Hmm, sorry, I'll be more careful about using strange features in the future :)

PR #558, opened 2016-10-12
ActiveRecord .uniq is deprecated for .distinct

kevindew:


Davidslv:
According to the docs it seems to be an alias: http://apidock.com/rails/v4.2.1/ActiveRecord/QueryMethods/uniq

kevindew:
| Is it raising warnings?

Yup, you get a: `DEPRECATION WARNING: uniq is deprecated and will be removed from Rails 5.1 (use distinct instead)`

Seems API Dock doesn't have rails 5.0.0, but it's in [Rails 5.0 release notes](http://edgeguides.rubyonrails.org/5_0_release_notes.html#active-record-deprecations)

PR #569, opened 2016-10-19
Expand parent links for ordered_related_items

dougdroper:
Introduce rules to allow for named link types to have other types to be
recursed on.

Before you could only expand on the same type as the first level, this
lets you expand different types by supplying an array of levels.

eg: ['parent', 'parent_taxons', 'parent'] would expand the parent at
level one, parent_taxons at level two, and parents for all levels
greater then n, (the size of the array), the last element being the
"sticky" recursive element. so for array of size 1, it would recurse on just
that element.

This also introduces the ability to expanded recursively for multiple
parent_taxons.

ruby bench/downstream_payload.rb
..........  4.160000   0.190000   4.350000 (  9.104915)
queries: 198

..........  2.240000   0.070000   2.310000 (  3.923168)
queries: 180

..........  0.120000   0.010000   0.130000 (  0.189995)
queries: 120

..........  0.180000   0.010000   0.190000 (  0.337268)
queries: 180

### DIFF
--- a/app/presenters/queries/expand_dependees.rb
+++ b/app/presenters/queries/expand_dependees.rb
@@ -1,0 +1,71 @@
+module Presenters
+  module Queries
+    class ExpandDependees
+      def initialize(content_id, controller)
+        @content_id = content_id
+        @controller = controller
+      end
+
+      def expand
+        parents(content_id)
+      end
+
+    private
+
+      attr_reader :content_id, :controller
+
+      def parents(content_id, type = nil, visited = [], level_index = 0)
+        visited << content_id
+        links = all_links(content_id, type)
+        cached_web_content_items = all_web_content_items(links)
+        level = links.each_with_object({}) do |link, memo|
+          link_type = link["link_type"].to_sym
+          memo[link_type] = expand_level(link, cached_web_content_items, visited, level_index).compact
+        end
+        level.select { |_k, v| v.present? }
+      end
+
+      def all_web_content_items(links)
+        uniq_links = links.flat_map { |l| JSON.parse(l["target_content_ids"]) }.uniq
+        controller.web_content_items(uniq_links).each_with_object({}) { |w, memo| memo[w.content_id] = w }
+      end
+
+      def expand_level(link, all_web_content_items, visited, level_index)
+        JSON.parse(link["target_content_ids"]).map do |target_id|
+          rules.expand_field(all_web_content_items[target_id]).tap do |expanded|
+            next_level = next_level(link, target_id, visited, level_index)
+            expanded.merge!(links: next_level) if expanded
+          end
+        end
+      end
+
+      def next_level(current_level, target_id, visited, level_index)
+        return {} if visited.include?(target_id)
+        link_type = current_level["link_type"]
+        return {} unless rules.recurse?(current_level["link_type"], level_index)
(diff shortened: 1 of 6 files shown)


app/queries/dependent_expansion_rules.rb
```diff
@@ -11,16 +11,27 @@ def expand_field(web_content_item)
...
+      end
     end
 
     def reverse_name_for(link_type)
       reverse_names[link_type.to_sym]
     end
 
     def recursive_link_types
-      reverse_names.keys - [:documents, :working_groups]
+      [
```
kevindew: As the rules this follows is somewhat opaque do you think it would benefit from a comment?

app/presenters/queries/expanded_link_set.rb
```diff
@@ -5,7 +5,7 @@ def initialize(content_id:, state_fallback_order:, locale_fallback_order: Conten
         @content_id = content_id
         @state_fallback_order = Array(state_fallback_order)
         @locale_fallback_order = Array(locale_fallback_order)
-        @visited = []
+        @level_index = {}
```
kevindew: This variable confused me a bit - as I understand it's used just in the recursive process on dependees and would cause problems if there were more multiple recursive processes. I wonder if it makes more sense being injected into children from the dependees method rather than being an instance variable?

app/queries/dependent_expansion_rules.rb
```diff
@@ -11,16 +11,27 @@ def expand_field(web_content_item)
       web_content_item.to_h.slice(*expansion_fields(web_content_item.document_type.to_sym))
     end
 
-    def recurse?(link_type)
-      recursive_link_types.include?(link_type.to_sym)
+    def recurse?(link_type, level_index = 0)
+      recursive_link_types.any? do |t|
+        t[level_index] == link_type.to_sym || t.last == link_type.to_sym
```
carlosmartinez: if you want to play code golf:

```
t.values_at(level_index, -1).include?(link_type.to_sym)
```

carlosmartinez:
approve38

app/queries/dependent_expansion_rules.rb
dougdroper: Added, although at some point it would be nice to make it not opaque.

dougdroper:
@kevindew I did a quick refactoring of this. Still could go further, but might be more beneficial for someone else to do that.

app/presenters/queries/expand_dependees.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      def all_links(content_id, link_type = nil)
+        sql = <<-SQL
+          select links.link_type,
+            json_agg(links.target_content_id order by links.link_type asc, links.position asc) as target_content_ids
+          from links
+          join link_sets on link_sets.id = links.link_set_id
+          where link_sets.content_id = '#{content_id}'
+          #{"and link_type = '#{link_type}'" if link_type}
```
kevindew: Super minor - but do we need to keep this comment?

dougdroper: it's not a comment.

kevindew: oh yeah - of course .... \*hides in corner\*

PR #571, opened 2016-10-19
Delete change notes when deleting content items

carlosmartinez:
Resolves [this bug](https://errbit.integration.publishing.service.gov.uk/apps/552637940da1157bfa000399/problems/58078bcb657863512cf50000)


kevindew:
Looks good - builds are failing at the mo due to an invalid SSL certificate.

Could you also add ChangeNote to https://github.com/alphagov/publishing-api/blob/master/db/migrate/helpers/delete_content_item.rb

carlosmartinez:
@kevindew - done.

PR #581, opened 2016-10-27
Remove Linkable reference from DeleteContentItem

gpeng:
`Linkable` was removed in #559


cbaines:
Thanks @gpeng , but @kevindew has beaten you too it https://github.com/alphagov/publishing-api/pull/579

PR #582, opened 2016-10-27
Render withdrawn parent

dougdroper:
This adds support for whitehall formats that require a link to a withdrawn parent 
eg `HtmlAttachment`

before: 
..........  3.940000   0.150000   4.090000 (  7.054913)
queries: 198

..........  2.130000   0.060000   2.190000 (  3.008070)
queries: 180

..........  0.100000   0.010000   0.110000 (  0.158590)
queries: 120

..........  0.150000   0.010000   0.160000 (  0.267114)
queries: 160

after: 
..........  4.190000   0.160000   4.350000 (  7.296519)
queries: 198

..........  2.120000   0.050000   2.170000 (  3.024749)
queries: 180

..........  0.090000   0.010000   0.100000 (  0.151902)
queries: 120

..........  0.150000   0.010000   0.160000 (  0.269801)
queries: 160

### DIFF
--- a/app/queries/dependent_expansion_rules.rb
+++ b/app/queries/dependent_expansion_rules.rb
@@ -7,7 +7,6 @@ module Queries
     end
 
     def expand_field(web_content_item)
-      return unless web_content_item
       web_content_item.to_h.slice(*expansion_fields(web_content_item.document_type.to_sym))
     end
 
--- a/app/queries/get_content_item_ids_with_fallbacks.rb
+++ b/app/queries/get_content_item_ids_with_fallbacks.rb
@@ -12,7 +12,7 @@ module Queries
       unpublishings = Unpublishing.arel_table
 
       fallback_scope = content_items.project(
-          content_items[:id],
+        content_items[:id],
             content_items[:content_id],
           )
           .join(states).on(states[:content_item_id].eq(content_items[:id]))
--- a/app/queries/get_content_item_ids_with_fallbacks.rb
+++ b/app/queries/get_content_item_ids_with_fallbacks.rb
@@ -9,9 +9,9 @@ module Queries
       content_items = ContentItem.arel_table
       states = State.arel_table
       translations = Translation.arel_table
+      unpublishings = Unpublishing.arel_table
 
-      fallbacks = cte(
-        content_items.project(
+      fallback_scope = content_items.project(
           content_items[:id],
             content_items[:content_id],
           )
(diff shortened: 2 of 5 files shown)


app/queries/get_content_item_ids_with_fallbacks.rb
```diff
@@ -9,25 +9,35 @@ def self.call(content_ids, locale_fallback_order: ContentItem::DEFAULT_LOCALE, s
...
           .where(content_items[:document_type].not_in(::ContentItem::NON_RENDERABLE_FORMATS))
-          .where(states[:name].in(state_fallback_order))
           .where(translations[:locale].in(locale_fallback_order))
-          .order(
-            order_by_clause(:states, :name, state_fallback_order),
-            order_by_clause(:translations, :locale, locale_fallback_order)
-          ),
-        as: "fallbacks"
+
+      if state_fallback_order.include?("withdrawn")
```
cbaines: Any reason for using `"withdrawn"` here, which is not a state, over `"unpublished"` which is a state?

app/presenters/queries/expand_dependents.rb
```diff
@@ -21,7 +21,9 @@ def dependents
         links.group_by(&:first).each_with_object({}) do |(type, link_array), hash|
           reverse = ::Queries::DependeeExpansionRules.reverse_name_for(type).to_sym
           link_ids = link_array.map(&:last)
-          items = all_web_content_items.select { |item| link_ids.include?(item.content_id) }
+          items = all_web_content_items.select do |item|
+            link_ids.include?(item.content_id)
+          end
```
cbaines: This looks to have nothing to do with what the commit is about? It also looks to be just a stylistic change (if I haven't missed anything)?

If it is just stylistic, I find that committing these separately can help make reviewing easier, as its more informative for those doing reviewing.

app/queries/dependent_expansion_rules.rb
```diff
@@ -7,7 +7,6 @@ def expansion_fields(link_type)
     end
 
     def expand_field(web_content_item)
-      return unless web_content_item
```
cbaines: I'm unsure why this is being removed?

spec/presenters/queries/expanded_link_set_spec.rb
```diff
@@ -101,13 +101,13 @@
...
-                  a_hash_including(base_path: "/e", links: {})
-                ]
-              })]
+              mainstream_browse_pages: [a_hash_including(
+                base_path: "/c",
+                links: {
+                  parent: [
+                    a_hash_including(base_path: "/e", links: {})
+                  ]
+                })]
```
cbaines: Same point about stylistic changes in their own commits, although for these changes, the diff is easy to read, so it doesn't harm readability.

app/queries/get_content_item_ids_with_fallbacks.rb
gpeng: We've used `withdrawn` here as a state because although it currently isn't a 'state' it should be. `unpublished` has unfortunately been overloaded (as it also is in Whitehall) and is actually a set of three or four states. We're explicitly targeting `withdrawn` content here rather than `redirected`, `gone` etc. At some point we need to revisit how unpublishing states are implemented but for now this at least should clarify the intent.

app/queries/dependent_expansion_rules.rb
gpeng: It's no longer required as the calling code guards against this.

gpeng:
Thanks @cbaines. I've moved (or removed) the 'stylistic' changes into their own commit as requested and responded to your comments. Could we have any further discussion in a review so as we can work towards getting this merged?

gpeng:
Thanks again @cbaines We did try that approach but it ended up in having to make a lot more changes elsewhere and also things got a bit less 'obvious'. We settled on this as the most explicit approach. Thanks for your review :+1:

PR #583, opened 2016-10-27
Reduce Sidekiq log verbosity

kevindew:
We're generating a lot of data by having sidekiq logging set to verbose. On the 26th of October we produced over 20G of log data on integration, the vast majority were a log of severity: DEBUG.

I don't believe we're making any use of the data it produces as we already log in the Rails application the parameters of a Content Item being sent downstream and if an error occurs we will still get all the information.

I've added a job id to our rails logging which should allow us to link our rails queuing to sidekiq logs.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,5 +1,5 @@
 ---
-:verbose: true
+:verbose: false
 :concurrency:  2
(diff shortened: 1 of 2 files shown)


cbaines:
Great, thanks @kevindew :)

PR #584, opened 2016-10-27
Migrate specialist-publisher change history

carlosmartinez:
Iterate specialist publisher change_history elements and create
corresponding ChangeNote entries. This enables us to move Specialist
Publisher over to the new way of managing change history (where
Publishing API rather than the app is responsible for maintaining
change history).

### DIFF
--- a/app/queries/get_change_history.rb
+++ b/app/queries/get_change_history.rb
@@ -1,0 +1,20 @@
+module Queries
+  class GetChangeHistory
+    def self.call(publishing_app)
+      Queries::GetLatest.
+        call(content_items(publishing_app)).
+        pluck(:id, "details->>'change_history'").
+        flat_map do |content_item_id, item_history|
+          JSON.parse(item_history).map do |item_history_element|
+            item_history_element.symbolize_keys.merge(content_item_id: content_item_id)
+          end
+        end
+    end
+
+    def self.content_items(publishing_app)
+      ContentItem.
+        where(publishing_app: publishing_app).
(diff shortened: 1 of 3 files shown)


carlosmartinez:
@danielroseman I guess it makes sense to hold off merging this until the equivalent Specialist Publisher work is deployed.

app/queries/get_change_history.rb
```diff
@@ -0,0 +1,20 @@
...
+          JSON.parse(item_history).map do |item_history_element|
+            item_history_element.symbolize_keys.merge(content_item_id: content_item_id)
+          end
+        end
+    end
+
+    def self.content_items(publishing_app)
+      ContentItem.
+        where(publishing_app: publishing_app).
+        where("json_array_length((details->>'change_history')::json) > 0")
```
cbaines: `json_array_length(details->'change_history') > 0` seems to work, and is a bit neater too.

db/migrate/20161027134135_add_specialist_publisher_change_notes.rb
```diff
@@ -0,0 +1,7 @@
+class AddSpecialistPublisherChangeNotes < ActiveRecord::Migration[5.0]
+  def change
```
cbaines: Why `change`, and not say `up` as I would guess that you cannot roll this back? (I did try locally, but nothing seemed to happen, which I guess could be fine?)

app/queries/get_change_history.rb
carlosmartinez: thanks.

PR #587, opened 2016-10-28
Fix usage of GovukSchemas::Schema

cbaines:
Change the use to match the previous schema file path generation in the
Publishing API.


cbaines:
Changes to the pact expectations are now up https://github.com/alphagov/gds-api-adapters/pull/618

PR #593, opened 2016-11-03
Add a benchmark script for get-expanded-links

cbaines:
Just in case this is useful in the future.

### DIFF
--- a/bench/get_expanded_links.rb
+++ b/bench/get_expanded_links.rb
@@ -1,0 +1,102 @@
+# /usr/bin/env ruby
+
+require ::File.expand_path('../../config/environment', __FILE__)
+require 'benchmark'
+
+require 'stackprof'
+
+abort "Refusing to run outside of development" unless Rails.env.development?
+
+large_reverse = Link.find_by_sql(<<-SQL).first[:target_content_id]
+  SELECT target_content_id
+  FROM links
+  WHERE links.link_type = 'parent'
+  GROUP BY target_content_id
+  ORDER BY COUNT (*) DESC
+  LIMIT 1
+SQL
+
+large_forward = LinkSet.find_by_sql(<<-SQL).first[:content_id]
+  SELECT content_id
+  FROM link_sets
+  WHERE id IN (
+    SELECT link_set_id
+    FROM links
+    GROUP BY link_set_id
+    ORDER BY COUNT(*) DESC
+    LIMIT 1
+  )
+SQL
+
+no_links = Link.find_by_sql(<<-SQL).first[:content_id]
+  SELECT content_id
+  FROM link_sets
+  LEFT JOIN links
+  ON links.link_set_id = link_sets.id
+  OR links.target_content_id = link_sets.content_id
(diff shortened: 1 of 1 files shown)


bench/get_expanded_links.rb
```diff
@@ -0,0 +1,97 @@
...
+
+benchmarks = {
+  'Many reverse dependencies' => get_content_id_and_locale(large_reverse),
+  'Many forward dependencies' => get_content_id_and_locale(large_forward),
+  'No dependencies' => get_content_id_and_locale(no_links),
+  'Single link each way' => get_content_id_and_locale(single_link),
+}
+
+#ActiveRecord::LogSubscriber::IGNORE_PAYLOAD_NAMES.delete("SCHEMA")
+#ActiveRecord::Base.logger = Logger.new(STDOUT)
```
dougdroper: Do we need these lines?

cbaines: Like this script, they are not needed, but are helpful to keep around.

dougdroper: Why are they helpful?

cbaines: Line 71 configures activerecord to log queries to STDOUT. Line 70 enables the logging of "SCHEMA" queries, which I believe is those queries that are related to validation rules.

When I was using this script to peek in to the performance of get-expanded-links, I was using this code to show the queries being performed, and the time they took to execute.

cbaines:
> Looks good, maybe include the output in the commit so we can compare it in the future.

I don't think including the output would be helpful, as the content items used depend entirely on the local database state, so the results from separate runs of the script are only comparable when using the same database state.

bench/get_expanded_links.rb
dougdroper: Sure, then can we activate them by passing in arguments or similar? Otherwise they will just be ignored, or deleted in the future by someone who thinks they are unnecessary.

dougdroper:
> I don't think including the output would be helpful

It is more for a rough guideline that a) allows me to see what this produces without having to run it myself, and b) shows that it works. 

I understand that it won't be very accurate for actual comparisons. 
Maybe just in the PR would be useful.

bench/get_expanded_links.rb
```diff
@@ -0,0 +1,97 @@
...
+        print "."
+      end
+    }
+    puts tms
+  end
+  puts "queries: #{queries}"
+  puts "total: #{queries.map(&:last).sum}"
+  puts "query time: #{query_total_duration_by_name}"
+  puts ""
+  File.open("#{name.parameterize}.output", 'w') { |file| file.write(JSON.pretty_generate(result)) }
```
dougdroper: can these .output files go into a tmp directory so that they are not tracked by git?

cbaines: Yep, I've put them in the log directory, as this seemed a bit more suitable.

cbaines:
@dougdroper Thanks for your comments, I've now got around to making some changes that you suggested.

PR #599, opened 2016-11-08
Fix issues with change notes

danielroseman:
This fixes a number of issues with change notes:
* Return change notes in the get_content action.
* Make change notes unique for a content item, and update an existing note on put_content rather than always creating a new one.
* Remove duplicates created since the original change went live.

Because publishing-api does not contain all the historical content items, we needed to break the relationship between change note and content item for those previous changes; we now compare on content_id rather than content_item_id when working out what change notes to include.

### DIFF
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -31,11 +31,9 @@ private
 
   def create_from_top_level_change_note
     return unless change_note
-    ChangeNote.create!(
-      content_item: content_item,
-      note: change_note,
-      public_timestamp: Time.zone.now
-    )
+    ChangeNote.
+      find_or_create_by!(content_item: content_item).
+      update!(note: change_note, public_timestamp: Time.zone.now)
   end
 
   def create_from_details_hash_change_note
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -4,6 +4,12 @@ class ChangeNote < ActiveRecord::Base
   def self.create_from_content_item(payload, content_item)
     ChangeNoteFactory.new(payload, content_item).build
   end
+
+  def self.join_content_items(content_item_scope)
+    content_item_scope.joins(
+      "LEFT JOIN change_notes ON change_notes.content_item_id = content_items.id"
+    )
+  end
 end
 
 class ChangeNoteFactory
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -33,13 +33,18 @@ private
     return unless change_note
     ChangeNote.
       find_or_create_by!(content_item: content_item).
-      update!(note: change_note, public_timestamp: Time.zone.now)
+      update!(
+        note: change_note,
+        content_id: content_item.content_id,
+        public_timestamp: Time.zone.now
+      )
   end
 
   def create_from_details_hash_change_note
     return unless note
     ChangeNote.create!(
       content_item: content_item,
+      content_id: content_item.content_id,
       public_timestamp: content_item.updated_at,
       note: note,
     )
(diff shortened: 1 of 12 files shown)


app/presenters/queries/content_item_presenter.rb
```diff
@@ -136,6 +140,7 @@ def parse_results(results)
             parse_state_history(result)
 
             result["warnings"] = get_warnings(result) if include_warnings
+            result.delete("change_note") if result["change_note"].nil?
```
carlosmartinez: could perhaps generalise this change by returning `result.compact`

app/presenters/change_history_presenter.rb
```diff
@@ -18,17 +18,24 @@ def details
...
+      UserFacingVersion.join_content_items(
+        ContentItem.where(
+          "content_items.content_id" => content_item.content_id
+        ))
+        .where("user_facing_versions.number <= ?", version_number)
+        .pluck(:id)
+    end
+
     def version_number
       UserFacingVersion.where(content_item_id: content_item.id).last.number
```
carlosmartinez: does this potentially have n+1 performance issues? i guess the same result could be achieved by joining UserFacingVersion on itself, but that's unnecessary if performance isn't a problem

app/models/change_note.rb
```diff
@@ -25,17 +31,20 @@ def build
...
   def create_from_top_level_change_note
     return unless change_note
-    ChangeNote.create!(
-      content_item: content_item,
-      note: change_note,
-      public_timestamp: Time.zone.now
-    )
+    ChangeNote.
+      find_or_create_by(content_item: content_item).
+      update(
```
dougdroper: Would it be better to raise here, might silently fail

app/presenters/change_history_presenter.rb
```diff
@@ -18,17 +18,24 @@ def details
...
         .pluck(:note, :public_timestamp)
         .map do |note, timestamp|
           { note: note, public_timestamp: timestamp }.stringify_keys
         end
     end
 
+    def content_item_ids
+      UserFacingVersion.join_content_items(
+        ContentItem.where(
+          "content_items.content_id" => content_item.content_id
```
dougdroper: I think you could just do the `ContentItem.where(content_id: content_item.content_id)` in this inner scope

db/migrate/20161104185652_unique_change_note_for_content_item.rb
```diff
@@ -0,0 +1,52 @@
+class UniqueChangeNoteForContentItem < ActiveRecord::Migration[5.0]
```
dougdroper: I have been toying with the idea of having a data_migration rake task namespace for things like this. WDYT?

dougdroper:
Some superficial bikeshedding comments, but main logic seems sound

app/presenters/change_history_presenter.rb
danielroseman: I can't see where the n+1 would be. This is just a single extra query for the whole presenter, no?

carlosmartinez: yeah, you're right

db/migrate/20161104185652_unique_change_note_for_content_item.rb
carlosmartinez: yes, rake tasks are probably a better place to have things like this

danielroseman: Yes, sounds sensible.

kevindew:
Looks good to me 👍, agree with Doug about the fail silently.

Outside of the code here, there are [two other ways](https://github.com/alphagov/publishing-api/blob/02c2b8573b5ea429ec8b8d8b19b09b7feb78ce0a/app/models/change_note.rb#L43-L62) we create a change note and I wonder if these are leaving us vulnerable to multiple change notes per content Item still?

It could be good to have a unique index on ChangeNote for content_item_id, which I believe won't raise an error on multiple nulls in the table.

danielroseman:
All comments addressed, rebased.

PR #600, opened 2016-11-08
[DO NOT MERGE] Change dependency resolution

cbaines:
As part of https://trello.com/c/iok78RXu/832-update-user-documentation-for-dependency-resolution-2 , I've been looking at how the dependency resolution code in the Publishing API works, and I think that it is going about things the wrong way around.

[DO NOT MERGE] as I want to do some more thinking on dependency resolution before trying to change things.

Message from the 1st commit:
Previously, GetDependees was used in dependency resolution. This means
that some of the basic assumed functionality of dependnecy resolution
does not work.

Consider for example two draft content items, A and B, where A has a
link to B. This link means that when A is presented to the content
store, the link to B will be "expanded", including information from B,
e.g. the title. Dependency resolution should trigger A to be presented
to the content store when changes are made to B, to keep the information
in the expanded link consistent with B.

In the terminology of dependency resolution, A is the "dependent" and B
is the "dependee". So, to view the above example in terms of dependency
resolution, upon changes to B (the dependee), any dependents of B must
be presented to the content store, as the information in there expanded
links could have become inconsistent with B.

Previously, GetDependees was used, which means that in the example
above, the dependency resolution worker would look for links from B to
other content items (the dependees). Now, GetDependents is used, which
means that the dependency resolution worker will look for content items
linking to B, fixing the above example.

### DIFF
--- a/app/workers/dependency_resolution_worker.rb
+++ b/app/workers/dependency_resolution_worker.rb
@@ -7,7 +7,7 @@ class DependencyResolutionWorker
   def perform(args = {})
     assign_attributes(args.deep_symbolize_keys)
 
-    content_item_dependees.each do |dependent_content_id|
+    content_item_dependents.each do |dependent_content_id|
       downstream_content_item(dependent_content_id)
     end
   end
@@ -23,10 +23,10 @@ private
     @payload_version = args.fetch(:payload_version)
   end
 
-  def content_item_dependees
+  def content_item_dependents
     Queries::ContentDependencies.new(content_id: content_id,
                                      fields: fields,
-                                     dependent_lookup: Queries::GetDependees.new).call
+                                     dependent_lookup: Queries::GetDependents.new).call
   end
 
   def downstream_content_item(dependent_content_id)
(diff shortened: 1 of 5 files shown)


dougdroper:
If you look at it like: 
`A -> B -> C`
When we change A's title , to say X, We want to update B and C automatically through Dependency resolution. Since B and C "breadcrumb/parent" representation should now show:
`X -> B`
`X -> B -> C`
respectively. 
So we need to "represent/expand" B and C.

Another example of:
`A -> B -> C`

If we change C's title, we should only update B (which we don't do now as you rightly describe, but can't remember if there were any valid reasons). But we should not update A. Since A will only expand it's children to one level, ie just B, and would have no "knowledge" of C. 

So the `GetDependents` isn't exactly what we want to use, since that does the opposite of `GetDependees` namely recursively look up the chain from `C -> B -> A`. But rather we should just find direct links to C, namely B, and represent those. 

Does that make sense?

cbaines:
Closing, as this is not sufficient to fully fix dependency resolution (and will make it worse for some use cases), may reopen if this is built upon.

PR #602, opened 2016-11-10
Use (hopefully) fixed version of sidekiq-unique-jobs

carlosmartinez:
Currently we're seeing a problem where job IDs are not removed from the
`uniquejobs` hash in Redis, leading to Redis memory running out.

The change in my version of the gem is to add the following method and call it from `Unlockable#after_unlock`:

```
    def ensure_job_id_removed
      Sidekiq.redis { |redis| redis.hdel("uniquejobs", jid) }
    end
```

If this works in integration, we can raise a pull request against the sidekiq-unique-jobs gem.

Alternative solutions:

- Ensure job uniqueness via some other means (Sidekiq Pro has it built in)
- Run a nightly job that clears out the uniquejobs hash.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ gem 'whenever', '0.9.4', require: false
 gem "govuk_sidekiq", "~> 1.0.1"
 gem "json-schema", require: false
 gem "hashdiff"
-gem "sidekiq-unique-jobs", require: false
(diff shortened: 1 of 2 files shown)


cbaines:
Looks good :) but lets test this on integration before merging.

PR #605, opened 2016-11-16
Remove invalid draft content item

mgrassotti:
Trello: https://trello.com/c/DIEDyH9j/507-change-withdrawal-unpublishing-explanation-broken-medium

The content item with content_id `d7ac9eb5-7d6a-420e-a298-04149cc59e96`
and id `1238944` has a `draft` state, which is not reflected in Whitehall.
Furthermore it prevents the unpublishing explanation update of the previous version from Whitehall, because this is not allowed if a draft is present.
As this draft is not present in Whitehall it is probably inconsistent.
So we decided to remove it, with all its associated supporting objects.

### DIFF
--- a/db/migrate/20161116114622_remove_invalid_draft_content_item.rb
+++ b/db/migrate/20161116114622_remove_invalid_draft_content_item.rb
@@ -1,0 +1,15 @@
+require_relative "helpers/delete_content_item"
+
+class RemoveInvalidDraftContentItem < ActiveRecord::Migration[5.0]
+  def up
+    id = 1238944
+    content_item = ContentItem.find_by(id: id)
+    if content_item
+      Helpers::DeleteContentItem.destroy_supporting_objects(content_item)
+      content_item.destroy
+    end
+  end
+
+  def down
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161114151042) do
+ActiveRecord::Schema.define(version: 20161116114622) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


kevindew:
@mgrassotti looks cool mostly, would you be able to modify it to handle the situation where the content item doesn't exist? Just for people running it on a test environment.

db/migrate/20161116114622_remove_invalid_draft_content_item.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "helpers/delete_content_item"
+
+class RemoveInvalidDraftContentItem < ActiveRecord::Migration[5.0]
+  def change
```
Davidslv: Am I right to think that this migration would not be reversible after you destroy this content_item?

http://api.rubyonrails.org/classes/ActiveRecord/IrreversibleMigration.html

mgrassotti: Yes, I think it could be a good idea to split this into `up` and `down`. But I am wondering if this should raise an IrreversibleMigration error on down or simply do nothing... If it manages also the situation where the content item doesn't exist, maybe doing nothing is acceptable too and it doesn't creates problems in case someone needs to rollback to a previous migration. What do you think?

carlosmartinez:
is there a case for just having a rake task with content item id as a parameter?

i don't think migrations are always the best place to do one-off data jobs.

mgrassotti:
@carlosmartinez we had a discussion about this. We think that probably it is better to have a migration for that, in order to keep a history of deletions. It is also automatically run during the deploy. 
With a rake task, yes we wouldn't pollute the code base, but as said we wouldn't keep track of data changes.

carlosmartinez:
@mgrassotti - cool, thanks

mgrassotti:
Do you guys think it is ok to merge now?

mgrassotti:
😂

PR #606, opened 2016-11-17
Allow env vars for database configuration

kevindew:
This makes it easier to define different credentials to run the application. The naming format is based off DATABASE_URL.

I've set 3 environment variables as I don't think there's a way to do it for dev and test by using a single one as per DATABASE_URL, if anyone knows an approach let me know.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,9 @@ default: &default
   encoding: unicode
   pool: 5
   template: template0
+  host: <%= ENV["DATABASE_HOST"] %>
+  username: <%= ENV["DATABASE_USERNAME"] %>
+  password: <%= ENV["DATABASE_PASSWORD"] %>
 
 development:
   <<: *default


danielroseman:
Why can't we use DATABASE_URL? The [docs](http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database) suggest this is possible.

kevindew:
@danielroseman that works for the dev environment but not the test environment, and annoyingly rails insists on creating both dev and test when you run tasks like db:create, db:setup and similar. 

I didn't know that we could specify a `url` parameter though, so it could be that we can set up a TEST_DATABASE_URL env var for test

cbaines:
What do you mean by defining different credentials to "run the application"? Do you mean just to get it to connect to PostgreSQL?

cbaines:
A possible alternative to this would be to use the standard PostgreSQL environment variables, which might not even require a configuration change https://www.postgresql.org/docs/9.4/static/libpq-envars.html

kevindew:
@cbaines that's cool - I didn't know you could do that. As we're going to have a few similar apps (eg mongo) that'll need similar an environment variable seems a more consistent approach.

kevindew:
Closing in favour of https://github.com/alphagov/publishing-api/pull/607

PR #607, opened 2016-11-17
Set up TEST_DATABASE_URL env var

kevindew:
This allows us to specify an env var for the test database which can be configured in parallel with DATABASE_URL.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -10,6 +10,7 @@ development:
 
 test:
   <<: *default
+  url: <%= ENV["TEST_DATABASE_URL"] %>
   database: publishing_api_test
 
 production:


kevindew:
FYI anyone reviewing @andrewgarner and I had a chat about this yesterday and we weren't able to resolve it with just `DATABASE_URL`

Reason:
When you don't specify a `RAILS_ENV` or specify `RAILS_ENV=development` and run a `db:create` or `db:seed` it creates both test and development databases [source](https://github.com/rails/rails/blob/796049ef3e0c526a2c97418d4ba266050bc38c0a/activerecord/lib/active_record/railties/databases.rake#L25)
The value of `ENV["DATABASE_URL"]` is only used then for the development environment and errors in the test environment.

boffbowsh:
#608 might be preferable

carlosmartinez:
Replaced by [https://github.com/alphagov/publishing-api/pull/608](608).

PR #608, opened 2016-11-21
Allow dev and test dbs to be configured with URL

boffbowsh:
For use in Docker environments. When running in development or test we
still want to be able to be able to set the database using Docker
environment variables. If a `DATABASE_URL` variable is provided it will
use it for the development environment. For test, it will strip the
`_development` suffix and replace it with `_test`.

If a `DATABASE_URL` is not supplied, Rails will fall back to using the
YAML configuration.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,10 +7,12 @@ default: &default
 development:
   <<: *default
   database: publishing_api_development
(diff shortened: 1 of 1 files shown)


dougdroper:
Yip makes sense to me

PR #610, opened 2016-11-21
Change default govuk-content-schemas branch

cbaines:
Change the branch from master to deployed-to-production. This will help
prevent changes being merged in to the Publishing API that don't work
with the currently deployed govuk-content-schemas (this changes comes
from GOV.UK RFC 59).


cbaines:
This is blocking on https://github.com/alphagov/govuk-app-deployment/pull/78 as there is no `deployed-to-production` branch at the moment.

PR #611, opened 2016-11-22
Add Dockerfile and .dockerignore

carlosmartinez:
The new `publishing-e2e-docker` project requires each dependent project
to have a Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential libpq-dev libxml2-dev libxslt1-dev && apt-get clean
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential libpq-dev libxml2-dev libxslt1-dev && apt-get clean
+
+# for capybara-webkit
```
dougdroper: Just double check if we need this

boffbowsh: Yeah, I can't think what this would be needed for.

carlosmartinez: Well spotted. Duly removed.

PR #615, opened 2016-11-29
Fix incorrectly named variable in rake task

dougdroper:


kevindew:
While you're fixing there's also a typo in the description

dougdroper:
@kevindew updated

PR #619, opened 2016-11-30
Dependency Resolution documentation

kevindew:
This is my first draft at dependency resolution documentation. I'm putting this up for a cursorary review prior to proof reading and adding better examples.

Any early feedback would be appreciated before I clean it up

### DIFF
--- a/doc/dependency-resolution.md
+++ b/doc/dependency-resolution.md
@@ -1,356 +1,111 @@
-## Dependency Resolution
+# Dependency Resolution
 
-**Problem Statement**
+## Contents
 
-We're undergoing a migration to the new publishing platform. The publishing
-platform doesn't currently prescribe how the following features should work:
-
-- Breadcrumbs
-
-These are links that appear at the top of the majority of GOV.UK pages and allow
-users to navigate up to a higher-level page
-
-- Related Links
-
-These are links that appear on the right hand side of the majority of GOV.UK
-pages and allow users to navigate to content that is related to the content they
-are currently viewing. When two pages are related, this relationship tends to
-be bidirectional.
-
-- Support for Govspeak
-
-Many of our publishing applications support a language called Govspeak, which is
-used for adding formatting structure to content, such as headers, lists and
-links. It is possible in Govspeak to add a link to another piece of content,
-which could be changed independently.
-
-- Collections of content
-
-There are a few pages on GOV.UK which are collections of other content. These
-pages are effectively directories of links to other content. Each of these links
-can change independently and its title should be updated on the collection page.
-
-- Centralised search indexing
-
-Currently, publishing applications manually speak to Rummager which indexes
-their content for search. We'd like to centralise this behaviour so that the
-Publishing API forwards this content in its expanded form onto a queue which is
-then consumed by Rummager.
-
-Note: In addition to these requirements, there are some non-functional
-requirements detailed below. This includes a architectural goal of moving
-towards a simpler content store that could be replaced with a simple storage
-technology rather than an application that must be maintained.
-
-**Summary**
-
-In all of these cases, there is a dependency between content items. In the first
-case, the name of a page might change or the hierarchy might change. This would
-mean that we have to update the breadcrumbs for many pages to reflect this.
-
-In the second case, the title or path of a related link could change which would
-require an update to any content item that linked to the changed item. This is
-very similar in the third and fourth cases where embedded links would need to be
-updated to reflect the new title / path of the content.
-
-**High-Level Approach**
-
-The agreed, high-level approach to cope with these scenarios is to have the
-Publishing API track and resolve dependencies on behalf of publishing
-applications. This is notionally a difficult thing to achieve and we'd rather
-solve this problem once as part of the platform than over and again in various
-publishing applications.
-
-We already capture some information about the dependencies of content items in
-their 'links' data. We intend to extend this behaviour to incorporate new
-business logic that will check these dependencies when content changes,
-determine which things need to be updated and send these content items
-downstream to the content store automatically.
-
-Until now, we haven't got much further than this in detailing how this solution
-will work. This write-up details the analysis of this problem and presents a
-solution that aims to meet business requirements.
-
-**Non-Functional Requirements**
-
-In addition to the functional requirements for the features described in the
-Problem Statement, there are additional non-functional requirements that we
-should bear in mind when considering a proposal.
-
-- Publishing API response times
-
-We need to ensure that response times to the Publishing API are kept within
-reasonable boundaries to ensure that we don't introduce problems with publishing
-applications timing out.
-
-- Network communications
-
-We should consider the number of communications between the Publishing API and
-the Content Store. It is anticipated that in some cases, a large number of items
-could be dependent on specific items. An example of this is the Home page, which
-is likely to appear in the breadcrumbs for a lot of pages. If we were to update
-the Home page, we should be careful not to unnecessarily send tens of thousands
-of updates unless this is absolutely necessary.
-
-- Choice in technology
-
-We should be careful about the choice in technology to solve this problem.
-Ideally, we should choose technology that is familiar and well-understood by the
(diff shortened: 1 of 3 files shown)


doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
+# Dependency Resolution
 
-**Problem Statement**
+Dependency Resolution is a concept within the Publishing API to describe two
+processes:
 
-We're undergoing a migration to the new publishing platform. The publishing
-platform doesn't currently prescribe how the following features should work:
+- Determining the links that are presented with a content item when it is
+  represented in the Content Store.
```
cbaines: I think of this as link expansion, which is a related, but separate concept to Dependency Resolution.

doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
+- Determining which content items require updating as a result of a change to
+  a different content item.
```
cbaines: I would normally phrase this as "re-presenting to the content store", rather than just updating. It also might be good to specify that this should only change the expanded links?

doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
-All of this behaviour should be completed out-of-band from the publishing
-application's request cycle with the Publishing API. It is proposed that we
-isolate all of this dependency resolution work into a separate Sidekiq worker
-that is responsible for determining the set of content items that need to be
-updated and for actually presenting and sending those updates downstream. When
-the Publishing API receives any interaction that can affect dependencies, it
-need only place this content item on a 'dirty' queue for processing by the
-dependency resolution worker.
+At the point of representing a content item to a Content Store we need to load
+the details of every content item that has an explicit or implicit link from
```
cbaines: I'm not sure I've heard of these transitive links, or dependencies referred to as "implicit" before. This is not how I think about them, as because of the way that link expansion can be configured, there may be no meaningful relationship between items that have some transitive dependency,

doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
-It is important to note that, although these groups are related, they could
-differ in their expansion fields, as well as the recursive nature of their
-expansion. To use the parent / child example, again – a child's parent will form
-part of its breadcrumbs and so its full ancestry would need to be expanded in
-order to render these. A group's children, however, could be presented one level
-at a time and need not be expanded recursively to form the full hierarchy of
-descendents.
+When a content item is updated we need not only update the representation of
+that item in the content store but also the representation of any item that has
+either an implicit or explicit link to a content item.
```
cbaines: It might be good to tweak this slightly, as it could be misleading. In terms of the raw details, you don't actually have to update every content item that you might consider needing updating just from looking at the graph, as it really depends if any of the data that goes in to the expanded links has changed. While at the moment the Publishing API does not do this, it is a possible change that may be made in the future.

I think it could be clearer to always talk about dependency resolution in terms of updating the expanded links of content items (and only talking about the actual links when you get to the implementation details).

doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
-      }
-    ]
-  }
-}
-```
+- As translations. If a content item is available with a different locale it
+  will automatically be linked as a translation of a content item.
+- Content items linking to this item as a parent item. Content items that are
+  linked to as a parent automatically implicitly reciprocate this linking as
+  children of this link.
```
cbaines: The implied "children" link type is only one of the 4 implicit link types, it might be good to try and generalise the description, and just use parent/children as an example.

kevindew:
Thanks very much @cbaines your suggestions all sound good.

doc/dependency-resolution.md
```diff
@@ -1,356 +1,129 @@
...
-in the links data. For now, I think we should include the expanded links as a
-separate field in the content item. Perhaps this can be called "expanded_links",
-or similar and we can send it alongside the unexpanded links. As we migrate
-applications to use dependency resolution, we should make them use this field
-instead of "links" and look at retiring the expansion of links in content store.
+Thus any time either 456 or 123 changes a new representation of the content
+item is created for the Content Store.
 
-Here is a rough example of what this **might** look like:
+In addition to explicit links there are also implicit links between content
```
tijmenb: As per our slack discussion, perhaps we could avoid implicit/explicit jargon altogether.

For example, instead of:

> In addition to explicit links there are also implicit links between content items.

You can explain:

> In addition to links that are set by publishing applications, content items items can be automatically linked by the publishing-api.

doc/link-expansion.md
```diff
@@ -0,0 +1,287 @@
...
+#### Available translations
+
+A content item can be available in multiple translations. This will be
+determined by content items sharing the same [content_id][content-id] and
+having different [locale][locale] values.
+
+The links for a content item will include all the translations of a content
+item, including the current locale.
+
+**eg**
```
thomasleese: Probably being really pedantic, but I would write ‘Example’ here just to make it 100% clear.

doc/link-expansion.md
```diff
@@ -0,0 +1,287 @@
...
+## Introduction
+
+Link expansion is a concept in the Publishing API which describes the process
+of converting the stored links of a content item into a JSON representation
+containing the details of these links. It is used in the Publishing API
+during the process of sending a content item downstream to the
+[Content Store][content-store].
+
+The process involves determining which content items should and can be linked
+to; the versions of them that are linked; and the fields that will be included
```
thomasleese: I would use commas here, looks strange to use a semi-colon to me.

doc/link-expansion.md
```diff
@@ -0,0 +1,287 @@
...
+`children` links which includes `C`.
+
+Reverse links are considered to be **dependents** of a content item and are
+determined using the [Presenters::Queries::ExpandDependents][expand-dependents]
+class.
+
+#### Available translations
+
+A content item can be available in multiple translations. This will be
+determined by content items sharing the same [content_id][content-id] and
```
thomasleese: I would wrap content_id and locale in backticks.

doc/link-expansion.md
```diff
@@ -0,0 +1,287 @@
...
+- `links` - Any [recursive links](#recursive-links) that are presented with a
+  link representation of a content item
+
+The fields can and are customised in some cases. This can be done on a
+`link_type` basis. These customisations can be performed in
+[Presenters::Queries::ExpandDependees][expand-dependees] and
+[Presenters::Queries::ExpandDependents][expand-dependents] - where the latter
+is for [reverse links](#reverse-links).
+
+[content-store]:https://github.com/alphagov/content-store
```
thomasleese: Missing whitespace between colon and URL.

doc/dependency-resolution.md
```diff
@@ -43,135 +52,51 @@ categories: Linked to a content item, reverse links, and translations.
...
 ### Linked to a content item
 
+The most simple type of link to a content item is an explicit one where for a
+content item `B` there is a link that targets a content item `C`. When
+dependency resolution occurs as the result of a change to `C` items that link
+to it, such as `B`, can be looked up simply.
+
 #### Recursive links
 
+Some link types are considered [recursive](link-expansion.md#recursive-links)
```
thomasleese: Would put a comma at the end of this line.

doc/dependency-resolution.md
```diff
@@ -43,135 +52,51 @@ categories: Linked to a content item, reverse links, and translations.
...
-Govuk-Dependency-Resolution-Source-Content-Id HTTP header if specified
-on a request made by the Publishing API indicates that this request is
-a result of dependency resolution. The value of the header is the
-`content_id` on which dependency resolution took place.
+If the content item that initiated dependency resolution is available in
+multiple locales each of these will be re-presented to the content store.
+
+## Updating Content Store
+
+As a result of the dependency resolution process content items identified to
```
thomasleese: I would put a comma between the word ‘process’ and ‘content’.

kevindew:
This has now been updated and squashed. Ready for further reviews and feedback.

PR #622, opened 2016-12-05
Import history

cbaines:
https://trello.com/c/7LsjNZM2/802-add-endpoints-for-importing-content-in-to-the-publishing-api-3

I'm looking at this as part of the work on importing Needs in to the Publishing API.


app/commands/v2/import.rb
```diff
@@ -1,27 +1,24 @@
 module Commands
   module V2
-    class Import
-      def initialize(params)
-        @params = params
-      end
+    class Import < BaseCommand
```
dougdroper: This might give a bit of weird results by inheriting from BaseCommand, BaseCommand wraps that `call` in a transaction, which will probably interfere with the transaction in this class, and the send_downstream needs to happen outside of that transaction, it also creates a Event (which you might want to do). but I didn't inherit from BaseCommand because of those hidden unexpected side effects.

danielroseman:
Seems like work is still ongoing on this. It might be best to close the PR until it's actually ready for review.

cbaines:
This is now at a point where we have importing needs in to the Publishing API working locally, so it would be good to get more review so this can be merged.

app/commands/v2/import.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      def send_downstream(content)
+        return unless content
+        DownstreamLiveWorker.perform_async_in_queue(
+          DownstreamLiveWorker::LOW_QUEUE,
+          content_id: content[:content_id],
+          locale: 'en',
+          message_queue_update_type: content[:payload][:update_type],
+          payload_version: 1 #event.id
```
dougdroper: This will need to be the Event#id otherwise the content store will reject it.

app/commands/v2/import.rb
```diff
@@ -0,0 +1,134 @@
+module Commands
+  module V2
+    class Import < BaseCommand
+
+      def self.call(payload)
+        logger.debug "#{self} called with payload:\n#{payload}"
```
dougdroper: do you still need this line?

app/commands/v2/import.rb
```diff
@@ -0,0 +1,134 @@
...
+            }
+          })
+        end
+
+        delete_all(payload[:content_id])
+        all_content_items.map.with_index do |event, index|
+          create_content_item(event, index, payload[:content_id])
+        end
+
+        send_downstream(all_content_items.find { |e| e[:action] == 'Publish' })
```
dougdroper: this will need to be wrapped in after_transaction_commit like https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/put_content.rb#L27

app/commands/v2/import.rb
cbaines: It was for consistency with the BaseCommand, but now we are just using its implementation.

app/commands/v2/import.rb
cbaines: Good spot, ec7bd1a

app/commands/v2/import.rb
```diff
@@ -0,0 +1,123 @@
...
+      end
+
+      def send_downstream(content)
+        return unless content
+        DownstreamLiveWorker.perform_async_in_queue(
+          DownstreamLiveWorker::LOW_QUEUE,
+          content_id: content[:content_id],
+          locale: 'en',
+          message_queue_update_type: content[:payload][:update_type],
+          payload_version: event.id
```
dougdroper: hmm, I still think this needs to be rethought. If we send two content_items (a redirect and the latest one) to the content_store with the same event.id it will error out and the second one will be lost.

cbaines: We should probably test for this if this is not covered, but I don't think this would be a problem. As the locking in the content store is done per base path, and providing that the Publishing API only sends one content item for each base path it wants to change (which I think is the behaviour), there should be no locking issues.

binaryberry: We don't expect there to be any redirects for needs. The way the rake task in govuk_need_api is being built, all versions of a Need have the [same base_path](https://github.com/alphagov/govuk_need_api/pull/114/files#diff-e6b7dadd98f8ffc2590ab131f4e70755R89). There is no functionality to redirect from one need to another in Maslow either.

dougdroper: ok great, can we remove the redirect stuff please, we will add it in when it is needed again, I am not confident in it yet without more testing.

cbaines: Done in  	f28b614 and 732c0c3

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+        unless UuidValidator.valid?(payload[:content_id])
+          raise CommandError.new(
+          code: 422,
+          error_details: {
+            error: {
+              code: 422,
+              message: "Content id not valid",
+              fields: "content_id",
+            }
+          })
```
kevindew: quibble - indentation fix for here

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+
+        delete_all(payload[:content_id])
```
kevindew: This could cause a content store inconsistency. If the items with this content_id had a base_path that didn't exist in the import the content_stores would have an orphaned item(s)

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+    private
+
+      def send_downstream(content)
+        return unless content
+        DownstreamLiveWorker.perform_async_in_queue(
+          DownstreamLiveWorker::LOW_QUEUE,
+          content_id: content[:content_id],
+          locale: 'en',
```
kevindew: We seem to be assuming that all imports will be for a locale of "en"?

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+          })
+        end
+
+        delete_all(payload[:content_id])
+        payload[:content_items].map.with_index do |event, index|
+          create_content_item(event, index, payload[:content_id])
+        end
+
+        after_transaction_commit do
+          send_downstream(payload[:content_items].find { |e| e[:action] == 'Publish' })
```
kevindew: From my understanding this will only update the live content store? and that is only when the `:action` is `Publish`, this seems to be the only action that will update the content store so no items will update the draft content store? and this usage seems somewhat confusing if the final state is unpublished


also quibble single quotes

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+      def create_content_item(event, index, content_id)
+        event_payload = event[:payload]
+        content_item_payload = event_payload.except(:state)
+
+        validate_content_item_payload(content_item_payload)
+
+        Services::CreateContentItem.new(
+          payload: content_item_payload.merge(content_id: content_id),
+          user_facing_version: index + 1,
+          lock_version: index + 1,
```
kevindew: probably wouldn't need to updating the lock version each time

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+              code: 422,
+              message: "Content id not valid",
+              fields: "content_id",
+            }
+          })
+        end
+
+        delete_all(payload[:content_id])
+        payload[:content_items].map.with_index do |event, index|
+          create_content_item(event, index, payload[:content_id])
```
kevindew: This could get us in a bit of a mess I think

As it creates a new user_facing_version each time this would only really work if you sent over a sequence of states that were "superseded" (x times), "publish". If you had "draft", "published" it'd get in a mess - I'd expect you'd get a validation message, but that'd be quite hard to decipher as it won't relate to which item in your import caused the problem.

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+        validate_schema(content_item_payload)
+      end
+
+      def attributes
+        @attributes ||=
+          [:base_path, :locale] + ContentItem.new.attributes.keys.map(&:to_sym)
+      end
+
+      def state(event)
+        event[:payload][:state] || 'superseded'
```
kevindew: quotes quibble

app/commands/v2/import.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      def publishing_app
```
kevindew: don't think this method is used anymore (assume it was part of redirect logic)

app/helpers/redirect_helper.rb
```diff
@@ -1,5 +1,5 @@
 module RedirectHelper
```
kevindew: Probably don't need edits to this file anymore

db/migrate/helpers/delete_content_item.rb
```diff
@@ -1,40 +1,8 @@
...
-      content_ids = Array(content_ids)
-
-      content_items = ContentItem.where(content_id: content_ids)
-      destroy_supporting_objects(content_items)
-      content_items.destroy_all
-
-      destroy_links(content_ids)
-    end
-
-    def self.destroy_supporting_objects(content_items)
```
kevindew: If we remove this method we should remove it from any migrations and update documentation

db/migrate/helpers/delete_content_item.rb
```diff
@@ -1,40 +1,8 @@
...
-
-      supporting_classes.each do |klass|
-        next unless ActiveRecord::Base.connection.data_source_exists?(klass.table_name)
-        klass.where(content_item: content_items).destroy_all
-      end
-
-      LockVersion.where(target: content_items).destroy_all
-    end
-
-    def self.destroy_links(content_ids)
```
kevindew: If we remove this method we should remove it from any migrations and update documentation

spec/commands/v2/import_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      }
+    end
+
+    let(:payload) do
+      {
+        content_id: content_id,
+        content_items: [
+          {
+            action: "PutContent",
+            payload: content_item,
```
kevindew: This usage here doesn't really make sense as we don't use the "PutContent" action for this and if we were using it we couldn't use it for "superseded"

spec/commands/v2/put_content_spec.rb
```diff
@@ -226,7 +226,7 @@
           described_class.call(payload)
           Commands::V2::Publish.call(content_id: content_id, update_type: "minor")
 
-          regex = /Content item user_facing_version=(\d+) and locale=en for content item=#{Regexp.quote(content_id)} conflicts/
+          regex = /Content item state=draft and locale=en for content item=#{Regexp.quote(content_id)} conflicts/
```
kevindew: Kinda weird this needs changing? not sure whats causing it - maybe it's concealing a bug?

app/commands/v2/import.rb
binaryberry: All needs have a locale of "en" - they don't have translations

db/migrate/helpers/delete_content_item.rb
cbaines: Migrations fixed in b40158e and README.md updated in ffdc131.

db/migrate/helpers/delete_content_item.rb
cbaines: Couldn't see anything using this, but I changed the documentation in ffdc131.

app/commands/v2/import.rb
cbaines: I've changed the import command to call out to the RepresentDownstream command, which should work much better d65e5c7.

app/commands/v2/import.rb
cbaines: I've got rid of this in a67d343, as the RepresentDownstream command will do a better job.

spec/commands/v2/import_spec.rb
cbaines: I've got rid of action now, which might help 99e250f.

app/helpers/redirect_helper.rb
cbaines: Sorry, misread you comment at first. Your right, and I have now reverted the changes to this file in 99e250f.

spec/commands/v2/put_content_spec.rb
cbaines: I've had a quick look, and it looks like its just dependent on the order in which the validation occurs. I would guess that this has changed due to the changes in the PutContent command.

I'm not sure this is the best time, as I realise there is quite a lot of churn in the database, but I can rewrite the locking code such that it serialises the command processing, and fails on the lock version instead if you think that is necessary?

app/commands/v2/import.rb
cbaines: I got rid of this code :)

app/commands/v2/import.rb
cbaines: Probably not, but it seems to me like the not doing so would look odd in the database?

app/commands/v2/import.rb
cbaines: I believe this is now fixed.

app/commands/v2/import.rb
cbaines: I've added this functionality now https://github.com/alphagov/publishing-api/pull/622/commits/a6835cce8e256f7811bef7c91ed3e4e5a72b518a

app/commands/v2/import.rb
cbaines: This is the error you get:
```
     CommandError:
       User facing version draft content item cannot be behind the live content item (1 < 2)
```

I can see a few options here, but I'm hesitant to make any changes while the changes to the database schema are underway.

cbaines:
What more needs doing on this now? I think I've addressed the feedback given, so let me know if I've missed something.

app/services/delete_content_item.rb
```diff
@@ -0,0 +1,40 @@
...
+        klass.where(content_item: content_items).destroy_all
+      end
+
+      LockVersion.where(target: content_items).destroy_all
+    end
+
+    def self.destroy_links(content_ids)
+      content_ids = Array(content_ids)
+      LinkSet.where(content_id: content_ids).destroy_all
+      Link.where(target_content_id: content_ids).destroy_all
```
boffbowsh: This will delete links from other content to this content, which might not be the desired thing. If we're replacing eg. a Publication and deleting all links TO it, that Publication will be removed from any Document Collections that were linking to it. I think only removing links FROM this content is required.

cbaines: Good spot :) This code is just being moved in this branch (from here: [1]), so you might want to make a separate Pull Request for this change?

1: https://github.com/alphagov/publishing-api/blob/c183b846d90c254953a81f049ed83f3e31ac272e/db/migrate/helpers/delete_content_item.rb#L37

boffbowsh: That might be the correct thing to do in that case, so perhaps the importer shouldn't use that delete helper, or should pass an argument saying not to delete links to this content.

cbaines: Ok, yeah, that makes sense. I've modified it now to not delete links b6da5da.

cbaines:
> What more needs doing on this now? I think I've addressed the feedback given, so let me know if I've missed something.

Bump again. @kevindew, @dougdroper could you take another look, as I think I've made changes since you last commented on this.

kevindew:
Sorry @cbaines I started a review the other day but didn't get a chance to finish it. We're just working out who'll have some time to go review it properly.

cbaines:
> Sorry @cbaines I started a review the other day but didn't get a chance to finish it. We're just working out who'll have some time to go review it properly.

Any news yet? Ideally this would be merged in and deployed to production before we get Maslow to a state where it can use the Publishing API, so that deploying Maslow using the Publishing API is just blocking on getting Maslow working, and not the Publishing API. We (@binaryberry @issyl0 and I) now have Maslow rendering content from the Publishing API in a rudimentary way, so I'm hopeful that getting Maslow working fully with the Publishing API is not too far off now.

kevindew:
@cbaines sorry nothing to report, are you free this morning? I can pop over and chat through some bits with you - if so let me know when your stand up finishes.

app/commands/v2/import.rb
```diff
@@ -0,0 +1,232 @@
...
+        end
+
+        not_draft_content_item = nil
+        content_items.reverse_each do |content_item|
+          if not_draft_content_item.present?
+            content_item.supersede
+          elsif content_item.state != "draft"
+            not_draft_content_item = content_item
+          end
+        end
```
kevindew: We're not sure what this code does or is needed?

app/commands/v2/import.rb
```diff
@@ -0,0 +1,232 @@
...
+          else
+            raise CommandError.new(
+              code: 422,
+              message: "Unrecognised state: #{state[:name]}."
+            )
+          end
+        end
+      end
+
+      def validate_history_entry(history_entry, index)
```
kevindew: should we validate states are in a correct form? and should this validation be in a separate class?

app/commands/v2/import.rb
```diff
@@ -0,0 +1,232 @@
...
+          if not_draft_content_item.present?
+            content_item.supersede
+          elsif content_item.state != "draft"
+            not_draft_content_item = content_item
+          end
+        end
+
+        new_content_store_base_paths = get_base_path_content_store_pairs
+
+        after_transaction_commit do
```
kevindew: Can we move this to a worker?

app/commands/v2/import.rb
cbaines: Yep, have removed it  	571288f

app/commands/v2/import.rb
cbaines: Yep, done in  	012af98

app/commands/v2/import.rb
```diff
@@ -0,0 +1,226 @@
...
+          raise CommandError.new(
+            code: 422,
+            error_details: {
+              error: {
+                code: 422,
+                message: "Content id not valid",
+                fields: "content_id",
+              }
+            }
+          )
```
cbaines: Yep, done in 1bb21d3687e67e821e83721d171f031ee140e582

app/commands/v2/import.rb
cbaines: I've added a test to check that no invalid states slip through in db2365b . I'm tempted to not add any explicit validation to the import endpoint, as it doesn't feel like the right place.

cbaines:
What is the progress on getting this merged? I'd like to start testing Maslow using the Publishing API on integration and staging over the next few days, and that will be easiest to do if this is merged.

From what I remember of our conversation last week @kevindew , we discussed some things about simplifying the way states are specified. I may have forgotten the details, but looking at the current implementation, I don't see any opportunities to simplify it.

kevindew:
@cbaines I've just added some documentation. If you can just check that over for me that'd be great then I'll approve.

cbaines:
> @cbaines I've just added some documentation. If you can just check that over for me that'd be great then I'll approve.

All fine by me.

cbaines:
:fireworks: :fireworks: :fireworks: :fireworks:

PR #624, opened 2016-12-05
Refactor Dependency Resolution

kevindew:
This is an attempt to fix the bugs listed in: https://trello.com/c/21wSjcbG/851-fix-dependency-resolution-2 , to make dependency resolution easier to test and to make it easier to understand.

The `ContentDependencies` class has been repurposed into a class that is given a content_id, locale and state fallback rules and from this can determine the content_id and locales of all items that require an update. Hopefully being a single class it will be easier to write tests against if/when future bugs are identified.


app/queries/content_dependencies.rb
```diff
@@ -1,40 +1,56 @@
...
+  #     in the item linked to. eg if our subject (A) has a parent of B, B would
+  #     automatically have a link to A of type children.
+  # 3 - Translations, for the subject all locales except the one provided is
+  #     added as dependencies, as well as all translations of the content_ids
+  #     found.
   class ContentDependencies
-    def initialize(content_id:, fields: [], dependent_lookup: GetDependents.new)
+    def initialize(content_id:, locale:, state_fallback_order:)
       @content_id = content_id
-      @fields = fields
```
kevindew: Yeah that'd be good to reduce the amount it'd need to be worked right across from the request -> worker point. Currently it's redundant as we only pass in the content_id field at a worker point. So I felt it was better to remove it than to leave it as something ambiguous.

kevindew:
ok @dougdroper should be good now 👍

PR #626, opened 2016-12-07
Remove '/grouped_content_and_links' endpoint

thomasleese:
Finding Things are no longer using this endpoint, therefore it should be removed as well as the associated tests.


dougdroper:
Might want to remove the route too. https://github.com/alphagov/publishing-api/blob/master/config/routes.rb#L32

app/controllers/v2/grouped_content_and_links_controller.rb
```diff
@@ -1,26 +0,0 @@
...
-  class GroupedContentAndLinksController < ApplicationController
-    #
-    # Query all content items and links. This can be used to extract data for
-    # batch processing tasks.
-    #
-    # Results are grouped by content id and are paginated using the
-    # last_seen_content_id parameter.
-    #
-    def index
-      query = Queries::GetGroupedContentAndLinks.new(
```
dougdroper: You can remove GetGroupedContentAndLinks

app/controllers/v2/grouped_content_and_links_controller.rb
```diff
@@ -1,26 +0,0 @@
...
-        last_seen_content_id: params[:last_seen_content_id],
-        page_size: params[:page_size],
-      )
-
-      if query.valid?
-        render json: Presenters::Queries::GroupedContentAndLinks.new(query.call).present
```
dougdroper: and this Presenters::Queries::GroupedContentAndLinks

PR #627, opened 2016-12-07
Include withdrawn flag in link expansion

dougdroper:
Passes the unpublishings type through to the web content item, and adds
the field to the link expansion of true when that unpublishing type is
withdrawn

Includes unpublishings of withdrawn in link expansion
Previously we explicitly didn't include withdrawn content item in the
dependents link expansion. We now include them, but there is a flag on
the expanded field that allows a frontend application to filter them out
if needed.

https://trello.com/c/Tv6uNqYq/857-expose-withdrawn-state-for-links


gpeng:
Beautiful... Almost poetic. :+1:

PR #628, opened 2016-12-08
WIP Stricter recursion rules in LinkedTo Query

kevindew:
We hit a problem (like one often does with recursion) that a case that
hadn't been considered properly causes the recursion to run a bit crazy.

The problem was caused by allowing matches of any recursive type and
then sorting out whether it matched a path at the end. When this was
used with `ordered_related_items` it was recursing on links between the
various `ordered_related_items` causing a large depth that was run for
every item and eventually killing PostgreSQL.

The remedy for this is to be more strict with how we check for a valid
link path. This has been done by checking that a link path either ends
with a valid sticky link path or has an exact non sticky path.
Unfortunately this has added to the queries complexity.

There are a couple of items in this commit that could be improved:

1) If there is a way to reference the values used in the recursive
select, we could make the query condition significantly simpler
2) I had a nightmare trying to work out whether an array is included
inside another array with the same order. Array equivalent of
"str".include?("st"). Surely Ruby has such a method?

### DIFF
--- a/app/queries/linked_to.rb
+++ b/app/queries/linked_to.rb
@@ -12,8 +12,11 @@ module Queries
       linked_to(content_id, expansion_rules)
     end
 
+  private
+
     def linked_to(content_id, expansion_rules)
       all_results = results(content_id, expansion_rules.recursive_link_types)
+      # This weeds out full matches from partial matches
       filtered = all_results.select do |result|
         link_type_path = JSON.parse(result["link_type_path"]).reverse
         link_type_path.count == 1 || expansion_rules.valid_link_recursion?(link_type_path)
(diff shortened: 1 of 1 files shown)


app/queries/linked_to.rb
```diff
@@ -12,59 +12,118 @@ def call(content_id:, expansion_rules:)
...
+
+    def quote_array(array)
+      quoted = array.map { |s| connection.quote(s) }.join(",")
+      "ARRAY[#{quoted}]"
+    end
+
+    def combination_inside_array(combination, array)
+      # FIXME: I want to do the equivalent of "str".include?("st") but with
+      # arrays but I don't know a nice method to do it. Please fix if you know
+      # a nicer way
```
steventux: It's six months too late but...
```ruby
array - (array - combination) == combination
```
?

PR #629, opened 2016-12-08
Remove deprecated v1 endpoints

thomasleese:
They should no longer be being used.


dougdroper:
https://github.com/alphagov/publishing-api/blob/master/app/presenters/downstream_presenter.rb#L109

danielroseman:
The tests are failing because the Pacts are still running against the old version of gds-api-adapters; they won't pass until we have bumped the version number there.

app/controllers/content_items_controller.rb
```diff
@@ -1,27 +0,0 @@
-class ContentItemsController < ApplicationController
-  before_action :validate_routing_key_fields, only: [:put_live_content_item]
-
-  def put_live_content_item
-    response = Commands::PutContentWithLinks.call(content_item)
```
dougdroper: should be able to remove PutContentWithLinks, PutDraftContentWithLinks

PR #644, opened 2016-12-13
Fix spec/integration/draft_substitution_spec.rb tests

thomasleese:


app/commands/v2/put_content.rb
```diff
@@ -117,10 +125,10 @@ def find_previously_drafted_content_item
         ).lock.first
       end
 
-      def clear_draft_items_of_same_locale_and_base_path(content_item, locale, base_path)
+      def clear_draft_items_of_same_locale_and_base_path(content_id, document_type, locale, base_path)
```
kevindew: Rather than this taking extra methods I think we could make it so this method has no inputs and put a instance method of `document type` since all the other methods are available on the class. This would then allow us to call this substitution by `clear_draft_items_of_same_locale_and_base_path if content_with_base_path?` - what do you think?

kevindew:
@thomasleese Looks cool 👍  aside from my comment would you be able to squash the merge commits?

app/commands/v2/put_content.rb
```diff
@@ -214,6 +202,10 @@ def content_id
         payload.fetch(:content_id)
       end
 
+      def document_type
+        content_item_attributes_from_payload[:document_type]
```
kevindew: Sorry to be a pain, could we swap this for just `payload[:document_type]` - just keeps it consistent with similar methods

PR #655, opened 2016-12-14
Remove format field

thomasleese:


thomasleese:
I wasn't sure if this change should just be internal or if it should also update the downstream presenter. As it is, I have updated the downstream presenter, but this should be confirmed before merging.

PR #657, opened 2016-12-14
Remove deprecated 'format' field

thomasleese:
I wasn't sure if this change should just be internal or if it should also update the downstream presenter. As it is, I have updated the downstream presenter, but this should be confirmed before merging.


kevindew:
Good stuff @thomasleese I'll have a look into it, could you update the documentation too? I know it's referenced in `doc/api.md` but may be mentioned in another one too.

thomasleese:
The tests should pass here when https://github.com/alphagov/gds-api-adapters/pull/634 is merged.

PR #674, opened 2016-12-16
Remove the special case for the master branch

cbaines:


danielroseman:
Why do we need to do this?

cbaines:
> Why do we need to do this?

The change is a general code improvement.

I'm doing it now, as its a step towards getting the gds-api-adapters build and test setup back in to a good state. I can't be sure, but I think that this change will fix the build for the master branch of gds-api-adapters [1]. I decided to go this way around it as I think treating branches consistently is better. Annoyingly, the gds-api-adapters change built successfully on the branch, but failed when merged in to master.

1: https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/master/

PR #680, opened 2016-12-20
Add benchmark for lookup-by-base-path

rubenarakelyan:
This commit adds a benchmark for `lookup-by-base-path`.

Trello: https://trello.com/c/oVMZWWXN/369-content-tagger-fix-bulk-tag-search-timing-out

### DIFF
--- a/bench/lookup_by_base_path.rb
+++ b/bench/lookup_by_base_path.rb
@@ -1,0 +1,31 @@
+# /usr/bin/env ruby
+
+require ::File.expand_path('../../config/environment', __FILE__)
+
+require 'benchmark'
+
+require 'stackprof'
+
+abort "Refusing to run outside of development" unless Rails.env.development?
+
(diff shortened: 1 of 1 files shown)


kevindew:
This looks cool - however we are on the verge of removing the code that this benchmarks (ContentItemFilter) so not sure if you want to hold fire for a bit or to benchmark against the upcoming branch?

rubenarakelyan:
Ah, well ideally we'd like before and after benchmarks for the same action to be able to compare performance once the code is changed. `ContentItemFilter` is being replaced with simple calls to `ContentItem` right?

kevindew:
Yup that's correct. We'll get an amended version of this into the PR for the change.

rubenarakelyan:
Great thanks!

rubenarakelyan:
Oops, closed by mistake!

PR #689, opened 2016-12-21
Clearer schema errors

kevindew:
Since our switch to remove format a couple of people have been caught out with the errors they have got from the Publishing API in regards to validating schemas.

In dev for a missing schema field you get an error of: "Unable to find schema for schema_name  Ensure GOVUK_CONTENT_SCHEMAS_PATH env variable is set and points to the root directory of govuk-content-schemas" which can be confusing as their schemas may be set up correctly and it is the missing field that causes the problem.

This change makes the reasons a user has hit an error clearer.

### DIFF
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -28,10 +28,12 @@ private
 
   def schema
     @schema || find_schema
+  rescue NoSchemaNameError
+    errors << no_schema_name_message
+    {}
   rescue Errno::ENOENT => error
     if Rails.env.development?
       errors << missing_schema_message
-      errors << dev_help if GovukSchemas::Schema.all.empty?
     end
     Airbrake.notify(error, parameters: {
       explanation: missing_schema_message,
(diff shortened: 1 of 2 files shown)


app/validators/schema_validator.rb
```diff
@@ -28,10 +28,13 @@ def valid?
...
   def schema
     @schema || find_schema
+  rescue NoSchemaNameError
+    errors << no_schema_name_message
+    {}
   rescue Errno::ENOENT => error
     if Rails.env.development?
       errors << missing_schema_message
-      errors << dev_help
+      errors << dev_help if GovukSchemas::Schema.all.empty?
```
cbaines: I think this will be redundant, as at least in the "development" RAILS_ENV, the Publishing API service will fail to start if this is the case https://github.com/alphagov/publishing-api/blob/master/config/environments/development.rb#L46

kevindew: @cbaines ok done.

cbaines: Could you get rid of the dev_help also, as I think that is unneeded now?

kevindew: d'oh! - thanks

PR #692, opened 2016-12-22
Create and dual-populate documents

thomasleese:
Create a table to store documents and dual-populate it with content item data when performing saves, but continue to use the data in the content items table as the canonical source in queries.

### DIFF
--- a/db/migrate/20161220134413_create_documents.rb
+++ b/db/migrate/20161220134413_create_documents.rb
@@ -1,0 +1,10 @@
+class CreateDocuments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :documents do |t|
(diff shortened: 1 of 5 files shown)


db/migrate/20161220134413_create_documents.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDocuments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :documents do |t|
+      t.string :content_id, null: false
```
kevindew: We can use a uuid type rather than a string type here

thomasleese:
This has diverted from the `documents-and-editions` a lot, will create a new PR.

PR #695, opened 2016-12-23
Ensure access limits are not sent for live items.

danielroseman:
Although the publish command destroys any access limit items, we have
seen occasions when the access_limited hash has been populated in the
live content store. This ensures that it will never be sent for
published items, and notifies airbrake if it encounters this situation.

### DIFF
--- a/app/presenters/downstream_presenter.rb
+++ b/app/presenters/downstream_presenter.rb
@@ -47,11 +47,19 @@ module Presenters
 
     def access_limited
       return {} unless access_limit
-      {
-        access_limited: {
-          users: access_limit.users
+      if web_content_item.state != 'draft'
+        Airbrake.notify(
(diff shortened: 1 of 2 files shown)


deanwilson:
Looks good.

app/presenters/downstream_presenter.rb
```diff
@@ -47,11 +47,19 @@ def links
 
     def access_limited
       return {} unless access_limit
-      {
-        access_limited: {
-          users: access_limit.users
+      if web_content_item.state == 'published'
```
kevindew: It's possible for an item to be unpublished here as well, I think it'd be safer to use `state != 'draft'`

danielroseman: Changed.

PR #704, opened 2017-01-04
Switch to using native UUID types in the database

thomasleese:
This was implemented during the _Documents and Editions_ work, but has been spun out into its own PR here to deploy in stages.

Migrations take about 5 minutes in total to run.

### DIFF
--- a/db/migrate/20161220090000_remove_invalid_content_ids.rb
+++ b/db/migrate/20161220090000_remove_invalid_content_ids.rb
@@ -1,0 +1,14 @@
+class RemoveInvalidContentIds < ActiveRecord::Migration[5.0]
+  def up
+    events_to_remove = [20, 41, 83, 51, 42, 61, 40, 44, 41, 39, 42, 40, 26, 83,
+                        89, 88, 90, 86, 88, 87, 89, 90, 87, 92, 26, 39, 61, 20,
+                        44, 51, 29, 18, 94, 86, 87, 94, 18, 29, 89, 90]
+
+    Event.where(content_id: events_to_remove).delete_all
+
+    link_sets_to_remove = [40, 42, 26, 39, 61, 20, 41, 83, 51, 44, 90, 88, 29,
+                           18, 86, 89, 87, 94, 92]
+
+    LinkSet.where(content_id: link_sets_to_remove).delete_all
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20161220090000_remove_invalid_content_ids.rb
```diff
@@ -0,0 +1,14 @@
...
+    events_to_remove = [20, 41, 83, 51, 42, 61, 40, 44, 41, 39, 42, 40, 26, 83,
+                        89, 88, 90, 86, 88, 87, 89, 90, 87, 92, 26, 39, 61, 20,
+                        44, 51, 29, 18, 94, 86, 87, 94, 18, 29, 89, 90]
+
+    Event.where(content_id: events_to_remove).destroy_all
+
+    link_sets_to_remove = [40, 42, 26, 39, 61, 20, 41, 83, 51, 44, 90, 88, 29,
+                           18, 86, 89, 87, 94, 92]
+
+    LinkSet.where(content_id: link_sets_to_remove).destroy_all
```
kevindew: It'd be better to use `delete_all` rather than `destroy_all` here. `delete_all` runs a "DELETE query" whereas `destroy_all` runs one "SELECT" and then individual DELETE queries one by one (as it also runs model callbacks).

db/migrate/20161220090001_make_content_id_uuid_type.rb
```diff
@@ -0,0 +1,15 @@
+class MakeContentIdUuidType < ActiveRecord::Migration[5.0]
+  def up
+    change_column :content_items, :content_id, 'UUID USING content_id::uuid'
+    change_column :events, :content_id, 'UUID USING content_id::uuid'
+    change_column :link_sets, :content_id, 'UUID USING content_id::uuid'
+    change_column :links, :target_content_id, 'UUID USING target_content_id::uuid'
```
kevindew: ah interesting that it's necessary to cast these. 👍

spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -364,7 +364,7 @@
 
     context "for a non-existent content item" do
       it "responds with 404" do
-        get :show, params: { content_id: "missing" }
+        get :show, params: { content_id: SecureRandom.uuid }
```
kevindew: What happens in the case that you do pass a non-uuid in as a parameter?

thomasleese: I've added a test for this -- you also get a `404` error code.

PR #683, opened 2016-12-20
Switch fully to documents and editions

thomasleese:
Second part of the documents and editions experience. Here we remove the `locale` and `content_id` columns from content items and switch to using only documents as the canonical source of data.

There are no migrations so this should be easy to deploy after #705.

PR #705, opened 2017-01-04
Create and dual populate documents

thomasleese:
This PR is the first set of changes from the _Documents and Editions_ work. It creates a new document model, runs a migration to populate it with existing data (which only takes a few seconds), and then dual-populates it with new data while still using the existing `content_id` and `locale` columns from `content_items` as the canonical source of that data.

Currently, complete migrations take about 5-10 minutes to run where, for approximately 1 minute, the table is locked.


spec/presenters/queries/link_set_presenter_spec.rb
```diff
@@ -2,15 +2,16 @@
 
 RSpec.describe Presenters::Queries::LinkSetPresenter do
   describe ".present" do
+    let(:content_id) { SecureRandom.uuid }
+    let(:link_set) { FactoryGirl.create(:link_set, content_id: content_id) }
+
     before do
       FactoryGirl.create(:lock_version, target: link_set, number: 101)
       @result = Presenters::Queries::LinkSetPresenter.present(link_set)
```
kevindew: I think it'd be more rspec idiomatic to move this to be the subject e.g `subject(:result) { Presenters... }`

db/migrate/20161221074640_link_documents_to_content_items.rb
```diff
@@ -0,0 +1,6 @@
+class LinkDocumentsToContentItems < ActiveRecord::Migration[5.0]
+  def change
+    add_column :content_items, :document_id, :integer
+    add_foreign_key :content_items, :documents
```
kevindew: Do we need to add a non-null constraint migration to go out as part of the documents and editions branch?

thomasleese: https://github.com/alphagov/publishing-api/pull/683/files#diff-2659c07fbcce1876783255b492a6d187R9 😄

app/models/content_item.rb
```diff
@@ -96,6 +100,20 @@ class ContentItem < ApplicationRecord
     end
   end
 
+  before_validation { ensure_document }
+  before_save { ensure_document }
+
+  def document_requires_updating?
+    !document || (changes[:locale] || changes[:content_id]) || (document.locale != locale || document.content_id != content_id)
```
kevindew: Do we need to check the changes hash when we also have the `document.` checks?

db/migrate/20170106122400_link_documents_and_editions.rb
```diff
@@ -0,0 +1,10 @@
+class LinkDocumentsAndEditions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :content_items, :document_id, :integer
+    add_foreign_key :content_items, :documents
+
+    add_index :content_items, [:document_id, :user_facing_version], unique: true
+    add_index :content_items, [:document_id, :content_store], unique: true
+    add_index :content_items, [:document_id, :state]
```
kevindew: These 3 will be locking as is, but can be changed to use the concurrent algorithm to not be so - though you have to disable the ddl transaction

db/migrate/20170106122400_link_documents_and_editions.rb
```diff
@@ -0,0 +1,10 @@
+class LinkDocumentsAndEditions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :content_items, :document_id, :integer
+    add_foreign_key :content_items, :documents
```
kevindew: Is this a fast query?

thomasleese: It took 38 seconds to run on my machine, it also locks the table which I didn't realise.

app/models/lock_version.rb
```diff
@@ -2,6 +2,11 @@ class LockVersion < ApplicationRecord
   include Version
   belongs_to :target, polymorphic: true
 
+  after_save do
+    item = lock_version_target
+    item.update_column(:lock_version, number) if number > item.lock_version
```
kevindew: probably wise to have a check that item is not nil here just in case?

db/migrate/20170106122400_link_documents_and_editions.rb
```diff
@@ -0,0 +1,12 @@
+class LinkDocumentsAndEditions < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :content_items, :document_id, :integer
+    add_foreign_key :content_items, :documents
```
kevindew: These queries should be within a transaction, so I think a separate migration for them would be better.

I believe we also don't get an index implicitly by adding a foreign key so should add an index here too

db/migrate/20170109161435_move_lock_version_data.rb
```diff
@@ -0,0 +1,30 @@
+class MoveLockVersionData < ActiveRecord::Migration[5.0]
+  def up
+    execute "UPDATE link_sets
+             SET lock_version = t.number
+             FROM (SELECT target_id, number FROM lock_versions) t
+             WHERE link_sets.id = t.target_id"
+
+    execute "UPDATE documents
```
kevindew: Do you think if we try run a find_each and loop through each record this might be less of a strain? so tons of short queries rather than one slow one?

spec/presenters/queries/link_set_presenter_spec.rb
```diff
@@ -5,17 +5,17 @@
     let(:content_id) { SecureRandom.uuid }
     let(:link_set) { FactoryGirl.create(:link_set, content_id: content_id) }
 
-    before do
+    subject(:result) do
       FactoryGirl.create(:lock_version, target: link_set, number: 101)
       @result = Presenters::Queries::LinkSetPresenter.present(link_set)
```
kevindew: can lose the `@result = ` here ^_^

app/models/lock_version.rb
```diff
@@ -2,6 +2,12 @@ class LockVersion < ApplicationRecord
   include Version
   belongs_to :target, polymorphic: true
 
+  after_save do
+    item = lock_version_target
+    return unless item
```
kevindew: I think usage of return here could have wider consequences than we expect. If I remember correctly (and I might not be) this would return from the method that is calling this block, whereas if we use next it will terminate this block.

fofr:
Where can I find more details on the _Documents and Editions_ work? ie What it entails and why we're doing it?

app/models/document.rb
```diff
@@ -0,0 +1,10 @@
+class Document < ApplicationRecord
+  has_many :content_item
+
+  validates :content_id, presence: true, uuid: true
```
dougdroper: -> uuid?

app/models/document.rb
```diff
@@ -0,0 +1,10 @@
+class Document < ApplicationRecord
+  has_many :content_item
```
dougdroper: content_items

app/models/content_item.rb
```diff
@@ -96,6 +100,20 @@ class ContentItem < ApplicationRecord
     end
   end
 
+  before_validation { ensure_document }
```
dougdroper: do we need this?

db/migrate/20170106122900_change_lock_version_defaults.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeLockVersionDefaults < ActiveRecord::Migration[5.0]
+  def up
+    change_column_default :link_sets, :stale_lock_version, 0
```
dougdroper: 0?

app/models/lock_version.rb
```diff
@@ -2,6 +2,12 @@ class LockVersion < ApplicationRecord
   include Version
   belongs_to :target, polymorphic: true
 
+  after_save do
+    item = lock_version_target
+    next unless item
+    item.update_column(:stale_lock_version, number) if number > (item.stale_lock_version || -1)
```
dougdroper: if we defaulted to 1, we could make this simpler

PR #717, opened 2017-01-09
Set locale to 'cy' for Welsh content from Publisher

davidbasalla:
We set the correct locale for these content items that are written in Welsh as a data migration.

We get the list of slugs from Publisher by running:
```
Artefact.where(language: "cy").order_by(:slug => "asc").map(&:slug)
```

We then get the content_ids from the PublishingApi by running:
```
welsh_slugs.each do |slug|
  content_ids << ContentItem.where(base_path: "/#{slug}").last.content_id
end
```

### DIFF
--- a/db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
+++ b/db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
@@ -1,0 +1,56 @@
+class FixWelshPublisherContentLocales < ActiveRecord::Migration[5.0]
+  def up
+    welsh_content_ids = [
+      "12cd2a97-e005-4f99-a4d1-a5e9fa1c4415",
+      "b394bbec-c8f8-4b50-adbc-840b630cc007",
+      "7fe4dd51-210d-4468-a988-99a742bfdc5b",
+      "d1d87169-3284-43c6-96e5-dd0cbb8cf89f",
+      "275a9cbc-62e9-48d8-82d6-ef460f66bc5f",
+      "976ba497-dc16-4a2c-b103-f3e153d51192",
+      "2176cab5-7600-493d-9e54-06c7f27297c5",
+      "139577cc-8eb6-4f1f-a170-7c456f03f6b1",
+      "4eda604d-76f1-4455-8281-5033c1c8596a",
+      "a661e67e-52bf-4d67-a0fc-5be231255a76",
+      "085e4c8b-0252-4cc3-af2f-db0edaebf539",
+      "3b4755b3-df53-4a5d-a8f1-8c8c20aadb46",
+      "6c1d8527-ea49-44c8-8525-21295889584e",
+      "0ae4f2a6-56d7-4728-b8de-18829475bf74",
+      "ee9da799-1f51-4397-8470-20af9e87fc12",
+      "15075444-4e7b-4096-b81a-6dcc9bece0b7",
+      "d732fb0d-1578-4d2d-8160-e3159338b554",
+      "70c0579a-87b5-49b8-b888-ad7135b97ca3",
+      "7cd5aa8a-b2b6-4c75-9fd1-447a96f8f303",
+      "867b6eea-0b95-4a48-84e3-960eca540521",
+      "e56a47a0-9dcb-413a-85ab-fa9980b3aa63",
+      "ff1a3eae-3daa-4167-a702-1898f32721ab",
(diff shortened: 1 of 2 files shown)


db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
```diff
@@ -0,0 +1,59 @@
...
+      "talu-gorchymyn-cynhaliaeth-ar-lein",
+      "taluch-bil-treth-hunanasesiad",
+      "treth-car",
+      "treth-incwm",
+      "twyll-budd-dal",
+      "yswiriant-gwladol",
+    ]
+
+    welsh_slugs.each do |slug|
+      content_items = ContentItem.where(base_path: "/#{slug}")
```
kevindew: It'd be good to including the `publishing_app` in this where condition - just to ensure this migration doesn't accidentally affect content from any other publishing apps?

db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
```diff
@@ -0,0 +1,59 @@
...
+      content_items.each do |c|
```
kevindew: I think `.find_each` would be a better choice here

db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
```diff
@@ -0,0 +1,59 @@
...
+        c.locale = 'cy'
+        c.save
```
kevindew: can probably get these to be just `update!(locale: "cy")` and save a line and also have it blow up if there's a validation conflict.

db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
davidbasalla: Sorry, might be obvious, but is it better because it does it in batches?

kevindew: Sorry - I got in a muddle and my comment doesn't apply. Please ignore.

My suggestion here was meant to be, grab all the items and loop them. eg `ContentItem.where(base_path: welsh_slugs, publishing_app: "your-app").each`  no need for `find_each` for a collection this size - not sure how my brain jumped to `f.ind_each` I blame Monday 😁

dougdroper: could do ContentItem.where(base_path: welsh_slugs).update_all(locale: 'cy')

kevindew: @dougdroper I was hesitant to recommend that due to our temporary change listeners.

dougdroper: cool

davidbasalla:
@kevindew have addressed your comments and tested this successfully in Integration!

kevindew:
Great thanks @davidbasalla 👍

davidbasalla:
@kevindew sorry could you have another look? We added a second commit which aims to send the updated content downstream as part of the data migration. Do you see any problems with this approach or would you prefer for us to run the `represent_downstream` rake task manually? If you're happy we'll squash the second commit into the first...

db/migrate/20170109151629_fix_welsh_publisher_content_locales.rb
```diff
@@ -0,0 +1,58 @@
...
+      "c970c404-8025-452f-afe9-c7cac002b047",
+      "eabe3d14-0e3c-4186-bf48-6c6ed11c58a7",
+    ]
+
+    welsh_content_ids.each do |content_id|
+      ContentItem.where(publishing_app: "publisher", content_id: content_id).order(user_facing_version: "asc").each do |c|
+        c.update!(locale: 'cy')
+      end
+
+      Commands::V2::RepresentDownstream.new.call(content_id)
```
dougdroper: You could do the RepresentDownstream and pass the welch_content_ids. and possible pass true to populate the draft content store? 
Also could do ContentItem.where(publishing_app: "publisher", content_id: welsh_content_ids) to stop the multiple queries?

davidbasalla: thanks @dougdroper, have addressed your comments! No objections to calling `RepresentDownstream` in the migration?

dougdroper: No objections 👍

PR #719, opened 2017-01-10
[WIP] Rename content items to editions

thomasleese:
The final part in the _documents and editions_ trilogy is to rename `ContentItems` to `Editions`.

This should be merged after #683.


thomasleese:
Will be re-opened when more appropriate to do so (code has diverted from `documents-and-editions` branch).

PR #722, opened 2017-01-11
Delete `document_collections` links

gpeng:
Whitehall used to send these but they are no longer required as the `document_collection` is inferred from the inverse `document` link type.


kevindew:
@gpeng could you include db/schema.rb with future migrations? We just got caught out here that this broke the end-to-end tests due to tests failing to run with pending migrations due to schema being behind this

PR #728, opened 2017-01-13
Second unpublishing doesn't set state to superseded

kevindew:
This fixes a bug that we found in the Publishing API where a 2nd
unpublishing of a content_item causes the state to be set to superseded.

### DIFF
--- a/spec/commands/v2/unpublish_spec.rb
+++ b/spec/commands/v2/unpublish_spec.rb
@@ -374,6 +374,11 @@ RSpec.describe Commands::V2::Unpublish do
 
       include_examples "creates an action"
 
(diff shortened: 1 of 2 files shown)


spec/commands/v2/unpublish_spec.rb
```diff
@@ -374,6 +374,12 @@
 
       include_examples "creates an action"
 
+      it "maintains the state of unpublished" do
+        described_class.call(payload)
+        unpublished_content_item.reload
+        expect(unpublished_content_item.state).to eq("unpublished")
```
dougdroper: could just do `expect(unpublished_content_item.reload.state).to eq("unpublished")` to drop the line above

PR #729, opened 2017-01-17
Put content cleanup

dougdroper:
Some refactoring of put content since PutContent as it stands is difficult to follow and reason about. 

Split validation into own object. 
Splitting out different responsibilities into separate classes. creating/updating/redirecting
Move some hidden logic in place to make things clearer. 

Hopefully this will make it easier to refactor more and understand the flow. There are still a few more things that we can do (observers, PORO etc), but will leave it for now.


app/models/lock_version.rb
```diff
@@ -30,9 +30,27 @@ def lock_version_target
     end
   end
 
+  def increment
+    self.number += 1
+  end
+
+  def increment!
+    self.number += 1
```
danielroseman: For neatness, this could call `increment`.

dougdroper: 👍

app/commands/v2/previously_published_item.rb
```diff
@@ -0,0 +1,77 @@
+module Commands
```
dougdroper: Yes definitely, will think of a better place

kevindew:
looks good 👍 - as saying in person it would be nice if we could define a hash of defined defaults for a content_item payload (potentially interspersed with previous values) to merge into as the payload. As opposed to the situation we have currently where we're using the ContentItem model defaults, then merging in payload, then mixing in previous values. Hopefully that sentence made sense.

app/commands/v2/previously_published_item.rb
danielroseman: You didn't move app/commands/v2/previously_published_item.rb, which was the one I was most concerned about...

PR #731, opened 2017-01-18
Add a migration to fix a whitehall Person object

h-lame:
For: https://trello.com/c/tP9fBbJU/92-people-profile-not-showing-on-gov-uk

This object was created and published, then deleted, then recreated
in whitehall.  Unfortunately it can't be published again because the
old object still lives in publishing-api with a different content_id
to the new object.  This migration updates the old object to share a
content_id with the new object and this allows the object to be published
from whitehall again.  We also fix the routes because the object is
a placeholder when published (routes not managed) and a gone when deleted
(routes are managed).

More detail about the context and approach is provided in the migration itself.

### DIFF
--- a/db/migrate/20170118104115_fix_john_cordery_content_items.rb
+++ b/db/migrate/20170118104115_fix_john_cordery_content_items.rb
@@ -1,0 +1,79 @@
+# The History
+# ===========
+#
+# In whitehall a Person object was created for John Corderoy on 29th Nov
+# 2016.  The base path for this Person is /government/people/john-corderoy,
+# and the content_id is 27b08e2d-1bd9-428d-a6c9-c6ffa40d166a.  Like all
+# Person objects this was sent to publishing-api immediately upon
+# creation and published, registering the route with router in the
+# process.  For some reason it was almost immediately deleted from
+# whitehall and publishing-api told to unpublish the content item (telling
+# router to 410 the path in the process).
+#
+# A couple of minutes later a new Person object was created for John which
+# has the same base path but a different content_id:
+#     eab84eb0-6a02-4d50-9644-0900e7b8cee7
+# When saved in whitehall this Person is drafted onto the publishing-api,
+# but cannot be published, because it's trying to reuse the same base_path
+# as an existing content_item with a different content_id.
+
+# If this new content item was published then the publishing-api and
+# content-store would be in sync with the whitehall db and things would
+# be better.  However, visiting the url for John's page would still get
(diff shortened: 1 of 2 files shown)


NeilvB:
This looks good to me but would be good to another pair of eyes too

db/migrate/20170118104115_fix_john_cordery_content_items.rb
```diff
@@ -0,0 +1,78 @@
...
+# same as the content_id of the new content_item (and delete the old Document).
+# This lets the whitehall, publishing-api, and content-store db's sync up, and
+# then we delete the old route (because it shouldn't be there any more).
+class FixJohnCorderyContentItems < ActiveRecord::Migration[5.0]
+  def up
+    old_content_id = '27b08e2d-1bd9-428d-a6c9-c6ffa40d166a'
+    new_content_id = 'eab84eb0-6a02-4d50-9644-0900e7b8cee7'
+
+    new_content_item = ContentItem.where(content_id: new_content_id).first
+    old_content_item = ContentItem.where(content_id: old_content_id).first
```
kevindew: Is there any risk that new ContentItems can be added for either of these items in the time between you writing this migration and deploying? just thinking that using `.first` would not get the latest one.

db/migrate/20170118104115_fix_john_cordery_content_items.rb
```diff
@@ -0,0 +1,78 @@
...
+#
+# The problme is that when it was deleted we published a "gone" item which
+# content-store does manage routes for.  When it is subsequently re-published
+# it goes back to being a "placeholder" and content-store doesn't manage the
+# route, leaving the 410 route in place.
+#
+# The solution is to change the content_id of the old content_item to be the
+# same as the content_id of the new content_item (and delete the old Document).
+# This lets the whitehall, publishing-api, and content-store db's sync up, and
+# then we delete the old route (because it shouldn't be there any more).
```
kevindew: 💯 for the story 😀

db/migrate/20170118104115_fix_john_cordery_content_items.rb
h-lame: I don't think there's a risk.  For the old one, there's no object in whitehall that could write to it, because deleting a Person fully deletes it from the db (no soft-delete).  For the new one, it can't be published because of the clash with the old one and so any updates that do happen would be written to the existing publishing api - based on my understanding that publishing api won't create a new content item if the current one is draft.

PR #733, opened 2017-01-20
Add Appsignal configuration

carvil:
This adds Appsignal to the app. We're evaluating to see if it is a good
replacement for Errbit, while also adding performance tracking.

It's only activated if APPSIGNAL_API_KEY is set, which is currently only in
integration (alphagov/govuk-puppet#5376
https://github.gds/gds/deployment/pull/1215).

https://trello.com/c/J3kDeZD7

cc/ @tijmenb

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,9 @@ GEM
       airbrake-ruby (~> 1.5)
     airbrake-ruby (1.5.0)
     amq-protocol (2.0.1)
+    appsignal (2.0.5)
+      rack
+      thread_safe
     arel (7.1.4)
     ast (2.3.0)
     awesome_print (1.7.0)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -437,4 +441,4 @@ DEPENDENCIES
   whenever (= 0.9.4)
 
 BUNDLED WITH
-   1.13.1
+   1.13.5
```
dougdroper: we might have to leave this at 1.13.1

carvil: Ok, I can put it back. What is the reason for that?

tijmenb:
@kevindew not sure yet! Shout if you want access to Appsignal.

Gemfile.lock
dougdroper: 1.13.1 is the version on CI

dougdroper: oh I lie, it is using 1.11.2. 

Maybe it is fine then.

carvil: I will put it back anyway, no need to change that version together with this change. Thanks for looking!

carvil:
@dougdroper I've updated the commit now

PR #734, opened 2017-01-20
Make 'Edition' available as an alias for 'ContentItem'

thomasleese:
This makes it easier to rollback from the `documents-and-editions`
branch by making sure that any newly created LockVersions with the
polymorphic target field set to a type of 'Edition' work.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -1,0 +1,4 @@
+# Needed to work around the fact that we use a polymorphic type and to avoid
+# including a migration which changed the class name in the database, we simply
+# make sure it is available under both names
+Edition = ContentItem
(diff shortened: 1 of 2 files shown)


app/models/content_item.rb
```diff
@@ -238,3 +238,8 @@ def requires_rendering_app?
     renderable_content? && document_type != "contact"
   end
 end
+
+# Needed to work around the fact that we use a polymorphic type and to avoid
+# including a migration which changed the class name in the database, we simply
+# make sure it is available under both names
+Edition = ContentItem
```
kevindew: Concern here is that if you try to load an Edition before any ContentItems have been loaded you'll get a `NameError: uninitialized constant ContentItem` since this is not on the edition autoloading path.

You could get around this by having this edition definition defined in a app/models/edition.rb file.

PR #735, opened 2017-01-23
Documents and Editions Documentation

thomasleese:
Work on improving the documentation following _Documents and Editions_ will go in this PR.


doc/link-expansion.md
```diff
@@ -279,16 +277,16 @@ By default links contain the following fields:
...
-- `description` - A short description of the content item
-- `document_type` - The document of the content item
-- `locale` - The language the content item linked to is in
-- `public_updated_at` - The date/time that a content item was last changed
-- `schema_name` - The [GOV.UK content schema][govuk-content-schema] that a
-  content item conforms to
-- `title` - The title of a content item
+- `content_id` - A UUID to represents the document
+- `description` - A short description of the content
+- `document_type` - A document type - often identical to schema_name
```
MatMoore: It would be helpful for this to explain the difference between the two, or link to somewhere else that already does.

In my mind document type influences how we present the document to users (publishers, and to a lesser extent end users) whereas schema name describes data. But the terminology is really confusing (especially considering "format" and "template" are also used a lot in different contexts)

MatMoore:
LGTM - I really like this change to the data model, it makes it much easier to understand.

One question: in https://github.com/alphagov/publishing-api/blob/e35f083e7830088d76f5dc91ec6af6f0cab0f340/doc/model/object-model.png
it shows a `LinkSet` belonging to an `Edition` rather than a `Document`. Is this correct?

I would have expected it to belong to the document because a link is only a relationship between two content ids and doesn't change when you create new editions.

doc/link-expansion.md
kevindew: @MatMoore Sure - A link sounds a good idea. We've struggled explaining document_type in the past as it's quite a loose concept. Would be good to get it defined well in the docs.

kevindew:
Thanks for checking this out @MatMoore and that diagram is incorrect indeed - thanks for pointing that out.

README.md
```diff
@@ -8,91 +8,94 @@ existing piece of content.
...
+- **Content Item**: A representation of content that can be sent to a
+  [content store][content-store].
+- **Link Set**: A collection of links to other documents. It is used to
+  capture relationships between pieces of content (e.g. parent/child)
+- **Unpublishing**: An object indicating a previously published edition
+  which has been removed from the live site.  Can be "gone", "withdrawal", or
+  "redirect".
 - **User**: A user of the system, which is used to restrict what content is
-returned from the API as well as prevent certain actions on content
+  returned from the API as well as prevent certain actions on content.
```
danielroseman: I don't think we restrict get_content by user any more.

doc/arch/adr-005-documents-and-editions.md
```diff
@@ -0,0 +1,106 @@
+# Decision Record: Resolve consistency and uniqueness in content
+
+## Context
+
+Publishing API has adopted the open/closed principal where the domain model is
```
danielroseman: typo: principle

doc/model.md
```diff
@@ -1,516 +1,247 @@
...
+[/put-content/:content_id][put-content-api] API.
 
-Content items will be published to the content store regardless of their
-`phase`.  If a content item has a `phase` of either *alpha* or *beta*, a visual
-component will be added to the front-end to show this (assuming the front-end
-application for that format supports this).
+Key fields that are set internally by the Publishing API are:
 
-There is more information on what each of the phases mean [here](https://www.gov.uk/service-manual/phases).
+- `state` - where an edition is in it's [publishing workflow](#workflow), can
```
danielroseman: its

doc/model.md
```diff
@@ -1,516 +1,247 @@
...
+  be "draft", "published", "unpublished" or "superseded".
+- `content_store` - indicates whether an edition is on the draft or live
+  content store, or empty to indicate that it is not on a content store.
```
danielroseman: Could probably do with some more detail here. If something has a value of "live", it may be in both content stores.

doc/model.md
```diff
@@ -1,516 +1,247 @@
...
-We based this object model on the needs of our users. In the Publishing Platform
-team, we are in a unique position such that the majority of our users are in
-fact other developers working on their respective publishing applications.
+A published edition can be unpublished, which will create an
+[`unpublishing`](#unpublishing) for the edition. The unpublished edition will
+be represented on the live content store.
 
-In summary, this object model aims to address the following needs:
+If a draft is published while there is already a published or unpublished
+edition. The previous edition will have it's `state` updated to "superseded"
```
danielroseman: its

doc/publishing-application-examples.md
```diff
@@ -1,30 +1,46 @@
...
 ```
   access_limited : ["8f3173bb-ee2a-48a2-aa86-f2f030830888"]
 ```
 
 ## Draft and Publish endpoint flow
 
-The Publishing API prescribes a draft-to-live workflow where the publishing application submits new draft content, this content may then be updated via the same endpoint and finally the content item can be published in a discreet endpoint request.
+The Publishing API prescribes a draft-to-live workflow where the publishing
+application submits new draft content, this content may then be updated via the
+same endpoint and finally can be published in a discreet
```
danielroseman: discrete

kevindew:
Thanks @danielroseman - I've addressed all your comments - and made a couple more tweaks.

doc/model.md
```diff
@@ -1,516 +1,253 @@
...
+
+- [Introduction](#introduction)
+  - [content_id](#content_id)
+  - [Diagram](#diagram)
+- [Content](#content)
+  - [Document](#document)
+  - [Edition](#edition)
+    - [Workflow](#workflow)
+    - [Uniqueness](#uniqueness)
+    - [Substitutiion](#substitution)
```
thomasleese: Should be 'Substitution'.

kevindew: Thanks, got it

PR #737, opened 2017-01-23
Remove all existing taxon tags

MatMoore:
The education taxonomy is currently very alpha - it's not visible to the public
yet.

The finding things team has just completed a bulk tagging exercise with
publishers so we want to clear all existing tags and import the new ones.

This change removes those tags (links with type `taxons`).

There will be a follow up change to clear out remaining taxon content items,
that do not belong to the current taxonomy (i.e. there is no chain of
`parent_taxon` links linking them).

### DIFF
--- a/db/migrate/20170123111649_clear_alpha_taxons.rb
+++ b/db/migrate/20170123111649_clear_alpha_taxons.rb
@@ -1,0 +1,5 @@
+class ClearAlphaTaxons < ActiveRecord::Migration[5.0]
+  def up
+    Link.where(link_type: "taxons").where("created_at < '2017-01-23'").delete_all
+  end
+end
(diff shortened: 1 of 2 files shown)


MatMoore:
@kevindew There's no schema changes, just a delete on the links table :)

kevindew:
@MatMoore despite that we still hit a problem if the db/schema.rb timestamp is not up to date with the last migration.

The scenario is as follows:
- We have a [config](https://github.com/alphagov/publishing-api/blob/master/config/environments/development.rb#L35) that prevents you running the app in development with pending migrations
- We have the [end-to-end test suite](https://github.com/alphagov/publishing-platform-docker) that runs in development environment
- This builds the db based on the timestamp in `db/schema.rb` and then these fail because Publishing API has a pending migration

which probably goes some way to explaining why we're hoping to separate out data migrations from schema ones 😉

MatMoore:
Oops - sorry, I missed that rails was still updating that version number. Have added that in.

kevindew:
Late to the party - but great branch name 😀

PR #738, opened 2017-01-23
Expose the unpublishing information through GetContentCollection

cbaines:
Adds a new supported field "unpublishing" to
GetContentCollection. If the content item is unpublished, this will
contain the type, explanation, alternative_path and unpublished_at
information.

This will enable publishing applications using the Publishing API to
get back the information they put in when they unpublish content, and
was motivated by this requirement for the Maslow application.

https://trello.com/c/MhXnrU1C/29-contentitempresenter-should-include-unpublishing-details

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -10,6 +10,7 @@ module Presenters
         *Edition::TOP_LEVEL_FIELDS,
         :publication_state,
         :content_id,
+        :unpublishing,
         :locale,
         :lock_version,
         :updated_at,
@@ -64,7 +65,13 @@ module Presenters
       end
 
       def join_supporting_objects(scope)
-        ChangeNote.join_editions(scope)
+        scope = ChangeNote.join_editions(scope)
+
+        if fields.include? :unpublishing
+          scope = Unpublishing.join_editions(scope)
+        end
+
+        scope
       end
 
       def order_and_paginate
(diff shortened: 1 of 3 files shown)


danielroseman:
Can you add some benchmark results before and after (from bench/get_content_items.rb) to ensure that this doesn't affect performance?

cbaines:
No problem, I've posted the results below. Looking at them, I don't think this change makes any difference to the performance of this benchmark.

Before:
```
..........  0.380000   0.100000   0.480000 (  8.297346)
..........  0.450000   0.120000   0.570000 (  8.941289)
..........  0.490000   0.140000   0.630000 (  8.759787)
```

After:
```
..........  0.380000   0.060000   0.440000 (  8.026165)
..........  0.400000   0.090000   0.490000 (  8.333924)
..........  0.420000   0.070000   0.490000 (  8.378935)
```

app/presenters/queries/content_item_presenter.rb
```diff
@@ -64,7 +65,16 @@ def latest
       end
 
       def join_supporting_objects(scope)
-        ChangeNote.join_editions(scope)
+        scope = ChangeNote.join_editions(scope)
+
+        if fields.include? :unpublishing
+          scope = scope.joins(
```
thomasleese: For consistency with other models, it might be good to implement this as an `Unpublishing.join_editions` method.

cbaines: Ok, I've updated this branch with that change :)

cbaines:
Bump, it would be useful to get this merged, and I'm not aware that it is waiting on anything?

PR #740, opened 2017-01-25
Run rcov by default

elliotcm:
I originally made rcov optional on local machines
(https://github.com/alphagov/publishing-api/commit/4874c5aaab9b12c2882414fa58456857b716f328) as it was causing a big delay at the end of
running any tests, and was rarely used locally.

Following some advice on the simplecov issue I raised back in May
(https://github.com/colszowka/simplecov/issues/499#issuecomment-275237948), 
adding a filter to exclude vendored (bundled) gems brings the speed
back up to acceptable levels.

This change restores the original behaviour, adds the filter, and
upgrades simplecov for good measure.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ group :development, :test do
   gem "timecop"
   gem "rspec"
   gem "rspec-rails", "~> 3.5"
-  gem "simplecov", "0.10.0", require: false
+  gem "simplecov", "0.13.0", require: false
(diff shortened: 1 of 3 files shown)


dougdroper:
thanks @elliotcm, this would still add a few seconds onto load time of the tests though, I am not sure we want to take that hit. We do have simplecov running in CI though, which is good. We can definitely upgrade simplecov, just not sure about it running locally.

PR #742, opened 2017-01-27
Add a script to check edition consistency with the content-store

thomasleese:
This PR adds a script to check for content consistency across the `publishing-api` and `content-store`s.

See alphagov/content-store#262 for the equivalent script in the `content-store`.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -183,6 +183,15 @@ class Edition < ApplicationRecord
     )
   end
 
+  def gone?
+    if state == "unpublished"
+      unpublishing = Unpublishing.find_by!(edition: self)
+      unpublishing.type == "gone"
+    else
+      document_type == "gone"
+    end
+  end
+
 private
 
   def check_document_fields_from_options(options)
--- a/lib/content_consistency_checker.rb
+++ b/lib/content_consistency_checker.rb
@@ -1,0 +1,76 @@
+require 'gds_api/content_store'
+
+class ContentConsistencyChecker
+  attr_reader :errors
+
+  def initialize(content_id, locale = "en")
+    @content_id = content_id
+    @locale = locale
+    @errors = []
+  end
+
+  def call
+    unless document
+      errors << "Document (#{content_id}, #{locale}) could not be found."
+      return errors
+    end
+
+    check_edition(draft, draft_content_store) if draft
+    check_edition(live, live_content_store) if live
+
+    errors
+  end
+
+private
+
+  attr_reader :content_id, :locale
+
+  def item_from_content_store(path, content_store)
+    begin
+      content_store.content_item(path).parsed_content
+    rescue GdsApi::ContentStore::ItemNotFound, GdsApi::HTTPGone, GdsApi::HTTPForbidden
+      nil
+    end
+  end
+
+  def check_edition(edition, content_store)
+    content_item = item_from_content_store(edition.base_path, content_store)
+
+    if edition.gone? && content_item
+      errors << "Content exists in the content store."
+    elsif content_item.nil?
+      errors << "Content is missing from the content store."
+      return
+    end
+
+    fields = [:rendering_app, :publishing_app, :schema_name, :document_type]
+    fields.each do |field|
+      edition_value = edition.send(field)
+      content_item_value = content_item[field.to_s]
+
+      if edition_value != content_item_value
+        errors << "Edition #{field} (#{edition_value}) does not match content store (#{content_item_value})."
+      end
+    end
+  end
+
+  def live
+    document.live
+  end
+
+  def draft
+    document.draft
+  end
+
+  def document
+    @document ||= Document.find_by(content_id: content_id, locale: locale)
+  end
+
+  def live_content_store
+    GdsApi::ContentStore.new(Plek.find("content-store"))
+  end
+
+  def draft_content_store
+    GdsApi::ContentStore.new(Plek.find("draft-content-store"))
+  end
+end
--- a/lib/tasks/check_content_consistency.rake
+++ b/lib/tasks/check_content_consistency.rake
@@ -1,0 +1,29 @@
+namespace :check_content_consistency do
+  def check_content(content_id, locale)
+    checker = ContentConsistencyChecker.new(content_id, locale)
+    errors = checker.call
+
+    if errors.any?
+      puts "#{content_id} 😱"
+      puts errors
+    end
+
+    errors.none?
+  end
+
+  desc "Check documents for consistency with the router-api and content-store"
+  task :one, [:content_id, :locale] => [:environment] do |_, args|
+    content_id = args[:content_id]
+    locale = args[:locale] || "en"
+    check_content(content_id, locale)
+  end
+
+  desc "Check all the documents for consistency with the router-api and content-store"
+  task all: :environment do
+    documents = Document.pluck(:content_id, :locale)
+    failures = documents.reject do |content_id, locale|
+      check_content(content_id, locale)
+    end
+    puts "Results: #{failures.count} failures out of #{documents.count}."
+  end
+end
--- a/spec/factories/edition.rb
+++ b/spec/factories/edition.rb
@@ -56,6 +56,8 @@ FactoryGirl.define do
     sequence(:base_path) { |n| "/dodo-sanctuary-#{n}" }
     schema_name "gone"
     document_type "gone"
+    state "superseded"
+    rendering_app nil
   end
 
   factory :access_limited_edition, aliases: [:access_limited_draft_edition], parent: :edition do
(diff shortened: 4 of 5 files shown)


app/models/edition.rb
```diff
@@ -242,6 +242,10 @@ def substitute
     )
   end
 
+  def gone?
+    schema_name == "gone" || document_type == "gone" || rendering_app.nil?
```
carlosmartinez: is this an accurate reflection of the nature of "gone"? if so, perhaps we should create a data sanitisation task; i'm not convinced three different types of goneness is a good idea.

lib/tasks/audit_item_consistency.rake
```diff
@@ -0,0 +1,35 @@
...
+    locale = args[:locale] || "en"
+    check_content(content_id, locale)
+  end
+
+  desc "Check all the documents for consistency with the router-api and content-store"
+  task check_all: :environment do
+    total_count = 0
+    failure_count = 0
+
+    Document.pluck(:content_id, :locale).each do |content_id, locale|
```
carlosmartinez: perhaps neater as something along the lines of:
```
docs = Document.pluck(:content_id, :locale)
successes, failures = docs.partition { |content_id, locale| check_content(content_id, locale) }
puts "Results: #{failures.count} failures out of #{docs.count}."
```

lib/tasks/check_content_consistency.rake
```diff
@@ -0,0 +1,29 @@
...
+    check_content(content_id, locale)
+  end
+
+  desc "Check all the documents for consistency with the router-api and content-store"
+  task all: :environment do
+    documents = Document.pluck(:content_id, :locale)
+    _, failures = documents.partition do |content_id, locale|
+      check_content(content_id, locale)
+    end
+    puts "Results: #{failures.count} failures out of #{docs.count}."
```
carlosmartinez: s/docs/documents

lib/tasks/check_content_consistency.rake
```diff
@@ -0,0 +1,29 @@
...
+  task :one, [:content_id, :locale] => [:environment] do |_, args|
+    content_id = args[:content_id]
+    locale = args[:locale] || "en"
+    check_content(content_id, locale)
+  end
+
+  desc "Check all the documents for consistency with the router-api and content-store"
+  task all: :environment do
+    documents = Document.pluck(:content_id, :locale)
+    _, failures = documents.partition do |content_id, locale|
```
carlosmartinez: might be a bit clearer with `failures=documents.reject ...` rather than `_, failures = documents.partition ...` (sorry, i made the stupid suggestion in the first place!)

lib/tasks/check_content_consistency.rake
```diff
@@ -0,0 +1,29 @@
...
+  desc "Check documents for consistency with the router-api and content-store"
+  task :one, [:content_id, :locale] => [:environment] do |_, args|
+    content_id = args[:content_id]
+    locale = args[:locale] || "en"
+    check_content(content_id, locale)
+  end
+
+  desc "Check all the documents for consistency with the router-api and content-store"
+  task all: :environment do
+    documents = Document.pluck(:content_id, :locale)
```
kevindew: might be safer to use a `find_each` to batch this rather than pulling everything out in one go.

app/models/content_consistency_checker.rb
```diff
@@ -0,0 +1,111 @@
+require 'gds_api/content_store'
+
+class ContentConsistencyChecker
+  attr_reader :content_id, :locale
```
kevindew: can move this to the `private` block unless they're intentionally public?

app/models/content_consistency_checker.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  def initialize(content_id, locale = "en")
+    @content_id = content_id
+    @locale = locale
+    @errors = []
+  end
+
+  def call
+    return @errors unless document
```
kevindew: be nice to have an `attr_reader` for errors

app/models/content_consistency_checker.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def draft
+    document.draft
+  end
+
+  def document
+    @document ||= Document.find_by!(content_id: content_id, locale: locale)
+  rescue ActiveRecord::RecordNotFound
+    @errors << "publishing-api: Content (#{content_id}) could not be found."
```
kevindew: We should include locale in the error message too

app/models/content_consistency_checker.rb
```diff
@@ -0,0 +1,111 @@
...
+  def live
+    document.live
+  end
+
+  def draft
+    document.draft
+  end
+
+  def document
+    @document ||= Document.find_by!(content_id: content_id, locale: locale)
```
kevindew: We could probably just have a `find_by` rather than the banged equivalent here and deal with a document of nil in the `call` method. Save a place where we mutate `errors`

app/models/content_consistency_checker.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  def check_edition_existence(edition, content_item)
+    if edition.gone? && content_item
+      @errors << "Content exists in the content store."
+    elsif content_item.nil?
+      @errors << "Content is missing from the content store."
+    end
+  end
+
+  def check_edition(edition, content_store)
```
kevindew: So I think here we're not actually checking consistency correctly. For each `base_path` the Publishing API has there should be the equivalent at a `base_path` in the Content Store. Within these these the routes and redirects should match - well everything should match. So you could use a downstream presenter to compare each aspect of these.

lib/tasks/check_content_consistency.rake
thomasleese: We've measured the RAM consumption of this as 56MB on the current integration data and therefore don't think this will be an issue.

app/models/content_consistency_checker.rb
thomasleese: Turns out we cannot even check the routes and the redirects here as the `content-store` does not expose such information via the API. Instead, these will be [checked in the script between the `content-store` and the `router-api`](https://github.com/alphagov/content-store/pull/262).

lib/content_consistency_checker.rb
```diff
@@ -0,0 +1,76 @@
...
+    content_item = item_from_content_store(edition.base_path, content_store)
+
+    if edition.gone? && content_item
+      errors << "Content exists in the content store."
+    elsif content_item.nil?
+      errors << "Content is missing from the content store."
+      return
+    end
+
+    fields = [:rendering_app, :publishing_app, :schema_name, :document_type]
```
kevindew: Are we able to check all the fields that content_store makes available? just so this can be a more general sync-check with content_store? 

An awkward field is links as you'd need to run the edition through the downstream presenter and would probably make this run way slower - but then that's probably the most useful field for a check, what do you think?

carlosmartinez:
fyi i squashed all the little commits into one. such power!

lib/content_consistency_checker.rb
thomasleese: Although this might be a useful test, I believe this is beyond the scope of this particular card (which is just focussing on the existence and location of content items). If we want to do this, I think we should pick it up as a separate issue.

thomasleese:
Some example output from me running it locally on my own database (therefore errors might not reflect production).

```
❯ bundle exec rake check_content_consistency:all
5f9d4d8d-7631-11e4-a3cb-005056011aef 😱
Content is missing from the content store.
9406e5c5-a8fd-4205-9326-0a8a03456380 😱
Edition rendering_app (government-frontend) does not match content store ().
Edition schema_name (statistics_announcement) does not match content store (redirect).
Edition document_type (national_statistics_announcement) does not match content store (redirect).
4cd4d22f-5bfe-4547-a898-02a066642ba7 😱
Edition rendering_app (government-frontend) does not match content store ().
Edition schema_name (statistics_announcement) does not match content store (redirect).
Edition document_type (national_statistics_announcement) does not match content store (redirect).
5e2e1997-7631-11e4-a3cb-005056011aef 😱
Edition rendering_app (whitehall-frontend) does not match content store ().
Edition schema_name (publication) does not match content store (redirect).
Edition document_type (guidance) does not match content store (redirect).
836fd55a-3797-45b5-802d-c90a5f39e995 😱
Content is missing from the content store.
dd7c745d-bf8b-46f5-b202-b5acc5028947 😱
Edition rendering_app () does not match content store (frontend).
Edition publishing_app (whitehall) does not match content store (frontend).
Edition schema_name (contact) does not match content store (special_route).
Edition document_type (contact) does not match content store (special_route).
68a37f8a-3976-4e43-83a2-e0a76f4d7b04 😱
Edition rendering_app (government-frontend) does not match content store ().
Edition schema_name (statistics_announcement) does not match content store (redirect).
Edition document_type (national_statistics_announcement) does not match content store (redirect).
310c1b63-3aef-4f02-92f5-ed0949c1ec28 😱
Content is missing from the content store.

```

PR #743, opened 2017-01-31
Database Cleanup

thomasleese:


app/models/edition.rb
```diff
@@ -67,71 +65,14 @@ class Edition < ApplicationRecord
...
-    end
-
-    if changes[:user_facing_version]
-      UserFacingVersion.find_or_initialize_by(content_item_id: id)
-        .update!(number: changes[:user_facing_version].last)
-    end
-  end
-
-  before_save { ensure_document }
+  def as_json(options = {})
```
kevindew: I think now we're removing the backwards compatible behaviour we should look at keeping this as just serializing edition as json and if we need different logic here we should have it coming through a presenter

PR #745, opened 2017-01-31
Expand the parent taxons of the taxon a content item is tagged to

suzannehamilton:
The links to the taxon hierarchy were being expanded correctly when the root node was a taxon, but not when the root node was a regular piece of content.

Trello: https://trello.com/c/QCxXilwI/360-get-expanded-taxons-in-the-link-section-of-content-pages

Are there any integration tests for link expansion? i.e. a test that when you send a link PATCH request, that it ultimately sends the right thing to the content store? If so, I'll update them with a test for this dependency tree.

### DIFF
--- a/app/queries/dependent_expansion_rules.rb
+++ b/app/queries/dependent_expansion_rules.rb
@@ -30,6 +30,7 @@ module Queries
       [
         [:parent],
         [:parent_taxons],
+        [:taxons, :parent_taxons],
         [:ordered_related_items, :mainstream_browse_pages, :parent],
       ]
     end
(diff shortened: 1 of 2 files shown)


tijmenb:
Looks good! Would indeed like an integration test that describes our use case.

I'll let someone from @alphagov/team-gov-uk-publishing-platform merge this.

dougdroper:
The closest thing we have to integration specs for this is: https://github.com/alphagov/publishing-api/blob/master/spec/presenters/queries/expanded_link_set_spec.rb#L84. That tests a similar use case. Let me know if you need help with that.

suzannehamilton:
Thanks! @dougdroper - are you happy for me to merge this as it is, with just unit tests? I'll look into the integration test as a separate step.

dougdroper:
yes sure

PR #748, opened 2017-01-31
[DISCUSS] allow lookup of redirected base paths

MatMoore:
This request is used in content tagger and publishing apps to lookup content ids, so that users can refer to pages by base paths when tagging to them.

For example:
- related links
- topics
- taxons

The request currently silently ignores some editions that have been redirected. The implementation looks for any editions with state `published` or `unpublished`.

There are different ways of unpublishing an edition:
- withdrawal (content is still shown, with a message)
- substitute (I'm ignoring this)
- gone (content is not there, HTTP gone)
- vanished (content is not there, HTTP not found)
- redirect (page redirects to another one)

This is an example of a redirected page that doesn't work with the request:
`/government/publications/impact-evaluation-of-the-send-pathfinder-programme` was
unpublished (consolidated) and redirected to
`/government/publications/send-pathfinder-programme-final-report`

As far as the frontend is concerned, everything works as expected, but I have
had trouble making sense of publishing api's record of this page and others
that have been redirected.

In this case, no non-superseded editions in publishing api are associated with
the **old URL**. The unpublished edition has the **new URL**. I haven't figured out
why this is, it could be intentional behaviour, or a historical bug (this
document was apparently unpublished multiple times).

I've had trouble understanding what the system is doing when unpublishing documents, because whitehall also behaves in an unusual way: every unpublish is followed by the creation of a new draft, using put content. This is so that an unpublished document can be "un-unpublished" by going through 2i or force publish again.

Put content can sometimes have the side effect of creating separate content
items with document type "redirect". I originally thought that I should expect
to see one of these "redirect" editions for the **old URL**, which would mean
that lookup_by_base_path would be able to successfully look up the content id
for this page. **This does not appear to be the case**.

In fact, in this example, there is a "redirect" edition, but it's a draft. I
don't understand how this was created or why.

Regardless of how the data got into this state, I think the implementation of this
request is imprecise. I've made two changes to it:

1. Add a parameter to request additional information about base paths that have
been redirected. This allows to display a more useful error to the user and
suggest they tag to the redirect target instead. (This is the problem I actually want to solve!)

I don't think this works at the moment, because it assumes that the unpublished
edition still has the old base path - which wasn't the case for the example I
looked at.

2. Don't return unpublished editions unless it was actually withdrawn (I
believe this was the original intent)

Tagging to a "gone" or "not found" page has no meaning, so I don't expect this to break anything.

**This shouldn't be merged as is** - I'd just like a review from someone with more understanding of how unpublishing actually works.

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,16 +1,42 @@
 class LookupsController < ApplicationController
   def by_base_path
-    # return content_ids for content that is visible on the live site
-    # withdrawn items are still visible
-    states = %w(published unpublished)
     base_paths = params.fetch(:base_paths)
 
-    base_paths_and_content_ids = Edition.with_document
-      .where(state: states, base_path: base_paths)
-      .pluck(:base_path, 'documents.content_id')
-      .uniq
+    base_paths_and_content_ids = Edition.with_document.includes(:unpublishing)
+      .where( base_path: base_paths)
 
-    response = Hash[base_paths_and_content_ids]
-    render json: response
+    published = {}
+    redirects = {}
+
+    base_paths_and_content_ids.each do |edition|
+      if redirected?(edition)
+        redirects[edition.base_path] = edition.unpublishing.alternative_path
+      elsif visible_on_live?(edition)
+        published[edition.base_path] = edition.document.content_id
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


MatMoore:
cc @gpeng @alphagov/team-gov-uk-finding-things

PR #749, opened 2017-02-01
Edition-level links

thomasleese:
Implementation of edition-level links which solves [the 'draft link' problem][0].

[0]: https://trello.com/c/Yd7RZaaT/870-edition-link-workflow-3

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -48,8 +48,8 @@ module Commands
 
         payload.fetch(:links, []).each do |link_type, target_link_ids|
           edition.links.create!(
-            target_link_ids.map do |target_link_id|
-              { link_type: link_type, target_content_id: target_link_id }
+            target_link_ids.map.with_index do |target_link_id, i|
+              { link_type: link_type, target_content_id: target_link_id, position: i }
             end
           )
         end
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -44,9 +44,11 @@ module Commands
       end
 
       def create_links(edition)
-        edition.copy_links_from(previously_published_item.links)
+        if payload[:links].nil?
+          edition.copy_links_from(previously_published_item.links)
+        end
 
-        payload.fetch(:links, []).each do |link_type, target_link_ids|
+        payload.fetch(:links, {}).each do |link_type, target_link_ids|
           edition.links.create!(
             target_link_ids.map.with_index do |target_link_id, i|
               { link_type: link_type, target_content_id: target_link_id, position: i }
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -46,14 +46,14 @@ module Commands
       def create_links(edition)
         if payload[:links].nil?
           edition.copy_links_from(previously_published_item.links)
-        end
-
-        payload.fetch(:links, {}).each do |link_type, target_link_ids|
-          edition.links.create!(
-            target_link_ids.map.with_index do |target_link_id, i|
-              { link_type: link_type, target_content_id: target_link_id, position: i }
-            end
-          )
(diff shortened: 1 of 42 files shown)


app/models/previously_published_item.rb
```diff
@@ -57,5 +65,8 @@ def previous_base_path
 
     def routes
     end
+
+    def previously_published_item
```
dougdroper: ?

app/commands/v2/put_content.rb
```diff
@@ -40,6 +40,19 @@ def update_content_dependencies(edition)
         update_last_edited_at(edition, payload[:last_edited_at])
         ChangeNote.create_from_edition(payload, edition)
         Action.create_put_content_action(edition, document.locale, event)
+        create_links(edition)
+      end
+
+      def create_links(edition)
+        edition.copy_links_from(previously_published_item.links)
```
kevindew: I think we should only copy links if `payload[:links]` is null. Treat it as a replacement if it is provided or fallback if not provided.

kevindew:
Ah [this double](https://github.com/alphagov/publishing-api/blob/master/spec/commands/v2/put_content_spec.rb#L5-L7) is why it's not tripping up schema tests - we should probably get rid of that.

kevindew:
Once you rebase this it'll go ❌  as we'll need to update the schemas I'm afraid.

app/commands/v2/put_content.rb
```diff
@@ -44,9 +44,11 @@ def update_content_dependencies(edition)
       end
 
       def create_links(edition)
-        edition.copy_links_from(previously_published_item.links)
+        if payload[:links].nil?
```
kevindew: I think it'd be more readable if this had an `if else` to make it clearer that copying and creating are exclusive

thomasleese:
Submitted a PR https://github.com/alphagov/govuk-content-schemas/pull/510 for the schema change.

app/presenters/queries/expanded_link_set.rb
```diff
@@ -3,10 +3,11 @@ module Queries
     class ExpandedLinkSet
       attr_reader :state_fallback_order
 
-      def initialize(content_id:, state_fallback_order:, locale_fallback_order: Edition::DEFAULT_LOCALE)
+      def initialize(content_id:, state_fallback_order:, locale_fallback_order: Edition::DEFAULT_LOCALE, edition_id: nil)
```
thomasleese: Although it is unnecessary coupling, we believe there would be performance issues with performing the query to find the `edition_id` when it is already available in the places this class is used.

app/presenters/queries/expand_dependents.rb
```diff
@@ -1,9 +1,10 @@
 module Presenters
   module Queries
     class ExpandDependents
-      def initialize(content_id, controller)
+      def initialize(content_id, controller, edition_id)
```
thomasleese: Similar to the comment above.

app/models/edition.rb
```diff
@@ -87,6 +88,14 @@ def base_path_present?
     base_path.present?
   end
 
+  def copy_links_from(other_links)
```
gpeng: Harking back to the original big object avoiding intent of the 'pillars', should this be a concern of `Edition`? As we've got command objects doing things to editions would this be better in there or in its own class? It looks like it's being tested within the `put_content` command tests already.

kevindew:
I haven't had a chance to read in depth so this is a bit of a fleeting comment - but it seems we're now taking locale into account with the edition level links? is this correct / intentional? I'm guessing it pulls in the links of every edition with that content_id.

I feel like since the behaviour of them is somewhat undefined/defined by us we should probably work out some integration tests that define it. What do you think?

kevindew:
closed as it's still a work in progress.

PR #750, opened 2017-02-01
Skip lookup of unpublished pages unless withdrawn

MatMoore:
This request currently returns the content ids for any base paths that belong to _published_ or _unpublished_ editions.

The intention of including unpublished pages was to support withdrawn content:
these pages are still live, but show a message saying the content is not current.

However, there are 5 types of unpublishing, and the others mean content is not
visible on live.

It's not useful to complete the lookup in these cases. This request is used for
tagging, and tagging to content that isn't visible has no meaning.

There are also separate content items with document types of "redirect" and
"gone" which do not represent content. This means that the current
implementation is not very strict about only including content ids that are
"visible", and this makes it more difficult to predict what will and won't work
with this request.

We noticed this when looking into redirected content. Ideally we should be able
to distinguish content that has been redirected to another page (consolidated)
from content that never existed, so that publishers or users of content tagger
don't get confused (just saying the content doesn't exist suggests they made a
mistake) and they can choose whether to tag to the new thing instead.

For now I think it's best to deliberately include this from the lookup, and
make an API change later to expose extra information about redirected content.
We are also considering whether we need to be able to query content ids
of base paths belonging to drafts.

This PR replaces https://github.com/alphagov/publishing-api/pull/748

Trello:
https://trello.com/c/rC2btscW/448-content-tagger-show-appropriate-error-messages-when-uploading-untaggable-content
https://trello.com/c/ivoxhbhT/459-decide-what-we-need-from-lookup-by-base-path-so-that-we-can-handle-draft-and-redirected-content-when-tagging

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -6,7 +6,10 @@ class LookupsController < ApplicationController
     base_paths = params.fetch(:base_paths)
 
     base_paths_and_content_ids = Edition.with_document
+      .left_outer_joins(:unpublishing)
       .where(state: states, base_path: base_paths)
+      .where("state = 'published' OR unpublishings.type = 'withdrawal'")
+      .where("document_type NOT IN ('gone', 'redirect')")
       .pluck(:base_path, 'documents.content_id')
       .uniq
 
--- a/spec/commands/v2/unpublish_spec.rb
+++ b/spec/commands/v2/unpublish_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe Commands::V2::Unpublish do
         }
       end
 
-      it "raises an error when expanation is blank" do
+      it "raises an error when explanation is blank" do
         msg = "Validation failed: Explanation can't be blank"
         expect { described_class.call(payload) }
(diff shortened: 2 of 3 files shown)


MatMoore:
@kevindew, what do you think?

app/controllers/lookups_controller.rb
```diff
@@ -6,7 +6,10 @@ def by_base_path
     base_paths = params.fetch(:base_paths)
 
     base_paths_and_content_ids = Edition.with_document
+      .joins("LEFT JOIN unpublishings ON unpublishings.edition_id = editions.id")
```
kevindew: I think you can simplify this to be `.left_outer_joins(:unpublishing)`

app/controllers/lookups_controller.rb
```diff
@@ -6,7 +6,10 @@ def by_base_path
...
       .where(state: states, base_path: base_paths)
+      .where("state = 'published' OR unpublishings.type = 'withdrawal'")
+      .where("document_type NOT IN ('gone', 'redirect')")
```
kevindew: 👍

PR #757, opened 2017-02-07
Link Expansion Refactor

kevindew:
This PR refactors the Link Expansion and Dependency Resolution concepts of the Publishing API. It was brought about due to a changing need of requirements so that recursive links were required in both direct and reverse links.

It aims to make greater use of sharing code between both: direct and reverse links; and link expansion and dependency resolution. It is also aims to make the features _hopefully_ easier to understand and debug.

# Things that have changed

## LinkExpansion and DependencyResolution as class level concepts
Hopefully this should make it easier to work out where to start with both concepts and make it easier to test how things should run in a developer console. 

## Unified dependee/dependent process
Rather than these being separate classes they are now performed together. The terminology has changed to direct and reverse links.

## Updated means to define allowed paths of links
Rather than having an array of link_types where the last one is considered _sticky_ we now have a means to define which item in a list of link types is recursive by a method on the symbol: `[:organisations, :parent.recurring]`

## Shared queries to access Links
The same queries: `Queries::LinksTo` and `Queries::LinksFrom` are used to access Links and DependencyResolution.

## Simplify state fallback
In many places we used to pass around a `state_fallback_order` field and became unnecessarily complicated with the usage of withdrawn links and similar.
Very rarely do we actually need the flexibility of a state fallback system and what we actually need to know is just whether drafts are a concern or not.

## Link Graph model
This PR introduces a light weight model called `LinkGraph` which is used to represent the relations of a link. This model in populated in both link expansion and dependency resolution as the source of the data. It can be accessed in a developer console to print out the links in either concept to ease debugging `DependencyResolution.new(content_id).link_graph.to_h`

## Separate determining links from adding content to them
Expanded links now build up a graph of links before populating them with content.


lib/link_expansion/content_repository.rb
```diff
@@ -0,0 +1,38 @@
+class LinkExpansion::ContentRepository
```
thomasleese: I think a better name might just be `ContentCache`.

PR #759, opened 2017-02-07
Support "search_in" parameter in get_content.

danielroseman:
This allows specifying which fields to search against, as a comma
separated list. We support a single level of nesting with a dot
seperator. So, to search against title and the value inside the
description hash, do:

    ?search_in=title,description.value

If not provided, it will default to the existing title and base_path.

Benchmarks, from bench/get_content_items.rb

without search:
..........  0.790000   0.260000   1.050000 (  3.063176)
queries: 30

with search
..........  0.800000   0.270000   1.070000 (  3.084760)
queries: 30

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -9,7 +9,8 @@ module V2
         fields: query_params[:fields],
         filters: filters,
         pagination: pagination,
-        search_query: query_params.fetch("q", "")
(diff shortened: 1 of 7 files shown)


danielroseman:
It occurred to me there might be an SQL injection vulnerability here; the top level fields are validated by the nested ones aren't, so it might be possible to craft something nasty. I'll take another look tomorrow.

PR #762, opened 2017-02-08
Fact check access limited

kevindew:
It appears we aren't storing fact_check_ids that can be received, it looks like this was just a missing piece of the fact_check_id work. 

Custom are hoping to start using these and hit an issue that they couldn't store these.

Migration takes 0.13 seconds
```
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] Migrating to AddFactCheckIds (20170208172603)
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] == 20170208172603 AddFactCheckIds: migrating ==================================
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] -- add_column(:access_limits, :fact_check_ids, :json, {:null=>false, :default=>[]})
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] -> 0.1345s
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] == 20170208172603 AddFactCheckIds: migrated (0.1347s) =========================
*** [err :: backend-1.backend.integration.publishing.service.gov.uk] 
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,16 +10,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170207155436) do
+ActiveRecord::Schema.define(version: 20170208172603) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_table "access_limits", force: :cascade do |t|
-    t.json     "users",      default: [], null: false
-    t.datetime "created_at",              null: false
-    t.datetime "updated_at",              null: false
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -49,6 +50,50 @@
     t.index ["edition_id"], name: "index_change_notes_on_edition_id", using: :btree
   end
 
+  create_table "content_items", force: :cascade do |t|
```
danielroseman: Not sure how these are being added.

kevindew: Oh yuck, thanks for catching that mess. I'll just fix it up.

PR #763, opened 2017-02-09
Fix documents reverse link name

thomasleese:
An 's' was accidentally added making it 'documents_collections'
when it should actually be 'document_collections'.


thomasleese:
See https://github.com/alphagov/publishing-api/blob/1d9f32c8261f7f9c945e8ff27fcee6938a91efe7/app/queries/dependent_expansion_rules.rb#L101 for how it used to be defined.

PR #764, opened 2017-02-09
Remove links from randomly generated edition

kevindew:
Links were causing this test to fail intermittently. This will fix the links issue.

I think this might well have been already an intermittent test so I've updated the errors so it's clearer what happened and hopefully we can look into fixing it fully if it occurs again.

### DIFF
--- a/spec/integration/message_queue_publishing_spec.rb
+++ b/spec/integration/message_queue_publishing_spec.rb
@@ -14,11 +14,11 @@ RSpec.describe "Message queue publishing" do
 
       put "/v2/content/#{content_id}", params: edition.to_json
 
-      expect(response).to be_ok
+      expect(response).to be_ok, "failed to put-content a randomly generated edition"
 
       post "/v2/content/#{content_id}/publish", params: { locale: edition["locale"] }.to_json
 
-      expect(response).to be_ok
+      expect(response).to be_ok, "failed to publish a randomly generated edition"
 
       ensure_message_queue_payload_validates_against_notification_schema
     end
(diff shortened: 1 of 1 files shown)


kevindew:
@danielroseman there doesn't seem to be a particular reason why we're using them there and does seem to add unnecessary baggage to this test that isn't concerned with it.

They have in the past helped us at catching problems where we don't properly support a schema. So might well make sense to remove them from this test and put together an integration test that is focused on random schemas.

tijmenb:
The reason to use random content items is that if the schema changes, the tests will automatically be updated. By using hard coded content items / fixtures we run the risk that if you update the schema, you also have to update fixtures across GOV.UK (which was one of the problems with gettin rid of `format`).

Maybe the `links` in the content schemas was merged too soon? What kind of problems has it caused?

tijmenb:
(For context, I wrote that test in https://github.com/alphagov/publishing-api/pull/523)

kevindew:
@tijmenb it wasn't probably merged too soon by us. We have an [error thrown](https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/put_content_validator.rb#L24-L41) if links is present. Which will be cool once https://github.com/alphagov/publishing-api/pull/749 is merged.

The problem we tend to hit with tests related to randomly generated schemas is it's not particularly easy to determine the problem was from our code not supporting a randomly generated schema, and then a rebuild goes green and it goes unresolved.

tijmenb:
Right, makes sense. An integration test that explicitly tests against randomised content items should surface those problems, assuming enough examples are run.

kevindew:
Merging this now so we have less build fails. Card to explore this at: https://trello.com/c/rZmA19cv/429-create-integration-test-for-randomly-generated-editions

PR #766, opened 2017-02-10
Remove web content item

carlosmartinez:
With the recent model changes, `WebContentItem` has become superfluous. In the name of making the codebase less confusing and easier to work with, we have removed it and pulled its core functionality into `Edition`. Therefore the downstream presenters now present an `Edition` rather than a `WebContentItem`.


app/queries/get_web_content_items.rb
```diff
@@ -1,79 +0,0 @@
-module Queries
```
kevindew: yay for removing this, only need to refactor one validator and we can lose the arel helpers.

app/workers/downstream_discard_draft_worker.rb
```diff
@@ -13,10 +13,10 @@ class DownstreamDiscardDraftWorker
   def perform(args = {})
     assign_attributes(args.symbolize_keys)
 
-    current_path = web_content_item.try(:base_path)
+    current_path = edition.try(:base_path)
```
kevindew: I don't think this needs to be a `try`

app/workers/downstream_discard_draft_worker.rb
```diff
@@ -46,11 +46,11 @@ def assign_attributes(attributes)
...
     if attributes[:locale]
       @locale = attributes[:locale]
-      @web_content_item = Queries::GetWebContentItems.for_content_store(content_id, locale, true)
+      @edition = Queries::GetEditionForContentStore.(content_id, locale, true)
     else
       content_item_id = attributes[:live_content_item_id]
-      @web_content_item = content_item_id ? Queries::GetWebContentItems.find(content_item_id) : nil
-      @locale = web_content_item.try(:locale)
+      @edition = content_item_id ? Edition.find(content_item_id) : nil
+      @locale = edition.try(:locale)
```
kevindew: again, should be fine as a method

app/queries/get_edition_for_content_store.rb
```diff
@@ -0,0 +1,18 @@
+module Queries
+  class GetEditionForContentStore
+    def self.call(content_id, locale, include_draft = false)
+      allowed_states = [:published, :unpublished]
+      allowed_states << :draft if include_draft
+
+      scope = Edition.with_document
+      scope = Unpublishing.join_editions(scope)
```
kevindew: probably nicer to have this as `.left_outer_join(:unpublishing)` I think we only have the `join_editions` because Edition didn't know about unpublishings

app/queries/get_edition_for_content_store.rb
```diff
@@ -0,0 +1,18 @@
+module Queries
+  class GetEditionForContentStore
+    def self.call(content_id, locale, include_draft = false)
+      allowed_states = [:published, :unpublished]
```
kevindew: rather than basing this on state can we base it on content_store value?

app/presenters/queries/available_translations.rb
```diff
@@ -28,7 +28,7 @@ def grouped_translations
 
       def expand_translation(id)
         expansion_rules = LinkExpansion::Rules
-        web_item = ::Queries::GetWebContentItems.call(id).first
+        web_item = Edition.where(id: id).first
```
kevindew: `find_by` ?

app/presenters/downstream_presenter.rb
```diff
@@ -97,14 +90,14 @@ def rendered_details
...
-        document_type: web_content_item.document_type
+        format: edition.schema_name,
+        schema_name: edition.schema_name,
+        document_type: edition.document_type
       }
     end
 
     def withdrawal_notice
-      unpublishing = Unpublishing.find_by(edition_id: web_content_item.id)
+      unpublishing = Unpublishing.find_by(edition_id: edition.id)
```
kevindew: `edition.unpublishing` instead?

app/models/edition.rb
```diff
@@ -192,6 +194,42 @@ def gone?
     end
   end
 
+  def to_h
```
kevindew: The changes in this class seem to introduce a bunch of presentation concerns into this model. 

I think it'd make sense that we have this logic in a presenter, what do you think? 

We also could do with adjusting `to_json` to not be using document fields and instead just be serializing the edition model.

app/workers/downstream_discard_draft_worker.rb
thomasleese: Just tried replacing it, got instead `undefined method `base_path' for nil:NilClass` which I wasn't expecting, but would appear that's why it's a `try`.

app/workers/downstream_discard_draft_worker.rb
thomasleese: I believe this is also required in the case when `edition` is `nil`.

app/queries/get_edition_for_content_store.rb
thomasleese: We tried this originally, but Rails wasn't having it because the foreign key relation is not defined in the database. We could create a migration to add this constraint though, what do you think?

app/workers/downstream_discard_draft_worker.rb
carlosmartinez: i guess if `edition` were `nil`, we should fail in a more transparent way.

kevindew: doh! sorry I misunderstood the code (and I'm pretty sure it was me that introduced the try in the first place)

app/models/content_store/content_item.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+private
+
+  def locale
+    edition.document.locale
+  end
+
+  def withdrawn?
+    edition.unpublishing.present? && edition.unpublishing.withdrawal?
```
thomasleese: I think we can just use `edition.withdrawn?` etc for these methods and the ones below. Or is the plan to remove those methods from `Edition` at some point?

app/models/content_store/content_item.rb
```diff
@@ -0,0 +1,54 @@
+class ContentItem
```
kevindew: I think it'd be make sense for this class to be within a content_store module, would save having to tweak the autoloader and feels less surprising.

carlosmartinez: yeah. it's all changing! i'll give you a shout when we've pushed.

dougdroper:
@kevindew 
> Main thing I've flagged is the presentation responsibility.

We took the `to_h` out of edition, but it turns out that it is not only responsible for presentation, it is also needed for link expansion. It could be pulled out into a separate object, but it didn't add any value. WDYT?

kevindew:
@dougdroper sure, yeah that'd kick in since we're presenting them as a content item in link expansion too - I can see the url methods are used in debug too so maybe they do belong at edition level.

Lets grab a few mins tomorrow to go through this just to try work out lines between Edition / Content Item concerns.

app/presenters/edition_presenter.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    def unexpanded_links
+      Queries::LinkSetPresenter.new(
+        LinkSet.find_by(content_id: edition.content_id)
+      ).links
+    end
+
+    def present
+      edition.to_h
+        .except(*%i{updated_at created_at document_id content_store last_edited_at id state user_facing_version api_url web_url withdrawn api_path unpublishing_type}) # only intended to be used by publishing applications
```
thomasleese: This is getting quite long now, might be worth extracting out into a constant at the top of the file with a name explaining what it is for. (I remember skipping past this line once before because I didn't know what it did when it was the exact thing I was looking for.)

app/presenters/queries/available_translations.rb
```diff
@@ -20,16 +17,19 @@ def translations
...
 
       def expand_translation(id)
         expansion_rules = LinkExpansion::Rules
-        web_item = ::Queries::GetWebContentItems.call(id).first
-        web_item.to_h.select { |f| expansion_rules.expansion_fields(:available_translations).include?(f) }
+        web_item(id).select { |f| expansion_rules.expansion_fields(:available_translations).include?(f) }
+      end
+
+      def web_item(id)
+        Edition.find_by(id: id).to_h
```
thomasleese: Can we use `find(id)` here?

carlosmartinez:
```
ActiveRecord::RecordNotFound:
Couldn't find all Editions with 'id': (11, en, published) (found 1 results, but was looking for 3)
```
bizarre!

PR #767, opened 2017-02-10
Improve print output

thomasleese:
Now shows if the failure was in live or draft.


thomasleese:
I've run the script on integration, I suspect it will take a few hours to finish, but this PR will improve its output.

PR #770, opened 2017-02-13
Fix audit script redirects and gones

thomasleese:
This was previously not working as it did not take into account unpublishings.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -185,10 +185,17 @@ class Edition < ApplicationRecord
 
   def gone?
     if state == "unpublished"
-      unpublishing = Unpublishing.find_by!(edition: self)
-      unpublishing.type == "gone"
+      unpublishing.gone?
     else
       document_type == "gone"
+    end
+  end
+
+  def redirect?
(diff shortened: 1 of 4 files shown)


app/models/edition.rb
```diff
@@ -185,13 +185,20 @@ def substitute
 
   def gone?
     if state == "unpublished"
-      unpublishing = Unpublishing.find_by!(edition: self)
-      unpublishing.type == "gone"
+      unpublishing.gone?
     else
       document_type == "gone"
     end
```
carlosmartinez: if you fancied a quick round of code golf, you could replace the method body with

`(state == "unpublished" && unpublishing.gone?) || document_type == "gone"`

app/models/edition.rb
```diff
@@ -185,13 +185,20 @@ def substitute
...
   end
 
+  def redirect?
+    if state == "unpublished"
+      unpublishing.redirect?
+    else
+      document_type == "redirect"
+    end
+  end
+
```
carlosmartinez: as above (approximately)

PR #771, opened 2017-02-13
Change search_in to array

rubenarakelyan:
This commit changes the `search_in` param to accept an array rather than a comma-separated string, in order to be consistent with other similar params like `fields[]`.

Depends on https://github.com/alphagov/gds-api-adapters/pull/664.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -347,11 +347,11 @@ and a state has been specified, the draft is returned.
 - `per_page` *(optional, default: 50)*
   - The number of results to be shown on a given page.
 - `q` *(optional)*
-  - Search term to match against the fields in `search_in`.
-- `search_in` *(optional)*
-  - Comma-seperated list of fields to search against. Each field can indicate a
-    single level of nesting via the dot operator. If not provided, will default
-    to `title` and `base_path`.
+  - Search term to match against the fields in `search_in[]`.
+- `search_in[]` *(optional)*
+  - Array of fields to search against. Each field can indicate a single level
+    of nesting via the dot operator. If not provided, will default to
+    `title` and `base_path`.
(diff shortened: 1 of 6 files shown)


spec/queries/get_content_collection_spec.rb
```diff
@@ -429,46 +429,46 @@
 
     context "search in" do
       context "with a single nested field" do
-        let(:search_in) { "details.body" }
+        let(:search_in) { [:'details.body'] }
```
kevindew: do we benefit from these being hashes? seems we have quite a lot of dot separated so might be easier as strings

doc/api.md
```diff
@@ -347,11 +347,11 @@ and a state has been specified, the draft is returned.
 - `per_page` *(optional, default: 50)*
   - The number of results to be shown on a given page.
 - `q` *(optional)*
-  - Search term to match against the fields in `search_in`.
-- `search_in` *(optional)*
-  - Comma-seperated list of fields to search against. Each field can indicate a
-    single level of nesting via the dot operator. If not provided, will default
-    to `title` and `base_path`.
+  - Search term to match against the fields in `search_in[]`.
```
kevindew: ❤️

spec/queries/get_content_collection_spec.rb
rubenarakelyan: Yup, I got taken in by the sheer beauty of symbols and vanity prevailed over legibility. Fixed.

kevindew:
looks good - any chance you could swap the single quotes to double quotes? 🙏 

I'm going to see if we can get Rails strong params in there so we can ensure array at controller point.

rubenarakelyan:
Done

PR #775, opened 2017-02-14
Allow search_in only in specific fields

kevindew:
This specifies which fields can be searched in for GetContentCollection
as root level fields and nested fields. These are limited since using
search_in on a field type that doesn't support `ilike` will result in a
500 error.

/cc @rubenarakelyan

### DIFF
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -82,19 +82,33 @@ module Queries
       presenter::DEFAULT_SEARCH_FIELDS
     end
 
+    def allowed_search_fields
+      presenter::SEARCH_FIELDS
+    end
+
+    def allowed_nested_search_fields
+      presenter::NESTED_SEARCH_FIELDS
+    end
+
     def search_fields
       return default_search_fields if search_in.blank?
-      search_in.map do |field|
-        elements = field.split('.')
-        unless permitted_fields.include?(elements.first) && elements.length <= 2
-          raise_error("Invalid search field: #{field}")
-        end
-        if elements.length == 2
-          "#{elements[0]}->>'#{escape_nested_field(elements[1])}'"
-        else
-          elements[0]
-        end
-      end
+      search_in.map { |field| search_field_for_query(field) }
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -121,6 +121,8 @@
     t.datetime "created_at",                    null: false
     t.datetime "updated_at",                    null: false
     t.integer  "position",          default: 0, null: false
+    t.integer  "content_item_id"
```
rubenarakelyan: Is this a stray schema change?

app/queries/get_content_collection.rb
```diff
@@ -82,19 +82,33 @@ def default_search_fields
       presenter::DEFAULT_SEARCH_FIELDS
     end
 
+    def allowed_search_fields
+      presenter::SEARCH_FIELDS
+    end
+
+    def nested_search_fields
```
rubenarakelyan: Would it be clearer to call this `allowed_nested_search_fields` or similar?

app/queries/get_content_collection.rb
```diff
@@ -82,19 +82,33 @@ def default_search_fields
...
+      raise_error("Invalid search field: #{field}") unless valid_search_field(field)
+      # special case as description is a JSON object that just contains a hash
+      # of value
+      return "description->>'value'" if field == "description"
+      return field unless field.include?(".")
+      elements = field.split(".")
+      "#{elements[0]}->>'#{escape_nested_field(elements[1])}'"
+    end
+
+    def valid_search_field(field)
```
dougdroper: `valid_search_field?`

db/schema.rb
kevindew: argh! I only made this mistake the other day. Thanks for catching

kevindew:
Ok, comments and mess up addressed.

tijmenb:
👍 perhaps an update to the docs would be good as well.

kevindew:
@tijmenb d'oh - of course. Does that read ok to you?

PR #777, opened 2017-02-16
Add support for deep-nested govspeak fields

whoojemaflip:
Documents with govspeak parts that are nested mutliple levels deep
in the details hash weren't getting rendered by publishing-api.

This adds recursion to the rendering presentor, so that these schema
types can be accomodated.

Pair-coded with @emmabeynon 

Trello: https://trello.com/c/r2gA1lxN/640-migrate-guide-format-to-rendered

### DIFF
--- a/app/presenters/details_presenter.rb
+++ b/app/presenters/details_presenter.rb
@@ -32,7 +32,7 @@ module Presenters
       return render_govspeak(value) if govspeak?(wrapped_array)
       return value if value.is_a?(String)
       return value if value.respond_to?(:has_key?) && value.has_key?(:content)
-      value.map {|o| recursively_transform_govspeak(o) }
+      value.map { |o| recursively_transform_govspeak(o) }
     end
 
     def recursively_transform_govspeak(obj)
--- a/app/presenters/details_presenter.rb
+++ b/app/presenters/details_presenter.rb
@@ -12,9 +12,7 @@ module Presenters
     def details
       @_details ||=
         begin
-          updated = content_item_details.each_with_object({}) do |(key, val), seed|
-            seed[key] = append_transformed_govspeak(val)
-          end
+          updated = recursively_transform_govspeak(content_item_details)
           updated[:change_history] = change_history unless change_history.blank?
           updated
         end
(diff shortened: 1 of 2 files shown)


davidbasalla:
Looks good to me 👍 Probably best to get some 👀 from Publishing Platform on this. @kevindew maybe?

h-lame:
FWIW I think this implementation does exactly what is needed, but I wonder if there's an opportunity to leverage the schemas so the system knows exactly which keys in the details hash need govspeak transformation rather than searching for them.  (I don't know if the schemas can be used like this, or if they contain this information).

whoojemaflip:
@h-lame breaking your comment down through the post-structural lens, here. I'm going to postulate that if it quacks like a duck, you can treat is as if it is a duck. Furthermore, the labels that a schema author might use to help outline the structure of a document can not be rigorously construed to have bearing on the purpose or intent of the associated objects.

Also, that sounds like hard work.

h-lame:
@whoojemaflip sure.  I should have said that I didn't really think what I was suggesting is something that should happen *as part of this PR* even if it is possible.  I agree it would probably be a lot of work.

whoojemaflip:
@h-lame it would probably take some aggressive policing of the schema field definitions to ensure they remained within an established taxonomy of types. Might be worth looking into at some point..

kevindew:
Hey @whoojemaflip - @dougdroper is just in the process of checking this out. Sorry for the delay.

app/presenters/details_presenter.rb
```diff
@@ -12,44 +12,75 @@ def initialize(content_item_details, change_history_presenter)
...
-          end
+          updated = recursively_transform_govspeak(content_item_details)
           updated[:change_history] = change_history unless change_history.blank?
           updated
         end
     end
 
   private
 
+    def recursively_transform_govspeak(obj)
```
dougdroper: I was just trying some refactoring and you might be able to do something like: 

```ruby
    def govspeak?(value)
      value.all? { |hsh| hsh.is_a?(Hash) } &&
        value.one? { |hsh| hsh[:content_type] == "text/govspeak" } &&
        value.none? { |hsh| hsh[:content_type] == "text/html" }
    end

    def process(value)
      return unless value
      wrapped_array = Array.wrap(value)
      return render_govspeak(value) if govspeak?(wrapped_array)
      return value if value.is_a?(String)
      return value if value.respond_to?(:has_key?) && value.has_key?(:content)
      value.map {|o| recursively_transform_govspeak(o) }
    end

    def recursively_transform_govspeak(obj)
      Array(obj).each_with_object({}) do |(key, value), memo|
        memo[key] = process(value)
      end
    end
```

it feels little neater then the multi elsif, (but only a little)

whoojemaflip: @dougdroper hey Doug - please commit & merge if you'd prefer.

app/presenters/details_presenter.rb
```diff
@@ -12,44 +12,56 @@ def initialize(content_item_details, change_history_presenter)
...
+        value.none? { |hsh| hsh[:content_type] == "text/html" }
+    end
+
+    def process(value)
+      return unless value
+      wrapped_array = Array.wrap(value)
+      return render_govspeak(value) if govspeak?(wrapped_array)
+      return value if value.is_a?(String)
+      return value if value.respond_to?(:has_key?) && value.has_key?(:content)
+      value.map { |o| recursively_transform_govspeak(o) }
```
kevindew: I might be being plain naive but I feel there's gotta be a less painful way through this. I'm going to have a tinker - may regret.

dougdroper: by all means, also I think it needs a rebase. which I will leave to you so that I don't mess up the branch.

kevindew:
@dougdroper had a spin, do you think that's an improvement?

whoojemaflip:
@dougdroper @kevindew maybe we should seek a 5th opinion...?

dougdroper:
@kevindew Nah, don't think it is an improvement :) but we are bikeshedding now.

PR #778, opened 2017-02-17
Fix Attachments and World Location News Articles

mgrassotti:
Trello: https://trello.com/c/BBPaIzcQ/588-check-htmlattachment-routes-3

Some attachment translations and WLNA were not correctly migrated. We are
deleting them from publishing API and republishing them in Whitehall.

Paired with @bevanloon, @gpeng, @kevindew


db/migrate/20170217095802_delete_problematic_translations.rb
```diff
@@ -0,0 +1,20 @@
+class DeleteProblematicTranslations < ActiveRecord::Migration[5.0]
+  def change
+  	require_relative '/var/apps/publishing-api/db/migrate/helpers/delete_content.rb'
```
kevindew: this shouldn't need the full absolute path and is more conventional at the top. See this [old one](https://github.com/alphagov/publishing-api/blob/319005711dd14a7f045895540a04f0cc6acf75d8/db/migrate/20160831103251_delete_duplicate_about_pages.rb) as an example

db/migrate/20170217095802_delete_problematic_translations.rb
```diff
@@ -0,0 +1,20 @@
...
+			"056ab5a5-2fb4-4b20-b028-eb182aa7e5c4",
+			"8d435ed2-8743-4e07-9800-5b24318e5d87",
+			"0243d2dd-0581-4186-b41a-b521b8368194",
+			"0e526f62-0bc7-4ff5-ae97-a3990551253d",
+			"755db6ab-b06b-48ba-ab0e-07ea9edb4d45",
+			"125c8fb9-04fe-4e64-a562-88e241e36db0",
+			"eb5a4ddb-ab6f-468b-80a7-5078db98861a"
+		]
+
+		Helpers::DeleteContent.destroy_document_with_links(ids)
```
kevindew: we got our method name wrong here - it should be destroy_document**s**_with_links. I'm just going to quickly fix that.

db/migrate/20170217095802_delete_problematic_translations.rb
```diff
@@ -0,0 +1,20 @@
+class DeleteProblematicTranslations < ActiveRecord::Migration[5.0]
+  def change
```
kevindew: should be just an `up` migration

mgrassotti: 👍 thanks @kevindew !

kevindew:
Just fixed that method name here: https://github.com/alphagov/publishing-api/pull/779#pullrequestreview-22472299

mgrassotti:
👍

kevindew:
Feel free to merge when you're ready.

PR #780, opened 2017-02-17
Edition links

kevindew:
This PR introduces links for an edition in addition to links associated with a link set.  Further details of the story available in [Trello](https://trello.com/c/Yd7RZaaT/870-edition-link-workflow-3-blocking-core)

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -6,6 +6,7 @@ class Link < ApplicationRecord
 
   validate :link_type_is_valid
   validate :content_id_is_valid
+  validate :link_set_xor_edition_presence
 
   def self.filter_editions(scope, filters)
     join_sql = <<-SQL.strip_heredoc
@@ -24,6 +25,12 @@ class Link < ApplicationRecord
   end
 
 private
+
+  def link_set_xor_edition_presence
+    unless link_set.blank? ^ edition.blank?
+      errors.add(:base, "must be associated with a link set or an edition")
+    end
+  end
 
   def link_type_is_valid
     unless link_type.match(/\A[a-z0-9_]+\z/) && link_type != "available_translations"
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -4,8 +4,8 @@ class Link < ApplicationRecord
   belongs_to :link_set
   belongs_to :edition, foreign_key: "content_item_id"
 
+  validates :target_content_id, presence: true
   validate :link_type_is_valid
-  validate :content_id_is_valid
   validate :link_set_xor_edition_presence
 
   def self.filter_editions(scope, filters)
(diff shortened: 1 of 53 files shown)


kevindew:
@gpeng This is ready for review now. If you wouldn't mind checking out the documentation to see if it seems like it does the job that'd be great. 

These are the appropriate commits to save you reading the whole lot: https://github.com/alphagov/publishing-api/pull/780/commits/2615364618f2b48effcaf4e24e20269d25a2ebbc https://github.com/alphagov/publishing-api/pull/780/commits/4e2b793002142c1afad32ec6ae58ba252b4d4131 https://github.com/alphagov/publishing-api/pull/780/commits/061a3011a2334916199ea15a2a5613d1db90655f https://github.com/alphagov/publishing-api/pull/780/commits/b487fd5dcb804eae7b360728f9f5937a5000aeab https://github.com/alphagov/publishing-api/pull/780/commits/46a6b3842dff39975a2dee944a7d1947c8ec7987

app/models/link.rb
```diff
@@ -24,15 +26,17 @@ def self.filter_editions(scope, filters)
...
+    elsif link_set.present? && edition.present?
+      errors.add(:base, "must be associated with a link set or an edition, not both")
     end
   end
 
-  def content_id_is_valid
-    unless target_content_id.is_a?(Hash) || UuidValidator.valid?(target_content_id)
-      errors.add(:link, "target_content_id must be a valid UUID")
+  def link_type_is_valid
+    unless link_type.match(/\A[a-z0-9_]+\z/) && link_type != "available_translations"
```
carlosmartinez: don't like magic regex, and don't like "unless and"

app/models/link_graph.rb
```diff
@@ -1,25 +1,41 @@
...
+    with_drafts: false,
+    link_reference:
+  )
     @root_content_id = root_content_id
+    @root_locale = root_locale
+    @with_drafts = with_drafts
     @link_reference = link_reference
   end
 
+  # An array of LinkGraph::Node objects
```
carlosmartinez: too many comments in this class. i don't think they should be necessary.

app/models/link_graph/node_collection_factory.rb
```diff
@@ -30,17 +51,34 @@ def link_reference
...
+      link_types_path: link_types_path,
     )
   end
 
-  def valid_link_nodes(link_type, link_content_ids)
-    links = link_content_ids.map do |link_content_id|
-      LinkGraph::Node.new(link_content_id, link_type, parent_node, link_graph)
+  def valid_link_nodes(link_type, links)
+    links = links.map do |link|
+      link_content_id, link_locale, edition_id = link.is_a?(Array) ? link : [link, nil, nil]
```
carlosmartinez: please-explain

app/models/link_graph.rb
kevindew: @carlosmartinez I added the comments in those classes to make it so people who weren't that familiar with them could use them to work out what's going on with them to try debug link expansion / dependency resolution. They are inconsistent so I agree with your comment but also wondered, as someone who's not as familiar with those classes, are they usable to an outsider - do they make any sense etc?

kevindew:
ha now I have Avril in my head. 

Are you finding the implementation or the concept complicated by the way? any scope you can see to make it easier to understand?

PR #782, opened 2017-02-20
Support route redirects in Unpublishing to same degree as a redirect document type

thomasleese:
[Related Trello card][trello].

[trello]: https://trello.com/c/fv1MfDsg/877-support-route-redirects-in-unpublishing-to-same-degree-as-a-redirect-document-type-2-blocking-custom


kevindew:
@thomasleese do we need to migrate over existing unpublishings with alternative_path fields?

thomasleese:
@kevindew No, [both types work](https://github.com/alphagov/publishing-api/pull/782/files#diff-4528583b0c3735512705cf90c37bfc55R95). Although new unpublishings will be stored with only the `redirects` field set.

app/models/unpublishing.rb
```diff
@@ -32,10 +48,4 @@ def redirect?
   def self.is_substitute?(edition)
     where(edition: edition).pluck(:type).first == "substitute"
   end
-
-  def self.join_editions(edition_scope)
-    edition_scope.joins(
-      "LEFT OUTER JOIN unpublishings ON editions.id = unpublishings.edition_id"
-    )
-  end
```
kevindew: 🎉

app/validators/routes_and_redirects_validator.rb
```diff
@@ -21,7 +27,7 @@ def validate(record)
       redirects_must_not_have_routes(record, routes)
       redirects_must_include_base_path(record, redirects)
     else
-      routes_must_include_base_path(record, routes)
+      routes_must_include_base_path(record, routes) unless is_unpublishing
```
kevindew: Could we make this simpler some how? 

I'm not sure why/if it is working on an unpublishing as wouldn't they fail the `return unless record.base_path.present?` guard?

It feels like this class should either know about unpublishings and editions (which kinda sucks) or maybe it should be tailored to the class and they use the shared logic of a different validator. What do you think?

app/validators/unpublishing_redirect_validator.rb
```diff
@@ -1,14 +0,0 @@
...
-    base_path = unpublishing.edition.base_path
-
-    if base_path == unpublishing.alternative_path
-      unpublishing.errors.add(
-        :alternative_path,
-        "base_path matches the unpublishing alternative_path #{base_path}"
-      )
-    end
-  end
-end
```
kevindew: With this gone, are we still catching this scenario?

db/migrate/20170217134810_add_foreign_key_for_unpublishings.rb
```diff
@@ -0,0 +1,5 @@
+class AddForeignKeyForUnpublishings < ActiveRecord::Migration[5.0]
+  def change
+    add_foreign_key :unpublishings, :editions
```
kevindew: yay 👍  does this take long to run? 

what do you think about having a delete: cascade on unpublishings?

spec/factories/unpublishing.rb
```diff
@@ -4,5 +4,6 @@
     type "gone"
     explanation "Removed for testing reasons"
     alternative_path "/new-path"
+    #redirects [{ path: "/", destination: "/new-path", type: "exact" }]
```
kevindew: is this comment meant to be here?

spec/models/unpublishing_spec.rb
```diff
@@ -50,11 +50,36 @@
       expect(subject).to be_valid
     end
 
-    it "requires an alternative_path for a 'redirect'" do
+    it "requires a redirects blob for a 'redirect'" do
       subject.type = "redirect"
       subject.alternative_path = nil
+      subject.redirects = nil
```
kevindew: Looks like this came before you so don't worry if this seems awful. But it'd be good to alias subject to a more meaningful name (unpublishing perhaps?) and not use it explicitly

spec/models/unpublishing_spec.rb
```diff
@@ -85,7 +110,7 @@
     end
   end
 
-  describe ".is_subtitute?" do
+  describe ".is_substitute?" do
```
kevindew: 😀

spec/requests/unpublishing_spec.rb
```diff
@@ -107,46 +119,60 @@
...
       end
     end
 
-    it "does not send to the message queue" do
-      allow(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
-      allow(PublishingAPI.service(:draft_content_store)).to receive(:put_content_item)
-      expect(PublishingAPI.service(:queue_publisher)).not_to receive(:send_message)
-
-      post "/v2/content/#{content_id}/unpublish", params: redirect_params
+    context "with a new-style payload" do
```
kevindew: probably want to use a different phrasing here than new and old. Explicitly whats different would be better

app/validators/unpublishing_redirect_validator.rb
kevindew: Ah saw the test that covers this.

app/validators/routes_and_redirects_validator.rb
thomasleese: I added a `base_path` method on `Unpublishing`s to support the `base_path` check. Personally I like the idea of a single validation class that supports both `Editions` and `Unpublishing`, perhaps I can add methods to both `Edition` and `Unpublishing` which supports the differences between the two classes while allowing the validator to be the same (therefore allowing us to validate against any new classes by only implementing those methods). What do you think?

db/migrate/20170217134810_add_foreign_key_for_unpublishings.rb
thomasleese: It took 0.9 seconds to run on my machine.

thomasleese:
Since we support both kinds of redirects, I was not intending on cleaning up the database (there are tests covering both scenarios, and we use the alternative_path field for `gone` unpublishings anyway). However, I could look at introducing another PR after this has been deployed if you think that's best.

app/validators/routes_and_redirects_validator.rb
kevindew: I'd be in favour of not adding to those models where possible. Very easy for them to have lots of logic concerns that they don't need and they are used in many contexts.

So if we were to have just one class that did the validation both, I think it'd be the responsibility of that class to know how to extract the information needed - since it likely won't be needed in the same way in a different context. 

The other option for a single class would be for the validator to be called in a lambda and the key information injected in from that, which would at least encapsulate the validation concerns to the lambda. Bummer there would be if it requires a reasonable amount of code.

kevindew:
Yeah I think it'd be best if our data was consistent, everything once this goes out will be in redirects. So if you were wanting to query for an `Unpublishing` that is to a particular destination you wouldn't want to have to do an or on the `redirects` hash and the `alternative_path` field.

thomasleese:
Sure, I will prepare another PR once this one is deployed.

app/models/unpublishing.rb
```diff
@@ -14,8 +16,22 @@ class Unpublishing < ApplicationRecord
   validates :edition, presence: true, uniqueness: true
   validates :type, presence: true, inclusion: { in: VALID_TYPES }
   validates :explanation, presence: true, if: :withdrawal?
-  validates :alternative_path, presence: true, if: :redirect?
-  validates_with UnpublishingRedirectValidator
+  validates :redirects, presence: true, if: :redirect?
+  validates_with RoutesAndRedirectsValidator, document_type: :type, if: :redirect?
```
kevindew: may as well just have "redirect" rather than call the :type method. Can we pass in base_path too to save needing the delegate method around once alternative_paths are migrated to redirects?

thomasleese: We can't pass the `base_path` here as we don't know anything about the instance (`type` worked as we passed the name of the method which was then called in `RoutesAndRedirectsValidator`, but we can't name the method as `:edition.base_path`). I think the `base_path` delegate is not too bad, you could argue that an unpublishing has a `base_path`, it just happens to be the same as the edition it was before it became unpublished.

thomasleese:
I've created two more PRs to support deployment in #787 and #788 .

PR #784, opened 2017-02-20
Allow setting first_published_at when updating a draft edition

thomasleese:
[Related Trello card][trello].

[trello]: https://trello.com/c/GuXm7xI1/884-allow-updating-first-published-at-blocking-core

### DIFF
--- a/app/models/update_existing_draft_edition.rb
+++ b/app/models/update_existing_draft_edition.rb
@@ -6,7 +6,6 @@ class UpdateExistingDraftEdition
     :locale,
     :created_at,
     :updated_at,
-    :first_published_at,
     :last_edited_at,
   ].freeze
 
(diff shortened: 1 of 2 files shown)


app/models/update_existing_draft_edition.rb
```diff
@@ -45,10 +44,17 @@ def assign_attributes_with_defaults
     edition.assign_attributes(new_attributes)
   end
 
+  def remove_first_published_at_if_not_explicitly_set(result)
+    result.delete(:first_published_at) unless payload.include?(:first_published_at)
```
dougdroper: can we use `tap`

spec/commands/v2/put_content_spec.rb
```diff
@@ -332,7 +332,17 @@
         expect(previously_drafted_item.content_store).to eq("draft")
       end
 
-      it "keeps the first_published_at timestamp if present" do
+      it "allows the setting of first_published_at" do
+        explicit_first_published = DateTime.new(2016, 05, 23, 1, 1, 1).rfc3339
+        payload[:first_published_at] = explicit_first_published
+
+        described_class.call(payload)
+        previously_drafted_item.reload
```
dougdroper: do we need this

PR #785, opened 2017-02-20
Use 'nonexistent-schema' as a schema_name in the tests

thomasleese:
A schema_name of type 'guide' was introduced which means all the
tests relying on it not existing stopped working. This is just a
quick fix to get the tests passing again and will be dealt with
properly at a later date.


kevindew:
is that relying on guidance not existing?

thomasleese:
@kevindew Yeah, it does.

kevindew:
@thomasleese shall we just go and name it something like nonexistant-schema ?

PR #786, opened 2017-02-20
Capture orphaned links before we delete them

dougdroper:
Given a -> b, if we change a -> c, b still thinks it has a child of a,
but since we severed this relation, b has no reference to a, which means
that it is missed by dependency resolution.

This captures those relations before we remove them, and passes them
down into the dependency resolution worker.

### DIFF
--- a/app/workers/dependency_resolution_worker.rb
+++ b/app/workers/dependency_resolution_worker.rb
@@ -11,6 +11,8 @@ class DependencyResolutionWorker
       send_downstream(content_id, locale)
     end
 
+    # FIXME: we need to do all locales not just 'en'
+    # make sure we update this when we come to edition level implementation.
     orphaned_content_ids.each { |content_id| send_downstream(content_id, "en") }
   end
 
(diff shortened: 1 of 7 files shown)


app/workers/dependency_resolution_worker.rb
```diff
@@ -8,17 +8,15 @@ def perform(args = {})
...
     dependencies.each do |(content_id, locale)|
-      if draft?
-        downstream_draft(content_id, locale)
-      else
-        downstream_live(content_id, locale)
-      end
+      send_downstream(content_id, locale)
     end
+
+    orphaned_content_ids.each { |content_id| send_downstream(content_id, "en") }
```
kevindew: I think we'd want all locales to be included in this? (If I remember right that's what content dependencies does)

dougdroper: Yes, I am sure it does. I was going to care a bit more about that with edition level links, didn't want to overly complicate it without knowing the edition implementation, but can hold off till that work is finished.

kevindew: Sure, do you want a jot a fixme in there?

dougdroper: good thinking

PR #787, opened 2017-02-21
Update old redirect unpublishings

thomasleese:
This should be deployed after #782 to migrate old-style `alternative_path` redirect unpublishings.

Takes about 85 seconds to run on my machine.

### DIFF
--- a/db/migrate/20170221093606_update_old_redirect_unpublishings.rb
+++ b/db/migrate/20170221093606_update_old_redirect_unpublishings.rb
@@ -1,0 +1,13 @@
+class UpdateOldRedirectUnpublishings < ActiveRecord::Migration[5.0]
+  def up
+    Unpublishing
+      .where(type: :redirect)
+      .where(redirects: nil)
+      .find_each do |unpublishing|
+        # this magical line works because the 'redirects' getter is overriden
+        # to generate a JSON blob automatically if it is nil within the
+        # database
+        unpublishing.update_attribute(:redirects, unpublishing.redirects)
+      end
+  end
+end


thomasleese:
Now that #782 has been deployed, we can look at merging this. It is a long running migration, though, so worth taking time to plan a strategy of deployment.

db/migrate/20170221093606_update_old_redirect_unpublishings.rb
```diff
@@ -0,0 +1,24 @@
...
+      .where(redirects: nil)
+      .find_each do |unpublishing|
+        # this magical line works because the 'redirects' getter is overriden
+        # to generate a JSON blob automatically if it is nil within the
+        # database
+        unpublishing.update_attribute(:redirects, unpublishing.redirects)
+      end
+  end
+
+  def down
```
carlosmartinez: this `down` method doesn't look like it actually reverses the effect of `up`. am i missing something?

thomasleese: True, it doesn't do the exact reverse. The `down` method here is more of a safety net in case we needed to rollback the code and any new unpublishings made while the new code was active, wouldn't have an `alternative_path` set.

carlosmartinez: others might have (better, more informed) opinions, but i don't think a `down` migration is the place to have that safety net. @dougdroper @kevindew whatchu reckon?

kevindew: Yeah I'd agree that since redirects isn't necessarily compatible with alternative path it might not be safe to do so. Presumably we should already have relatively good faith that the redirects collection works before this goes out too?

thomasleese:
Currently on integration, 6379 are using the new `redirects` hash and 26602 are not.

kevindew:
@thomasleese any chance you could push up a schema.rb for this? @michaelihejirika can't run his tests 😞  - we'll have to sort this out on e2e

thomasleese:
@kevindew I can, but it won't have any changes. When I created this PR I updated `schema.rb` but a subsequent PR (which added another migration) was merged and when rebased, the date of this migration was before what was in `schema.rb` in master, so `schema.rb` didn't need changing.

kevindew:
@thomasleese ah sorry, it might be that his repo is not up to date - I jumped to a conclusion. Thanks for having a look

PR #788, opened 2017-02-21
Remove backwards compatible redirect unpublishing code

thomasleese:


spec/factories/unpublishing.rb
```diff
@@ -4,5 +4,17 @@
...
+    after(:build) do |unpublishing|
+      unless unpublishing.redirects
+        unpublishing.redirects = [
+          {
+            path: unpublishing.edition.base_path,
+            type: :exact,
+            destination: "/new-path"
+          }
+        ]
+      end
```
kevindew: should this only be inserted when the unpublishing type is redirect?

thomasleese: Yep, fixed. 👍

spec/models/unpublishing_spec.rb
```diff
@@ -26,6 +25,11 @@
       expect(subject).to be_invalid
       expect(subject.errors[:type].size).to eq(1)
 
+      # because we haven't given any 'redirects'
```
carlosmartinez: given we have four sets of assertions in this test, perhaps it'd be a good idea to separate them out into separate contexts? that way we could also dispense with the comment.

PR #794, opened 2017-02-23
Destroy some WLNA that are failing sync checks.

bevanloon:


db/migrate/20170223082047_destroy_some_more_bad_wlna.rb
```diff
@@ -0,0 +1,17 @@
+require_relative "helpers/delete_content"
+
+class DestroySomeMoreBadWlna < ActiveRecord::Migration[5.0]
+  def change
```
thomasleese: I think this should only be run during an `up` operation.

bevanloon: 👍 fixed in 085a48d1c350cf319799d4bbc09065ff18cbef35

bevanloon:
Checked in `schema.rb` as well in 085a48d1c350cf319799d4bbc09065ff18cbef35

PR #796, opened 2017-02-23
Index edition on (document_type, updated_at)

MatMoore:
In `GET linkables` we check to see when the latest updated was to documents of a
particular type. This query can take some time to complete at the moment
because the document_type may not be updated very frequently, so the database
would have to scan through a lot of records if just using the updated_at index.

I think this means we can also remove the `document_type` index

cc @kevindew


kevindew:
Thanks @MatMoore would you be able to create the index concurrently: https://robots.thoughtbot.com/how-to-create-postgres-indexes-concurrently-in as it looks like this will lock the table for about 30 seconds or so.

PR #797, opened 2017-02-23
Use a single query, with no Edition objects for linkables endpoint

cbaines:
Benchmarking this using the linkables benchmark, but just the 'need' document type gives the following results.

Before:
24.380000   0.150000  24.530000 ( 32.389552)
queries: 45

After:
1.460000   0.060000   1.520000 (  3.584924)
queries: 22

### DIFF
--- a/app/queries/get_linkables.rb
+++ b/app/queries/get_linkables.rb
@@ -1,17 +1,4 @@
 module Queries
-  LinkablePresenter = Struct.new(:title, :content_id, :publication_state, :base_path, :internal_name) do
-    def self.from_hash(hash)
-      fields = [
-        hash[:title],
-        hash[:content_id],
-        hash[:state],
-        hash[:base_path],
-        hash[:details].fetch(:internal_name, hash[:title]),
-      ]
-      new(*fields)
-    end
-  end
-
   class GetLinkables
     def initialize(document_type:)
       @document_type = document_type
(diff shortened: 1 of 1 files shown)


kevindew:
Thanks @cbaines this looks good, I did some [benchmarking](https://gist.github.com/kevindew/592635ff85d48eb8d2bc53b74c21ad80) and it does make a decent difference to just pluck.

An extra bonus is I know about `FULL OUTER JOIN` 👍 

A couple of things though:

This seems to change behaviour from the previous iteration, in regards to this queries based off content store rather than state, and this returns all locales rather than just en. I don't think this optimisation should aim to change behaviour - though it is a shame that we don't test for this behaviour. So I'd suggest here that we add a test that makes this explicit and we fix this against that.

Also, I wonder if you tried the `DISTINCT` and `ORDER BY` technique that is used in [get_edition_ids_with_fallbacks](https://github.com/alphagov/publishing-api/blob/master/app/queries/get_edition_ids_with_fallbacks.rb). I'm thinking that using that technique this could be switched from being 2 queries to 1 but naturally could perform worse than two. The other advantage that if these used the same technique would be the consistency in how we solve this problem. What do you think?

cbaines:
> Also, I wonder if you tried the DISTINCT and ORDER BY technique that is used in get_edition_ids_with_fallbacks. I'm thinking that using that technique this could be switched from being 2 queries to 1 but naturally could perform worse than two. The other advantage that if these used the same technique would be the consistency in how we solve this problem. What do you think?

I've just tried [1] the most minimal changes I could think of to get it down to one query, but it does indeed perform worse when testing with the benchmark (however, because this avoids two whole round trips with big payloads in the worst case, it might actually perform better in a production like environment).

The semantics in this make sense to me, but as you say, the tests are very permissive, and I'm not aware that the finer details of what this endpoint should do are documented anywhere, so I'm inclined to abandon this change, at least for now.

1: 8d2a5dd6ef54ebdeb81b4e1903c2215fad7d4b20

PR #801, opened 2017-02-27
Delete some corrupt content items

gpeng:
This commit adds a migration to delete two content items that have become corrupt leading to Whitehall not being able to republish them.

The associated Whitehall documents will be republished once this migration has run.


kevindew:
@gpeng can you include the schema.rb please?

gpeng:
Doh... Sorry. I've updated it. I always think of these as data migrations a la Whitehall and they don't update the schema version. I only speak Whitehall these days 🤕

PR #804, opened 2017-02-28
Support expanding nested fields

danielroseman:
This came out of an investigation about how we would render the travel advice index via link expansion. To do that, we would need the country and change description elements from the details hashes to be present in the expanded links - but we wouldn't want the entire details, as that would result in 225 entire documents being embedded in the index.

The proposal is to allow elements in the list of CUSTOM_FIELDS_FOR_DOCUMENT_TYPE to be arrays; the expand_fields method would then follow the array and extract the element, using the last name in the array as the key in the expanded hash.

Opinions welcome (on the principle, not necessarily the implementation).

### DIFF
--- a/lib/link_expansion/rules.rb
+++ b/lib/link_expansion/rules.rb
@@ -109,13 +109,16 @@ module LinkExpansion::Rules
   end
 
   def potential_expansion_fields(document_type)
-    find_custom_expansion_fields(document_type) || DEFAULT_FIELDS
+    (find_custom_expansion_fields(document_type) || DEFAULT_FIELDS).map do |field|
+      Array(field).first
+    end
   end
 
   def expand_fields(edition, link_type)
-    edition.to_h.slice(
-      *expansion_fields(edition.document_type, link_type)
-    )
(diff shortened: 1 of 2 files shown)


kevindew:
Principle looks good to me. 👍 

Thinking aloud... I wonder if it might be easier to grok if they are defined as dot separated strings eg "details.country" but I don't know if we could bitten by period being a valid JSON character. Maybe %i(details country) probably in bikeshedding territory here.

danielroseman:
I've rebased this and updated to work with the recent changes to link expansion.

In terms of the syntax, if they were dot-separated strings we'd have to do some ugly conversion to and from symbols, not sure it would be worth it.

lib/link_expansion/rules.rb
```diff
@@ -113,9 +113,10 @@ def potential_expansion_fields(document_type)
   end
 
   def expand_fields(edition, link_type)
```
thomasleese: The method above, `potential_expansion_fields`, will need to be modified to return a flattened list of fields in this case, i.e. `:details` should be returned in this list directly.

danielroseman: Done.

PR #806, opened 2017-02-28
[DO NOT MERGE] Allow lookup requests to return draft content

MatMoore:
We want to be able to tag to draft content in content tagger.

For example, content tagger is now the only way to tag certain things to mainstream content (eg need ids).

This breaks at the moment because we can't lookup the content, which affects both bulk tagging and the regular tagging workflow.

I've made it return all drafts that don't have an access limit.

https://trello.com/c/6VPQcc7C/507-content-tagger-should-be-able-to-tag-to-the-first-draft-of-a-document

We shouldn't merge this until we've made content tagger handle drafts (show the state, and replace live links with draft-origin links).

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,17 +1,20 @@
 class LookupsController < ApplicationController
   def by_base_path
-    # return content_ids for content that is visible on the live site
-    # withdrawn items are still visible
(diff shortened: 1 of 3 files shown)


MatMoore:
Closing this in favour of https://github.com/alphagov/publishing-api/issues/812

PR #808, opened 2017-03-01
Link expansion on document type

thomasleese:
Allow ability to perform custom field link expansion based on the link type. Also, expand the finder details hash on `finder` link types.

### DIFF
--- a/lib/link_expansion/rules.rb
+++ b/lib/link_expansion/rules.rb
@@ -53,6 +53,7 @@ module LinkExpansion::Rules
     { document_type: :placeholder_organisation,   fields: DEFAULT_FIELDS_WITH_DETAILS },
     { document_type: :taxon,                      fields: DEFAULT_FIELDS_WITH_DETAILS },
     { document_type: :need,                       fields: DEFAULT_FIELDS_WITH_DETAILS },
+    { document_type: :finder, link_type: :finder, fields: DEFAULT_FIELDS_WITH_DETAILS },
   ].freeze
 
   def root_reverse_links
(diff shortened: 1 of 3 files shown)


tijmenb:
Would you be able to add an example for this? I think I understand what this is doing but want to make sure.

thomasleese:
@tijmenb If the `link_type` is `finder` and the `document_type` of the target is also `finder`, this code will expand the `details` hash during link expansion in order to allow `specialist-frontend` to render the pages from a single content item.

app/presenters/queries/available_translations.rb
```diff
@@ -25,7 +25,7 @@ def grouped_translations
 
       def expand_translation(id)
         expansion_rules = LinkExpansion::Rules
-        web_item(id).select { |f| expansion_rules.expansion_fields(:available_translations).include?(f) }
+        web_item(id).select { |f| expansion_rules.expansion_fields(:available_translations, nil).include?(f) }
```
kevindew: maybe this parameter could be optional or passed as a hash? The nil lacks any contextual meaning

lib/link_expansion/rules.rb
```diff
@@ -42,7 +42,8 @@ def recurring
     :withdrawn,
   ].freeze
 
-  CUSTOM_FIELDS_FOR_DOCUMENT_TYPE = {
+  CUSTOM_EXPANSION_FIELDS = {
```
kevindew: maybe these could be defined as hashes? ie
```
{ document_type: :redirect, fields: [] } 
```
then the code might do the talking rather than the comments? though downside is it could seem overly verbose?

thomasleese: I'll give it a go, see what it looks like.

thomasleese: I don't think it's overly verbose, makes everything nice and clear.

PR #811, opened 2017-03-01
Ensure all documents have link sets

MatMoore:
PutContent used to create a LinkSet when creating a draft, but this got removed in 050ca98b44f2460b1668061a2d5f4dfe4bfcf01a.

This affects `GET links` and `GET expanded-links`

Both of these now return 404 for documents that have never had links set, which complicates building tagging interfaces that work with draft content.

`LinkSet` is an implementation detail: we expect to be able to get links at any time regardless of publishing status of the document.

There are different ways this could work.

- Rather than create the `LinkSet`s, the requests that read them could handle the case where the `LinkSet` is missing, but the document exists.
- We could eliminate `LinkSet` completely: I think now that we have a `Document` model, we don't need it: conceptually a `Link` belongs to either an `Edition` or a `Document` (via its `content_id`).
- Or we could just restore the previous implementation, which is what I've done in this PR.

https://trello.com/c/6VPQcc7C/507-content-tagger-should-be-able-to-tag-to-the-first-draft-of-a-document

### DIFF
--- a/app/queries/get_expanded_links.rb
+++ b/app/queries/get_expanded_links.rb
@@ -1,31 +1,42 @@
 module Queries
   class GetExpandedLinks
     def self.call(content_id, locale)
-      link_set = find_link_set(content_id)
+      if (link_set = LinkSet.find_by(content_id: content_id))
+        expanded_link_set(link_set, locale)
+      elsif Document.where(content_id: content_id).exists?
+        empty_expanded_link_set(content_id)
+      else
+        error_details = {
+          error: {
+            code: 404,
+            message: "Could not find link set with content_id: #{content_id}"
+          }
+        }
+
+        raise CommandError.new(code: 404, error_details: error_details)
+      end
+    end
+
+    def self.expanded_link_set(link_set, locale)
       expanded_link_set = Presenters::Queries::ExpandedLinkSet.new(
-        content_id: content_id,
+        content_id: link_set.content_id,
         draft: true,
         locale: locale,
       )
 
       {
-        content_id: content_id,
+        content_id: link_set.content_id,
         expanded_links: expanded_link_set.links,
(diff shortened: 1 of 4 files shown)


thomasleese:
This implementation seems fine to me.

I would prefer, however, if we went with the first option: handling the lack of a link set case at a presenter/controller level. At the very least, it means we avoid having to run that migration; but it would be nice if we could eventually remove `LinkSet`s which I think that way encourages.

danielroseman:
Agree that in the long term we should probably get rid of Linksets, I've added to the list of things to consider in https://trello.com/c/pU7beFpK/32-review-other-entities-in-the-system-timebox

MatMoore:
@thomasleese do you think it's worth reworking it to follow option 1? I'd want to make sure it's applied consistently - is there anything else besides GET links and GET expanded-links that is affected?

thomasleese:
@MatMoore Yeah, I think on balance that would be best. I don't believe there are any other endpoints affected.

MatMoore:
Cool, I'll rework it then

MatMoore:
I've reimplemented this as an extra case in get_links and get_expanded_links. If it's ok I can squash the commits.

app/queries/get_expanded_links.rb
```diff
@@ -1,31 +1,42 @@
 module Queries
   class GetExpandedLinks
     def self.call(content_id, locale)
-      link_set = find_link_set(content_id)
+      if (link_set = LinkSet.find_by(content_id: content_id))
+        expanded_link_set(link_set, locale)
+      elsif (Document.where(content_id: content_id).exists?)
```
thomasleese: I think the brackets on the `if` look a bit weird here.

MatMoore:
@thomasleese I've removed the earlier commits.

PR #814, opened 2017-03-02
Fix only superseded editions

thomasleese:
[Trello card](https://trello.com/c/QZn5BbDV/879-fix-editions-which-are-all-marked-as-superseded-2)

### DIFF
--- a/db/migrate/20170302134438_fix_only_superseded_editions.rb
+++ b/db/migrate/20170302134438_fix_only_superseded_editions.rb
@@ -1,0 +1,35 @@
+require_relative "helpers/delete_content"
+
+class FixOnlySupersededEditions < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  def up
+    content_ids_to_delete = []
+
+    Document
+      .joins(:editions)
+      .where(%q(
+        NOT EXISTS (
+          SELECT id FROM editions AS e
+          WHERE e.state IN ('draft', 'published', 'unpublished')
+            AND e.document_id = documents.id
+        )
+        AND editions.publishing_app IN ('service-manual-publisher', 'whitehall')
+        AND editions.updated_at < '2017-01-18'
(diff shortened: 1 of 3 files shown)


db/migrate/20170302134438_fix_only_superseded_editions.rb
```diff
@@ -1,16 +1,20 @@
...
-          .update_column(:state, "unpublished")
-      end
+      ))
+
+    docs.find_each do |document|
+      document.editions.order(updated_at: :desc).first
+        .update_column(:state, "unpublished")
+    end
+
+    Commands::V2::RepresentDownstream.new.(docs.map(&:content_id))
```
kevindew: This will probably go wrong as the transaction is not committed at the time this is called. Is it also a pain point running locally?

db/migrate/20170302134438_fix_only_superseded_editions.rb
```diff
@@ -1,16 +1,20 @@
...
         )
-      )).find_each do |document|
-        document.editions.order(updated_at: :desc).first
-          .update_column(:state, "unpublished")
-      end
+      ))
+
+    docs.find_each do |document|
+      document.editions.order(updated_at: :desc).first
+        .update_column(:state, "unpublished")
```
kevindew: If I recall correctly only a few of these have corresponding unpublishing objects and I believe we expect them to exist in other parts of the code, so would have thought this might go boom? you could try altering the unpublished edition factory to not include an unpublishing object and see what blows up in the tests?

thomasleese: Hmm, yes, I think you're right. We don't know anything about what should be in the `Unpublishing` instance (`type`, `explanation`, `unpublished_at`, etc) though. Do you think a better fix would be to make the system cope with unpublished editions without an `Unpublishing` or just pick a suitable `type` and leave it at that.

PR #816, opened 2017-03-02
Fix dependency resolution difference on finder link types

thomasleese:
This is a bug that was accidentally missed by early merge of #808 .

### DIFF
--- a/lib/edition_diff.rb
+++ b/lib/edition_diff.rb
@@ -1,5 +1,3 @@
-require 'hashdiff'
-
 class EditionDiff
   attr_reader :current_edition, :version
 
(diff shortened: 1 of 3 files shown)


lib/edition_diff.rb
```diff
@@ -10,16 +10,21 @@ def initialize(current_edition, version: nil, previous_edition: nil)
...
-    HashDiff.best_diff(
+    hash_diff(
       @previous_edition ? @previous_edition.to_h.deep_symbolize_keys : previous_edition.to_h.deep_symbolize_keys,
-      current_edition.to_h.deep_symbolize_keys)
+      current_edition.to_h.deep_symbolize_keys
+    )
+  end
+
+  def hash_diff(a, b)
+    a.size > b.size ? a.to_a - b.to_a : b.to_a - a.to_a
```
carlosmartinez: sneaky

lib/edition_diff.rb
```diff
@@ -10,16 +10,21 @@ def initialize(current_edition, version: nil, previous_edition: nil)
...
-      diff.map { |_, field, _| field.to_sym }
+    LinkExpansion::Rules.finder_expansion_fields(current_edition.document_type) &
+      diff.map { |key, value| key }
   end
 
 private
 
   def diff
-    HashDiff.best_diff(
+    hash_diff(
```
dougdroper: we should remove the require at the top

PR #817, opened 2017-03-02
Update draft content store on Publish

brenetic:
Previously the publishing api was not updating the draft content store
when a piece of content was published with a change. This resulted in disparity
between the actual and expected content in the draft content store.

Specifically we noticed this because the publish command removes the
`access_limited` hash in the live content store but did not in the draft
content store.

We now will trigger the `DownstreamDraftWorker` if there is a access
limited hash.

Mobbed with Christopher Baines and Emma Beynon

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -6,8 +6,11 @@ module Commands
 
         publish_edition
 
-        after_transaction_commit do
-          send_downstream(document.content_id, document.locale, update_type)
+        if downstream
+          after_transaction_commit do
+            send_downstream_live
+            send_downstream_draft if access_limit
+          end
         end
 
         Success.new(content_id: content_id)
@@ -35,8 +38,12 @@ module Commands
         Action.create_publish_action(edition, document.locale, event)
       end
 
+      def access_limit
+        @_access_limit ||= AccessLimit.find_by(edition: edition)
+      end
+
       def remove_access_limit
-        AccessLimit.find_by(edition: edition).try(:destroy)
+        access_limit.try(:destroy)
       end
 
       def validate
@@ -171,19 +178,35 @@ module Commands
         EditionDiff.new(edition).field_diff.present?
       end
 
-      def send_downstream(content_id, locale, update_type)
-        return unless downstream
-
+      def send_downstream_live
         queue = update_type == 'republish' ? DownstreamLiveWorker::LOW_QUEUE : DownstreamLiveWorker::HIGH_QUEUE
-
         DownstreamLiveWorker.perform_async_in_queue(
           queue,
+          live_worker_params
+        )
+      end
+
+      def send_downstream_draft
+        queue = update_type == 'republish' ? DownstreamDraftWorker::LOW_QUEUE : DownstreamDraftWorker::HIGH_QUEUE
+        DownstreamDraftWorker.perform_async_in_queue(
+          queue,
+          worker_params
+        )
+      end
+
+      def live_worker_params
+        {
+          message_queue_update_type: update_type,
+          update_dependencies: update_dependencies?
+        }.merge(worker_params)
+      end
+
+      def worker_params
+        {
           content_id: content_id,
           locale: locale,
-          message_queue_update_type: update_type,
           payload_version: event.id,
-          update_dependencies: update_dependencies?,
-        )
+        }
       end
     end
   end
--- a/spec/requests/content_item_requests/downstream_requests_spec.rb
+++ b/spec/requests/content_item_requests/downstream_requests_spec.rb
@@ -218,8 +218,8 @@ RSpec.describe "Downstream requests", type: :request do
     end
 
     it "doesn't send draft dependencies to the live content store" do
-      expect(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
-        .with(a_hash_including(base_path: '/a'))
+      allow(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
+      allow(PublishingAPI.service(:draft_content_store)).to receive(:put_content_item)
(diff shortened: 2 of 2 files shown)


app/commands/v2/publish.rb
```diff
@@ -171,19 +174,29 @@ def update_dependencies?
...
           queue,
+          worker_params
+        )
+      end
+
+      def send_downstream_draft
+        queue = update_type == 'republish' ? DownstreamDraftWorker::LOW_QUEUE : DownstreamDraftWorker::HIGH_QUEUE
+        DownstreamDraftWorker.perform_async_in_queue(
+          queue,
+          worker_params.merge(update_dependencies: update_dependencies?)
```
kevindew: Whereas here I think we want to have `update_dependencies` as false as if only concern is access limits we don't need to update any dependencies

app/commands/v2/publish.rb
```diff
@@ -6,8 +6,11 @@ def call
 
         publish_edition
 
-        after_transaction_commit do
-          send_downstream(document.content_id, document.locale, update_type)
+        if downstream
+          after_transaction_commit do
+            send_downstream_live
+            send_downstream_draft
```
kevindew: we could just do this only if access limits exist. Code might be a little ugly to do it as we'd have to check for existence pre publish_edition, but would save an unnecessary represent downstream call

spec/requests/content_item_requests/downstream_requests_spec.rb
```diff
@@ -218,15 +218,16 @@
     end
 
     it "doesn't send draft dependencies to the live content store" do
-      expect(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
-        .with(a_hash_including(base_path: '/a'))
+      allow(PublishingAPI.service(:live_content_store)).to receive(:put_content_item)
+      allow(PublishingAPI.service(:draft_content_store)).to receive(:put_content_item)
```
kevindew: did we lose part of the test here?

app/commands/v2/publish.rb
```diff
@@ -171,19 +174,29 @@ def update_dependencies?
...
+          queue,
+          worker_params.merge(update_dependencies: update_dependencies?)
+        )
+      end
+
+      def worker_params
+        {
           content_id: content_id,
           locale: locale,
           message_queue_update_type: update_type,
```
kevindew: It's not of major consequence but the downstream draft worker doesn't accept this attribute by the way.

app/commands/v2/publish.rb
```diff
@@ -171,19 +174,29 @@ def update_dependencies?
...
 
-      def send_downstream(content_id, locale, update_type)
-        return unless downstream
-
+      def send_downstream_live
         queue = update_type == 'republish' ? DownstreamLiveWorker::LOW_QUEUE : DownstreamLiveWorker::HIGH_QUEUE
-
         DownstreamLiveWorker.perform_async_in_queue(
           queue,
+          worker_params
```
cbaines: The intent here is to not update dependencies through the DownstreamLiveWorker, and instead update them through the DownstreamDraftWorker. I believe the decision to do it this way was made because the DownstreamDraftWorker resolves dependencies for both content stores, whereas the DownstreamLiveWorker only resolves dependencies for the live content store.

However, thinking about this again, I think I see the problem. Dependency resolution for the draft content store would have happened when the put content request was made, so its unnecessary to do it a 2nd time upon publish. Therefore the better way is as you say to keep the DownstreamDraftWorker doing the dependency resolution.

app/commands/v2/publish.rb
cbaines: Yep, I've replied here https://github.com/alphagov/publishing-api/pull/817#discussion_r104128148

spec/requests/content_item_requests/downstream_requests_spec.rb
cbaines: Yes, an `expect` was changed to an `allow` as the expectation didn't relate to what the test said it would test.

app/commands/v2/publish.rb
kevindew: cool - last sentence do you mean DownstreamLiveWorker?

spec/requests/content_item_requests/downstream_requests_spec.rb
kevindew: Ah sure, might be clearer as a separate commit but no biggie

app/commands/v2/publish.rb
cbaines: Yep, good spot :)

app/commands/v2/publish.rb
cbaines: @brenetic has now made this change.

app/commands/v2/publish.rb
cbaines: @brenetic has now fixed this.

brenetic:
@kevindew  Have updated the PR 🚀

app/commands/v2/publish.rb
```diff
@@ -171,19 +178,35 @@ def update_dependencies?
...
 
-      def send_downstream(content_id, locale, update_type)
-        return unless downstream
-
+      def send_downstream_live
         queue = update_type == 'republish' ? DownstreamLiveWorker::LOW_QUEUE : DownstreamLiveWorker::HIGH_QUEUE
-
         DownstreamLiveWorker.perform_async_in_queue(
           queue,
+          live_worker_params.merge(worker_params)
```
kevindew: I feel like this merge belongs in the live_worker_params method, what do you think?

kevindew:
Thanks @brenetic - sorry one nag!

kevindew:
@brenetic feel free to merge

PR #820, opened 2017-03-03
Make potential expansion fields a supported scenario

thomasleese:


lib/link_expansion/rules.rb
```diff
@@ -92,23 +92,24 @@ def next_dependency_resolution_link_types(link_types_path)
...
 
-  def find_custom_expansion_fields(document_type, link_type)
+  def find_custom_expansion_fields(document_type, options = {})
+    should_check_link_type = options.include?(:link_type)
+    link_type = options[:link_type]
+
     condition = CUSTOM_EXPANSION_FIELDS.find do |cond|
-      cond[:document_type] == document_type &&
-        cond.fetch(:link_type, link_type) == link_type
+      return false if should_check_link_type && cond.fetch(:link_type, link_type) != link_type
```
kevindew: this will return from the actual method I believe rather than the block - is that intentional? you probably want a next or break?

lib/link_expansion/rules.rb
```diff
@@ -92,23 +92,24 @@ def next_dependency_resolution_link_types(link_types_path)
...
+      cond[:document_type] == document_type
     end
     condition[:fields] if condition
   end
 
   def expansion_fields(document_type, link_type = nil)
-    find_custom_expansion_fields(document_type, link_type) || DEFAULT_FIELDS
+    find_custom_expansion_fields(document_type.to_sym, link_type: link_type.try(:to_sym)) ||
```
kevindew: can we handle the to_sym aspect within the `find_custom_expansion_fields`?

PR #821, opened 2017-03-06
Add a migration to fix and represent specialist documents

thomasleese:
Migration takes 15 minutes to run.

### DIFF
--- a/db/migrate/20170306135152_represent_specialist_documents.rb
+++ b/db/migrate/20170306135152_represent_specialist_documents.rb
@@ -1,0 +1,63 @@
+class RepresentSpecialistDocuments < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  DOCUMENT_TYPE_CONTENT_IDS = {
+    aaib_report: "b7574bba-969f-4c49-855a-ae1586258ff6",
+    asylum_support_decision: "581b4bba-e58f-4d07-8e0a-03c8c00165cc",
+    business_finance_support_scheme: "12fecc5b-c02a-405e-b96e-7aa32ceaf504",
+    cma_case: "fef4ac7c-024a-4943-9f19-e85a8369a1f3",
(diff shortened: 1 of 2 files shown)


db/migrate/20170306135152_represent_specialist_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def up
+    content_ids_to_represent = []
+
+    Edition
+      .with_document.includes(:document)
+      .where(document_type: DOCUMENT_TYPE_CONTENT_IDS.keys)
+      .where(publishing_app: "specialist-publisher")
+      .where.not(state: "superseded")
```
kevindew: is it problematic if we do this to superseded items? feels like it'd be more complete if we can do this if we can.

db/migrate/20170306135152_represent_specialist_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+      .find_each do |edition|
```
kevindew: we might be able to do this through find_each.flat_map and return a value rather than mutate one. 

Not sure if that'll make it much more readable or not though?

db/migrate/20170306135152_represent_specialist_documents.rb
thomasleese: No problem, in fact this is how I had it originally. I decided to ignore them because the migration took 20 minutes to run before I included this restriction. It's not a blocking migration, but I remember when we were doing the documents and editions work we had issues with timeouts. If we really want this behaviour, I could look at structuring the entire migration differently, or splitting it up by document type.

thomasleese:
There are no cases where items have a `parent` link, so there is no risk of duplicate links.

```
SELECT links.* FROM editions INNER JOIN documents ON documents.id = editions.document_id INNER JOIN link_sets ON link_sets.content_id = documents.content_id INNER JOIN links ON links.link_set_id = link_sets.id WHERE publishing_app = 'specialist-publisher' AND link_type = 'parent' AND links.created_at < '2017-03-06';
```

Issue #812, opened 2017-03-02
Proposal for new version of lookup-by-base-path


MatMoore:
# Background

In regular publishing apps, we refer to documents by content ID, and using base paths as an identifier is discouraged. `GET /v2/content` can be used to search for editions.

In content tagger, things work a bit differently. We work at the document level rather than with specific editions, and use `PATCH /v2/links` to add links via the LinkSet.

For the basic tagging workflow, users enter a URL or base path of anything on GOV.UK, and we lookup the content id with `POST /lookup-by-base-path`.

This request may take multiple content ids to lookup simultaneously, which we use for bulk tagging workflows.

# Problem

Currently if a document has not been published yet, we can't look up the item in content tagger.

We want to allow content tagger to work with any draft content, while continuing to support live content (which may be unpublished and withdrawn). We should distinguish between the two so the user can either view their changes immediately (live) or preview them on the draft-origin (draft).

For error handling, it would also be useful to distinguish between content that doesn't exist, and content that used to exist, but has been redirected/removed.

This is something we noticed when bulk tagging to the education taxonomy, because while identifying content to tag to the taxonomy, education publishers also improved the content and consolidated a lot of pages. In this case the bulk imports triggered a lot of "Content not on GOV.UK" errors.

# Proposal

I'd like to create a new version of `/lookup-by-base-path`, that is less restrictive and returns more context about what content store it's in and whether it's been unpublished.

## Request

```
POST /v2/lookup-by-base-path

base_path[]=/foo&base_path[]=/bar
```

## Response

In the response there is now an object representing the current state of each content store.

```
{
  "/foo" => {
    {
     draft: {
      content_id: "123123-12312-12312",
     }
     live: {
       content_id: "1213123-12312-12552",
       unpublishing: {
         type: "redirect"
         alternative_path: "/foo",
       },
      },
    }
  }
  "/bar" => {
    {
     draft: {
      content_id: "123123-12312-12312",
     }
     live: {
       content_id: "1213123-12312-12552",
     }
    }
  }
}
```

- If the content id never existed, the base path is not returned in the response.
- `draft` returns details of the content item currently in the draft content store. There should always be a draft object.
- `live` returns details of the content item currently in the live content store. There will only be a live object for published and unpublished content.
- `unpublishing` is returned for the `live` content item if the live content item
is an unpublished document.
- anything with a nil value is omitted from the response

### Usage

We'd add a new gds-api-adapter to return the response.

In content tagger we'd use it like this:

[...]

PR #822, opened 2017-03-06
Lookup by base paths v2

MatMoore:
This request replaces `/lookup-by-base-path`. It allows draft and unpublished documents to be looked up in addition to live content.

More background in https://github.com/alphagov/publishing-api/issues/812

Trello: https://trello.com/c/MHBTiA65/544-publishing-api-support-for-lookups-of-draft-content-split-from-allow-users-to-tag-draft-content

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ You can run the pact verification tests on their own using:
 $ bundle exec rake pact:verify
 ```
 
-See [doc/pacts.md](doc/pacts.md) for more details about the pacts and the pact
+See [doc/pact_testing.md](doc/pact_testing.md) for more details about the pacts and the pact
 broker.
 
 ## Example API requests
--- a/app/controllers/v2/lookups_controller.rb
+++ b/app/controllers/v2/lookups_controller.rb
@@ -1,0 +1,17 @@
+module V2
+  class LookupsController < ApplicationController
+    def lookup_by_base_paths
+      if base_paths.length > 100
+        raise CommandError.new(code: 400, message: "base_paths must contain less than 100 items")
+      end
+
+      render json: Queries::LookupByBasePaths.call(base_paths)
+    end
+
+  private
+
+    def base_paths
+      params.fetch(:base_paths)
+    end
+  end
+end
--- a/app/controllers/v2/lookups_controller.rb
+++ b/app/controllers/v2/lookups_controller.rb
@@ -11,7 +11,11 @@ module V2
   private
 
     def base_paths
-      params.fetch(:base_paths)
+      permitted_params.require(:base_paths)
+    end
+
+    def permitted_params
+      params.permit(base_paths: [])
     end
   end
 end
--- a/app/queries/lookup_by_base_paths.rb
+++ b/app/queries/lookup_by_base_paths.rb
@@ -1,39 +1,43 @@
 module Queries
   module LookupByBasePaths
     def self.call(base_paths)
-      base_paths_and_content_ids = Edition.with_document
-        .left_outer_joins(:unpublishing)
-        .left_outer_joins(:access_limit)
-        .where(base_path: base_paths)
-        .where("access_limits.edition_id IS NULL")
-        .where("state != 'superseded'")
-        .order(:base_path)
-        .order(:state)
-        .order(user_facing_version: :desc)
+      # Ensure we have a key for every requested base path
+      response = {}
+      base_paths.each do |base_path|
+        response[base_path] = nil
+      end
+
+      grouped_rows = Edition
+        .with_document
+        .with_unpublishing
+        .with_access_limit
+        .where(base_path: base_paths, "access_limits.edition_id": nil)
+        .where.not(content_store: nil)
+        .order(:base_path, :content_store, user_facing_version: :desc)
         .pluck(
           %{
-            DISTINCT ON (editions.base_path, editions.state)
+            DISTINCT ON (editions.base_path, editions.content_store)
             editions.base_path,
             documents.content_id,
             documents.locale,
             editions.document_type,
-            CASE editions.state WHEN 'draft' THEN 'draft' ELSE 'live' END,
+            editions.content_store,
             unpublishings.type,
-            unpublishings.alternative_path
+            unpublishings.redirects
           }
-        )
+        ).group_by(&:first)
 
-      lookups = base_paths_and_content_ids.group_by(&:first)
-
-      lookups.each_with_object({}) do |group, result|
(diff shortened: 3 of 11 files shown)


MatMoore:
I've had to park this ticket for a while but the publishing api part is ready for review.

README.md
```diff
@@ -126,7 +126,7 @@ You can run the pact verification tests on their own using:
 $ bundle exec rake pact:verify
 ```
 
-See [doc/pacts.md](doc/pacts.md) for more details about the pacts and the pact
+See [doc/pact_testing.md](doc/pact_testing.md) for more details about the pacts and the pact
```
kevindew: 👍

app/controllers/v2/lookups_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+        raise CommandError.new(code: 400, message: "base_paths must contain less than 100 items")
+      end
+
+      render json: Queries::LookupByBasePaths.call(base_paths)
+    end
+
+  private
+
+    def base_paths
+      params.fetch(:base_paths)
```
kevindew: maybe we could use [Strong Parameters](https://github.com/rails/strong_parameters) here (disclaimer: it's a been a while since I used them so this could be a duff suggestion) - perhaps they'd allow us to require an array of base_paths which would save us worrying about a couple of edge cases of bad usage?

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+        .order(:state)
+        .order(user_facing_version: :desc)
+        .pluck(
+          %{
+            DISTINCT ON (editions.base_path, editions.state)
+            editions.base_path,
+            documents.content_id,
+            documents.locale,
+            editions.document_type,
+            CASE editions.state WHEN 'draft' THEN 'draft' ELSE 'live' END,
```
kevindew: you probably want to use editions.content_store here

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
+module Queries
+  module LookupByBasePaths
+    def self.call(base_paths)
+      base_paths_and_content_ids = Edition.with_document
+        .left_outer_joins(:unpublishing)
```
kevindew: we have a `with_unpublishing` scope here, whether that makes it feel more readable or not though `¯\_(ツ)_/¯`

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+        .left_outer_joins(:access_limit)
+        .where(base_path: base_paths)
+        .where("access_limits.edition_id IS NULL")
```
kevindew: I _think_ we can group these together like `where(base_path: base_paths, "access_limits.edition_id": nil)`

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+        .where("state != 'superseded'")
```
kevindew: touch more Railsy to be `.where.not(state: "superseded")` you might want that to actually be `.where.not(content_store: nil)` though

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+        .order(:base_path)
+        .order(:state)
+        .order(user_facing_version: :desc)
```
kevindew: can we group these?

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+        .pluck(
+          %{
+            DISTINCT ON (editions.base_path, editions.state)
+            editions.base_path,
+            documents.content_id,
+            documents.locale,
+            editions.document_type,
+            CASE editions.state WHEN 'draft' THEN 'draft' ELSE 'live' END,
+            unpublishings.type,
+            unpublishings.alternative_path
```
kevindew: If you're looking for where the redirect unpublishings have moved to this value is not the one to use anymore (unfortunately as it was more simple) we now have a hash of redirects which has each type of redirect route in there.

bench/lookup_by_base_path.rb
```diff
@@ -8,24 +8,44 @@
 
 abort "Refusing to run outside of development" unless Rails.env.development?
 
-benchmarks = Location.order("RANDOM()").limit(10).pluck(:base_path)
+benchmarks = Edition.order("RANDOM()").limit(10).pluck(:base_path)
```
kevindew: Thanks 👍

app/queries/lookup_by_base_paths.rb
```diff
@@ -0,0 +1,57 @@
...
+          }
+        )
+
+      lookups = base_paths_and_content_ids.group_by(&:first)
+
+      lookups.each_with_object({}) do |group, result|
```
kevindew: maybe here we would do better looping through the base_paths and adding results, which would allow us to return nil for base_paths that aren't in the Publishing API rather than the more opaque omission?

app/queries/lookup_by_base_paths.rb
MatMoore: :man_facepalming: that's a lot simpler 😄

app/queries/lookup_by_base_paths.rb
MatMoore: Ah ok. I can rework this to return the hash instead.

app/queries/lookup_by_base_paths.rb
MatMoore: I just went with omission here to match the semantics of v1. Do you think its useful to see the nils?

MatMoore:
@kevindew thanks for reviewing. I might be a while fixing the suggestions as I've had to pick up another ticket, but will make the changes once I have a free moment.

app/queries/lookup_by_base_paths.rb
kevindew: I think so, since you're providing it with a list of things to look up it feels more semantically helpful to return results that correspond to the list you provided

MatMoore:
@kevindew I've made the suggested changes.

MatMoore:
FYI I've just pushed an update to the doc/api.md documentation, since I decided not to add a pact test for this endpoint.

I'll wait for https://github.com/alphagov/gds-api-adapters/pull/678 to be reviewed before merging.

tijmenb:
We're picking up https://trello.com/c/6VPQcc7C next week, closing in the meantime.

PR #826, opened 2017-03-08
Fix flakiness caused by random content 

tijmenb:
This PR fixes the flakiness caused by the random content generation. 

To do we first add a proper error message that outputs the failures. With that I found two cases where the random content, which is valid according to the schema, is not valid according to this app. I've put in temporary fixes.

Also added an explicit test that tests 50 random items against the put content & publish endpoints. To verify there are no more issues I've run it 2000 times.

More details in the commits.

### DIFF
--- a/spec/integration/message_queue_publishing_spec.rb
+++ b/spec/integration/message_queue_publishing_spec.rb
@@ -1,35 +1,20 @@
 require "rails_helper"
-require "govuk_schemas"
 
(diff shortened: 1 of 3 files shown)


tijmenb:
For titles, it seems logical to enforce a minimum string length. I think that's what we originally intended by making them `required` - it's a bit counterintuitive coming from rails that an empty string satisfies a presence requirement.

PR #829, opened 2017-03-09
Represent downstream

kevindew:
This fixes some of the oddities with representing downstream. See commits for further details.


lib/tasks/represent_downstream.rake
```diff
@@ -1,11 +1,16 @@
 namespace :represent_downstream do
+  def represent_downstream(scope)
+    scope.distinct.in_batches.each do |batch|
+      content_ids = batch.pluck(:content_id)
+      Commands::V2::RepresentDownstream.new.call(content_ids)
+      sleep 5
```
thomasleese: 🤢

PR #830, opened 2017-03-10
Add migration to fix parted Publisher redirects

davidbasalla:
A while back we switched Publisher to use the Publishing API for unpublishing with redirects. At that point the Publishing API did not support unpublishing with prefix routes, so the routes were set incorrectly to 'exact'. This issue has since been fixed, however we added a CSV to router-data to fix these routes (to make absolutely sure, we just included ALL routes for unpublished parted formats, but the list of affected items is much smaller). This migration aims to fix the affected content items so that we can remove the CSV from router-data.

See https://github.gds/gds/router-data/pull/619

### DIFF
--- a/db/migrate/20170309165912_fix_redirects_for_publisher_multipart_content.rb
+++ b/db/migrate/20170309165912_fix_redirects_for_publisher_multipart_content.rb
@@ -1,0 +1,46 @@
+class FixRedirectsForPublisherMultipartContent < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  BASE_PATHS_FOR_PARTED_CONTENT_WITH_EXACT_REDIRECT_TYPE = [
+    "/accredited-electrical-and-electronic-equipment-producer-compliance-scheme-scotland",
+    "/air-conditioning-system-energy-assessor-accreditation-england-wales",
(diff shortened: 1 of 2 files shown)


kevindew:
@davidbasalla if you want to represent downstream in this you'll need to disable the ddl transaction (otherwise it'll start downstreaming before the content is committed to the db. Because of this you'll also want to be sure that your migration can handle being terminated in the middle if something goes wrong as you won't have the db rollback safety net. 

Any risk that items could change in the time from you putting this together and deploying? we often put safety guards on these types of migration by restricting them to a particular publishing_app and maybe a time stamp

davidbasalla:
Thanks @kevindew, have updated based on your comments, what do you think?

kevindew:
looks good 👍 have you been able to run it locally?  Can you include the schema.rb file please once it has run?

Once good can you try it on integration prior to us merging just to check?

davidbasalla:
Ah of course, no luck locally yet but will include schema change once I've got it working (need to wait for data replication) and try in integration!

davidbasalla:
@kevindew Have just tested this on Integration and it all appears to work! Sorry, took a while to get my dev vm set up with correct data so that I could push the schema changes...

PR #831, opened 2017-03-10
Allow exclusion of drafts from GetExpandedLinks

rubenarakelyan:
Currently, GetExpandedLinks returns all links, whether live or draft. This commit adds a `with_drafts` option, mirroring that already available on `LinkSetExpansion`, to allow the user to choose whether or not drafts should be returned.

Trello: https://trello.com/c/LmVHv7Jv/537-build-design-for-tagging-to-draft-taxons-in-whitehall

cc @kevindew

### DIFF
--- a/app/controllers/v2/link_sets_controller.rb
+++ b/app/controllers/v2/link_sets_controller.rb
@@ -5,7 +5,11 @@ module V2
     end
 
     def expanded_links
-      render json: Queries::GetExpandedLinks.call(content_id, params[:locale])
+      render json: Queries::GetExpandedLinks.call(
+        content_id,
+        params[:locale],
+        with_drafts: with_drafts?,
+      )
     end
 
     def patch_links
@@ -22,6 +26,12 @@ module V2
     end
 
   private
+
+    def with_drafts?
+      # Cast the `with_drafts` query param to a real boolean, and default to
+      # `true` to preserve existing behaviour
+      ActiveModel::Type::Boolean.new.cast(params.fetch(:with_drafts, true))
+    end
 
     def links_params
       payload.merge(content_id: content_id)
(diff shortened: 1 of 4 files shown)


app/controllers/v2/link_sets_controller.rb
```diff
@@ -23,6 +27,12 @@ def get_linked
 
   private
 
+    def with_drafts?
+      # Cast the `with_drafts` query param to a real boolean, and default to
+      # `true` to preserve existing behaviour
+      ActiveModel::Type::Boolean.new.cast(params.fetch(:with_drafts, true))
```
kevindew: Is the type case done as ?with_drafts=false is interpreted as true (being a string)?

kevindew:
looks good, any chance you can drop a change on doc/api.md too?

rubenarakelyan:
I've thought of a better way of doing this in https://github.com/alphagov/publishing-api/pull/833 so I'll close this PR.

app/controllers/v2/link_sets_controller.rb
MatMoore: "false" and other false-ish strings are interpreted as false. Happy to change it if there's a convention for this.
http://edgeapi.rubyonrails.org/classes/ActiveModel/Type/Boolean.html

MatMoore:
Tests are now passing, so ready for another review.

app/controllers/v2/link_sets_controller.rb
kevindew: Nah no convention, I did some googling after commenting before and looks like we need something like this here, bit of a bummer really since ActiveModel seems a totally unrelated dependency in this context but I don't know a better way.

PR #833, opened 2017-03-13
Mark draft expanded links

rubenarakelyan:
This commit marks expanded links to draft editions with `draft: true` so that they can be differentiated from links to live editions.

Trello: https://trello.com/c/LmVHv7Jv/537-build-design-for-tagging-to-draft-taxons-in-whitehall

Paired with @Davidslv 

cc @kevindew

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -483,6 +483,9 @@ were created.
 Retrieves the expanded link set for the given `content_id`. Returns arrays of
 details for each linked edition in groupings of `link_type`.
 
+Expanded links from both the live and draft content stores are returned.
+Links to draft editions are marked with `draft: true`.
+
 ### Path parameters
 
 - [`content_id`](model.md#content_id)
--- a/lib/link_expansion.rb
+++ b/lib/link_expansion.rb
@@ -49,7 +49,10 @@ private
     rules.expand_fields(edition, node.link_type).tap do |expanded|
       links = populate_links(node.links)
       auto_reverse = auto_reverse_link(node)
-      expanded.merge!(links: (auto_reverse || {}).merge(links))
+      expanded.merge!(
+        links: (auto_reverse || {}).merge(links),
+        draft: draft?(edition),
+      )
     end
   end
 
@@ -70,6 +73,10 @@ private
     # this requires we assess impact on the rendering applications first.
     %i(children parent related_statistical_data_sets).include?(link_type) ||
       edition.state != "unpublished"
+  end
+
+  def draft?(edition)
+    edition.content_store == "draft"
   end
 
   def rules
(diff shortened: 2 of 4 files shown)


kevindew:
Hey Reuben, just had a chance to have a nosey now.

This implementation does raise some concerns for me in comparison to the last implementation:

1. This would impact on the representation in the content store eg viewing [foreign-travel-advice/thailand](https://www.gov.uk/api/content/foreign-travel-advice/thailand) would have every link marked as `draft: false`. I don't really think this information should be exposed as it shouldn't be used in frontend apps. 
2. Things marked as `draft: true`would also include items that are published and have a draft available. This doesn't look like it'd work with your mock up where you have a selection of items that are just draft.
3. We'd still be stuck on how to access expanded links without drafts from expanded link sets endpoint (not sure if there is a need or not but would be nice to sort it)

I think the previous implementation helped on points 1 and 3 but I can understand it still left us short of 2 without a second request that was diffed against the first, so can understand that's not an ideal.

rubenarakelyan:
Indeed it would have been good to have a parameter but as you mention, this would leave us having to make two requests rather than one and finding the difference. We could, rather than a boolean flag, have a draft/live/both (default) parameter to give more flexibility, but we'd still have to make two requests.

1. Would it help to reduce the scope of this field to only be returned if it is asked for?
2. Here we could change from looking at `content_store` to looking at the last state in the state history - would that help?
3. We could still implement the previous solution but with the shortfall of not being able to get only draft links

What do you think?

kevindew:
Hey @rubenarakelyan sorry I didn't see the notification for this - got a ping from @Davidslv instead.

> 1. Would it help to reduce the scope of this field to only be returned if it is asked for?

Ideally we'd want to avoid this as the field existing introduces additional logic into link expansion, and gives them multiple ways they can be represented - thus you no longer have the canonical way. 

As it stands currently this endpoint gives you access to the same representation as you get in a content item. Once this differs from that then we've got a bit of a fork in the path with different things in different contexts - which feels bad from a system consistency perspective.

> 2. Here we could change from looking at content_store to looking at the last state in the state history - would that help?

As the endpoint stands you currently get the last state in the state history as it is, you just don't have it reported back.

> 3. We could still implement the previous solution but with the shortfall of not being able to get only draft links

Yup.

Probably best we have a catchup? do you want to ping me or @Davidslv when you have a chance so we can work out how to move this forward?

MatMoore:
After discussing with @kevindew offline, we decided to revisit https://github.com/alphagov/publishing-api/pull/831 instead of this one.

Including state information in these responses is difficult because
- it breaks consistency with content store representations, so it's hard to conceptualise what the response really is
- expanded links also includes edition links - so in the general case, whenever we traverse a link, the source can be draft, the link itself can be draft (if an edition link) and the target can be draft 🙀

PR #834, opened 2017-03-13
Solve the case of the missing change notes

thomasleese:
We were previously only adding change notes during a put content request when the `update_type` was set to `major`. However, in our API docs we state that `update_type` can be left out, meaning no change notes would be created.

This PR changes the behaviour by also creating change notes on publish, where the `update_type` is not an optional field.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -32,10 +32,15 @@ module Commands
         edition.publish
         remove_access_limit
         create_publish_action
+        create_change_note if payload[:update_type].present?
       end
 
       def create_publish_action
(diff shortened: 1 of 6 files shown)


app/commands/v2/publish.rb
```diff
@@ -32,12 +32,17 @@ def publish_edition
         edition.publish
         remove_access_limit
         create_publish_action
+        create_change_note if payload[:update_type] == "major"
```
kevindew: I imagine we only care if they provided a `payload[:update_type]`? I'm guessing we're mixing in logic from another domain caring if it is major?

PR #835, opened 2017-03-13
Remove Specialist Publisher parent links

kevindew:
Parent links were added to all specialist publisher documents in
https://github.com/alphagov/publishing-api/pull/821 . This has had an
unfortunate side effect - as parent links are represented as "children"
links, this caused finder content items such as /aaib-reports or
/dfid-research-outputs to have links to every child which was making the
content items huge in size.

The effects of this seem to be:
- 413 errors writing to content store
- slow responses from content store
- potentially (I hope this was the cause) a huge increase in memory
usage in content store.

This migration removes these parent links, which seems to be the best
immediate action we can take. As an intermediary message apps that plan
to consume this link can use the finder link which is the same as the
parent.

cc @fofr - this would undo the work we've done to put parent links into specialist documents, would you be able to use the finder link for the breadcrumbs in the mean time until we find a solution?

### DIFF
--- a/db/migrate/20170313184559_remove_specialist_document_finder_links.rb
+++ b/db/migrate/20170313184559_remove_specialist_document_finder_links.rb
@@ -1,0 +1,59 @@
+class RemoveSpecialistDocumentFinderLinks < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  DOCUMENT_TYPE_CONTENT_IDS = {
+    aaib_report: "b7574bba-969f-4c49-855a-ae1586258ff6",
+    asylum_support_decision: "581b4bba-e58f-4d07-8e0a-03c8c00165cc",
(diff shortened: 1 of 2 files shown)


fofr:
Do we intend to fix this, so that parents can still exist without huge content items? cc @tijmenb

danielroseman:
A first step towards fixing this might be to stop automatically expanding parents within children links, since that appears to be pointless.

kevindew:
@fofr the approach on our planning board for this is to [spike into whether we can define reverse links at a schema level](https://trello.com/c/A2yy9Lmo/443-spike-can-reverse-links-be-defined-as-part-of-a-schema) which would allow us to make `children` only definable when appropriate. A simpler thing would be to cut out the magic. 

@danielroseman do we know who, if anyone, uses the `children` type?

danielroseman:
Anything with reverse dependencies uses children, as far as I know: eg policies, guidance, and the travel-advice-index stuff I've been working on.

PR #836, opened 2017-03-14
Add instrumentation for cache hit/miss

tijmenb:
This subscribes to the active support notifications to increment a cache hit/miss counter. We can use this to see whether or not our linkables cache is doing anything useful, and how often the cache is invalidated.

Tested by running the tests in the VM and listening to UDP 8125 (statsd) in another terminal:

```
vagrant@development:~/govuk/publishing-api$ nc -u -l 8125
govuk.app.publishing-api.cache.cache_miss:1|c
```

Part of an effort to speed up content-tagger.

### DIFF
--- a/config/initializers/instrumentation.rb
+++ b/config/initializers/instrumentation.rb
@@ -1,0 +1,7 @@
+statsd = Statsd.new
+statsd.namespace = "govuk.app.publishing-api"
+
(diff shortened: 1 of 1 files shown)


tijmenb:
@kevindew I wasn't planning to remove this actually, would be good to keep tracking stuff like this. The more stats & timers the better!

PR #837, opened 2017-03-14
Include orphaned_content_ids in worker unique credentials

kevindew:
It was reported that our clearing of orphaned_content_ids was behaving
inconsistently. A reason this could be happening is that jobs to clear
them could be rejected as orphaned_content_ids do not appear in the
sidekiq uniq jobs criteria.

This adds them in so they will not be discarded.


kevindew:
@suzannehamilton this should fix your issue, do let me know if you still get it.

PR #842, opened 2017-03-16
Add auth_bypass_id

brenetic:
Content needs to be previewable by users without signon accounts in most
stages of a publishers workflow, not just during fact check.

This adds a migration that renames the fact_check_ids column in the AccessLimits
table to auth_bypass_ids.

Also renames occurrences of fact_check_ids to auth_bypass_ids.

https://trello.com/c/7oRBYwnn/690-rename-fact-check-id-to-auth-bypass-id

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'dalli'
 if ENV["API_DEV"]
   gem "gds-api-adapters", path: "../gds-api-adapters"
 else
-  gem "gds-api-adapters", "38.1.0"
+  gem 'gds-api-adapters', "~> 41.0"
 end
 
 gem "gds-sso", "13.0.0"
--- a/db/migrate/20170316160949_add_auth_bypass_id.rb
+++ b/db/migrate/20170316160949_add_auth_bypass_id.rb
@@ -1,0 +1,5 @@
+class AddAuthBypassId < ActiveRecord::Migration[5.0]
+  def change
+    add_column :access_limits, :auth_bypass_ids, :json, null: false, default: []
+  end
+end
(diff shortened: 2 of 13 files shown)


Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'dalli'
 if ENV["API_DEV"]
   gem "gds-api-adapters", path: "../gds-api-adapters"
 else
-  gem "gds-api-adapters", "38.1.0"
+  gem 'gds-api-adapters', "~> 41.0.0"
```
kevindew: probably want to loosen this to 41.0

db/migrate/20170316160949_add_auth_bypass_id.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthBypassId < ActiveRecord::Migration[5.0]
+  def change
+    add_column :access_limits, :auth_bypass_ids, :json, null: false, default: []
```
kevindew: is this migration quick? adding default ones can sometimes be quite painful.

Also will this cause any problems for existing data in fact_check_ids?

binaryberry: @kevindew when running the migration locally, it's quick: it takes 0.299s. 
For the fact_check_ids column, we'll only need to keep it for a few hours, until we've republished all draft content so they all have an auth_bypass_id.

kevindew: sounds good 👍

Gemfile
binaryberry: Thanks, amended!

PR #846, opened 2017-03-20
Add a migration to republish manuals documents

thomasleese:
This is required as a result of alphagov/manuals-publisher@23eb39f31f7516296288dad0706dddfe1b3746dd.

[Trello card](https://trello.com/c/dN27kLGQ/898-resolve-inconsistent-documents-2)


thomasleese:
Better to just run the rake task.

PR #847, opened 2017-03-20
Append EXECUTOR_NUMBER to test database name

kevindew:
This is to allow for concurrent builds in Jenkins without the database
name colliding.

There was an example failure of this but sadly that's out of our recent history now.

/cc @tijmenb

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,8 +45,9 @@ node {
       govuk.setEnvar("RAILS_ENV", "test")
       govuk.setEnvar("RCOV", "1")
       govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+      sh('bundle exec rake db:drop db:create')
       sh('bin/rails db:environment:set')
-      sh('bundle exec rake db:drop db:create db:schema:load')
+      sh('bundle exec rake db:schema:load')
     }
 
     stage("Lint") {
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,12 +7,14 @@ default: &default
 development:
   <<: *default
   database: publishing_api_development
-  url: <%= ENV["DATABASE_URL"]%>
+  url: <%= ENV["DATABASE_URL"] %>
 
 test:
   <<: *default
-  database: publishing_api_test
-  url: <%= ENV["DATABASE_URL"].try(:sub, /([-_]development)?$/, '_test')%>
+  # EXECUTOR_NUMBER is a env var added by Jenkins, this allows concurrent builds
+  # on the same box without the database colliding.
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -45,8 +45,9 @@ node {
       govuk.setEnvar("RAILS_ENV", "test")
       govuk.setEnvar("RCOV", "1")
       govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+      sh('bundle exec rake db:drop db:create')
       sh('bin/rails db:environment:set')
-      sh('bundle exec rake db:drop db:create db:schema:load')
```
thomasleese: Why was this change necessary?

config/database.yml
```diff
@@ -7,12 +7,14 @@ default: &default
...
+  url: <%= ENV["DATABASE_URL"] %>
 
 test:
   <<: *default
-  database: publishing_api_test
-  url: <%= ENV["DATABASE_URL"].try(:sub, /([-_]development)?$/, '_test')%>
+  # EXECUTOR_NUMBER is a env var added by Jenkins, this allows concurrent builds
+  # on the same box without the database colliding.
+  database: publishing_api_test<%= "_executor_#{ENV['EXECUTOR_NUMBER']}" if ENV["EXECUTOR_NUMBER"] %>
+  url: <%= ENV["DATABASE_URL"].try(:sub, /([-_]development)?$/, "_test") %>
```
thomasleese: I see this was like this before, but I wonder why we're using `try` here when if we don't have that environment variable set we might the error to be more obvious.

cbaines:
Is this useful now that https://github.com/alphagov/govuk-puppet/pull/5641 has been merged?

kevindew:
@cbaines thanks - didn't know @tijmenb had solved the problem too. 

We'll still need something in this project since we don't (and as I understand can't) use that task in Publishing API - would be good to be consistent though.

config/database.yml
kevindew: the url parameter here will override the values set in database, username, password (and probably a few others). If it is nil no overriding will occur, which is what will normally be the case when you're running these tests.

Whereas if you want to run these tests in the docker environment for instance you'll need a url set and will want to run the substitution.

tl;dr - the env var is normally nil

Jenkinsfile
kevindew: the `db:environment:set` command requires the database to exist, we've never run into this problem as presumably the database existed before this was added.

config/database.yml
thomasleese: A-ha yes, that makes sense!

cbaines:
> @cbaines thanks - didn't know @tijmenb had solved the problem too.
> 
> We'll still need something in this project since we don't (and as I understand can't) use that task in Publishing API - would be good to be consistent though.

Ah, yeah, ok.

PR #851, opened 2017-03-21
Support redirects including fragments

cbaines:
https://trello.com/c/RfRbIDB9/694-fix-unpublish-with-fragment-redirect-in-publisher

### DIFF
--- a/app/validators/absolute_path_validator.rb
+++ b/app/validators/absolute_path_validator.rb
@@ -11,7 +11,7 @@ private
     return false unless path.present? && path.starts_with?("/")
 
     uri = URI.parse(path)
-    uri.path == path && path !~ %r{//} && path !~ %r{./\z}
+    "#{uri.path}#{uri.fragment}" == path && path !~ %r{//} && path !~ %r{./\z}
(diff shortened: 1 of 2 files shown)


kevindew:
Ahh this is interesting. I can understand your use case but I think this is going to need a deeper fix than this.

We use this validator in a few places where the schemas define `absolute_path` in those instances we don't want to accept a fragment. 

For redirects we'd want to support what the schemas describe as [`absolute_fullpath`](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/case_study/publisher_v2/schema.json#L341-L345) so I think we'd probably want a new validator for that and to test redirect destinations against that instead?

cbaines:
> We use this validator in a few places where the schemas define absolute_path in those instances we don't want to accept a fragment.
> 
> For redirects we'd want to support what the schemas describe as absolute_fullpath so I think we'd probably want a new validator for that and to test redirect destinations against that instead?

Ok, I'll hopefully have time to look again at this today.

cbaines:
I'm replacing this with https://github.com/alphagov/publishing-api/pull/859

PR #852, opened 2017-03-21
Send an Airbrake error when inconsistent documents are found

thomasleese:
This will allow the task to run automatically and send us an alert when inconsistent documents are found.

### DIFF
--- a/lib/tasks/check_content_consistency.rake
+++ b/lib/tasks/check_content_consistency.rake
@@ -4,8 +4,14 @@ namespace :check_content_consistency do
     errors = checker.call
 
     if errors.any?
-      puts "#{content_id} #{locale} 😱"
-      puts errors
+      Airbrake.notify(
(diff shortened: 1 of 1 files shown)


lib/tasks/check_content_consistency.rake
```diff
@@ -4,8 +4,14 @@ namespace :check_content_consistency do
     errors = checker.call
 
     if errors.any?
-      puts "#{content_id} #{locale} 😱"
-      puts errors
+      Airbrake.notify(
+        "Found an inconsistent document: #{content_id} #{locale} 😱",
```
tijmenb: Not a 100% sure you can pass it a string, I remember vaguely that it has to be an exception object. Could be wrong though.

In any case, if you use the same message/string for all errors they'll be rolled up into a single exception. Using a message per-error might clutter errbit and cause an overload of emails, which you may not want.

PR #853, opened 2017-03-21
Don't allow random redirect placeholder schemas

kevindew:
These don't really work as placeholder isn't suitable for a redirect
document type and cause intermittent test failures.

### DIFF
--- a/spec/support/random_content_helpers.rb
+++ b/spec/support/random_content_helpers.rb
@@ -3,29 +3,20 @@ require "govuk_schemas"
 module RandomContentHelpers
   def generate_random_edition(base_path)
     random = GovukSchemas::RandomExample.for_schema(publisher_schema: "placeholder")
-
-    item = random.merge_and_validate(
+    payload = random.payload
+    random.merge_and_validate(
       base_path: base_path,
-
-      # TODOs:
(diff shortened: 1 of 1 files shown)


tijmenb:
I'm wondering if we can fix a lot of the weirdness if the publishing-api validates stuff against `schema_name` instead of `document_type`. The document type should be allowed to be anything and shouldn't have an effect.

kevindew:
@tijmenb Yeah I think we have a document_type == redirect conditional that'd probably be fine being based off schema_name. We'd have to do a bit of noseying around to see what implications are.

spec/support/random_content_helpers.rb
```diff
@@ -3,29 +3,20 @@
...
   def generate_random_edition(base_path)
     random = GovukSchemas::RandomExample.for_schema(publisher_schema: "placeholder")
-
-    item = random.merge_and_validate(
+    payload = random.payload
+    random.merge_and_validate(
       base_path: base_path,
-
-      # TODOs:
+      # TODO: don't allow placeholder redirect, it doesn't make sense
```
thomasleese: Can we get rid of the TODO comment now?

kevindew:
closing in favour of: https://github.com/alphagov/publishing-api/pull/858

PR #805, opened 2017-02-28
Add document supertypes to frontend & message queue

tijmenb:
The "document supertype" is a grouping for document types. It will allow us to segment pages by more types, like "Publication" or "Announcement" (used in Whitehall).

alphagov/govuk_document_types#1

https://trello.com/c/07lAcDtC

PR #855, opened 2017-03-21
Run content store pact tests as part of build

kevindew:
We publish pact tests for content store but don't actually test against it, this adds in the testing against content store to help us ensure we're not breaking our pacts with an individual build.

This was prompted off the back of https://github.com/alphagov/publishing-api/pull/805 which broke in content store.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,7 @@
 
 REPOSITORY = "publishing-api"
 DEFAULT_SCHEMA_BRANCH = "deployed-to-production"
+DEFAULT_CONTENT_STORE_BRANCH = "master"
 
 node {
   def govuk = load("/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy")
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -52,7 +59,7 @@ node {
       govuk.rubyLinter("app config Gemfile lib spec")
     }
 
-    // Prevent a project"s tests from running in parallel on the same node
+    // Prevent a project's tests from running in parallel on the same node
```
thomasleese: hehe

Jenkinsfile
```diff
@@ -90,7 +97,35 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == "master") {
+    stage("Checkout content store") {
+      sh('rm -rf tmp/content-store')
```
thomasleese: I think we should stick to `"` here, probably just overlooked.

kevindew:
@thomasleese comments addressed if you'd like to re-review

PR #859, opened 2017-03-22
Use one redirect destination validation approach

cbaines:
https://trello.com/c/RfRbIDB9/694-fix-unpublish-with-fragment-redirect-in-publisher

### DIFF
--- a/spec/commands/v2/unpublish_spec.rb
+++ b/spec/commands/v2/unpublish_spec.rb
@@ -112,6 +112,28 @@ RSpec.describe Commands::V2::Unpublish do
             a_hash_including(destination: "/something-amazing")
           ])
         end
+
+        context "including a destination with a fragment" do
+          let(:redirects) do
+            [
+              {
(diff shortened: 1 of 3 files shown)


cbaines:
> Aces! that looked a lot less painful than I thought it'd be.

Yeah, the code just turned out to be a bit weird...

spec/commands/v2/unpublish_spec.rb
```diff
@@ -112,6 +112,28 @@
...
+                destination: "/something-amazing#foo",
+              }
+            ]
+          end
+
+          it "should populate the redirects hash" do
+            described_class.call(payload)
+
+            unpublishing = Unpublishing.first
+            puts payload
```
kevindew: @cbaines doh just noticed this, could you kill it?

cbaines: Oh dear, sorry https://github.com/alphagov/publishing-api/pull/860

PR #862, opened 2017-03-22
Expose an endpoint for document actions

tijmenb:
This adds an experimental endpoint to expose the "actions" (update, publish, etc) of a document. We'll use it in content-tagger to show a rudimentary content history.

Inspired by @commuterjoy's excellent talk yesterday about sustainable development, this feature will stop working on April 22nd. This will make sure that we'll either properly build this endpoint (with documentation) or remove the code.

https://trello.com/c/HQ28fhrF

### DIFF
--- a/app/controllers/v2/actions_controller.rb
+++ b/app/controllers/v2/actions_controller.rb
@@ -1,6 +1,8 @@
 module V2
   class ActionsController < ApplicationController
     def index
+      TimedFeature.check!(owner: "Tijmen", expires: "2017-04-22")
+
       actions = Action
         .where(content_id: params[:content_id])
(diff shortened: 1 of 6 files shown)


app/controllers/v2/actions_controller.rb
```diff
@@ -1,5 +1,15 @@
 module V2
   class ActionsController < ApplicationController
+    def index
+      TimedFeature.check!(owner: "Tijmen", expires: "2017-04-22")
+
+      actions = Action
+        .where(content_id: params[:content_id])
+        .as_json(only: %i[action user_uid created_at])
```
kevindew: Could we add a limit here? pagination would be better of course, but understand it's pretty bare bones

PR #864, opened 2017-03-23
Rabbitmq issues

kevindew:
Hopefully a fix for the `undefined method `next_channel_id' for nil:NilClass` error and fixes configuration for e2e tests


thomasleese:
Will be nice to get rid of that error! 😄

PR #865, opened 2017-03-23
Restructure how consistency checker works

thomasleese:
It now takes a dump of the content from the `content-store` to compare the data. This means there is a dramatic decrease in how long it takes to complete.


thomasleese:
@kevindew This is ready for review. 😄

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
 class ContentConsistencyChecker
```
kevindew: might be that this class belongs in the data_hygiene module?

lib/tasks/check_content_consistency.rake
```diff
@@ -1,32 +1,31 @@
...
   end
+end
 
-  desc "Check documents for consistency with the router-api and content-store"
-  task :one, [:content_id, :locale] => [:environment] do |_, args|
-    content_id = args[:content_id]
-    locale = args[:locale] || "en"
-    check_content(content_id, locale, false)
-  end
+desc "Check all the documents for consistency with the router-api and content-store"
```
kevindew: is this actually just content-store? also this implies that it does the work independently whereas it's actually given the data

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
+  end
+
+  def hash_field(object)
+    Digest::SHA1.hexdigest(
+      JSON.generate(object)
+    )
+  end
+
+  def draft?
+    content_store == "draft"
```
kevindew: defensive but you might want to run `.to_s` on this incase this gets called with a symbol.

lib/tasks/check_content_consistency.rake
```diff
@@ -1,32 +1,31 @@
...
-    errors = checker.call
+require "content_consistency_checker"
 
-    if errors.any?
-      puts "#{content_id} #{locale} 😱"
-      puts errors
-    end
+def report_errors(errors, content_store)
+  Airbrake.notify(
+    "Inconsistent #{content_store} documents",
```
kevindew: maybe name this more explicitly - "Documents inconsistent with the #{content_store} content store" ?

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
+      unless content_item.gone?
+        errors[path] << "Content is not gone in the content store."
       end
     else
-      fields = [:rendering_app, :publishing_app, :schema_name, :document_type]
+      fields = [:content_id, :locale, :rendering_app, :publishing_app, :schema_name, :document_type]
       fields.each do |field|
         edition_value = edition.send(field)
-        content_item_value = content_item[field.to_s]
+        content_item_value = content_item.send(field)
```
kevindew: I think `public_send` would be better as we only want to use public methods, and when you see `send` in ruby it tends to trigger a concern that someones trying to use a protected/private method

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
+      check_edition(edition)
     end
+  end
+
+  def check_content
+    unchecked_content.each do |path|
+      content_item = content_dump.fetch(path)
+      next if content_item.content_id.nil?
+      next if content_item.gone?
+      next unless content_item.schema_name.present?
```
kevindew: maybe put these three conditions into a method?

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
-    check_edition("Draft", draft, draft_content_store) if draft
-    check_edition("Live", live, live_content_store) if live
+      edition = Edition.find_by(
+        content_store: content_store,
+        base_path: content_item.base_path
+      )
+      next if edition
 
-    errors
+      errors[content_item.base_path] << "No edition available."
```
kevindew: can probably just have the above condition on this line and avoid the next?

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
+    @checked_content = Set.new
+    @errors = Hash.new { |hash, key| hash[key] = [] }
   end
 
-  def call
-    unless document
-      errors << "Document (#{content_id}, #{locale}) could not be found."
-      return errors
+  def check_editions
+    editions_to_check.find_each do |edition|
```
kevindew: maybe this'll fit on a single line?

lib/content_consistency_checker.rb
```diff
@@ -3,92 +3,124 @@
...
+
+  def get_content_item(path)
+    begin
+      path = path.to_sym
+      content_item = content_dump.fetch(path)
+      checked_content.add?(path)
+      content_item
+    rescue KeyError
+      nil
+    end
```
kevindew: I don't think you need this in a begin and end, you can just have the rescue in the method

thomasleese:
@kevindew Ready for re-review. 😄

PR #866, opened 2017-03-23
Delete fact check attribute

binaryberry:
Removes fact-check-ids from AccessLimits table. They are no longer necessary as they have been replaced by auth_bypass_ids.

[DO NOT MERGE] Do not merge until https://github.com/alphagov/publishing-api/pull/842 and its related PRs have been deployed.

https://trello.com/c/7oRBYwnn/690-rename-fact-check-id-to-auth-bypass-id


db/migrate/20170322172057_remove_fact_check_ids_from_access_limits.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveFactCheckIdsFromAccessLimits < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :access_limits, :fact_check_ids, :json, null: false, default: []
```
thomasleese: Do we not need to rename this field so we keep the existing data?

PR #873, opened 2017-04-10
Refactor and change redirect validation

cbaines:
https://trello.com/c/4vQuPUn3/15-3-update-publishing-api

This is a prerequisite for moving some of the redirects currently managed through the router-data tool in to the Publishing API

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -130,35 +130,43 @@ private
 
       return unless errors.empty?
 
-      return if destination.starts_with?("/")
-
       begin
-        destination_uri = URI.parse(destination)
+        uri = URI.parse(destination)
       rescue URI::InvalidURIError
         errors << "destination is an invalid url"
         return
       end
 
-      unless url_constructed_as_expected?(destination, destination_uri)
-        errors << "destination url is not constructed as expected"
-        return
-      end
-
-      errors << "external redirects must use https" unless uri.scheme == "https"
-
-      unless valid_govuk_url?(destination_uri)
-        errors << "is not a valid redirect destination"
-      end
+      validate_external_redirect(destination, uri) if external?(destination)
     end
 
   private
 
-    def valid_govuk_url?(uri)
-      host = uri.host
-      if host =~ /\A.+\.gov\.uk\z/i
-        label = host.split(".").first
-        label.present? && valid_subdomain?(label)
+    def external?(destination)
+      !destination.starts_with?("/")
+    end
+
+    def validate_external_redirect(destination, uri)
+      unless url_constructed_as_expected?(destination, uri)
+        errors << "destination url is not constructed as expected"
+        return
       end
+
+      if uri.host.nil?
+        errors << "missing host for external redirect"
+        return
+      end
+
+      errors << "external redirects only accepted within the gov.uk domain" unless
+        uri.host.end_with?(".gov.uk")
+
+      errors << "internal redirect should not be specified with full url" if
+        %w(gov.uk www.gov.uk).include? uri.host
+
+      errors << "external redirects must use https" unless uri.scheme == "https"
+
+      errors << "subdomain invalid" unless
+        uri.host.split(".").all? { |subdomain| valid_subdomain?(subdomain) }
     end
 
     def valid_subdomain?(label)
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -13,7 +13,7 @@ class RoutesAndRedirectsValidator < ActiveModel::Validator
 
     redirects.each do |redirect|
       RouteValidator.new.validate(record, :redirects, redirect, base_path)
-      RedirectValidator.new.validate(record, redirect)
+      RedirectValidator.new(record, redirect).validate
     end
 
     must_have_unique_paths(record, routes, redirects)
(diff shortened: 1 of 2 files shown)


app/validators/routes_and_redirects_validator.rb
```diff
@@ -113,76 +113,78 @@ def additional_keys(route, attribute)
...
+
+    def validate
       path = redirect[:path]
       destination = redirect[:destination]
-      type = redirect[:type]
 
-      unless path.present?
-        record.errors[:redirects] << "path must be present"
-      end
+      errors << "path must be present" unless path.present?
```
pmanrubia: I would dry-it up into:

```ruby
      errors << "path must be present" unless path.present?
      errors << "destination must be present" unless destination.present?
      errors << "path cannot equal the destination" if path == destination

      return unless errors.empty?
```

It reads better and removes duplication

app/validators/routes_and_redirects_validator.rb
```diff
@@ -113,76 +113,78 @@ def additional_keys(route, attribute)
...
-      unless valid_exact_redirect_target?(destination)
-        record.errors[:redirects] << "is not a valid redirect destination"
+      begin
+        uri = URI.parse(destination)
+      rescue URI::InvalidURIError
+        errors << "destination is an invalid url"
+        return
       end
+
+      return if destination.starts_with?("/")
```
pmanrubia: it is better to avoid a `return` in the middle of a method:

```ruby
    validate_external_redirect(destination, uri) if external_redirect?
  end

private

  def external_redirect?(destination)
    !destination.starts_with?("/")
  end

```

By extracting the method it also reads a bit better

app/validators/routes_and_redirects_validator.rb
```diff
@@ -113,76 +113,78 @@ def additional_keys(route, attribute)
...
+      validate_external_redirect(destination, uri)
     end
 
   private
 
-    def acceptable_destination?(target)
-      target.starts_with?("/") || valid_govuk_campaign_url?(target)
-    end
+    def validate_external_redirect(destination, uri)
+      unless url_constructed_as_expected?(destination, uri)
```
pmanrubia: `url_constructed_as_expected` does not sound well to me. I would explore other names? `valid_uri?`, `accepted_uri?`

I would also get rid of `url` and use `uri` instead.

It also feels to me that it would be better to perform the validation listed above in this method as well. 
As side effect we could get rid of one parameter in the method signature.

```ruby
  begin
    uri = URI.parse(destination)
  rescue URI::InvalidURIError
    errors << "destination is an invalid url"
  end
```

app/validators/routes_and_redirects_validator.rb
```diff
@@ -123,7 +123,6 @@ def initialize(record, redirect)
     def validate
       path = redirect[:path]
       destination = redirect[:destination]
-      type = redirect[:type]
```
pmanrubia: If this is something that was missing in previous PRs, I would add this commit to the first position explaining that we are fixing a previous issue and that all tests are passing in the commit notes.
If this is not the case, I would squash into proper commit.

spec/support/routes_and_redirects_validator.rb
```diff
@@ -179,11 +179,11 @@
       end
 
       it "is valid if the url is a wellformed gov.uk campaign external url" do
-       %w(
-         https://new-vat-rates.campaign.gov.uk/
-         https://new-vat-rates.campaign.gov.uk/path/to/your/new/vat-rates
-         https://new-vat-rates.campaign.gov.uk/path/to/your/new/vat-rates?q=123&&a=23344
-       ).each do |destination|
+        %w(
```
pmanrubia: I would squash this indentation issue with previous commit.

app/validators/routes_and_redirects_validator.rb
```diff
@@ -148,16 +148,16 @@ def validate
 
       errors << "external redirects must use https" unless uri.scheme == "https"
 
-      unless valid_govuk_campaign_url?(destination_uri)
```
pmanrubia: What about adding a test sample that covers a subdomain that is not a campaign?

app/validators/routes_and_redirects_validator.rb
cbaines: It's not something that has become unused through other changes made here, so I've moved it to be the earliest commit in the branch.

app/validators/routes_and_redirects_validator.rb
cbaines: Good point, done in 878f5ea8ae8a0d866a71c63fb7ee12b40a12c3f2

app/validators/routes_and_redirects_validator.rb
cbaines: I've done this now.

app/validators/routes_and_redirects_validator.rb
cbaines: I went with `external?` so let me know if that works for you?

app/validators/routes_and_redirects_validator.rb
cbaines: I've changed things in 293edb5

app/validators/routes_and_redirects_validator.rb
pmanrubia: 👍

pmanrubia:
Great job @cbaines

PR #874, opened 2017-04-10
Support passing segments_mode through to the router/router-api

cbaines:


kevindew:
Would we be able to include redirect_type too? Seems a good opportunity to get feature parity with the GDS API Adapter https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/router.rb#L33-L44

cbaines:
> Would we be able to include redirect_type too? Seems a good opportunity to get feature parity with the GDS API Adapter https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/router.rb#L33-L44

Good idea, I've added that too now.

PR #875, opened 2017-04-12
Fix link in docs to dependency resolution library

andrewgarner:


tijmenb:
This PR and https://github.com/alphagov/publishing-api/pull/876 is failing because there's a slash in the branch name, which trips up the pact branch testing. @kevindew will see about URL-encoding this in the Jenkinsfile.

PR #876, opened 2017-04-12
Fix heading hierarchy in pact testing docs

andrewgarner:
Use primary and secondary headings.


tijmenb:
This PR and https://github.com/alphagov/publishing-api/pull/875 is failing because there's a slash in the branch name, which trips up the pact branch testing. @kevindew will see about URL-encoding this in the Jenkinsfile.

PR #880, opened 2017-04-19
Add a migration to repair missing attachment URLs

thomasleese:
We've identified some attachments with `nil` URLs which this migration should fix.


kevindew:
Looks cool, have you given it a spin on integration?

thomasleese:
I've run the migration on integration, still need to check it does what we want in the content store, but I opened the PR for review early.

PR #882, opened 2017-04-20
Change Edition description to string

emmabeynon:
We only use Edition `description` as a string field, so we convert it from JSON to string.

This is now covered in separate pull requests in order to handle deployments better:
- https://github.com/alphagov/publishing-api/pull/885
- https://github.com/alphagov/publishing-api/pull/886
- https://github.com/alphagov/publishing-api/pull/887


app/models/edition.rb
```diff
@@ -228,31 +227,6 @@ def web_url
...
-      attributes.merge("description" => description.fetch("value"))
-    else
-      attributes
-    end
-  end
-
-  def self.column_defaults
-    super.merge("description" => nil)
-  end
-
```
thomasleese: 🎉

emmabeynon:
> Do you know how long the migrations take? Do we need to first make a deploy which first dual-populates the description2 field, and then another deploy to remove the old field?

They only took a couple of minutes locally, but I think it would be sensible to do two separate deploys.

app/presenters/queries/content_item_presenter.rb
```diff
@@ -106,7 +106,7 @@ def select_fields(scope)
           when :lock_version
             "documents.stale_lock_version AS lock_version"
           when :description
-            "description->>'value' AS description"
+            "description"
```
kevindew: I reckon we could kill both these lines as this would get caught in the else

PR #883, opened 2017-04-20
Another bunny battle

kevindew:
Wrap Bunny connection in a mutex

Starting a connection on Bunny is not threadsafe - hopefully this
problem is why we have seen the "NoMethodError: undefined method
'next_channel_id' for nil:NilClass"

see: https://github.com/ruby-amqp/bunny/issues/498 for further information.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem "gds-sso", "13.0.0"
 gem "govuk_schemas", require: false
 gem "govuk_document_types", "~> 0.1.4"
 
-gem 'bunny', '2.5.1'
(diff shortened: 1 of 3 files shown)


michaelklishin:
As was explained in https://github.com/ruby-amqp/bunny/issues/498, `NoMethodError: undefined method 'next_channel_id' for nil:NilClass` has nothing to do with concurrent access.

As of 2.7.1 (https://github.com/ruby-amqp/bunny/commit/1c117c58abd7150dfaf69497937e37b14a7e61fe), Bunny will throw a clearer exception.

PR #885, opened 2017-04-20
2) Copy description values to description2

emmabeynon:

This task takes c.31 minutes on integration.

### DIFF
--- a/db/migrate/20170420160820_copy_description_values_to_description_string.rb
+++ b/db/migrate/20170420160820_copy_description_values_to_description_string.rb
@@ -1,0 +1,11 @@
+class CopyDescriptionValuesToDescriptionString < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  def up
+    Edition.in_batches do |relation|
+      execute("UPDATE editions
+               SET description_string=description ->> 'value'
+               WHERE id IN (#{relation.pluck(:id).join(', ')})")
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170420154120) do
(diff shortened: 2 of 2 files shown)


app/models/edition.rb
```diff
@@ -232,6 +232,7 @@ def web_url
   # be removed when the column is fixed.
   def description=(value)
     super("value" => value)
+    self.description2 = value
```
kevindew: I think this should be in the first PR (might also be useful to number them)

db/migrate/20170420160820_copy_description_values_to_description_string.rb
```diff
@@ -0,0 +1,12 @@
+class CopyDescriptionValuesToDescriptionString < ActiveRecord::Migration[5.0]
+  def up
+    Edition.all.each do |edition|
```
kevindew: you'll probably find this is a bit quicker if you use `find_each` as that'll pull out batches of 1000.

db/migrate/20170420160820_copy_description_values_to_description_string.rb
```diff
@@ -0,0 +1,12 @@
+class CopyDescriptionValuesToDescriptionString < ActiveRecord::Migration[5.0]
```
kevindew: I think we'll want to disable ddl transaction on this so that we don't lock up all editions while this runs

db/migrate/20170420160820_copy_description_values_to_description_string.rb
```diff
@@ -0,0 +1,12 @@
...
+  def up
+    Edition.all.each do |edition|
+      string_description = edition.description.to_s
+      edition.update_attribute(:description_string, string_description)
+    end
+  end
+
+  def down
+    Edition.update_all(description_string: nil)
```
kevindew: I don't think we'll actually need a down method here, particularly as it would affect more than just this migration (all the values set from https://github.com/alphagov/publishing-api/pull/886)

kevindew:
woop! 30 minutes is a lot better 👍

PR #886, opened 2017-04-20
1) Add description_string to Edition temporarily

emmabeynon:
In order to prepare for converting `description`'s type to string, we
temporarily create a string attribute called `description_string`.  We also ensure
that any new values saved to `description` are automatically saved to
`description_string`.

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -8,6 +8,7 @@ module Presenters
       :document_id,
       :id,
       :last_edited_at,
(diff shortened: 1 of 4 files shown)


app/presenters/edition_presenter.rb
```diff
@@ -8,6 +8,7 @@ class EditionPresenter
       :document_id,
       :id,
       :last_edited_at,
+      :description_string,
```
kevindew: 👍

PR #887, opened 2017-04-20
[DO NOT MERGE] 3) Rename edition description and description string

emmabeynon:


app/models/edition.rb
```diff
@@ -57,7 +57,6 @@ class Edition < ApplicationRecord
     in: %w(alpha beta live),
     message: 'must be either alpha, beta, or live'
   }
-  validates :description, well_formed_content_types: { must_include: "text/html" }
```
kevindew: yay not even sure this did anything anyway

app/models/edition.rb
```diff
@@ -227,32 +227,6 @@ def web_url
     Plek.current.website_root + base_path
   end
 
-  # FIXME These are required for the special 'description' column but should
-  # be removed when the column is fixed.
-  def description=(value)
```
kevindew: If we want to allow full rollback ability we'd probably want to keep this method and have it up date description_json, which would probably for the best as it would allow us to back out of this without lost data (hopefully anyway)

kevindew:
All looks good 👍  - nice commit messages too! I think once the comments are addressed we should give it a spin on integration

emmabeynon:
This is working as expected on integration 🎇

app/presenters/edition_presenter.rb
```diff
@@ -49,6 +50,13 @@ def rendered_details
       { details: details_presenter.details }
     end
 
+    # We're keeping this until such time as we decide to remove description_json
+    # entirely, so that we don't lose the data in case we decide to revert.
+    def description=(value)
+      super(value)
+      self.description_json = { "value" => value }
+    end
+
```
kevindew: I think this has ended up in the wrong class? should be in the edition model

emmabeynon:
Whoops, good spot, will fix now!

Emma Beynon
Junior Technologist, GOV.UK <http://gov.uk/> Custom Formats
Government Digital Service

app/presenters/edition_presenter.rb
emmabeynon: Whoops, good spot, will fix now!

PR #888, opened 2017-04-21
Add documentation for link filtering

cbaines:


cbaines:
> Thanks :+1: I didn't even realise we did this - that's a strange format for a query string.

Neither did I! I was looking again at getting organisation filtering in to Maslow, which was previously removed as I didn't think it was possible, but it turns out it always was :)

PR #889, opened 2017-04-21
Distributed lock when writing to content store

kevindew:
https://trello.com/c/T9OE6SSJ/916-distributed-lock-to-resolve-race-conditions-updating-content-store

This uses the [with_advisory_lock](https://github.com/mceachen/with_advisory_lock) gem to use PostgreSQL's advisory lock feature to perform a distributed lock when writing to the content store.

See individual commits for further details.

### DIFF
--- a/app/clients/content_store_writer.rb
+++ b/app/clients/content_store_writer.rb
@@ -6,11 +6,15 @@ require "active_support/core_ext/hash/keys"
 # to the content store.
 class ContentStoreWriter < GdsApi::ContentStore
   def put_content_item(base_path:, content_item:)
-    put_json(content_item_url(base_path), content_item)
+    with_lock(base_path, :content_item) do
+      put_json(content_item_url(base_path), content_item)
+    end
   end
 
   def put_publish_intent(base_path:, publish_intent:)
-    put_json(publish_intent_url(base_path), publish_intent)
+    with_lock(base_path, :publish_intent) do
+      put_json(publish_intent_url(base_path), publish_intent)
+    end
   end
 
   def get_publish_intent(base_path)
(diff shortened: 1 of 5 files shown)


cbaines:
How do you see this fitting in with the payload versions (this overlaps with the comment on the Trello ticket)?

I see the payload versions as preventing older payloads from overwriting newer payloads, causing the Publishing API to retry until the Content Store is correctly updated. Whereas using a lock means that there should be no issues with requests overlapping, so the use of payload versions is now redundant?

kevindew:
Sorry @cbaines - I missed your comment on the Trello board.

I see this and the UpdateLock (`payload_versions`) as being two separate things. This lock ensures that we aren't writing to the content store at the same path at the same time, whereas the payload_version is used to ensure that an older item doesn't replace a newer one.

I'd guess the following situation could occur:
- two items with content_id: a with different payload versions start being processed 
- newer payload version gets a lock first
- older item waits until that lock is free, and then requests content store

It would be good though to find a way to kill off the payload version concern as it isn't ideal to have the concerns split across the apps. I'll include this on my wishlist for the downstream rewrite.

cbaines:
> Sorry @cbaines - I missed your comment on the Trello board.
> 
> I see this and the UpdateLock (payload_versions) as being two separate things. This lock ensures that we aren't writing to the content store at the same path at the same time, whereas the payload_version is used to ensure that an older item doesn't replace a newer one.
> 
> I'd guess the following situation could occur:
> 
>     two items with content_id: a with different payload versions start being processed
>     newer payload version gets a lock first
>     older item waits until that lock is free, and then requests content store

Ah yes, I missed that the problems are indeed separate, and can overlap.

app/clients/content_store_writer.rb
```diff
@@ -6,28 +6,46 @@
...
   def publish_intent_url(base_path)
     "#{endpoint}/publish-intent#{base_path}"
   end
+
+  def lock_name_prefix
+    options[:lock_name_prefix] || ""
+  end
+
+  def with_lock(*args)
+    lock_name = ([lock_name_prefix] + args).map(&:to_s).join("_")
```
cbaines: This could be simplified by using the `endpoint` as the lock_name_prefix?

kevindew: That's a much better idea - will implement

PR #891, opened 2017-04-27
[DO NOT MERGE] 4) Remove description json

emmabeynon:


app/presenters/edition_presenter.rb
```diff
@@ -5,8 +5,6 @@ class EditionPresenter
       :api_url,
       :content_store,
       :created_at,
-      :description_json,
-      :description_string,
```
thomasleese: Ah, I guess the removal of `:description_string` was missed in an earlier commit?

PR #893, opened 2017-04-27
Throw an error if the first publish does not use a 'major' update_type

leenagupte:
Trello card: https://trello.com/c/3S4m5wAJ

Only accept an update type of `major` if a document is being published for the first time

Related to: PR https://github.com/alphagov/gds-api-adapters/pull/698

### DIFF
--- a/spec/commands/v2/publish_spec.rb
+++ b/spec/commands/v2/publish_spec.rb
@@ -323,6 +323,12 @@ RSpec.describe Commands::V2::Publish do
           end
 
           it "uses the stored timestamp for major or minor" do
+            FactoryGirl.create(:unpublished_edition,
+              document: draft_item.document,
+              public_updated_at: public_updated_at,
+              base_path: base_path,
+            )
+
             expect(DownstreamLiveWorker)
               .to receive(:perform_async_in_queue)
               .with(
@@ -341,7 +347,13 @@ RSpec.describe Commands::V2::Publish do
           payload[:update_type] = "minor"
           ChangeNote.create(edition: draft_item)
         end
+
         it "deletes associated ChangeNote records" do
+          FactoryGirl.create(:live_edition,
+            document: draft_item.document,
+            base_path: base_path,
+          )
+
           expect { described_class.call(payload) }
             .to change { ChangeNote.count }.by(-1)
         end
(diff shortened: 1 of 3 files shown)


spec/commands/v2/publish_spec.rb
```diff
@@ -341,7 +347,13 @@
           payload[:update_type] = "minor"
           ChangeNote.create(edition: draft_item)
         end
+
         it "deletes associated ChangeNote records" do
+          FactoryGirl.create(:live_edition,
+            document: draft_item.document,
+            base_path: base_path,
+          )
+
```
kevindew: Is this and the above item to ensure we have a previous item for a minor type? If so is there any chance we could refactor things around so this is within a before block for a context so we can try keep the `it` blocks related focused on the test?

cbaines:
As far as I'm aware, this behaviour (of crashing if the update_type for the first publish is minor) was added in c97f916761e8b04477cdcc98bb501ae0db839f7b . I'm uncertain what value forcing the initial update_type to be `major` has (apart from the not crashing bit). Would this still be a good rule to enforce even if the Publishing API didn't crash?

spec/commands/v2/publish_spec.rb
leenagupte: Yes. Exactly. But yes, I can move it to the `before` block for neatness.

kevindew:
@cbaines yeah I think it makes sense semantically that, with the lack of an update type that specifies initial creation, that a document beginning it's life published is something major rather than minor.

tijmenb:
Important to note is that "self-published" apps often republish on deploy with a minor update type, even if the page is new (there's no logic to see if it's new).  For example, smart-answers would crash when creating a new smart answer: https://github.com/alphagov/smart-answers/blob/master/app/services/content_item_publisher.rb#L6.

kevindew:
@tijmenb thanks for the heads up. Ugh that's a pain though. We must be building up so much bad history from it.

cbaines:
> @cbaines yeah I think it makes sense semantically that, with the lack of an update type that specifies initial creation, that a document beginning it's life published is something major rather than minor.

Ok, that sounds fine.

leenagupte:
This issue seems a lot more complex than first anticipated. As a starting point, I'm going to close this PR and create a new one stops publishing api from crash if a `minor` update type is passed on first publish.

We should revisit this at some point and decide what the different update types mean.
For example,  at the moment "republish" just means put it on a lower priority queue.
Should "major" just mean that email is sent (and the changelog is updated), or should there be a way to opt out of the email?

PR #894, opened 2017-04-28
Prevent redirect destination ending in slash

cbaines:


kevindew:
This makes sense - my first thought was that it should be covered by schemas but I'm guessing we that doesn't get caught when unpublishing?

cbaines:
> This makes sense - my first thought was that it should be covered by schemas but I'm guessing we that doesn't get caught when unpublishing?

That looks to be the case, I spotted this error https://errbit.publishing.service.gov.uk/apps/537de1230da115fb25004224/problems/5901d9726578631e4bb50a00

PR #896, opened 2017-05-02
Set public_updated_at for auto published items

leenagupte:
Trello card: https://trello.com/c/3S4m5wAJ
Supersedes: PR #893 

## Changes

Stop the publishing-api from crashing in cases where an Edition has never had a `major` update (the `public_updated_at` value has never been set) and the update type is `minor`.

## Context

If an Edition is published in batches by an application rather than a human then the update_type is usually set to minor.

See [Smartanswers](https://github.com/alphagov/smart-answers/blob/master/app/services/content_item_publisher.rb#L6), where all Smartanswers are republished when a single smart answer is updated.

Setting the update_type to minor stop users from receiving multiple emails, potentially about topics they have no interest in.

The `publish` function expects that the Edition has a value in `public_updated_at` if a minor update is being published. If it doesn’t, the publishing-api crashes.

Most applications get around this my setting `public_updated_at` themselves.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -154,7 +154,7 @@ module Commands
       def set_public_updated_at
         return if edition.public_updated_at.present?
 
-        if update_type == "major"
+        if update_type == "major" || previous_item.blank?
           edition.update_attributes!(public_updated_at: Time.zone.now)
         elsif update_type == "minor"
           edition.update_attributes!(public_updated_at: previous_item.public_updated_at)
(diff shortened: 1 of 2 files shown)


app/commands/v2/publish.rb
```diff
@@ -154,7 +154,7 @@ def clear_published_items_of_same_locale_and_base_path
       def set_public_updated_at
         return if edition.public_updated_at.present?
 
-        if update_type == "major"
+        if update_type == "major" || previous_item.blank?
```
thomasleese: I see that we also compare the `update_type` with `"major"` in a few other places in this command. Do those needs to be updated as well to do this?

leenagupte: @thomasleese The other place that `update_type == "major"` is used in the publish function is to check whether to delete the change log. I think for self/auto publishers it's ok not to keep the change log.

thomasleese: @leenagupte Yep, that makes sense 😄

PR #899, opened 2017-05-04
Link to withdrawn available translations

thomasleese:
This PR adds the ability to allow editions to link to withdrawn available translations.

[Trello card](https://trello.com/c/EWBxn7q6/931-can-t-link-to-withdrawn-available-translations-2)

### DIFF
--- a/app/presenters/queries/available_translations.rb
+++ b/app/presenters/queries/available_translations.rb
@@ -38,17 +38,24 @@ module Presenters
       end
 
       def state_fallback_order
-        with_drafts ? %i[draft published] : %i[published]
+        return %i[draft published unpublished] if with_drafts
+        %i[published unpublished]
       end
 
       def edition_scope
-        Edition
+        scope = Edition
           .with_document
           .with_unpublishing
           .where(
             documents: { content_id: content_id },
             state: state_fallback_order,
           )
+
+        # filter out unpublishings which aren't withdrawals (i.e. gone, redirect, etc)
+        scope
+          .where("
+            editions.state != 'unpublished' OR unpublishings.type = 'withdrawal'
+          ")
       end
 
       def pluck_and_sort_editions(scope)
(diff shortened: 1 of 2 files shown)


app/presenters/queries/available_translations.rb
```diff
@@ -16,28 +16,58 @@ def translations
...
+        pluck_and_sort_editions(
+          filter_unpublishings(editions)
+        )
       end
 
       def expand_translation(id)
-        expansion_rules = LinkExpansion::Rules
-        web_item(id).select { |f| expansion_rules.expansion_fields(:available_translations).include?(f) }
+        web_item(id).select do |field|
+          expansion_rules
```
kevindew: might be slightly easier to follow to have `LinkExpansion::Rules` here instead of a method

app/presenters/queries/available_translations.rb
```diff
@@ -16,28 +16,58 @@ def translations
...
+        end
       end
 
       def state_fallback_order
-        with_drafts ? %i[draft published] : %i[published]
+        return %i[draft published unpublished] if with_drafts
+        %i[published unpublished]
+      end
+
+      def editions
```
kevindew: since we're using this as a scope could be better to call it `edition_scope`?

app/presenters/queries/available_translations.rb
```diff
@@ -16,28 +16,58 @@ def translations
...
+          .where(
+            documents: { content_id: content_id },
+            state: state_fallback_order,
+          )
+      end
+
+      def filter_unpublishings(scope)
+        scope.where("
+          editions.state != 'unpublished' OR unpublishings.type = 'withdrawal'
+        ")
```
kevindew: does this ever run outside of edition scope? could be a part of the editions method to save you having to jump around while reading?

thomasleese: Sure, I originally put it in the edition scope, but I found it looks a bit confusing why it was there. I opted for putting it in a separate method rather than having a comment, but maybe it is fine in the edition scope.

kevindew: Sure I think a comment is a pretty good idea because what we're doing is pretty weird

thomasleese:
@kevindew I believe I have addressed your comments. 😄

PR #900, opened 2017-05-04
Sync public_updated_at with first_published_at on specific Whitehall formats

steventux:

This migration updates the low risk formats/schema types which can use the `first_published_at` value as the `public_updated_at` timestamp where it currently has an older time.

There's some additional work to be done in Whitehall to shore up invalid date combinations so that more bad data isn't sent to the Publishing API.
Trello card here: https://trello.com/c/TPwh5GYI/491-whitehall-should-provide-a-public-updated-at-value-if-when-sending-a-first-published-at-value-to-the-publishing-api

### DIFF
--- a/db/migrate/20170512090530_sync_whitehall_public_updated_at.rb
+++ b/db/migrate/20170512090530_sync_whitehall_public_updated_at.rb
@@ -1,0 +1,31 @@
+class SyncWhitehallPublicUpdatedAt < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  def up
+    schema_names = %w(gone redirect special_route speech statistical_data_set take_part topical_event_about_page)
+
+    schema_names.each do |schema_name|
+      scope = Edition.where(publishing_app: "whitehall")
+                     .where(schema_name: schema_name)
+                     .where("first_published_at AT TIME ZONE 'Europe/London' > public_updated_at")
+
+      # Sync all the timestamps
+      scope.update_all("public_updated_at = first_published_at")
+      puts "#{scope.count} records updated for schema '#{schema_name}'"
+      puts "(#{scope.pluck(:id).join(",")})"
+
+      # Narrow the scope to find current content
+      current_editions_scope = scope.where(state: %w(draft published))
+      current_edition_content_ids = current_editions_scope.map(&:content_id)
+
+      if Rails.env.production? && current_editions_scope.any?
+        # Represent current editions downstream
+        Commands::V2::RepresentDownstream.new.call(current_edition_content_ids, with_drafts: true)
+
+        # Output what has been updated downstream
+        puts "The following items have been represented downstream:"
+        puts current_editions_scope.map(&:base_path).uniq.sort
+      end
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170510090933) do
+ActiveRecord::Schema.define(version: 20170512090530) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
```diff
@@ -0,0 +1,29 @@
+class SyncWhitehallPublicUpdatedAt < ActiveRecord::Migration[5.0]
+  def change
```
thomasleese: I think this should be `up` so it doesn't happen on a rollback.

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
```diff
@@ -0,0 +1,29 @@
...
+    schema_names = %w(gone redirect special_route speech statistical_data_set take_part topical_event_about_page)
+
+    schema_names.each do |schema_name|
+      scope = Edition.where(publishing_app: "whitehall")
+                     .where(schema_name: schema_name)
+                     .where("first_published_at AT TIME ZONE 'Europe/London' > public_updated_at")
```
thomasleese: Do we need the `AT TIME ZONE` bit for `public_updated_at` as well, or is that implied, or simply not required?

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
```diff
@@ -0,0 +1,29 @@
...
+      puts "#{scope.count} records updated for schema '#{schema_name}'"
+      puts "(#{scope.pluck(:id).join(",")})"
+
+      # Narrow the scope to find current content
+      current_editions_scope = scope.where(state: %w(draft published))
+      current_edition_content_ids = current_editions_scope.map(&:content_id)
+
+      if current_editions_scope.any?
+        # Represent current editions downstream
+        Commands::V2::RepresentDownstream.new.call(current_edition_content_ids, with_drafts: true)
```
thomasleese: Because you are doing this, you'll want to include the following line at the top of the class:

```ruby
disable_ddl_transaction!
```

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
```diff
@@ -0,0 +1,29 @@
+class SyncWhitehallPublicUpdatedAt < ActiveRecord::Migration[5.0]
+  def change
+    schema_names = %w(gone redirect special_route speech statistical_data_set take_part topical_event_about_page)
```
thomasleese: I'm wondering what advantage limiting by `schema_name` gives us, what if there is a mismatched time in an edition that is not part of these schemas?

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
steventux: This is defensive thinking on my part, the `first_published_at` can be set by the admin user (eg. speeches) which may contain a value with an implicit BST offset.
The record count differs with an without this, the scope is narrowed slightly by doing the query with the timezone, which might mean some genuine < 1 hour discrepancies still exist. I'd expect to have another pass at this given the volume of data.

thomasleese: Ah sure, yeah, that makes sense. 👍

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
steventux: The complete number of Whitehall records where the `first_published_at` is more recent that the `public_updated_at` is ~600,000, I'm addressing the simple cases first here on a per-format basis.
Formats like `gone` and `redirect` are fairly low risk and the others I've included here either don't render these dates or have a correct `first_published_at` value which can be used as the `public_updated_at`. Basically this should cover the quick wins.

thomasleese: Ah cool, yeah, that sounds good, get the easy ones out of the way! 👍

db/migrate/20170504145330_sync_whitehall_public_updated_at.rb
steventux: :+1:

kevindew: And you only want it in production as it won't run on dev vms

kevindew:
I think I said this somewhere before but can't remember the outcome. Do we have first_public_at for these formats? 

And secondly are we at risk of setting different first_published_at for each editions of documents with this?

steventux:
`statistical_data_set` uses first public at but this [appears to be in sync with `first_published_at`](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/payload_builder/first_public_at.rb#L5)

steventux:
On the second point I think it's possible we'd set different values as there's nothing to stop this happening, I'll take a look at the data and see if this is likely to happen.

PR #901, opened 2017-05-05
Fix link expansion concurrency bug

thomasleese:
This PR fixes a concurrency bug with link expansion.

[Trello card](https://trello.com/c/kzYq1ceG/927-concurrent-bug-problem-producing-inconsistent-documents-3)

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -99,7 +99,7 @@ class Edition < ApplicationRecord
   def draft_cannot_be_behind_live
     return unless document
 
-    if state == "draft"
+    if draft?
       draft_version = user_facing_version
       published_unpublished_version = document.published_or_unpublished.try(:user_facing_version)
     end
@@ -210,6 +210,10 @@ class Edition < ApplicationRecord
 
   def substitute?
     unpublished? && unpublishing.substitute?
+  end
+
+  def draft?
+    content_store == "draft"
   end
 
   def api_path
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -83,10 +83,9 @@ module Presenters
     end
 
     def expanded_link_set_presenter
-      @expanded_link_set_presenter ||= Presenters::Queries::ExpandedLinkSet.new(
-        content_id: edition.content_id,
-        draft: draft,
-        locale: edition.locale
+      @expanded_link_set_presenter ||= Presenters::Queries::ExpandedLinkSet.by_edition(
+        edition,
+        with_drafts: draft,
       )
     end
 
--- a/app/presenters/queries/available_translations.rb
+++ b/app/presenters/queries/available_translations.rb
@@ -1,9 +1,16 @@
 module Presenters
   module Queries
     class AvailableTranslations
-      def initialize(content_id, with_drafts: false)
-        @content_id = content_id
-        @with_drafts = with_drafts
+      def self.by_edition(edition)
+        self.new(edition: edition)
+      end
+
+      def self.by_content_id(content_id, with_drafts: false)
+        self.new(content_id: content_id, with_drafts: with_drafts)
+      end
+
+      def initialize(options)
+        @options = options
       end
 
       def translations
@@ -13,7 +20,19 @@ module Presenters
 
     private
 
-      attr_reader :content_id, :with_drafts, :expanded_translations
+      attr_reader :options
+
+      def edition
+        @edition ||= options[:edition]
+      end
+
+      def content_id
+        edition ? edition.content_id : options.fetch(:content_id)
+      end
+
+      def with_drafts
+        edition ? edition.draft? : options.fetch(:with_drafts)
+      end
 
       def grouped_translations
         pluck_and_sort_editions(edition_scope)
@@ -27,8 +46,13 @@ module Presenters
         end
       end
 
+      def edition_for_id(id)
+        return edition if edition && edition.id == id
+        Edition.find_by(id: id)
+      end
+
       def web_item(id)
-        Edition.find_by(id: id).to_h
+        edition_for_id(id).to_h
       end
 
       def expanded_translations
--- a/app/presenters/queries/available_translations.rb
+++ b/app/presenters/queries/available_translations.rb
@@ -62,6 +62,7 @@ module Presenters
         scope.pluck(:id, :locale, :state)
           .sort_by { |(_, _, state)| state_fallback_order.index(state.to_sym) }
           .group_by { |(_, locale)| locale }
+          .each_with_object({}) { |(key, value), hash| hash[key] = value.first }
       end
     end
   end
--- a/app/presenters/queries/available_translations.rb
+++ b/app/presenters/queries/available_translations.rb
@@ -1,8 +1,8 @@
 module Presenters
   module Queries
     class AvailableTranslations
-      def self.by_edition(edition)
-        self.new(edition: edition)
+      def self.by_edition(edition, with_drafts: false)
+        self.new(edition: edition, with_drafts: with_drafts)
       end
 
       def self.by_content_id(content_id, with_drafts: false)
@@ -11,6 +11,7 @@ module Presenters
 
       def initialize(options)
         @options = options
+        @with_drafts = options.fetch(:with_drafts)
       end
 
       def translations
@@ -20,7 +21,7 @@ module Presenters
 
     private
 
-      attr_reader :options
+      attr_reader :options, :with_drafts
 
       def edition
         @edition ||= options[:edition]
@@ -28,10 +29,6 @@ module Presenters
 
       def content_id
         edition ? edition.content_id : options.fetch(:content_id)
-      end
-
-      def with_drafts
-        edition ? edition.draft? : options.fetch(:with_drafts)
       end
 
       def grouped_translations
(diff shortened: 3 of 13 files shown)


lib/link_expansion/content_cache.rb
```diff
@@ -32,8 +44,7 @@ def editions(content_ids)
       locale_fallback_order: locale_fallback_order,
       state_fallback_order: state_fallback_order,
     )
-    Edition.with_document.includes(:document).where(id: edition_ids)
-      .each_with_object(results) { |item, memo| memo[item.content_id] = item }
+    Edition.with_document.includes(:document).where(id: edition_ids).all
```
kevindew: assume we no longer need the results variable?

lib/link_expansion/content_cache.rb
```diff
@@ -17,8 +17,20 @@ def find(content_id)
 
   attr_reader :store, :with_drafts, :locale
 
+  def preload_content(editions, content_ids)
+    store = Hash[editions.compact.map { |edition| [edition.content_id, edition] }]
```
kevindew: is there a reasonable likelihood of having nil items in the array?

lib/link_expansion/content_cache.rb
```diff
@@ -17,8 +17,20 @@ def find(content_id)
...
+
+    editions(content_ids).each do |edition|
+      if !store.has_key?(edition.content_id)
+        store[edition.content_id] = edition
+      end
+    end
+
+    store
```
kevindew: I think there's scope to refactor this a bit to make it easier to understand.

method name might make sense more like "build_store" or similar as another `preload_*` gets a bit tricky to grok.

could we do something like:
```
to_preload = content_ids - existing_editions.map(&:content_id)
preloaded_editions = editions(to_preload)
// group them all by content_id to create the store
```

not sure if it only seems more readable to me? could be there's a different way to do it that's easier still - wdyt?

app/presenters/queries/available_translations.rb
```diff
@@ -32,7 +32,7 @@ def web_item(id)
       end
 
       def expanded_translations
-        @expanded_translations ||= grouped_translations.map do |_, (id)|
+        @expanded_translations ||= grouped_translations.map do |_, ((id, _, _))|
```
thomasleese: We do need the extra brackets here.

What was happening before was that an array like `[1, "en", "draft"]` was being passed to `expand_translation` and then Rails was running a query like `id IN (1, 'en', 'draft')`. This happens because `grouped_translations` returns a hash mapping locale to an array containing one item which is the array containing id, locale and state. An alternative fix might be to change `grouped_translations` to not return the superfluous one item array?

app/presenters/queries/expanded_link_set.rb
```diff
@@ -3,10 +3,11 @@ module Queries
     class ExpandedLinkSet
       attr_reader :draft
 
-      def initialize(content_id:, draft: false, locale: Edition::DEFAULT_LOCALE)
+      def initialize(content_id: nil, locale: nil, edition: nil, draft: false)
```
thomasleese: I suppose we don't need to, but I allowed it for debugging.

lib/link_expansion.rb
```diff
@@ -7,9 +7,9 @@
 class LinkExpansion
   attr_reader :content_id, :locale, :with_drafts
 
-  def initialize(content_id, locale: Edition::DEFAULT_LOCALE, with_drafts: false)
+  def initialize(content_id, locale: nil, with_drafts: false)
     @content_id = content_id
-    @locale = locale
+    @locale = locale || Edition::DEFAULT_LOCALE
```
thomasleese: Yeah, because we changed constructor for `ExpandedLinkSet` to accept a `nil` locale (rather than a default one of `en`), this method now gets either a `nil` locale or a set one, and if it's `nil` and we haven't given an `edition` we want it to default to `en`.

lib/link_expansion.rb
```diff
@@ -5,11 +5,14 @@
...
 #
 class LinkExpansion
-  attr_reader :content_id, :locale, :with_drafts
+  attr_reader :content_id, :locale, :edition, :with_drafts
 
-  def initialize(content_id, locale: nil, with_drafts: false)
-    @content_id = content_id
-    @locale = locale || Edition::DEFAULT_LOCALE
+  def initialize(content_id: nil, locale: nil, edition: nil, with_drafts: false)
+    raise "Only (content_id, locale) or edition should be set." if ((content_id || locale) && edition) || (!(content_id || locale) && !edition)
```
thomasleese: It's used elsewhere in the app without an edition, for the `GetExpandedLinks` query.

app/presenters/queries/expanded_link_set.rb
thomasleese: Actually, as per above, it's used elsewhere in the app without an edition, for the `GetExpandedLinks` query.

lib/link_expansion/content_cache.rb
thomasleese: There is, but I will fix this so it doesn't.

app/presenters/queries/available_translations.rb
kevindew: sure, yeah it'd be good if we could simplify it as it's a lot of brackets and underscores so quite intimidating.

app/presenters/queries/expanded_link_set.rb
kevindew: Ah dang - it bugs me a bit having both interfaces in as it ups the complexity and chances of inconsistency. I suppose since you can use GetExpandedLinks without an edition (via our endpoint) we can't fix that though.

kevindew: How about if we make it all more explicit?

So rather than attr_reader for draft etc have those as methods which are like

```
def draft
  edition ? edition.draft? : options.fetch(:draft)
end
```

And maybe initialize via static methods that are either `by_edition` or `by_content_id` (or similar)

Just trying wondering how we can unburden it a bit and cut back on ability to have edition conflict with options

thomasleese: Yeah, that sounds quite nice actually. I'll give it ago!

thomasleese:
@kevindew I believe I have addressed the issues you raised.

app/models/edition.rb
```diff
@@ -212,6 +212,10 @@ def substitute?
     unpublished? && unpublishing.substitute?
   end
 
+  def draft?
+    state == "draft"
```
kevindew: I'd dare say content_store == "draft" is our de-facto method rather than state, what do you think?

app/presenters/queries/available_translations.rb
```diff
@@ -62,6 +62,7 @@ def pluck_and_sort_editions(scope)
         scope.pluck(:id, :locale, :state)
           .sort_by { |(_, _, state)| state_fallback_order.index(state.to_sym) }
           .group_by { |(_, locale)| locale }
+          .each_with_object({}) { |(key, value), hash| hash[key] = value.first }
```
kevindew: I think the commit message and the code are out of sync here

app/presenters/edition_presenter.rb
```diff
@@ -83,10 +83,10 @@ def access_limited
...
 
     def expanded_link_set_presenter
-      @expanded_link_set_presenter ||= Presenters::Queries::ExpandedLinkSet.new(
-        content_id: edition.content_id,
-        draft: draft,
-        locale: edition.locale
+      @expanded_link_set_presenter ||= Presenters::Queries::ExpandedLinkSet.by_content_id(
+        edition.content_id,
+        locale: edition.locale,
+        with_drafts: draft,
```
kevindew: hmm can we do this by_edition somehow? (sorry I'm sure this interface is the bain of your life) I'm guessing this was wrote before with_drafts was an option?

app/presenters/queries/expanded_link_set.rb
```diff
@@ -15,17 +22,37 @@ def links
...
         LinkExpansion.new(content_id,
-          with_drafts: draft,
           locale: locale,
+          with_drafts: with_drafts,
         ).links_with_content
       end
 
+      def available_translations
+        if edition
+          AvailableTranslations.by_edition(edition)
```
kevindew: do we want to pass with_drafts here?

app/queries/get_expanded_links.rb
```diff
@@ -18,10 +18,10 @@ def self.call(content_id, locale, with_drafts: true)
     end
 
     def self.expanded_link_set(link_set, locale, with_drafts:)
-      expanded_link_set = Presenters::Queries::ExpandedLinkSet.new(
-        content_id: link_set.content_id,
-        draft: with_drafts,
+      expanded_link_set = Presenters::Queries::ExpandedLinkSet.by_content_id(
+        link_set.content_id,
         locale: locale,
+        with_drafts: with_drafts,
```
kevindew: 👍

thomasleese:
@kevindew Cool, those should be fixed now!

kevindew:
Great - looks good. There's just this comment that might still need something: https://github.com/alphagov/publishing-api/pull/901/files/04791cae538701c377c9cb410098b004ad845a1f..79ee45ccba40302db657d44706005ea2a179ac9a#diff-5c16970016358021adb7211c8b9a79f6

kevindew:
Ah either I missed that it was sorted or it was sorted since I posted that link (I'm assuming the former :wink:)

thomasleese:
Haha, I think GitHub was doing weird things and showing old commits, yeah, I sorted that! 😄

PR #902, opened 2017-05-08
Add missing change notes for specialist documents

steventux:
https://trello.com/c/hBz4UCa3/936-add-change-notes-for-specialist-documents-that-are-missing-them

Adds a migration to create missing change notes for specialist documents.
Uses the earliest relevant timestamp from the earliest edition in the document sequence.
Took 47 seconds to run on a dev VM.

### DIFF
--- a/db/migrate/20170508101720_add_missing_change_notes_to_specialist_documents.rb
+++ b/db/migrate/20170508101720_add_missing_change_notes_to_specialist_documents.rb
@@ -1,0 +1,28 @@
+class AddMissingChangeNotesToSpecialistDocuments < ActiveRecord::Migration[5.0]
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher",
+                          schema_name: "specialist_document",
+                          state: "published")
+      .with_document
+      .where("NOT EXISTS (SELECT * FROM change_notes WHERE content_id::uuid = documents.content_id)")
+
+    scope.each do |edition|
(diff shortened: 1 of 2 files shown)


db/migrate/20170508101720_add_missing_change_notes_to_specialist_documents.rb
```diff
@@ -0,0 +1,28 @@
+class AddMissingChangeNotesToSpecialistDocuments < ActiveRecord::Migration[5.0]
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher",
+                          schema_name: "specialist_document",
+                          state: "published")
+      .joins(:document)
```
thomasleese: You can use `.with_document` instead here.

db/migrate/20170508101720_add_missing_change_notes_to_specialist_documents.rb
```diff
@@ -0,0 +1,28 @@
...
+
+private
+
+  def earliest_date_for(edition)
+    pub_date = earliest(edition.first_published_at, edition.public_updated_at)
+    earliest(pub_date, edition.created_at)
+  end
+
+  def earliest(d1, d2)
+    d1 < d2 ? d1 : d2
```
kevindew: super minor and no need to actually make a change, but `[d1, d2].min` is probably a cleaner way to do this

PR #903, opened 2017-05-09
Use the change note time if earlier than first published at

steventux:

There are around 300 specialist-publisher editions which have a first published time
later than the timestamp recorded in their corresponding change note. Update these
to use the time from the change note.


steventux:
@thomasleese I tried to do this with AR but updating a table with a value from another column doesn't appear to be well supported, I think it can be achieved with Arel but not with `joins` or `includes` that I could make work.

thomasleese:
@steventux Yep, I figured that would be the case, seems fine to me.

PR #904, opened 2017-05-09
Repair attachments with missing content_type

emmabeynon:
A bug was discovered in Specialist Publisher where a CMA case could not be
saved as it included attachments that did not have a `content_type`, causing a
`HTTPUnprocessableEntity` error in the Publishing API.  This adds the correct
`content_type` for all attachments that had it set to nil.


emmabeynon:
@thomasleese all looks good on integration 🎉

PR #905, opened 2017-05-10
Sent the govuk_request_id field to the Content Store

thomasleese:
Currently it is only sent to the message queue but it was decided
that it would be useful to have in the Content Store as well.

Not to be merged until after https://github.com/alphagov/govuk-content-schemas/pull/613 has gone in.

[Trello Card](https://trello.com/c/JZ5Zabe9/702-remove-publishing-request-id-from-travel-advice-publisher-details-hash-2)

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -28,7 +28,6 @@ module Presenters
     def for_message_queue(update_type)
       present.merge(
         update_type: update_type,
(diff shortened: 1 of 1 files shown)


cbaines:
I would guess that the request id is already being sent to the Content Store, but as a HTTP header. I think there is some code in the gds-api-adapters that does this.

thomasleese:
@cbaines Yep, you're right, it will already be accessible in the content store.

PR #907, opened 2017-05-11
Set the publishing_request_id when publishing an edition

thomasleese:
This will be send down to the Content Store so it is usable by frontend apps.

This shouldn't be merged until https://github.com/alphagov/content-store/pull/290 has gone though.

[Trello Card](https://trello.com/c/JZ5Zabe9/702-remove-publishing-request-id-from-travel-advice-publisher-details-hash-2)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ else
 end
 
 gem "gds-sso", "13.0.0"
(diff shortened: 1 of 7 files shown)


thomasleese:
@kevindew Would you be able to approve this? 😄

PR #908, opened 2017-05-11
Populate change history for specialist documents without it

steventux:

There are ~200,000 specialist documents which have an empty change history, this won't
get presented downstream because a memoizing clause in how the change history is presented.
So identify these records (they can be anything but draft), generate an initial change note
if none exists, then update the change history in the details hash from the change notes
for the edition
This migration took 395 seconds to run on a dev VM.

### DIFF
--- a/db/migrate/20170512103531_remove_empty_specialist_publisher_change_history.rb
+++ b/db/migrate/20170512103531_remove_empty_specialist_publisher_change_history.rb
@@ -1,0 +1,53 @@
+class RemoveEmptySpecialistPublisherChangeHistory < ActiveRecord::Migration[5.0]
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher",
+                          schema_name: "specialist_document")
+                   .where.not(state: "draft")
+                   .where("details ->> 'change_history' = '[]'")
+                   .order("user_facing_version ASC")
+
+    scope.each do |edition|
+      change_notes = ChangeNote
+        .where(content_id: edition.content_id)
+        .where("edition_id IS NULL OR edition_id IN (?)", edition_ids(edition))
+        .order(:public_timestamp)
+        .pluck(:note, :public_timestamp)
+        .map { |note, timestamp| { note: note, public_timestamp: timestamp } }
+
+      if change_notes.empty?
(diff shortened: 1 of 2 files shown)


db/migrate/20170511100331_remove_empty_specialist_publisher_change_history.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def earliest_date_for(edition)
+    pub_date = earliest(edition.first_published_at, edition.public_updated_at)
+    earliest(pub_date, edition.created_at)
+  end
+
+  def earliest(d1, d2)
+    return d1 unless d2
+    return d2 unless d1
+    [d1, d2].min
```
kevindew: if we do `[d1, d2].compact.min` can we skip the returns?

kevindew:
Looks like you've had fun here @steventux 

Anyway... if I'm reading this correct, I'd think it has the potential to add the change note of "First Published." to every edition of a document, which might create change histories like:

- First Published.
- First Published.
- First Published.
- First Published.

Whereas we'd want it just done on the first publishing. 

Perhaps that can be achieved without too much pain, by doing a group by document_id and order by lowest user facing version?

db/migrate/20170511100331_remove_empty_specialist_publisher_change_history.rb
```diff
@@ -0,0 +1,54 @@
+class RemoveEmptySpecialistPublisherChangeHistory < ActiveRecord::Migration[5.0]
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher",
+                          schema_name: "specialist_document")
+                   .where.not(state: "draft")
+                   .where("details ->> 'change_history' = '[]'")
+                   .group(:document_id, :id)
```
kevindew: by having the id in here does this not mean that we get every edition still?

steventux:
@kevindew as discussed the amendment to look up existing `ChangeNote` records relating to any edition in the series (with a lower `user_facing_version`) means that when the first `Edition` without a `ChangeNote` has this populated the subsequent editions in the series will skip creating the dupes you mentioned above.

PR #924, opened 2017-05-19
Make change notes document field not nullable

thomasleese:
This is the fifth part of linking change notes with documents. It should not be deployed until #923 has been deployed.

[Trello Card](https://trello.com/c/7GRaupoY/919-since-changehistory-is-associated-with-a-content-id-it-should-be-associated-with-a-locale-1)


thomasleese:
@kevindew It took only a few seconds locally, it will only be painful if there are `null` values, which would suggested a wider problem if there are.

```
irb(main):001:0> ChangeNote.where(document_id: nil).count
=> 0
```

PR #926, opened 2017-05-23
Data migration docs

thomasleese:
Add some docs on writing data migrations to avoid the two main pitfalls that occur with them.

[Trello Card](https://trello.com/c/HWvyiuxQ/942-data-migrations-providing-devs-with-a-tool-2)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,56 @@
+# Contributing
+
+## Are you writing a migration to change Publishing API data?
+
+We have a lot of migrations added to the Publishing API to change data, our
+long term plan is to reduce these so that these can be done via the API itself
+and not need migrations. Until we are at that point please use the following
+guidelines when creating migrations.
+
+### Run the migration on your machine first
+
+At the very least this will ensure that the code runs and that your data looks
+correct, but it also allows you to get an idea for how long the migration will
+take to finish. You should then include that information in the PR to help with
+whoever is doing the deployment.
+
+We also encourage you to run and check the results of the migration on
+integration, since the data there will be closer to production.
+
+### Always include the `schema.rb` file
+
+Even if the change is just the timestamp, it's important to include this file
+in your commit otherwise when trying to use the app in testing an error saying
+there are pending migrations will appear.
+
+### Target only content related to the application you are affecting
+
+Mistakes in migrations can and do happen, by at least targeting the application
+you are affecting (with `.where(publishing_app: )`) you can avoid unnecessary
+fallout.
+
+### Don't assume the records you are altering exist
+
+Migrations get run in test environments, on old databases or when a developer
+first clones the project. You should always make sure that your migrations run
+fine without the data there that you are expecting.
+
+### Do you need to represent this data downstream?
+
+Representing the data downstream won't work on local machines, so you will need
+to have a check in the migration making sure it is running in the right
+environment (`Rails.env.production?`). Alternatively, and ideally, you can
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
...
+  This prevents the need to have a check for production since the Sidekiq
+  workers are not available in the other environments, and means you do not
+  have to use `disable_ddl_transaction!`.
+- If the record you are trying to modify doesn't exist, simply move on to the
+  next one rather than failing. This prevents new clones of the
+  `publishing-api` from having to replicate their data from production before
+  being able to run the migrations.
+
+We also recommend that every so often, old data migrations are deleted since
+they are most likely no longer useful.
```
kevindew: The content here looks good but I think it maybe implies quite a high understanding.

I'd maybe try going more along this sort of lines:

## Are you writing a migration to change Publishing API data?

We have a lot of migrations added to the Publishing API to change data, our long term plan is to reduce these so that these can be done via the API itself and not need migrations. Until we are at that point please follow the following guidelines when creating migrations.

### Firstly run the migration first on your machine first
Why
Include details in PR of how long it took against reasonably fresh set of data

### Always include the schema.rb file
Even if it's just the timestamp
Why

### Target only content related to the application you are affecting
e.g include where "publishing_app: whitehall" to reduce risk of affecting another app, even if it seems unlikely

### Don't assume the records you are altering exist
Migrations are run in test environments, on old dbs etc

### Do you need to represent this data downstream?
This will break local machines
Use rake task ideally separate
if not disable transaction and use


### If you are disabling transaction ensure the migration is idempotent
Why

What do you think?

As a side idea - rather than using the if env == production we could consider having an ENV var for checking whether you can represent_downstream

thomasleese: Yes this is much better!

thomasleese:
@kevindew This is ready for re-review!

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+Even if the change is just the timestamp, it's important to include this file
+in your commit otherwise when trying to use the app in testing an error saying
+there are pending migrations will appear.
+
+### Target only content related to the application you are affecting
+
+Be as specific as possible when updating content, for example, include a
+condition on the `publishing_app` even if it seems unlikely there would be
+content in any other state.
```
kevindew: I'd be a bit firmer here, more that mistakes can and do happen the least you can do is scope your query just to the publishing_app.

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+Migrations get run in test environments, on old databases or when a developer
+first clones the project. You should always make sure that your migrations run
+fine without the data there that you are expecting.
+
+### Do you need to represent this data downstream?
+
+Representing the data downstream won't work on local machines, so you will need
+to have a check in the migration making sure it is running in the right
+environment. Alternatively, and ideally, you can use the
+`represent_downstream:` class of rake tasks in Jenkins to achieve the same
```
kevindew: might want to link to the file with the tasks in here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+result. If you do decide to represent downstream in the migration itself, you
+will need to have `disable_ddl_transaction!` at the top to make sure the data
```
kevindew: "you must disable transactions for this migration (running the `disable_ddl_transaction!` method) as otherwise you will be representing downstream before the data is committed to the database."

Was hoping to find a link for usage of disable_ddl_transaction! but there doesn't seem to be a good one.

PR #929, opened 2017-05-23
Increase DB pool size

kevindew:
We now have increased number of sidekiq workers from 2 to 6 so this
would require an increase in the pooled connections. 6 should be all
that is required here but 10 gives us a little wiggle room.

I've given this a spin on integration and no problems have occurred, of
course I said that on the PR that this is to fix. This time I've sent 6000
things downstream though

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,7 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  pool: 5
+  pool: 10
   template: template0
 
 development:


h-lame:
I remember seeing lots of timeouts on the dev VM in the past until we reduced the rummager concurrency (see: https://github.com/alphagov/rummager/pull/734).  Worth checking back to see if the same problem occurs again now there is more contention for connections on the db again.

kevindew:
@h-lame Ah did that fix the problem [on this card](https://trello.com/c/WKaO9Nls/389-work-out-the-development-postgresql-connection-problem-1-2-day) ? I'm well confused as to how they're related though :-/

h-lame:
I think it did because the rummager sidekiq workers all connected to the publishing-api to do whatever it is they do and with 16 of them at once they were taking up all the available connections for the publishing-api.  I reckon this'll be fine though; the pool size was 5 now it's 10.  If it was ok before then adding 4 new workers with 5 new pooled connections should mean you have even more wiggle room than you did before.

kevindew:
Ok cool - well fingers crossed :crossed_fingers: 

It seems pretty weird as you should only get the error from Publishing API if somehow Publishing API has more processes running than connections pooled, and since Rummager doesn't Publishing API processes (as far as I'm aware) I don't get how that affected it. Very strange.

PR #932, opened 2017-05-25
Allow multiple content ids for represent downstream

kevindew:
Can be used by running commands such as:

`rake represent_downstream:content_id[46fb4dec-d076-473a-8c42-f24d03e04b30 46fb4dec-d076-473a-8c42-f24d03e04b30]`

### DIFF
--- a/lib/tasks/represent_downstream.rake
+++ b/lib/tasks/represent_downstream.rake
@@ -55,13 +55,14 @@ namespace :represent_downstream do
   end
 
   desc "
-  Represent an individual edition downstream
+  Represent an individual or multiple editions downstream
   Usage
   rake 'represent_downstream:content_id[57a1253c-68d3-4a93-bb47-b67b9b4f6b9a]'
   "
   task :content_id, [:content_id] => :environment do |_t, args|
+    content_ids = args[:content_id].split(" ")
     represent_downstream(
-      Edition.with_document.where(documents: { content_id: args[:content_id] })
+      Edition.with_document.where(documents: { content_id: content_ids })
     )
   end
 end


lib/tasks/represent_downstream.rake
```diff
@@ -55,13 +55,14 @@ namespace :represent_downstream do
   end
 
   desc "
-  Represent an individual edition downstream
+  Represent an individual or multiple editions downstream
   Usage
   rake 'represent_downstream:content_id[57a1253c-68d3-4a93-bb47-b67b9b4f6b9a]'
   "
   task :content_id, [:content_id] => :environment do |_t, args|
+    content_ids = args[:content_id].split(" ")
```
thomasleese: It seems more intuitive to me that the separator would be a comma (if I did't read the docs/code, I would assume it to be a comma).

kevindew: Yeah - I based it on specialist publishers similar commands: https://github.com/alphagov/specialist-publisher/blob/master/lib/tasks/republish.rake and did notice (rather lately) that there had been confusion over their usage [here](https://deploy.publishing.service.gov.uk/job/run-rake-task/1155/console) (which may mean I tried to fix way more content_ids than needed)

The problem I have with comma is that it's a small step to ", " and then more whitespace concerns. Do we have anything in Publishing API tasks that sets a consistency here?

kevindew: I guess the other thing is you can use comma to separate arguments in the `[]` so space might be more friendly for not conflicting with that  ¯\_(ツ)_/¯

thomasleese: Hmm, yeah, that seems reasonable. That's true about the arguments in `[]`, perhaps Rake might not even let you have commas in that string.

PR #933, opened 2017-05-25
Better support for republish update_type

kevindew:
This band aids a couple of the holes that we have in the system for using an update_type of republish - as evidenced from running the specialist publisher republish task. 

I think we should aim to remove this update_type as it adds junk to the history and all feels complicated and prone to bugs, but until then it's probably best to keep it going.

### DIFF
--- a/app/models/previously_published_item.rb
+++ b/app/models/previously_published_item.rb
@@ -19,7 +19,7 @@ class PreviouslyPublishedItem
     previously_published_item.user_facing_version + 1
   end
 
-  def set_first_published_at?
+  def has_first_published_at?
     true
   end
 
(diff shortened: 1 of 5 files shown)


app/models/previously_published_item.rb
```diff
@@ -27,6 +27,14 @@ def first_published_at
     previously_published_item.first_published_at
   end
 
+  def set_last_edited_at?
```
steventux: I don't really understand the name for this method though I can see there's a `set_first_published_at?` method so I can see it's consistent.
My understanding is that it is used to determine whether the previous item has a `last_edited_at` timestamp? In which case `has_last_edited_at?` would make more sense as `set_` prefixed methods are usually mutators*.

`* The biased opinion of an ex-Java developer.`

kevindew: Yup I agree. Any methods with get/set prefixes get me twitchy. I'll see if I can change both of them without the application burning.

kevindew:
@steventux Just addressed that comment.

PR #934, opened 2017-05-25
Widen scope of specialist publisher republish fix

kevindew:
The previous scope didn't catch items that had been edited since the data was altered.


kevindew:
Thanks @steventux 👍

PR #935, opened 2017-05-26
Refactor unit tests

steventux:
https://trello.com/c/a6izDPez/930-separate-integration-tests-from-unit-tests-for-valid-schemas-3

An initial attempt at separating larger integration scenarios from unit tests.
I've focused on the main command specs which are extensive, some of these were good candidates for moving to integration specs and calling via the endpoints.

There's also some work done to DRY up common examples and build a common context for the new integration specs for the `/v2/content` endpoint


spec/commands/v2/discard_draft_spec.rb
```diff
@@ -8,9 +8,11 @@
     end
 
     let(:expected_content_store_payload) { { base_path: "/vat-rates" } }
+    let(:content_id) { SecureRandom.uuid }
```
thomasleese: Since `content_id` is not actually used in the file anywhere on its own, we can just leave it to FactoryGirl to create one for our document.

spec/commands/v2/put_content_spec.rb
```diff
@@ -31,20 +31,17 @@
...
+      validator = double(:validator)
+      expect(Commands::V2::PutContentValidator).to receive(:new)
+        .with(payload, instance_of(described_class))
+        .and_return(validator)
+      expect(validator).to receive(:validate)
+      expect(PathReservation).to receive(:reserve_base_path!)
+      expect { described_class.call(payload) }.not_to raise_error
     end
 
+
```
thomasleese: Extra blank line

spec/integration/put_content/new_edition_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+RSpec.describe "PUT /v2/content when the payload is for a brand new edition" do
+  include PutContentIntegrationHelper
+
+  let(:payload) { default_payload }
+
+  it "creates an edition" do
+    put "/v2/content/#{content_id}", params: payload.to_json
+
+    edition = Edition.last
```
thomasleese: I wonder if we can use `subject` somehow to avoid having to put `Edition.last` in all the examples?

steventux:
Thanks @thomasleese I think this is rebased and up to date now.

PR #940, opened 2017-06-01
De-duplicate embedded specialist edition change history

steventux:
https://trello.com/c/NbATcan2/956-de-duplicate-specialist-publisher-embedded-change-history.-(2)

This migration identifies editions with change history embedded in the details hash
which contain duplicate notes. Some of these may be legitimate entries so we clear
the  entry in the details hash and represent the edition downstream.
The presenter will regenerate this history correctly based on the ChangeNote records.
Migration took ~58 seconds on a dev VM

### DIFF
--- a/db/migrate/20170612115845_dedupe_embedded_specialist_change_history.rb
+++ b/db/migrate/20170612115845_dedupe_embedded_specialist_change_history.rb
@@ -1,0 +1,95 @@
+class DedupeEmbeddedSpecialistChangeHistory < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    # Find ChangeNotes which have no edition id and
+    # have duplicate documents / notes, these need to be removed first.
+    # The only publishing apps affected are specialist publisher and service manual publisher.
+    dupe_change_note_scope = ChangeNote.where(edition_id: nil)
+                                       .group(:document_id, :note)
+                                       .having("COUNT(*) > 1")
+
+    dupe_change_notes_ids = dupe_change_note_scope.pluck("array_agg(id)").flatten
+    last_dupe_change_notes_ids = dupe_change_note_scope.pluck("max(id)").flatten
+    change_note_ids_to_delete = dupe_change_notes_ids - last_dupe_change_notes_ids
+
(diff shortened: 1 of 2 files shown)


db/migrate/20170612115845_dedupe_embedded_specialist_change_history.rb
```diff
@@ -0,0 +1,102 @@
...
+  def up
+    # Find ChangeNotes which have no edition id and
+    # have duplicate documents / notes, these need to be removed first.
+    # The only publishing apps affected are specialist publisher and service manual publisher.
+    dupes_sql = <<-SQL
+      SELECT max(id)
+      FROM change_notes
+      WHERE edition_id IS NULL
+      GROUP BY document_id, note
+      HAVING (count(*) > 1)
```
kevindew: I'm a bit confused by this query and the loop - maybe I'm totally missing something. I'm assuming we're pulling out one ID each time, if so why not all in one, and/or why the max one?

steventux: I struggled to find a way to give me back all the change note ids where there was duplication (defined by grouping document and note) as the grouping clause won't allow the query to return all change note ids. Using max(id) gives me back one dupe only, by looping over this query we delete the dupes by reduction/repetition.
I attempted this by trying the grouping in a sub-query but the outer select could to my mind always return non-duplicates.
If there's a handy way in SQL of determining the dupes then giving back the ids for all but one in the series it would improve this massively.

kevindew: Ah! I understand it better now thanks. So it returns just the first id from each collision - I thought it returned just one row at a time.

You can do this to get them all back: `ChangeNote.where(edition_id: nil).group(:document_id, :note).having("COUNT(*) > 1").pluck("array_agg(id)")`

steventux: That's much better - thanks will update

steventux:
@kevindew updated, I had to make one small change to this. I'm happy it works now.

PR #941, opened 2017-06-02
Add a rake task to discard drafts

leenagupte:
Trello card: https://trello.com/c/PSlBhEMi

### DIFF
--- a/doc/admin-tasks.md
+++ b/doc/admin-tasks.md
@@ -1,11 +1,54 @@
-## Admin Tasks
+# Admin Tasks
 
 This is a place to list all of the admin tasks available in Publishing API.
 
-### Discarding drafts
+## Discarding drafts
 
 If you need to discard a draft of a document, run the `discard_draft` rake task:
 
 ```
 bundle exec rake discard_draft['some-content-id']
 ```
+
+## Representing data downstream
+
+Sometimes you need to re-send content to the Content Store to ensure consistency.
+
+The following tasks will allow you to specify which content items/editions to add to the queue that updates the Content Store.
+
+* Represent all editions downstream
+```
+bundle exec represent_downstream:all
+```
+N.B. This task will take several hours and should be used with caution.
+
+* Represent downstream for a specific document_type
+```
+bundle exec represent_downstream:document_type['a-document-type']
+```
+
+* Represent downstream for a rendering application
+```
+bundle exec represent_downstream:rendering_app['application-name']
+```
+
+* Represent downstream for a publishing application
+```
+bundle exec represent_downstream:publishing_app['application-name']
+```
+
+* Represent downstream content which has at least one link of type `taxon`
+```
+bundle exec represent_downstream:tagged_to_taxon
+```
+
+* Represent an individual edition downstream
+```
+bundle exec represent_downstream:content_id['some-content-id']
+```
+
+* Represent multiple editions downstream
+```
+bundle exec represent_downstream:content_id['some-content-id some-other-content-id']
+```
+N.B. The content ids are separated by a space.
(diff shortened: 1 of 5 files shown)


spec/rails_helper.rb
```diff
@@ -71,6 +72,7 @@
   end
 
   config.before(:suite) do
+    Rails.application.load_tasks
```
cbaines: It would be useful to split these changes out in to another commit, that describes why they are happening.

leenagupte: You can't really, because the first test I added won't run without this change.

cbaines: Adding a new commit with these changes before this commit would work though?

leenagupte: It would work, but then the change by itself isn't necessary. I can update the commit note with why this line had to be added to make things a bit clearer.

doc/admin-tasks.md
```diff
@@ -0,0 +1,11 @@
+## Admin Tasks
+
+This is a place to list all of the admin tasks available in publishing-api.
```
kevindew: bit nitpicky but we wouldn't have a dash in Publishing API.

doc/admin-tasks.md
```diff
@@ -0,0 +1,11 @@
+## Admin Tasks
```
kevindew: Thanks for writing this up as documentation 👍  I'm wondering if we actually need it though - as it doesn't really tell you any more than running `rake -T` it might be prudent to just suggest running that in the main readme rather than linking to this file? A concern of mine is that we'd want this file to have all the rake tasks listed pretty much and then we might just be duplicating (and potentially) falling out of sync with the rake documentation.

leenagupte: I added it to a this file so that it would appear in the developer docs. At the moment the main README is not pulled, only contents of the docs folder are. See: https://docs.publishing.service.gov.uk/apis/publishing-api.html

kevindew: @leenagupte Ah sure - makes sense (especially given the acceptance criteria) - I don't think it's too pressing for this to be available to 2nd line. I've no memory of this ever coming up as an issue since I've been here, so probably more a specialist task. Guess it'd be nicer if we had a way of documenting the rake tasks automatically for documentation, but not sure how to do that.

Seems cool though if you think it'll be useful, can just re-asses later if it becomes a pain. Would you be able to include documentation for represent_downstream too since that's a set of common tasks?

leenagupte: @kevindew I can do that.

doc/admin-tasks.md
```diff
@@ -0,0 +1,52 @@
...
+```
+bundle exec represent_downstream:rendering_app['application-name']
+```
+
+* Represent downstream for a publishing application
+```
+bundle exec represent_downstream:publishing_app['application-name']
+```
+
+* Represent downstream content tagged to a parent taxon
```
kevindew: It would be good to rephrase this one slightly: 'Represent downstream content which has at least one link of type "taxon"' - what do you think?

leenagupte: Are `taxons` defined as links?

doc/admin-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+
+Sometimes you need to re-send content to the Content Store to ensure consistency.
+
+The following tasks will allow you to specify which content items/editions to add to the queue that updates the Content Store.
+
+* Represent all editions downstream
+```
+bundle exec represent_downstream:all
+```
+N.B. This task will several hours and should be used with caution.
```
kevindew: Think we missed a word here? - all good besides that

PR #942, opened 2017-06-05
Add timestamps to Documents

emmabeynon:
Documents should have timestamps so we know when they were created or updated.
We add timestamps and populate them primarily using Edition data, however there
are some cases where a Document has no editions, in which case we use Event
data instead. This is only possible for documents with an `en` locale, so for
non-en locales we use data from S3.

https://trello.com/c/XloUGpgm/953-store-timestamps-on-documents-(1)

### DIFF
--- a/db/migrate/20170602111120_add_timestamps_to_documents.rb
+++ b/db/migrate/20170602111120_add_timestamps_to_documents.rb
@@ -1,0 +1,122 @@
+class AddTimestampsToDocuments < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    add_timestamps(:documents, null: true) unless column_exists?(:documents, :created_at)
+    puts "> Timestamp fields added."
+    puts "> Updating multiple-locale documents without editions..."
+
+    update_multiple_locale_documents_without_editions
+    puts "> Multiple-locale documents without editions updated."
+    puts "> Updating remaining documents..."
+
+    remaining_documents = Document.where(created_at: nil).count
+
+    Document.where(created_at: nil).find_each.with_index do |document, index|
+      editions = Edition.where(document_id: document.id).order(:user_facing_version)
+      if editions.exists?
+        update_timestamps_from_editions(document, editions)
+      else
+        update_timestamps_from_events(document)
+      end
+      puts "> #{index + 1}/#{remaining_documents} documents updated." if (index + 1) % 100 == 0
+    end
+
+    puts "> All remaining documents updated."
+    puts "> Setting NOT NULL constraint on created_at and updated_at..."
+    change_column_null :documents, :created_at, false
+    change_column_null :documents, :updated_at, false
+  end
+
+  def down
+    remove_column :documents, :created_at
+    remove_column :documents, :updated_at
+  end
+
+private
+
+  MULTIPLE_LOCALE_DOCUMENTS_WITHOUT_EDITIONS = [
+    {
+      content_id: "f091d705-8221-40c4-bc5d-2914c2d0763f",
+      locale: "en",
+      created_at: "Sun, 22 Jan 2017 10:46:37 UTC +00:00",
+      updated_at: "Sun, 22 Jan 2017 10:55:43 UTC +00:00"
+    },
+    {
+      content_id: "f091d705-8221-40c4-bc5d-2914c2d0763f",
+      locale: "ar",
+      created_at: "Sun, 22 Jan 2017 10:50:19 UTC +00:00",
+      updated_at: "Sun, 22 Jan 2017 10:55:43 UTC +00:00"
+    },
+    {
+      content_id: "0c0640e1-83e5-49a7-9da6-7ec3dfdf9936",
+      locale: "es",
+      created_at: "Mon, 13 Mar 2017 12:42:32 UTC +00:00",
+      updated_at: "Mon, 13 Mar 2017 14:29:05 UTC +00:00"
+    },
+    {
+      content_id: "ca74cbbc-7c20-4e28-a3e0-b003490a162d",
+      locale: "en",
+      created_at: "Wed, 12 Apr 2017 05:51:27 UTC +00:00",
+      updated_at: "Wed, 12 Apr 2017 08:12:07 UTC +00:00"
+    },
+    {
+      content_id: "ca74cbbc-7c20-4e28-a3e0-b003490a162d",
+      locale: "ar",
+      created_at: "Wed, 12 Apr 2017 07:19:55 UTC +00:00",
+      updated_at: "Wed, 12 Apr 2017 08:12:07 UTC +00:00"
+    },
+    {
+      content_id: "5f525430-7631-11e4-a3cb-005056011aef",
+      locale: "ar",
+      created_at: "Thu, 13 Apr 2017 17:13:09 UTC +00:00",
+      updated_at: "Thu, 13 Apr 2017 17:13:20 UTC +00:00"
+    },
+    {
+      content_id: "602c5c4a-7631-11e4-a3cb-005056011aef",
+      locale: "cy",
+      created_at: "Wed, 22 Feb 2017 16:07:01 UTC +00:00",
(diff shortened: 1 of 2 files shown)


cbaines:
> are some cases where a Document has no editions, in which case we use Event
> data instead. This is only possible for documents with an en locale, so for
> non-en locales we use data from S3.

I'd be interested to know why it's only possible for documents with an en locale, and how Amazon S3 fits in?

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,108 @@
...
+    change_column_null :documents, :created_at, false
+    change_column_null :documents, :updated_at, false
+  end
+
+private
+
+  NON_EN_DOCUMENTS_WITHOUT_EDITIONS = [
+    {
+      content_id: "f091d705-8221-40c4-bc5d-2914c2d0763f",
+      locale: "en",
```
cbaines: It's a bit contradictory that the locale here is "en" but the name of the array would suggest otherwise?

emmabeynon: Ah yes, my mistake this should be documents with multiple locales, not just non-EN.  Will update the naming 👍

emmabeynon:
> are some cases where a Document has no editions, in which case we use Event
> data instead. This is only possible for documents with an en locale, so for
> non-en locales we use data from S3.
> I'd be interested to know why it's only possible for documents with an en locale, and how Amazon S3 fits in?

@kevindew has more knowledge about this than I do

kevindew:
@cbaines When we have a document without any editions we can use the event table to pull out the timestamps. The payload field of these are archived to S3 after 30 (if I remember correctly) days but content_id and timestamps remain in the Event table. So if we have only one locale of the document we can pull the created_at & updated_at just from the last timestamps in Event table. Whereas if we have multiple locales it's trickier.

@emmabeynon Does this take long to run? I'm expecting it fully locks up the Document table while it runs. I imagine with a couple of modifications (conditional on whether table has columns) it could run idempotently and have `disable_ddl_transaction!` which should improve the locking.

cbaines:
> @cbaines When we have a document without any editions we can use the event table to pull out the timestamps. The payload field of these are archived to S3 after 30 (if I remember correctly) days but content_id and timestamps remain in the Event table. So if we have only one locale of the document we can pull the created_at & updated_at just from the last timestamps in Event table. Whereas if we have multiple locales it's trickier.

Ah, ok, that makes sense. Like the variable name, you might want to update the commit message to talk about multiple locales rather than en locales @emmabeynon

emmabeynon:
> @emmabeynon Does this take long to run? I'm expecting it fully locks up the Document table while it runs. I imagine with a couple of modifications (conditional on whether table has columns) it could run idempotently and have disable_ddl_transaction! which should improve the locking.

It took about 1.75 hours locally.  I've added in `disable_ddl_transaction!`, and will test how long it takes on integration.

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,110 @@
...
+    puts "> Timestamp fields added."
+    puts "> Updating multiple-locale documents without editions..."
+
+    update_non_en_documents_without_editions
+    puts "> Multiple-locale documents without editions updated."
+    puts "> Updating remaining documents..."
+
+    Document.where(created_at: nil).find_each do |document|
+      editions = Edition.where(document_id: document.id).order(:user_facing_version)
+      if editions.present?
```
kevindew: if you change this to `.exists?` it might run a bit faster - `present?` http://www.ombulabs.com/blog/benchmark/performance/rails/present-vs-any-vs-exists.html

kevindew:
@emmabeynon Sure it'll probably take just as long (maybe longer), the difference will be that Publishing API won't get locked while it's running.

You'll need to add an `if` on line 7 otherwise if this migration fails and you retry it it still won't work.

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def update_timestamps_from_editions(document, editions)
+    created_at = editions.first.created_at
+    updated_at = editions.last.created_at
+    update_timestamps(document, created_at, updated_at)
+  end
+
+  def update_timestamps(document, created_at, updated_at)
+    document.update_attributes(created_at: created_at, updated_at: updated_at)
```
kevindew: We might want to try here having this be 

```ruby
Document.transaction do
  document.reload
  document.created_at ||= created_at
  document.updated_at ||= updated_at
  document.save
end
```

which will save it overwriting any values set during this (hopefully that doesn't make it way slower)

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,110 @@
...
+      created_at: "Tue, 28 Mar 2017 06:21:57 UTC +00:00",
+      updated_at: "Wed, 29 Mar 2017 06:38:39 UTC +00:00"
+    }
+  ]
+
+  def update_multiple_locale_documents_without_editions
+    MULTIPLE_LOCALE_DOCUMENTS_WITHOUT_EDITIONS.each do |document|
+      Document
+        .find_by(content_id: document[:content_id], locale: document[:locale])
+        .update_attributes(created_at: document[:created_at], updated_at: document[:updated_at])
```
kevindew: Might be good for this to call the update_timestamps method so everything goes through the same.

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,110 @@
...
+    puts "> Multiple-locale documents without editions updated."
+    puts "> Updating remaining documents..."
+
+    Document.where(created_at: nil).find_each do |document|
+      editions = Edition.where(document_id: document.id).order(:user_facing_version)
+      if editions.exists?
+        update_timestamps_from_editions(document, editions)
+      else
+        update_timestamps_from_events(document)
+      end
```
kevindew: Might be an idea to `puts` something about the progress here - could do a count first and then use `find_each.with_index` to have the number? - just for a bit more feedback while it runs.

emmabeynon: I originally used `puts '.'` but canned it as it the output was so long! But I agree that feedback is useful, and a count seems more useful than a full stop :)

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,117 @@
...
+    remaining_documents = Document.where(created_at: nil).count
+
+    Document.where(created_at: nil).find_each.with_index do |document, index|
+      editions = Edition.where(document_id: document.id).order(:user_facing_version)
+      if editions.exists?
+        update_timestamps_from_editions(document, editions)
+      else
+        update_timestamps_from_events(document)
+      end
+      puts "> #{index += 1}/#{remaining_documents} documents updated."
```
kevindew: 👍  - You probably want this to just be `index + 1` rather than mutating it. You could also add an `if (index + 1) % 100 == 0` so it only outputs every 100 records (or similar)

db/migrate/20170602111120_add_timestamps_to_documents.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    puts "> All remaining documents updated."
+    puts "> Setting NOT NULL constraint on created_at and updated_at..."
+    change_column_null :documents, :created_at, false
+    change_column_null :documents, :updated_at, false
+  end
```
kevindew: should we have a `down` where it sets change_column back and removes the timestamps? could be good for retrying this.

emmabeynon: Do you think the timestamps should be set to nil, or just remove the columns entirely?

kevindew: Remove columns entirely 👍

PR #943, opened 2017-06-12
Track document owner

steventux:
https://trello.com/c/u2ty6FdR/939-3-keep-track-of-redirects-created-automatically

Adds an optional `Document#owner` for redirects. This is an association with the `Document` of the edition which is being redirected. Because redirects are created with a new Document and content id, this will help us track where redirects originated.

### DIFF
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -1,13 +1,20 @@
 module Commands
   class BaseCommand
-    attr_reader :callbacks
+    attr_reader :callbacks, :options
 
-    def self.call(payload, downstream: true, nested: false, callbacks: [])
+    def self.call(payload, downstream: true, nested: false, callbacks: [], **options)
       logger.debug "#{self} called with payload:\n#{payload}"
 
       response = EventLogger.log_command(self, payload) do |event|
         PublishingAPI.service(:statsd).time(self.name.gsub(/:+/, '.')) do
-          new(payload, event: event, downstream: downstream, callbacks: callbacks, nested: nested).call
+          new(
+            payload,
+            event: event,
+            downstream: downstream,
+            callbacks: callbacks,
+            nested: nested,
+            **options
+          ).call
         end
       end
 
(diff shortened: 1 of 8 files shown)


app/helpers/redirect_helper.rb
```diff
@@ -63,14 +44,19 @@ def previous_base_path
...
+        {
+          path: route[:path],
+          type: route[:type],
+          destination: route[:path].gsub(old_base_path, new_base_path)
+        }
+      end
+    end
+
+    def owner
+      previously_published_item.try(:document) ||
```
kevindew: is this because this could be null? - which makes sense to me. I'm just a bit confused why line 28 doesn't have this problem?

app/commands/base_command.rb
```diff
@@ -1,13 +1,20 @@
 module Commands
   class BaseCommand
-    attr_reader :callbacks
+    attr_reader :callbacks, :options
 
-    def self.call(payload, downstream: true, nested: false, callbacks: [])
+    def self.call(payload, downstream: true, nested: false, callbacks: [], **options)
```
kevindew: 👍

app/commands/base_command.rb
```diff
@@ -18,12 +25,13 @@ def self.call(payload, downstream: true, nested: false, callbacks: [])
       raise_validation_command_error(e)
     end
 
-    def initialize(payload, event:, downstream: true, nested: false, callbacks:)
+    def initialize(payload, event:, downstream: true, nested: false, callbacks:, options:)
```
kevindew: Do we want to double splat here too?

app/helpers/redirect_helper.rb
steventux: `previously_published_item` can be a `PreviouslyPublishedItem::NoPreviousPublishedItem` which doesn't implement a `#document` method.
I don't really understand the need for `NoPreviousPublishedItem`, it feels like it doesn't succinctly answer the question of whether a thing has been previously published.

kevindew: Ah yes that makes sense. Yeah it's the nil object pattern but it feels like there's too much data on it.

PR #944, opened 2017-06-13
Create exporter for a tagging event log

tijmenb:
This creates a rake task that exports a CSV with a "tagging history". It contains all the taxonomy-related tagging events. We use the CSV in content tagger to show the history of how much was tagged to a taxon when, and by whom. We use a CSV rather than an actual API to postpone a decision on what the API should look like.

The tests for the exporter really need some de-duping, but I'm not sure yet which way is best.

https://trello.com/c/GRPCKmbU

### DIFF
--- a/lib/tasks/taxonomy_event_log.rb
+++ b/lib/tasks/taxonomy_event_log.rb
@@ -1,0 +1,69 @@
+class TaxonomyEventLog
+  def export
+    exported = []
+    previous_events = {}
+
+    raw_events.each do |event|
+      previous_taxons = previous_events[event.content_id].to_a
+      current_taxons = event.payload[:links][:taxons]
+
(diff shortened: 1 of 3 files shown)


lib/tasks/taxonomy_event_log.rb
```diff
@@ -0,0 +1,59 @@
...
+    @title_cache[content_id] ||= begin
+      document = Document.find_by(content_id: content_id) || raise("#{content_id} not found")
+      latest_edition = document.editions.last
+      latest_edition ? latest_edition.title : "[Discarded draft]"
+    end
+  end
+
+  def raw_events
+    Event
+      .where('payload IS NOT NULL')
```
thomasleese: I think it's best to use double quotes to stick with `publishing-api` conventions.

PR #945, opened 2017-06-13
Content history prototype

thomasleese:
Implementation of a basic content history prototype based on what the new content API might look like.

You must have the environment variable `CONTENT_API_PROTOTYPE` set to something to test it out.

[Trello Card](https://trello.com/c/01I7Xn05/88-historic-content-store-simple-publishing-api-prototype)

### DIFF
--- a/app/controllers/content_api_prototype/content_items/content_controller.rb
+++ b/app/controllers/content_api_prototype/content_items/content_controller.rb
@@ -31,13 +31,43 @@ module ContentApiPrototype
           .new(edition, draft: edition.draft?)
           .for_content_store(0)
 
-        resolve_text_html(presented_edition)
+        presented_edition = resolve_text_html(presented_edition)
+
+        inject_links(presented_edition, edition)
       end
 
       def resolve_text_html(presented_edition)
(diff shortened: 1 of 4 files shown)


app/controllers/v2/content_api_prototype/content_items/content_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+          presented_edition[:details] = resolver.resolve(presented_edition[:details])
+          presented_edition[:description] = resolver.resolve(presented_edition[:description])
+          presented_edition
+        end
+
+        def inject_links(presented_edition, edition)
+          change_history_presenter = Presenters::ChangeHistoryPresenter.new(edition)
+          change_notes = change_history_presenter.send(:change_notes)
+
+          presented_edition[:details][:change_history] = presented_edition[:details][:change_history].each_with_index.map do |presented_change_note, index|
```
kevindew: Thinking further about this I think you're right to have this as change notes. Maybe we should just have links to previous and next editions and allow you to paginate that way rather than having a full list?

Ideally it feels like there should be an endpoint that exposes a paginated list of the history but I don't think that fits into this prototype

PR #948, opened 2017-06-14
[Do not merge] Downgrade Airbrake gem

leenagupte:
Trello card: https://trello.com/c/wg6F6Gk3

The inconsistent document checker task is no longer sending an error report to Errbit. 
This seems to have started happening when the app was upgraded to Rails 5. As it's known that version 5 of Airbrake is not compatible with Rails 5, the hope is that simply changing the version of Airbrake used to the alphagov fork will fix the problem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,7 @@ gem "arel", "8.0"
 gem "unicorn", "~> 4.9.0"
 gem "logstasher", "0.6.2"
 gem "plek", "~> 1.10"
-gem 'airbrake', '~> 5.5'
-gem 'airbrake-ruby', '1.5'
+gem 'airbrake', github: 'alphagov/airbrake', branch: 'silence-dep-warnings-for-rails-5'
 gem "pg"
 gem 'dalli'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,13 @@
+GIT
+  remote: git://github.com/alphagov/airbrake.git
+  revision: ad9c926a56535c48f95c33824a76e10bc8f91179
+  branch: silence-dep-warnings-for-rails-5
+  specs:
+    airbrake (4.3.8)
+      builder
+      multi_json
+      rails (>= 5.0)
+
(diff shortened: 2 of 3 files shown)


cbaines:
Looking at the Publishing E2E Tests failure, there looks to be a Airbrake configuration issue, maybe https://github.com/alphagov/publishing-api/blob/master/config/initializers/airbrake.rb needs changing.

thomasleese:
Yeah, perhaps it needs to be more like https://github.com/alphagov/manuals-publisher/blob/master/config/initializers/airbrake.rb which is using our Airbrake branch.

kevindew:
> This seems to have started happening when the app was upgraded to Rails 5. As it's known that version 5 of Airbrake is not compatible with Rails 5

I'm not too sure this is accurate. We've been using Rails 5 with Airbrake 5 for [9 months](https://github.com/alphagov/publishing-api/commit/2756fac852529490ec683af515cb5c9159cc0feb) - and we're able to get airbrake notifications

h-lame:
> > This seems to have started happening when the app was upgraded to Rails 5. As it's known that version 5 of Airbrake is not compatible with Rails 5

> I'm not too sure this is accurate. We've been using Rails 5 with Airbrake 5 for 9 months - and we're able to get airbrake notifications

Airbake 5+ does work with Rails 5 and doesn't really work out of the box with Rails 4 (lots of dep warnings rather than anything being broken as far as I can tell).  What it doesn't work 100% with is our errbit install.  It no longer gives you *deploy* notifications, which are fairly useful.  Our branch is an Airbrake 4 fork with all the Rails 5 dep warnings silenced (maybe not all yet, but that's the intention).

leenagupte:
@h-lame Thanks for clarifying. I'm trying to figure out why some error reports from publishing-api and content-store aren't being sent to Errbit. The first step is to see if downgrading Airbrake works.

thomasleese:
Closing since this doesn't solve the problem.

PR #951, opened 2017-06-20
Add rake task to find redirected DFID reports

suzannehamilton:
Add a temporary rake task which finds all published redirects with the DFID base path (`/dfid-research-outputs/`). These were not removed from the search API when the redirects were configured. This commit will allow us to build a list of them in order to remove them from the search API in bulk.

This rake task will be removed once it has been run in production.

https://trello.com/c/ECoFt3Yq/136-dfid-research-output-finder-issues-to-investigate

### DIFF
--- a/lib/tasks/dfid_redirects.rake
+++ b/lib/tasks/dfid_redirects.rake
@@ -1,0 +1,14 @@
+require "csv"
+
+desc "Find DFID research report redirects"
+task dfid_redirects: :environment do
+  dfid_redirects = Edition
+    .where(document_type: "redirect")
+    .where("base_path LIKE :prefix", prefix: "/dfid-research-outputs/%")
+
+  csv_out = CSV.new($stdout)
+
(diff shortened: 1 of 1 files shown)


kevindew:
Are you concerned at all with dfid-research-outputs that are unpublished with type redirect?

suzannehamilton:
🤦‍♀️ I was concentrating on fixing the duplicate search results, but of course there could be deleted reports lurking in the search index too.

I can't find any `unpublished` DFID redirects in my dev DB, but it doesn't hurt to include it in the query in case there are some in production. 👍

kevindew:
Sure - they wouldn't actually appear in this query as they'd not have a document_type of redirect. I checked the db and there isn't actually any though (there is only one unpublished [dfid-research-output](https://www.gov.uk/dfid-research-outputs/the-market-for-training-services-a-demand-experiment-with-bangladeshi-garment-factories) and that's a gone)

PR #954, opened 2017-06-23
Add missing `public_updated_at` dates for specialist-publisher docs

suzannehamilton:
There are around 1146 published specialist-publisher docs with a missing `public_updated_at` date. This prevents them from being republished as a minor update because specialist-publisher requires a `public_updated_at` date to generate a `public_timestamp` in the search index.

This migration populates that field with the best data available, which is the `first_published_at` date. This is populated for all published specialist-publisher documents.

There are long-term plans to simplifiy the document and edition timestamps, so this is just a temporary workaround to allow all specialist-publisher documents to be republished.

https://trello.com/c/ECoFt3Yq/136-dfid-research-output-finder-issues-to-investigate

cc @kevindew

### DIFF
--- a/db/migrate/20170623091950_add_missing_public_updated_timestamps.rb
+++ b/db/migrate/20170623091950_add_missing_public_updated_timestamps.rb
@@ -1,0 +1,17 @@
+class AddMissingPublicUpdatedTimestamps < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    editions_to_update = Edition
+      .where(publishing_app: "specialist-publisher")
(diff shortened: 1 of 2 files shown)


kevindew:
looks good - does it run fast enough that it locking isn't a problem?

suzannehamilton:
Hmm, it takes around 30 seconds. Should I add `disable_ddl_transaction!`? I think it should be safe in this case: the migration is idempotent and if it failed in the middle it wouldn't leave inconsistent data.

kevindew:
Sure - that sounds good

suzannehamilton:
Done. 👍

PR #955, opened 2017-06-26
Add expanded links caching

tijmenb:
This adds caching for the expanded links endpoint. It is unfortunately needed since the `/world` taxon is huge and takes about 20s to build. A more permanent solution is hopefully forthcoming via https://trello.com/c/mOdpAGVj.

### DIFF
--- a/app/controllers/v2/link_sets_controller.rb
+++ b/app/controllers/v2/link_sets_controller.rb
@@ -5,11 +5,15 @@ module V2
     end
 
     def expanded_links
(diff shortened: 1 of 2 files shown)


tijmenb:
Wrote up why this is a bad solution on the Trello: https://trello.com/c/mOdpAGVj/961-improve-performance-of-the-expanded-link-set-endpoint-3#comment-59523f83cd2c7225e9fc9e62

PR #957, opened 2017-06-27
Improve taxonomy events export

tijmenb:
https://trello.com/c/HYg8SgMU

### DIFF
--- a/lib/tasks/taxonomy_event_log.rb
+++ b/lib/tasks/taxonomy_event_log.rb
@@ -7,9 +7,9 @@ class TaxonomyEventLog
       previous_taxons = previous_events[event.content_id].to_a
       current_taxons = event.payload[:links][:taxons]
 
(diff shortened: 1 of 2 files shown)


lib/tasks/taxonomy_event_log.rb
```diff
@@ -7,9 +7,9 @@ def export
       previous_taxons = previous_events[event.content_id].to_a
       current_taxons = event.payload[:links][:taxons]
```
whoojemaflip: Might want to `event.payload.fetch(:links, {})[:taxons]` in case there's no `link` hash

whoojemaflip: now the `raw_event` query has changed

tijmenb: There's always a `links` in `PatchLinksSet`, I think it's part of the validation (and the export works atm).

PR #959, opened 2017-06-28
Use Airbrake.notify_sync instead of Airbrake.notify

thomasleese:
When running in a rake task, Airbrake.notify doesn't properly
finish triggering before the rake task ends and therefore the
message never makes it to Airbrake.

[Trello Card](https://trello.com/c/wg6F6Gk3/947-3-investigate-why-the-inconsistent-checker-is-not-working-all-the-time)


leenagupte:
That was it! Amazing!

PR #961, opened 2017-07-03
Pre-cache expanded links

steventux:
Part of https://trello.com/c/mOdpAGVj/961-improve-performance-of-the-expanded-link-set-endpoint-3

Paired with @emmabeynon on this work.

Pre-cache expanded links when writing links from the various Publishing API commands.
Uses Rails.cache to store expanded links, no decisions have yet been made about underlying cache implementation though it would be preferable to use a shared memcached for best performance.

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -29,7 +29,9 @@ module Commands
 
         after_transaction_commit do
           send_downstream(orphaned_content_ids)
+          ExpandedLinkSetCacheWorker.perform_async(content_id)
         end
+
 
         Action.create_patch_link_set_action(link_set, event)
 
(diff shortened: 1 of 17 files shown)


kevindew:
Haven't dug into this in too much detail so could be very wrong. But I wonder if this approach is flawed. 

I'm thinking that this caches links each time the links of a content id changes, but it does not update cache when one of links changes. 

eg. A links to B, I update A, A links are cached, I update B, A's cache is not cleared

steventux:
@kevindew Are you saying pre-caching expanded links is inherently flawed or that we've missed some endpoints or interactions which would need to refresh the cache?

steventux:
Closing for now, a couple of things to investigate from speaking with Kevin might inform how we cache.

PR #962, opened 2017-07-03
De dupe more specialist publisher change notes

steventux:
This fell out of this story: https://trello.com/c/NbATcan2/956-de-duplicate-specialist-publisher-embedded-change-history-2

There are approx. 570 change notes which are duplicates but
do not have a corresponding edition id, their duplicates do have
this id so delete them and represent the affected editions downstream.

### DIFF
--- a/db/migrate/20170703153415_dedupe_more_specialist_publisher_change_notes.rb
+++ b/db/migrate/20170703153415_dedupe_more_specialist_publisher_change_notes.rb
@@ -1,0 +1,31 @@
+class DedupeMoreSpecialistPublisherChangeNotes < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    dupe_change_note_ids = ChangeNote.group(:public_timestamp, :note)
+                                     .having("COUNT(*) > 1")
+                                     .pluck("array_agg(id)")
+
+    document_ids = ChangeNote.where(id: dupe_change_note_ids.flatten,
+                                    edition_id: nil)
+                             .pluck(:document_id).flatten.uniq
+
+    # ~ 572 specialist-publisher change notes.
+    ChangeNote.where(id: dupe_change_note_ids.flatten,
+                     edition_id: nil).delete_all
+
+    # Remove Edition#details[:change_history] and represent the Edition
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -129,7 +129,6 @@
     t.index ["link_set_id", "target_content_id"], name: "index_links_on_link_set_id_and_target_content_id"
     t.index ["link_set_id"], name: "index_links_on_link_set_id"
     t.index ["link_type"], name: "index_links_on_link_type"
-    t.index ["target_content_id", "link_type"], name: "index_links_on_target_content_id_and_link_type"
```
thomasleese: I'm not sure this should be here!

db/migrate/20170703153415_dedupe_more_specialist_publisher_change_notes.rb
```diff
@@ -0,0 +1,29 @@
+class DedupeMoreSpecialistPublisherChangeNotes < ActiveRecord::Migration[5.1]
+  def up
+    dupe_change_note_ids = ChangeNote.group(:public_timestamp, :note)
+                                     .having("COUNT(*) > 1")
+                                     .pluck("array_agg(id)")
+
```
thomasleese: Empty line.

db/migrate/20170703153415_dedupe_more_specialist_publisher_change_notes.rb
```diff
@@ -0,0 +1,29 @@
...
+    # Remove Edition#details[:change_history] and represent the Edition
+    # downstream, the history will be reconstructed from ChangeNotes.
+    Edition.where(document_id: document_ids).each do |edition|
+      edition_details = edition.details
+      edition_details.delete(:change_history)
+      edition.update!(details: edition_details)
+
+      if Rails.env.production?
+        Commands::V2::RepresentDownstream.new.call(edition.content_id)
```
thomasleese: I think we can pass this an array of IDs, it might be quicker doing that than calling it in each iteration.

steventux:
@thomasleese updated (the empty line comment is still showing up I guess this is because it was made against the one above the redundant line)

steventux:
good catch! done

PR #963, opened 2017-07-04
Allow lookup by base path filter criteria to be specified

dwhenry:
In order to migrate content out of router-data we need 
to be able to find the content ID for unpublished
redirects. 

This change allow us to remove the filtering  required 
by mainstream applications for unpublished.type and 
document_type fields.

We have inverted the existing logic around unpublishing_type
filters based on the fact that content must be `published`
or `unpublished` and only `unpublished` content will
have an unpublishing record.

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -5,13 +5,18 @@ class LookupsController < ApplicationController
     states = %w(published unpublished)
     base_paths = params.fetch(:base_paths)
 
-    base_paths_and_content_ids = Edition.with_document
-      .left_outer_joins(:unpublishing)
+    scope = Edition.left_outer_joins(:unpublishing)
+
+    # where not in (..) does not return records that where the field is null
+    scope = scope.where(unpublishings: { type: nil }).or(
+      scope.where.not(unpublishings: { type: params.fetch(:exclude_unpublishing_types, %w{vanish redirect gone}) })
+    )
+
+    scope = scope.with_document
       .where(state: states, base_path: base_paths)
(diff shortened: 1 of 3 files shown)


app/controllers/lookups_controller.rb
```diff
@@ -5,13 +5,18 @@ def by_base_path
...
+    scope = Edition.left_outer_joins(:unpublishing)
+
+    # where not in (..) does not return records that where the field is null
+    scope = scope.where(unpublishings: { type: nil }).or(
+      scope.where.not(unpublishings: { type: params.fetch(:exclude_unpublishing_types, %w{vanish redirect gone}) })
+    )
+
+    scope = scope.with_document
       .where(state: states, base_path: base_paths)
-      .where("state = 'published' OR unpublishings.type = 'withdrawal'")
```
cbaines: From just reading this query, I think this would have had also provided content ids for content that was superseded, provided it was unpublished. I think this is undesirable, I guess because it allows there to be multiple content ids that could be considered valid in the case where there are multiple superseded and unpublished editions with different content ids.

 The changes you have made keep this as a possibility, but I think they also open up the possibility of superseded items which haven't been unpublished being selected.

My suggestion would be to add the condition `.where(content_store: 'live')` to the query, as that will ensure that no superseded content is selected. What do you think?

dwhenry: This below line ensures that we don't get `draft` or `superseded` content in the returned results.
```
.where(state: states, base_path: base_paths)
```

cbaines: Ah, great, that looks good :+1:

PR #964, opened 2017-07-06
Restrict by_base_path to 'live' content

cbaines:
Restricting by only editions in the 'published' or 'unpublished'
states is not sufficient at the moment, as there are sometimes
multiple editions with published/unpublished states for a single base
path.

Currently, it's possible to get in to this problomatic situation when
an edition is published, where the previous edition is unpublished,
and the edition being published has a different content id to the
unpublished edition. The content store for the unpublished edition
will be NULL, but it will still be in the unpublished state.

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -13,7 +13,7 @@ class LookupsController < ApplicationController
     )
 
     scope = scope.with_document
-      .where(state: states, base_path: base_paths)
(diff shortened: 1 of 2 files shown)


cbaines:
@thomasleese I've now updated the test data, as the tests were failing due to not having the content_store set.

cbaines:
Thanks for your review @thomasleese :)

PR #966, opened 2017-07-11
Add an index on edition state

thomasleese:
It seems that the various indexes on state with another field can't be used in this case.

Previous:

```
 Seq Scan on editions  (cost=0.00..308447.38 rows=380037 width=541) (actual time=0.039..1555.785 rows=375091 loops=1)
   Filter: ((state)::text = 'published'::text)
   Rows Removed by Filter: 2170939
 Total runtime: 2092.298 ms
```

Now:

```
 Bitmap Heap Scan on editions  (cost=8761.72..304727.41 rows=380037 width=541) (actual time=74.664..1126.715 rows=375091 loops=1)
   Recheck Cond: ((state)::text = 'published'::text)
   Rows Removed by Index Recheck: 921250
   ->  Bitmap Index Scan on editions_state_idx  (cost=0.00..8666.71 rows=380037 width=0) (actual time=72.198..72.198 rows=375091 loops=1)
         Index Cond: ((state)::text = 'published'::text)
 Total runtime: 1493.419 ms
```

### DIFF
--- a/db/migrate/20170711122824_add_index_on_edition_state.rb
+++ b/db/migrate/20170711122824_add_index_on_edition_state.rb
@@ -1,0 +1,7 @@
+class AddIndexOnEditionState < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :editions, :state, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170703153415) do
+ActiveRecord::Schema.define(version: 20170711122824) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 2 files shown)


kevindew:
I notice that the optimised query is still pretty slow (1.5s). I'm just wondering if this is an optimisation that we actually benefit from in a place? and/or if we still need the state and base_path index.

Just being quite protective of them as it's easy to have loads of indexes and not know if we gain from them or not.

thomasleese:
I'm not sure about the index on `state` and `base_path`, but we do gain benefit (although it is small like in the query example above) from this index when running a query like below which was taken from a slow get content request:

```sql
SELECT *, COUNT(*) OVER () as total FROM (SELECT DISTINCT ON(editions.document_id) editions.document_id, documents.content_id as content_id, to_char(public_updated_at, 'YYYY-MM-DD"T"HH24:MI:SS"Z"') as public_updated_at FROM "editions" INNER JOIN "documents" ON "documents"."id" = "editions"."document_id" WHERE "documents"."locale" = 'en' AND "editions"."state" = 'published' ORDER BY editions.document_id, user_facing_version DESC) subquery ORDER BY public_updated_at desc LIMIT 50 OFFSET 0;
```

thomasleese:
Reworking this idea.

PR #968, opened 2017-07-12
Improve get content performance

thomasleese:
Before:

```
SELECT  *, COUNT(*) OVER () as total FROM (SELECT DISTINCT ON(editions.document_id) editions.document_id, documents.content_id as content_id, to_char(public_updated_at, 'YYYY-MM-DD"T"HH24:MI:SS"Z"') as public_updated_at FROM "editions" INNER JOIN "documents" ON "documents"."id" = "editions"."document_id" WHERE "documents"."locale" = 'en' AND "editions"."state" = 'published' ORDER BY editions.document_id, user_facing_version DESC) subquery ORDER BY public_updated_at desc LIMIT 50 OFFSET 0

Took: 2.418973426 seconds
```

After:

```
SELECT  *, 363332 as total FROM (SELECT documents.content_id as content_id, to_char(public_updated_at, 'YYYY-MM-DD"T"HH24:MI:SS"Z"') as public_updated_at, public_updated_at as order_public_updated_at FROM "editions" INNER JOIN "documents" ON "documents"."id" = "editions"."document_id" WHERE "documents"."locale" = 'en' AND "editions"."state" = 'published') subquery ORDER BY order_public_updated_at desc LIMIT 50 OFFSET 0

Took: 0.002072432 seconds
```

Individually the following changes do very little to help on their own, so I provide a justification for each change and how they work together to introduce this speed increase:

## Using raw fields

We already have an index on `public_updated_at` but instead we are ordering on the `to_char` result of converting that field to an ISO8601 date, this meant that the index wasn't getting used. On its own this makes very little difference as the bottleneck came from calculating the total.

## Skip finding the latest version of an edition if possible

Avoids a `DISTINCT ON` query and some ordering. On its own this sped up the query by about 0.9 seconds, with the bottleneck still being calculating the total.

## Estimate the total

We can estimate the total number of rows using the query planner rather than performing a count over the entire set of results which can sometimes be quite large. Here is where the big speed increases come in, but on its own this wouldn't speed up the query as we would be ordering on a `to_char`'d field and still have a `DISTINCT` in the query which would become the new bottlenecks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,8 @@ gem 'airbrake-ruby', '1.5'
 gem "pg"
 gem 'dalli'
 
+gem "pg-eyeballs"
+
 gem "colorize", "~> 0.8"
 
 if ENV["API_DEV"]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -223,6 +223,9 @@ GEM
     parser (2.3.1.4)
       ast (~> 2.2)
     pg (0.19.0)
+    pg-eyeballs (1.1.0)
+      activerecord (>= 4.0, < 6.0)
+      pg
     plek (1.12.0)
     powerpack (0.1.1)
     pry (0.10.4)
@@ -415,6 +418,7 @@ DEPENDENCIES
   pact
   pact_broker-client
   pg
+  pg-eyeballs
   plek (~> 1.10)
   pry
   pry-byebug
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -22,6 +22,12 @@ module Presenters
       SEARCH_FIELDS = %w(title base_path description).freeze
       NESTED_SEARCH_FIELDS = %w(details).freeze
 
+      ORDER_FIELD_MAPPINGS = {
+        public_updated_at: :order_public_updated_at,
+        last_edited_at: :order_last_edited_at,
+        first_published_at: :order_first_published_at,
+      }.freeze
+
       def self.present_many(edition_scope, params = {})
         new(edition_scope, params).present_many
       end
@@ -59,9 +65,16 @@ module Presenters
         @results ||= execute_query(query)
       end
 
+      def mapped_order
+        Hash[order.map do |key, value|
+          next [ORDER_FIELD_MAPPINGS[key], value] if ORDER_FIELD_MAPPINGS.include?(key)
+          [key, value]
+        end]
+      end
+
       def query
         ordering_query = Edition.select("*, COUNT(*) OVER () as total").from(fetch_items_query)
-        ordering_query = ordering_query.order(order.to_a.join(" ")) if order
+        ordering_query = ordering_query.order(mapped_order.to_a.join(" ")) if order
         ordering_query = ordering_query.limit(limit) if limit
         ordering_query = ordering_query.offset(offset) if offset
         ordering_query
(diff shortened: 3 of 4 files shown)


kevindew:
Great numbers 👏  🎉 

What do we know about the accuracy of the estimate? Using an estimate makes me nervous as the API response states that value as a fact - I imagine you share similar feelings on it. 

The total value is used for Specialist Publisher for pagination and we can have a huge range of queries (free text search for instance) so it strikes me that for the things closest to a user it's likely to be inaccurate.

Do you know if this is all still slow when we have the document_type in by the way? another thought is that if you're looking through the entirety of Pub API (which is what I think content tools might be doing) is that paginating through that might be a nightmare due to the number of items probably changing while you work through it.

While we're considering this sort of stuff and the other option of this card being a dedicated endpoint, what would we do about this for a different endpoint? not return a total and just links to prev & next? use a cursor to paginate? Any visibility on what other APIs do with big sets like this.

Oh and another thought. You remember how it's fast when we have a LIMIT 50. If we have a huge offset (eg were getting the last page of all editions) is it super slow again? 

Sorry for big rambling comment, the implications of this are making me wonder how we should think through other endpoints which have a lot of data

thomasleese:
> What do we know about the accuracy of the estimate? Using an estimate makes me nervous as the API response states that value as a fact - I imagine you share similar feelings on it.

I ran through a few queries during development to see what this was like. I couldn't find a query that returned a lower estimate than actual value so there shouldn't be an issue about losing data. In terms of accuracy, I was surprised at how good the estimates were, here are some examples:

```
publishing_api_development=> EXPLAIN SELECT * FROM editions WHERE title LIKE '%Tax%';
                            QUERY PLAN                             
-------------------------------------------------------------------
 Seq Scan on editions  (cost=0.00..308447.38 rows=51699 width=541)
   Filter: ((title)::text ~~ '%Tax%'::text)
(2 rows)

publishing_api_development=> SELECT COUNT(*) FROM editions WHERE title LIKE '%Tax%';
 count 
-------
 47512
(1 row)

publishing_api_development=> EXPLAIN SELECT * FROM editions WHERE title LIKE '%abc%';
                           QUERY PLAN                            
-----------------------------------------------------------------
 Seq Scan on editions  (cost=0.00..308447.38 rows=242 width=541)
   Filter: ((title)::text ~~ '%abc%'::text)
(2 rows)

publishing_api_development=> SELECT COUNT(*) FROM editions WHERE title LIKE '%abc%';
 count 
-------
    97
(1 row)
```

> Do you know if this is all still slow when we have the document_type in by the way? another thought is that if you're looking through the entirety of Pub API (which is what I think content tools might be doing) is that paginating through that might be a nightmare due to the number of items probably changing while you work through it.

Hmm, that's true about the items changing. Limiting by `document_type` definitely makes the original query faster (fewer rows to hold in memory), but doesn't have any difference on kind of queries generated using this PR.

> While we're considering this sort of stuff and the other option of this card being a dedicated endpoint, what would we do about this for a different endpoint? not return a total and just links to prev & next? use a cursor to paginate? Any visibility on what other APIs do with big sets like this.

Yeah I did have a look at how some other APIs work. Atlasssian didn't include any total in the result, whereas GitHub didn't include a total but it was able to include a link to the first and last page, as well as next and previous. Twitter includes the total in pagination results, whereas Facebook only provides a next and previous link.

> Oh and another thought. You remember how it's fast when we have a LIMIT 50. If we have a huge offset (eg were getting the last page of all editions) is it super slow again?

Yeah, sadly it does get progressively slower, but still faster than before due to the other improvements in this PR. The final query took just under 1.5 seconds when I tested it this morning.

> Sorry for big rambling comment, the implications of this are making me wonder how we should think through other endpoints which have a lot of data

Yeah, definitely something worth discussing and I wasn't even sure this PR would make it into the current endpoint as it is!

thomasleese:
Closing this since using an estimate is not good enough.

PR #971, opened 2017-07-17
Notify downstream services on unpublish

thomasleese:
[Trello Card](https://trello.com/c/5a7bQyHI/963-notify-downstream-services-search-indexing-when-documents-are-unpublished-3)

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -104,7 +104,7 @@ module Commands
           queue,
           content_id: content_id,
           locale: locale,
-          message_queue_update_type: "links",
+          message_queue_event_type: "links",
           payload_version: event.id,
           orphaned_content_ids: orphaned_content_ids,
         )
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -219,7 +219,7 @@ module Commands
 
       def live_worker_params
         {
-          message_queue_update_type: update_type,
+          message_queue_event_type: update_type,
           update_dependencies: update_dependencies?,
           orphaned_content_ids: orphaned_content_ids,
         }.merge(worker_params)
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -30,6 +30,7 @@ module Commands
         set_public_updated_at
         set_first_published_at
         set_publishing_request_id
+        set_update_type
         edition.publish
         remove_access_limit
         create_publish_action
@@ -175,6 +176,11 @@ module Commands
         edition.update_attributes!(
           publishing_request_id: GdsApi::GovukHeaders.headers[:govuk_request_id]
         )
+      end
+
+      def set_update_type
+        return if edition.update_type
+        edition.update_attributes!(update_type: update_type)
       end
 
       def publish_redirect(previous_base_path, locale)
--- a/app/commands/v2/represent_downstream.rb
+++ b/app/commands/v2/represent_downstream.rb
@@ -48,7 +48,7 @@ module Commands
             content_id: content_id,
             locale: locale,
             payload_version: event.id,
-            message_queue_update_type: "links",
+            message_queue_event_type: "links",
             update_dependencies: false,
           )
         end
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -40,10 +40,10 @@ class DownstreamPayload
     end
   end
 
-  def message_queue_payload(update_type)
+  def message_queue_payload
     Presenters::EditionPresenter.new(
       edition, draft: draft
-    ).for_message_queue(update_type || edition.update_type)
+    ).for_message_queue
   end
 
 private
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -31,19 +31,23 @@ class DownstreamPayload
   end
 
   def content_store_payload
-    return content_payload unless unpublished?
+    return content_payload_for_content_store unless unpublished?
 
     case unpublishing.type
-    when "redirect" then redirect_payload
+    when "redirect" then redirect_payload_for_content_store
     when "gone" then gone_payload
-    else content_payload
+    else content_payload_for_content_store
     end
   end
 
   def message_queue_payload
-    Presenters::EditionPresenter.new(
-      edition, draft: draft
-    ).for_message_queue
+    return content_payload_for_message_queue unless unpublished?
+
+    case unpublishing.type
+    when "redirect" then redirect_payload_for_message_queue
+    #when "gone" then gone_payload
+  else content_payload_for_message_queue
+    end
   end
 
 private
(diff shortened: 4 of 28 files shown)


thomasleese:
Content schema change in https://github.com/alphagov/govuk-content-schemas/pull/638.

app/commands/v2/represent_downstream.rb
```diff
@@ -48,7 +48,7 @@ def downstream_live(content_id, locale)
             content_id: content_id,
             locale: locale,
             payload_version: event.id,
-            message_queue_update_type: "links",
+            message_queue_event_type: "links",
```
kevindew: maybe we should rename this - I think "links" was only used as a hack as we didn't have an alternative option whereas it's a type of republish

app/commands/v2/patch_link_set.rb
```diff
@@ -104,7 +104,7 @@ def downstream_live(content_id, locale, orphaned_content_ids)
           queue,
           content_id: content_id,
           locale: locale,
-          message_queue_update_type: "links",
+          message_queue_event_type: "links",
```
kevindew: I like this 👍

app/commands/v2/publish.rb
```diff
@@ -177,6 +178,10 @@ def set_publishing_request_id
         )
       end
 
+      def set_update_type
+        edition.update_attributes!(update_type: update_type)
```
kevindew: I'm not sure if this is safe, we could lose minor/major ones to republish. Admittedly it's but a drop in the bucket of our problems on update_type. Nethertheless I'd probably suggest we work on this as part of the [epic](https://trello.com/c/K164C5u8/5-ensure-content-entering-publishing-api-has-an-update-type-of-major-minor)

app/downstream_payload.rb
```diff
@@ -52,27 +44,40 @@ def unpublishing
...
+  end
+
+  def message_queue_presenter
+    return content_presenter unless unpublished?
+
+    case unpublishing.type
+    when "redirect" then redirect_presenter
+    when "gone" then gone_presenter
+    when "vanish" then vanish_presenter
+    else content_presenter
```
kevindew: will Search be able to determine if something is withdrawn or not from this?

app/presenters/gone_presenter.rb
```diff
@@ -1,5 +1,38 @@
...
+      locale: edition.document.locale,
+    )
+  end
+
+  def for_content_store(payload_version)
+    present.merge(payload_version: payload_version)
+  end
+
+  def for_message_queue
+    present.merge(content_id: content_id, locale: locale)
```
kevindew: will these lose the `govuk_request_id` field?

spec/requests/unpublishing_spec.rb
```diff
@@ -154,10 +155,18 @@
         end
       end
 
-      it "does not send to the message queue" do
+      it "does sends to the message queue" do
```
kevindew: s/sends/send

app/downstream_payload.rb
thomasleese: Yes, it will fall through to the content presenter, which is fine for withdrawn since the payload format is the same for a published item.

app/commands/v2/represent_downstream.rb
thomasleese: I just wanted to avoid making a breaking change in this PR, I'll speak to Search and see what they think.

thomasleese: We've spoken to the Search team and they would like to leave this as it is for the moment.

app/downstream_payload.rb
thomasleese: The routing key will also be `*.unpublish`.

app/commands/v2/represent_downstream.rb
kevindew: cool - can we get a card in the backlog for it? (with any information as to what would block the change)

app/downstream_payload.rb
```diff
@@ -52,27 +44,40 @@ def unpublishing
...
-  def redirect_payload
-    RedirectPresenter.present(
-      base_path: base_path,
-      publishing_app: edition.publishing_app,
-      public_updated_at: unpublishing.created_at,
-      redirects: unpublishing.redirects,
-    ).merge(payload_version: payload_version)
+  def vanish_presenter
+    VanishPresenter.from_edition(edition)
   end
```
kevindew: Just noticed something here that I think may have slipped in a while ago (may be more appropriate to do in a separate PR if no other changes are needed here)

In [0b736c9](https://github.com/alphagov/publishing-api/commit/0b736c9717e6015363630b6cf131754fa5ff243e) we dropped memoizing the presenter, which (if I understand correctly) probably means we're doing link expansion twice on a publish unnecessarily. I think some memoizing of the presenters would stop this.

app/presenters/redirect_presenter.rb
```diff
@@ -1,5 +1,46 @@
...
+  end
+
+  def for_content_store(payload_version)
+    present.merge(payload_version: payload_version)
+  end
+
+  def for_message_queue
+    present.merge(
+      content_id: content_id,
+      locale: locale,
```
kevindew: This strikes me as a little strange, since there is a discrepancy between what we send to the content store and what we send to the message queue.

Does search need this? and if they do is there any scope for them to not. I'm just a little concerned as the content we send with this is no longer the content (In my mind it's effectively replaced rather than existing in a different form) and not sure this propagates a different view.

thomasleese: Hmm, the reason I did this was before the following which was in the card:

> Use the content store payloads for `redirect`, `gone`, but add `content_id` as a field.
> Introduce new payloads for `vanish`, `substitute`, containing `base_path` and `content_id`.

I don't know for sure if the search team actually need this though, so I will clarify.

kevindew: cool thanks - do you also know if they're bothered about things not having a base_path?

app/downstream_payload.rb
thomasleese: This is a good point, but I think it should be for a separate card. I'll create one.

app/helpers/redirect_helper.rb
```diff
@@ -12,12 +12,13 @@ def initialize(previously_published_item, previously_drafted_item, payload, call
...
 
-      redirect_payload = RedirectPresenter.present(
+      redirect_payload = RedirectPresenter.new(
         base_path: previous_base_path,
         public_updated_at: Time.zone.now,
         redirects: redirects_for(previous_routes, previous_base_path, payload[:base_path]),
-        publishing_app: payload[:publishing_app]
-      ).merge(content_id: SecureRandom.uuid)
+        publishing_app: payload[:publishing_app],
+        content_id: SecureRandom.uuid,
```
kevindew: Since pulling content_id out of the presenter should this still be here? or it should be merged in like previously?

app/presenters/gone_presenter.rb
```diff
@@ -1,5 +1,36 @@
...
+  def for_message_queue
+    present.merge(
+      govuk_request_id: GdsApi::GovukHeaders.headers[:govuk_request_id],
+    )
+  end
+
+private
+
+  attr_reader :base_path, :publishing_app, :alternative_path, :explanation,
+    :content_id, :locale
```
kevindew: I'm guessing we don't need content_id and locale here anymore?

app/helpers/redirect_helper.rb
thomasleese: Hmm, I think I prefer it like this because we'd have to make the `present` method public and I like the distinction of the different `for_` methods.

app/presenters/gone_presenter.rb
thomasleese: Yep!

app/helpers/redirect_helper.rb
kevindew: Sure, that makes sense. Do you need to do something with the locale in RedirectPresenter?

thomasleese: I guess, potentially, this is a wider bug, because `locale` was never sent before. Do you reckon it does need to be included?

kevindew: It'll just be falling back to en so isn't a problem.

However by introducing it into the input of RedirectPresenter and not including it in the output seems a bit prone to a confusing in scenario. I.e. you can now create a "fr" RedirectPresenter and if you pass it to redirect helper it will turn to "en"

Not that redirects need a locale ¯\_(ツ)_/¯

thomasleese: Yeah okay, I think I'll add it back in.

thomasleese: Hmm actually looks like `locale` isn't always available and it breaks the schemas... I reckon if we wanted to fix this it should be a separate card.

kevindew: maybe add content_id as an argument to for_redirect_helper and then the class doesn't need to know about content_id or locale?

thomasleese: Yep, I agree, that's done now!

PR #973, opened 2017-07-18
Implement new /editions endpoint

thomasleese:
[Trello Card](https://trello.com/c/iAdftXss/962-help-content-tools-get-content-out-of-publishing-api-faster-more-get-content-woes-3)


doc/api.md
```diff
@@ -537,6 +538,33 @@ which is `details.internal_name` and falls back to `title`.
...
+  - Used to restrict documents to a given locale.
+- `key` *(optional, defaults to `updated_at,id`)
+  - The pagination key to use when looking through results.
+- `page` *(optional)*
+  - The page of the pagination key to start on.
+- `count` *(optional, default: 100)*
+  - The number of results to be shown on a given page.
+- `publishing_app` *(optional)*
+  - Used to restrict editions to those for a given publishing app.
+- `states` *(optional)*
```
kevindew: We should probably specify that this is an array (I guess it got missed on get content too)

doc/api.md
```diff
@@ -537,6 +538,33 @@ which is `details.internal_name` and falls back to `title`.
...
+parameters.
+
+### Query string parameters
+
+- `fields[]` *(optional)*
+  - Accepts an array of: "analytics_identifier", "base_path",
+    "content_id", "description", "document_type", "locale",
+    "public_updated_at", "schema_name", "title"
+  - Determines which fields will be returned in the response, if omitted all
+    fields will be returned.
```
kevindew: I think we probably need details as an option? the docs for get content seem inconsistent with that actually returns (which is, of course, everything ever)

app/presenters/keyset_pagination_presenter.rb
```diff
@@ -0,0 +1,40 @@
+module Presenters
+  class KeysetPaginationPresenter
+    def initialize(pagination, request_url)
```
kevindew: might be slightly clearer if this was called pagination_query

app/presenters/keyset_pagination_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+        results: results
+      }
+    end
+
+  private
+
+    attr_reader :results, :pagination, :request_url, :present_record_filter
+
+    def links
+      [next_link]
```
kevindew: Could we have self and previous links? 

fun challenge here: is can we know if there are actually next or previous items?

app/queries/get_editions.rb
```diff
@@ -0,0 +1,82 @@
...
+    def validate_fields!
+      return unless fields
+
+      invalid_fields = fields - permitted_fields
+      return unless invalid_fields.any?
+
+      raise CommandError.new(code: 400, error_details: {
+        error: {
+          code: 400,
+          message: "Invalid column name(s): #{invalid_fields.to_sentence}"
```
kevindew: is it intentional that we have code: 400 twice here?

doc/api.md
```diff
@@ -537,6 +538,33 @@ which is `details.internal_name` and falls back to `title`.
...
+- `locale` *(optional)*
+  - Accepts: An available locale from the [Rails I18n gem][i18n-gem]
+  - Used to restrict documents to a given locale.
+- `key` *(optional, defaults to `updated_at,id`)
+  - The pagination key to use when looking through results.
+- `page` *(optional)*
+  - The page of the pagination key to start on.
+- `count` *(optional, default: 100)*
```
kevindew: could we change this to be `per_page` since we have a [precedent](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#user-content-query-string-parameters)

doc/api.md
```diff
@@ -537,6 +538,33 @@ which is `details.internal_name` and falls back to `title`.
...
+- `fields[]` *(optional)*
+  - Accepts an array of: "analytics_identifier", "base_path",
+    "content_id", "description", "document_type", "locale",
+    "public_updated_at", "schema_name", "title"
+  - Determines which fields will be returned in the response, if omitted all
+    fields will be returned.
+- `locale` *(optional)*
+  - Accepts: An available locale from the [Rails I18n gem][i18n-gem]
+  - Used to restrict documents to a given locale.
+- `key` *(optional, defaults to `updated_at,id`)
```
kevindew: Can we change this to order and use it in the same way as [get content](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#user-content-query-string-parameters)

I think the id can be implicit and not something that needs be part of a users concern.

doc/api.md
```diff
@@ -537,6 +538,33 @@ which is `details.internal_name` and falls back to `title`.
...
+  - The pagination key to use when looking through results.
+- `page` *(optional)*
```
kevindew: I don't think we should use page as it's used elsewhere with an integer. I think we might be best to go back to "before" and "after"

spec/controllers/v2/editions_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        end
+      end
+    end
+
+    context "all editions" do
+      it "returns the correct list for the first page" do
+        get :index
+        expect(parsed_response["results"].count).to eq(100)
+        expect(parsed_response["links"]).to eq([
+          { "href" => "http://test.host/v2/editions?page=2017-01-01T09%3A00%3A00Z%2C100", "rel" => "next" },
```
kevindew: Could be good to have a urlencode function on an unencoded string so it's easier for the human brain to parse.

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,69 @@
...
+    def call
+      paginated_query = query.order(order_clause)
+      paginated_query = paginated_query.where(where_clause, *page) if page
+      paginated_query.limit(count)
+    end
+
+    def key_for_record(record)
+      values = key.keys.map do |k|
+        value = record[k.to_s]
+        if value.instance_of?(ActiveSupport::TimeWithZone)
```
kevindew: Might be slightly simpler if we have 
```
next value.iso8601 if value.respond_to?(:iso8601)
value.to_s
```

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,69 @@
+module Queries
+  class KeysetPagination
+    attr_reader :query, :order, :key, :count, :page
+
+    def initialize(query, key: nil, order: nil, count: nil, page:)
+      @query = query
+      @key = key || DEFAULT_KEY
+      @order = order || DEFAULT_ORDER
+      @count = (count || DEFAULT_COUNT).to_i
+      @page = page
```
kevindew: I wonder if these need to be constants? 

In trying to understand this I've struggled with what we might expect these should be in terms of type format, and that might help.

As this is meant to be something that can be reused might be good to add some param documentation to this class too.

app/controllers/v2/editions_controller.rb
```diff
@@ -0,0 +1,69 @@
+module V2
+  class EditionsController < ApplicationController
+    def index
+      query = Queries::KeysetPagination.new(
+        Queries::GetEditions.new(
+          fields: query_params[:fields],
+          filters: filters,
+        ).call,
+        **pagination_params
```
thomasleese: Made this method use a new class to parse the parameters.

app/controllers/v2/editions_controller.rb
```diff
@@ -0,0 +1,64 @@
+module V2
+  class EditionsController < ApplicationController
+    def index
+      query = Queries::KeysetPagination.new(
+        Queries::GetEditions.new(
+          fields: query_params[:fields],
+          filters: filters,
+        ),
+        **pagination_params
+      )
```
kevindew: Were we swapping out this Queries::GetEditions class? I thought we were going to use a different class as we'd be using a pluck?

thomasleese: Good point, I've pushed some commits for this now.

Also I've implemented the feature to only show other pagination links when they are results available!

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,130 @@
...
+        raise "Number of previous values does not match the number of fields."
+      end
+    end
+
+    def call
+      results
+    end
+
+    def next_before_key
+      key_for_record(results.first)
```
kevindew: will this be running an additional query?

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,130 @@
...
+  class KeysetPagination
+    attr_reader :client, :order, :key, :count, :previous
+
+    # Initialises the keyset pagination class.
+    # Params:
+    # +client+:: the pagination client to query over
+    # +key+:: +Hash+ a hash containing the pagination key, mapped from presented name to internal name, i.e. { id: "editions.id" }
+    # +order+:: +Symbol+ either :asc for ascending and :desc for descending
+    # +count+:: +Fixnum+ the number of records to return in each page
+    # +after+:: +Array+ the current page to paginate after, an array containing a value for each field in the key
```
kevindew: also might be good to rename key to pagination_key it gets a bit confusing seeing `key.keys`

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+
+      if before
+        @previous = before
+        @order = order == :asc ? :desc : :asc
+        @should_reverse = true
+      else
+        @previous = after
+        @should_reverse = false
+      end
```
kevindew: I think you're already thinking about how to present this clearer - so just a nudge to carry on 👍

kevindew: I'm also a little confused by @previous -  I think I've got it now. Might be easier though to just deal with befores and afters though.

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,130 @@
...
+    def where_clause_lhs
+      key.values.join(", ")
+    end
+
+    def where_clause_rhs
+      (["?"] * key.count).join(", ")
+    end
+
+    def where_clause
+      "(#{where_clause_lhs}) #{order_character} (#{where_clause_rhs})"
```
kevindew: Can these just be variables within this method rather than method calls? I don't think they're used elsewhere and are one liners so might be easier to read.

app/queries/keyset_pagination/get_editions.rb
```diff
@@ -0,0 +1,62 @@
...
+      validate_fields!
+    end
+
+    def call
+      editions
+    end
+
+  private
+
+    attr_writer :fields, :filters, :pagination
```
kevindew: do we write to these?

spec/controllers/v2/editions_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+        50.times do |index|
+          FactoryGirl.create(
+            :unpublished_edition,
+            id: index + 101,
+            document: FactoryGirl.create(:document),
+            base_path: "/content#{index + 101}",
+            publishing_app: "test",
+          )
+        end
+      end
```
kevindew: This looks like a lot of DB IO to run before each test. Can we maybe reduce it so we only need to do this whole lot for something testing pagination and then just add a single item for when we're checking other aspects?

app/queries/keyset_pagination.rb
thomasleese: No, it's just taking stuff from the hash to build the key.

kevindew: does results not run a query? 😕

thomasleese: Oh yeah... good point.

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    def next_after_key
+      key_for_record(results.last)
+    end
+
+    def has_next_before?
+      KeysetPagination.new(
+        client, key: key, order: order, count: 1,
+        before: next_before_key, after: nil
+      ).call.count >= 1
```
thomasleese: I'm not sure that it is overkill.

There would be a lot of duplication when generating the `where_next_page` query like that. It wouldn't be possible to use `paginated_query` as it is right now on both because one goes forwards and one goes backwards, so all the logic which reverses the query would also need to be passed into that method rather than in the constructor of the `KeysetPagination` class -- in such a way that the `paginated_query` method ends up becoming as complex as the `KeysetPagination` class itself.

app/queries/keyset_pagination/get_editions.rb
```diff
@@ -0,0 +1,86 @@
...
+      :updated_at,
+      :created_at,
+    ].freeze
+
+    def pagination_field
+      @pagination_field ||= (if order.first == "-"
+                               order[1..order.length]
+                             else
+                               order
+                             end).to_sym
```
kevindew: You might find this is easier to format in a `begin ... end` block then you can have multiple statements

kevindew: Also is there something that restricts which fields they can order by? Just since it's very slow without the index.

app/queries/keyset_pagination.rb
```diff
@@ -0,0 +1,134 @@
...
+  # Example usage:
+  #   client = KeysetPagination::GetEditions.new
+  #   query = KeysetPagination.new(client, per_page: 100, before: [10])
+  #
+  # The `before` and `after` parameters represent the current pagination key as
+  # an array containing the last values of the previous page.
+  # For example, to get the next page of a page that returned 10 items with the
+  # last ID being 10, the after key would be [10]. It is an array because some
+  # pagination clients may use pagination keys across multiple fields, for
+  # example, `GetEditions` uses [date, id].
```
kevindew: 👌

app/queries/keyset_pagination/get_editions.rb
thomasleese: Have added a restriction on the order by fields.

Not sure about using `begin ... end` as I don't think it'll make it any clearer or succinct, it'll look like this:

```
begin
  if order.first == "-"
    field = order[1..order.length]
  else
    field = order
  end
  field.to_sym
end
```

kevindew: I'd have thought you could do
```
@pagination_field ||= begin
  field = order.first == "-" ? order[1..order.length] : order
  field.to_sym
end
```

thomasleese: True... not a fan of ternary operators, but that's done now!

kevindew:
This looks good to me, the docs for it just need an update (order, page)

thomasleese:
@kevindew Yep, for the presentation tomorrow!

kevindew:
@thomasleese that might be cutting it a bit fine for get content 😉

PR #974, opened 2017-07-19
[DO NOT MERGE] Send first_published_at to GA when publishing

elenatanasoiu:
For: [Trello card](https://trello.com/c/7HwMeMfX/177-custom-dimension-first-published-date)

This will send a GA hit whenever an edition is published for the first time.
The information will be sent along as an `event`, using the `trackerId` currently in
use in govuk_admin_template (`tid: UA-26179049-6`).

The client id we're sending (`cid: 660ad712-9753-4cb9-97a7-c9e9f13c318e`) is
randomly generated and can be used to identify that the hits are coming from the
publishing-api rather than govuk_admin_template or govuk_frontend_toolkit (both
apps are normally used to send analytics information).

### DIFF
--- a/app/workers/downstream_live_worker.rb
+++ b/app/workers/downstream_live_worker.rb
@@ -1,4 +1,6 @@
 require 'sidekiq-unique-jobs'
+require 'net/http'
+require 'uri'
 
 class DownstreamLiveWorker
   include DownstreamQueue
@@ -42,6 +44,8 @@ class DownstreamLiveWorker
       DownstreamService.broadcast_to_message_queue(payload, update_type)
     end
 
+    send_first_time_being_published_ga_event if first_time_being_published?
+
     enqueue_dependencies if update_dependencies
   rescue AbortWorkerError => e
     notify_airbrake(e, args)
@@ -80,5 +84,15 @@ private
 
   def notify_airbrake(error, parameters)
     Airbrake.notify(error, parameters: parameters)
+  end
+
+  def send_first_time_being_published_ga_event
+    uri = URI.parse('http://www.google-analytics.com/collect')
+    params = { 'v': '1', 'tid': 'UA-26179049-1', 'cid': '660ad712-9753-4cb9-97a7-c9e9f13c318e', 't': 'event', 'cd90': edition.first_published_at }
+    Net::HTTP.post_form(uri, params)
+  end
+
+  def first_time_being_published?
+    edition.document.editions.where(state: 'substituted').count == 0
   end
 end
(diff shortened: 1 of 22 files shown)


kevindew:
This feels like something that belongs in the sidekiq worker for downstreaming live content: https://github.com/alphagov/publishing-api/blob/master/app/workers/downstream_live_worker.rb

Ideally it'd be something that listened to the Publishing API message queue rather than integrating directly - but not sure how much that messes with your scope. We have email alert api and search that listen to the message queue currently.

elenatanasoiu:
@kevindew I've moved the sending of the event to the DownstreamLiveWorker

app/workers/downstream_live_worker.rb
```diff
@@ -81,4 +85,14 @@ def enqueue_dependencies
   def notify_airbrake(error, parameters)
     Airbrake.notify(error, parameters: parameters)
   end
+
+  def send_ga_event
```
thomasleese: I think it would be better if this was named `send_first_time_being_published_ga_event`.

app/workers/downstream_live_worker.rb
```diff
@@ -81,4 +85,14 @@ def enqueue_dependencies
...
+    uri = URI.parse('http://www.google-analytics.com/collect')
+    params = { 'v': '1', 'tid': 'UA-26179049-1', 'cid': '660ad712-9753-4cb9-97a7-c9e9f13c318e', 't': 'event', 'cd90': edition.first_published_at }
+    Net::HTTP.post_form(uri, params)
+  end
+
+  def first_time_being_published?
+    %w(major minor).include? edition.update_type
```
thomasleese: I'm not sure if this is a good way of determining if this was the first publish. I think a better way would be to check if the document that the edition has any `published_or_unpublished` edition.

spec/commands/v2/import_spec.rb
```diff
@@ -36,6 +36,11 @@
       }
     end
 
+    before do
```
thomasleese: Since we're having to duplicate this in a lot of places, I wonder if there is some way we can improve this. Perhaps have it in a class method called `GoogleAnalyticsTestHelper` or similar, and then use `include GoogleAnalyticsTestHelper` in the tests which need it?

elenatanasoiu: I've moved this into a helper

app/workers/downstream_live_worker.rb
elenatanasoiu: `published_or_unpublished` was always returning an edition because the point where the GA event had to be sent was after an edition had been published. 

I'm using this to check if there's already a published edition present: 

`edition.document.editions.where(state: 'substituted').count == 0`

elenatanasoiu:
Following a conversation with Tim from analytics, it seems that sending this info via the publishing-api won't provide the analytics team with enough information, so I'm closing this PR.

PR #976, opened 2017-07-20
Cache edition hash

steventux:
https://trello.com/c/mOdpAGVj/961-improve-performance-of-the-expanded-link-set-endpoint-3

Pluck and cache a hash of relevant Edition attributes instead of inflating the entire model.

```
Benchmark.measure { LinkExpansion.by_content_id("91b8ef20-74e7-4552-880c-50e6d73c2ff9").links_with_content }
```
Before:
```
=> #<Benchmark::Tms:0x007f82b1f36778 @cstime=0.0, @cutime=0.0, @label="", @real=5.256469395999375, @stime=0.6900000000000002, @total=3.6900000000000004, @utime=3.0>
```
After:
```
=> #<Benchmark::Tms:0x007f02172f1668 @cstime=0.0, @cutime=0.0, @label="", @real=4.324912767999194, @stime=0.6699999999999999, @total=2.74, @utime=2.0700000000000003>
```


lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
+module LinkExpansion::HashedEdition
+  def edition_hash(values)
+    return nil unless values.present?
+    hash = values.is_a?(Array) ? hash_for(values) : values
+    hash = SymbolizeJSON.symbolize(hash)
+    hash = hash.slice(*edition_fields)
+    hash[:api_path] = api_path(hash) unless hash[:base_path].nil?
+    hash[:withdrawn] = withdrawn?(hash)
+    hash.delete(:id)
```
thomasleese: Reckon we could do `hash.except(:id)` and merge these two lines into one.

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
...
+    return nil unless values.present?
+    Hash[edition_fields.zip(values)]
+  end
+
+
+  def api_path(attrs)
+    "/api/content" + attrs[:base_path]
+  end
+
+  def withdrawn?(attrs)
```
kevindew: Can we include this check in the original query without major pain?

steventux: In the case of a large taxonomy tree doing the LEFT OUTER JOIN onto unpublishings appears to be slower by a couple of hundred ms but I guess taxons can't be unpublished/withdrawn.
```
#<Benchmark::Tms:0x007fb235e553f0 @cstime=0.0, @cutime=0.0, @label="", @real=4.603115538000566, @stime=0.7899999999999998, @total=2.88, @utime=2.09>
```

thomasleese: Looks like they can be unpublished, but there aren't many which are.

```
[5] pry(main)> Edition.distinct.where(state: "unpublished", document_type: "taxon").count
=> 186
[6] pry(main)> Edition.distinct.where(document_type: "taxon").count
=> 4047
```

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
...
+    hash.except(:id)
+  end
+
+private
+
+  def edition_fields
+    fields = LinkExpansion::Rules::DEFAULT_FIELDS_WITH_DETAILS.dup
+    fields << :id << :state << :"unpublishings.type"
+    fields -= %i[api_path withdrawn]
+    fields
```
kevindew: Could make this cleaner by doing
```
LinkExpansion::Rules::DEFAULT_FIELDS_WITH_DETAILS
 +    %i(id state unpublishings.type) - %i(api_path withdrawn)
```

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
...
+    Hash[edition_fields.zip(values)]
+  end
+
+
+  def api_path(hash)
+    "/api/content" + hash[:base_path]
+  end
+
+  def withdrawn?(hash)
+    unpublishing_type = hash.delete(:"unpublishings.type")
```
kevindew: should this be deleting here? feels fine to just access it rather than mutate the hash

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
+module LinkExpansion::HashedEdition
+  def edition_hash(values)
+    return nil unless values.present?
+    hash = values.is_a?(Array) ? hash_for(values) : values
```
steventux: Yeah this can be an array of values from the AR `pluck` https://github.com/alphagov/publishing-api/pull/976/files#diff-d21d7da8d7d146227416a146fde15973R51 but can also be `Edition#attributes` from here https://github.com/alphagov/publishing-api/pull/976/files#diff-d21d7da8d7d146227416a146fde15973R23

lib/link_expansion/hashed_edition.rb
steventux: I was conscious of not changing the expected output for the sake of confidence in the existing integration tests, but now I've compared the outputs of the master branch to this one I'm happy to amend, we'd have duplicate values in the hash if that matters?

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
+module LinkExpansion::HashedEdition
+  def edition_hash(values)
+    return nil unless values.present?
```
steventux: I believe it happens within a test here: 
https://github.com/alphagov/publishing-api/blob/master/spec/lib/link_expansion/content_cache_spec.rb#L19
Not sure if the `ContentCache#find` method could be called with a non-existent `content_id`...
https://github.com/alphagov/publishing-api/pull/976/files#diff-d21d7da8d7d146227416a146fde15973R33

lib/link_expansion/hashed_edition.rb
steventux: :+1:

lib/link_expansion/content_cache.rb
```diff
@@ -40,7 +43,12 @@ def editions(content_ids)
       locale_fallback_order: locale_fallback_order,
       state_fallback_order: state_fallback_order,
     )
-    Edition.with_document.includes(:document).where(id: edition_ids).all
+    Edition
+      .with_document
+      .with_unpublishing
```
steventux: :+1:

lib/link_expansion/hashed_edition.rb
kevindew: ah sure, I'd think to drop the field you'd probably want to do that in the `except` call? My concern is that this cleaning up isn't really related to what this method is determining. Unless I'm missing something (which is quite possible)

kevindew:
I meant to query, do we still need the `to_h` method on edition once this work is done? Hope we can kill it as I never liked that method 😀

steventux:
It looks like it's used elsewhere https://github.com/alphagov/publishing-api/search?utf8=%E2%9C%93&q=to_h&type= eg. https://github.com/alphagov/publishing-api/blob/5256788d3037b433eb35967cb100355b9c1d9c8b/app/presenters/edition_presenter.rb#L38

steventux:
@kevindew I think this is up to date now, and it's faster (admittedly on a different machine)
```
#<Benchmark::Tms:0x007fcd33ae4b80 @cstime=0.0, @cutime=0.0, @label="", @real=3.9112095829987084, @stime=0.8200000000000001, @total=2.62, @utime=1.8>
```

lib/link_expansion/content_cache.rb
```diff
@@ -40,10 +41,25 @@ def editions(content_ids)
...
+    Edition
+      .joins(
+        <<-SQL.strip_heredoc
+          LEFT OUTER JOIN unpublishings
+          ON unpublishings.edition_id = editions.id
+          AND editions.state = 'unpublished'
+        SQL
+      )
+      .with_document
+      .includes(:document)
```
kevindew: I think this might be redundant with a pluck? the with_document should add the join.

lib/link_expansion/edition_hash.rb
```diff
@@ -0,0 +1,36 @@
...
+      hash = values.is_a?(Array) ? hash_for(values) : values
+      hash = SymbolizeJSON.symbolize(hash)
+      hash = hash.slice(*edition_fields)
+      hash[:api_path] = api_path(hash) unless hash[:base_path].nil?
+      hash[:withdrawn] = withdrawn?(hash)
+      hash.except(:id, :"unpublishings.type")
+    end
+
+    def edition_fields
+      LinkExpansion::Rules::DEFAULT_FIELDS_WITH_DETAILS.dup +
```
thomasleese: I don't think we need the `.dup` here since that'll happen implicitly when using `+` and `-`.

lib/link_expansion/hashed_edition.rb
```diff
@@ -0,0 +1,37 @@
+module LinkExpansion::HashedEdition
```
thomasleese: What's the difference between `HashedEdition` and `EditionHash`?

steventux: @kevindew questioned whether the `HashedEdition` module shouldn't be a utility class with a class methods, at this point I decided to rename it to `EditionHash` as the original name felt a bit ambiguous.

thomasleese: Cool, I think maybe you forgot to remove `hashed_edition.rb` from the commit as it's still showing up in GitHub.

steventux: The product of a dodgy rebase :+1:

kevindew:
@steventux are we missing a push here? notice we still have the HashedEdition and EditionHash classes

PR #977, opened 2017-07-24
Check whether there are child links when looking up links

kevindew:
Trello: https://trello.com/c/mOdpAGVj/961-improve-performance-of-the-expanded-link-set-endpoint-3

Firstly apologies that this is such a big PR 😢 

This is an approach to speeding up the building of link graphs for expanded link sets, to reduce the number of empty queries by checking at the point of looking up one level of links whether there are any on the next level. This can reduce something ~2200 queries to ~200.

## Benchmarks

master branch (before):
```
irb(main):002:0> puts Benchmark.measure { LinkExpansion.by_content_id("91b8ef20-74e7-4552-880c-50e6d73c2ff9").link_graph.links_content_ids.count }
  9.740000   0.960000  10.700000 ( 16.155548)
=> nil
irb(main):003:0> LinkExpansion.by_content_id("91b8ef20-74e7-4552-880c-50e6d73c2ff9").link_graph.links_content_ids.count
=> 2530
irb(main):004:0> puts Benchmark.measure { DependencyResolution.new("91b8ef20-74e7-4552-880c-50e6d73c2ff9").dependencies.count }
 12.500000   0.650000  13.150000 ( 18.276691)
=> nil
irb(main):005:0> DependencyResolution.new("91b8ef20-74e7-4552-880c-50e6d73c2ff9").dependencies.count
=> 5327
```

this branch (after):
```
irb(main):005:0> puts Benchmark.measure { LinkExpansion.by_content_id("91b8ef20-74e7-4552-880c-50e6d73c2ff9").link_graph.links_content_ids.count }
  1.240000   0.110000   1.350000 (  1.969484)
=> nil
irb(main):006:0> LinkExpansion.by_content_id("91b8ef20-74e7-4552-880c-50e6d73c2ff9").link_graph.links_content_ids.count
=> 2530
irb(main):007:0> puts Benchmark.measure { DependencyResolution.new("91b8ef20-74e7-4552-880c-50e6d73c2ff9").dependencies.count }
  6.710000   0.510000   7.220000 ( 11.260492)
=> nil
irb(main):008:0> DependencyResolution.new("91b8ef20-74e7-4552-880c-50e6d73c2ff9").dependencies.count
=> 5327
```

Unfortunately there has been a pretty big compromise towards performance over ease of understanding as it does get a bit difficult to reason about the checks for the next level of links.

## Other changes

Introducing this change required a fair bit of responsibility shifting in the classes that build up a link graph as dumber objects needed to be more aware of whether they had classes or not so there's been a bunch of refactoring across the various classes that make up LinkExpansion and DependencyResolution, the individual commit messages should explain the reasonings behind these.

Also I'll probably rebase this after https://github.com/alphagov/publishing-api/pull/976 has been merged.


app/models/link_graph/node.rb
```diff
@@ -68,4 +80,20 @@ def to_h
       links: children,
     }
   end
+
+  def might_have_own_links?
```
steventux: Why the shift from a definite sounding attribute `has_own_links` to a less-than-certain method name?

app/queries/edition_links.rb
```diff
@@ -0,0 +1,110 @@
...
+          THEN editions.content_store = 'draft'
+          ELSE editions.content_store = 'live'
+        END
+      SQL
+    end
+
+    def group_results(results)
+      grouped = results
+        .group_by(&:first)
+        .map do |type, values|
```
thomasleese: Might be nicer to use `each_with_object` here.

kevindew:
Thanks for the review @steventux  👍

app/queries/links.rb
```diff
@@ -0,0 +1,243 @@
...
+    end
+
+    def link_content_id_field
+      @mode == :from ? "links.target_content_id" : "link_sets.content_id"
+    end
+
+    def group_results(results)
+      grouped = results
+        .group_by(&:first)
+        .map do |type, values|
```
thomasleese: Might be nicer to use `each_with_object` here.

lib/expansion_rules.rb
```diff
@@ -59,25 +59,25 @@ def recurring
...
+  def reverse_links
     REVERSE_LINKS.values
   end
 
   def reverse_link_type(link_type)
     REVERSE_LINKS[link_type.to_sym]
   end
 
-  def un_reverse_link_type(link_type)
+  def unreverse_link_type(link_type)
```
kevindew: yeah it's a bummer.

So we have something like `child_taxons` which is a reverse link type, and when we look that up in the db we need to unreverse that name to be `parent_taxons` - so there's the context and why forward doesn't quite fit.

There is probably hope for a better name though, I'm just not sure how to make it succinct. 

eg could be `db_column_for_reverse_link_type` or `field_for_reverse_link_type` do you have any ideas too?

app/models/link_graph/node.rb
kevindew: well it's because the field can be true|false|null. When it is true we have_own_links when false we don't have them, and when it's null it's a maybe.

lib/expansion_rules.rb
steventux: That explanation is really helpful, I think I'm struggling with the word `unreverse`.
Given that the purpose of the method is to return the opposing type of link `opposing_link_type` might be enough, that would map `children` -> `parent` and vice versa, do we really care that it's relevant to to reverse links? Like I said, not important for this PR.

kevindew: yeah it's a crappy term since it's almost a double negative. 

It only unreverses reverse link types though, say if you passed it parent, it wouldn't swap that round to child for instance.

app/queries/links.rb
```diff
@@ -0,0 +1,243 @@
...
+        where = { "link_sets.content_id": content_id }
+      else
+        where = { "links.target_content_id": content_id }
+      end
+      where[:link_type] = allowed_link_types if allowed_link_types
+      where
+    end
+
+    def where_not
+      if mode == :from
```
kevindew: good point 👍

app/queries/links.rb
```diff
@@ -0,0 +1,243 @@
...
+
+    def fields
+      base_fields = [:link_type, link_content_id_field]
+      base_fields << has_own_links_field if check_for_from_children?
+      base_fields << is_linked_to_field if check_for_to_children?
+      base_fields
+    end
+
+    def link_content_id_field
+      @mode == :from ? "links.target_content_id" : "link_sets.content_id"
```
kevindew: Thanks!

app/queries/links.rb
kevindew: Yeah I'd agree, I'll give that a spin

kevindew:
Thanks @thomasleese I'll get those addressed tomorrow.

kevindew:
@thomasleese @steventux comments addressed. Codebase is scrubbed of "unreverse" - though I'm not sure the new name is much better, I blame the concept.

PR #981, opened 2017-07-27
Persist expanded links when downstreaming

kevindew:
Trello: https://trello.com/c/mOdpAGVj/961-improve-performance-of-the-expanded-link-set-endpoint-3

This stores expanded links to the database when content items are downstreamed. This is to allow for a much faster lookup of the expanded-links endpoint and avoid the need for a long cache to be on it.

It has the downside that the result of a /expanded-links call may return stale data as it is only updated when the content store is updated. However you can specify a generate=true query string which will generate the expanded-links with the controller, you just have to deal with it potentially taking a second or two.

## Breaking changes

This changes the output of the expanded-links endpoint.

Previously it returned data like so:
```
{
  "content_id": "some-content-id",
  "expanded_links": {},
  "version": 32
}
```
Now it returns data like so:
```
{
  "generated": "2017-07-28T08:52:44Z"
  "expanded_links": {}
}
```

`content_id` is no longer returned as it seemed superfluous (why not all the other parameters that query this request).
`version`, which was the LinkSet stale_lock_version value, is no longer returned as it didn't adequately reflect the version. There is both Document and LinkSet stale_lock_version numbers that reflect the expanded_links - and the expanded links could change without this number changing.

I've not been able to find any evidence that these values are used:
- [content_tagger](https://github.com/alphagov/content-tagger/blob/a658875f7a0ce66da048e462cf674ea9c8cef39b/app/services/legacy_taxonomy/client/publishing_api.rb#L22-L23)
- [govuk_taxonomy_helpers](https://github.com/alphagov/govuk_taxonomy_helpers/blob/947e2ec9d7a20fe31fd594926a91ce95a9a46a6d/lib/govuk_taxonomy_helpers/publishing_api_response.rb#L39-L41)
- [whitehall](https://github.com/alphagov/whitehall/blob/e9fece297555adc1576ffa27af382e830d390dd2/lib/taxonomy/publishing_api_adapter.rb#L42-L44)
- [rummager](https://github.com/alphagov/rummager/blob/cc0841de674d765cef247bae8cd861067e8e2d3d/lib/indexer/links_lookup.rb#L77)

### DIFF
--- a/app/controllers/v2/link_sets_controller.rb
+++ b/app/controllers/v2/link_sets_controller.rb
@@ -5,13 +5,12 @@ module V2
     end
 
     def expanded_links
-      json = Rails.cache.fetch ['expanded-links', content_id, params[:with_drafts], params[:locale]], expires_in: 1.hour do
-        Queries::GetExpandedLinks.call(
-          content_id,
-          params[:locale],
-          with_drafts: with_drafts?,
-        )
-      end
+      json = Queries::GetExpandedLinks.call(
+        content_id,
+        locale,
+        with_drafts: with_drafts?,
+        generate: generate?,
+      )
 
       render json: json
     end
@@ -37,12 +36,20 @@ module V2
       ActiveModel::Type::Boolean.new.cast(params.fetch(:with_drafts, true))
     end
 
+    def generate?
+      ActiveModel::Type::Boolean.new.cast(params.fetch(:generate, false))
+    end
+
     def links_params
       payload.merge(content_id: content_id)
     end
 
     def content_id
       params.fetch(:content_id)
+    end
+
+    def locale
+      params.fetch(:locale, Edition::DEFAULT_LOCALE)
     end
   end
 end
--- a/app/models/expanded_links.rb
+++ b/app/models/expanded_links.rb
@@ -1,0 +1,3 @@
+class ExpandedLinks < ApplicationRecord
+  include FindOrCreateLocked
+end
--- a/app/models/expanded_links.rb
+++ b/app/models/expanded_links.rb
@@ -1,3 +1,26 @@
 class ExpandedLinks < ApplicationRecord
   include FindOrCreateLocked
+
+  def self.locked_update(
+    content_id:,
+    locale:,
+    with_drafts:,
+    payload_version:,
+    expanded_links:
+  )
+    transaction do
+      entry = find_or_create_locked(
+        content_id: content_id,
+        locale: locale,
+        with_drafts: with_drafts,
+      )
+
+      next unless entry.payload_version <= payload_version
+
+      entry.update!(
+        payload_version: payload_version,
+        expanded_links: expanded_links,
+      )
+    end
+  end
 end
--- a/app/queries/get_expanded_links.rb
+++ b/app/queries/get_expanded_links.rb
@@ -1,42 +1,69 @@
 module Queries
   class GetExpandedLinks
-    def self.call(content_id, locale, with_drafts: true)
-      if (link_set = LinkSet.find_by(content_id: content_id))
-        expanded_link_set(link_set, locale, with_drafts: with_drafts)
-      elsif Document.where(content_id: content_id).exists?
-        empty_expanded_link_set(content_id)
-      else
-        error_details = {
-          error: {
-            code: 404,
-            message: "Could not find link set with content_id: #{content_id}"
-          }
-        }
+    def self.call(content_id, locale, with_drafts: true, generate: false)
+      self.new(content_id, locale, with_drafts, generate).call
+    end
 
-        raise CommandError.new(code: 404, error_details: error_details)
+    def call
+      expanded_links = generate ? nil : links_from_storage
+
+      if expanded_links
+        stored_links_response(expanded_links)
+      else
+        generate_links_response
       end
     end
 
-    def self.expanded_link_set(link_set, locale, with_drafts:)
-      expanded_link_set = Presenters::Queries::ExpandedLinkSet.by_content_id(
-        link_set.content_id,
+  private
+
+    attr_reader :content_id, :locale, :with_drafts, :generate
+
+    def initialize(content_id, locale, with_drafts, generate)
+      @content_id = content_id
+      @locale = locale
+      @with_drafts = with_drafts
+      @generate = generate
+    end
+
+    def links_from_storage
+      ExpandedLinks.find_by(
+        content_id: content_id,
+        locale: locale,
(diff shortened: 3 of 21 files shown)


app/queries/get_expanded_links.rb
```diff
@@ -1,42 +1,75 @@
...
-            message: "Could not find link set with content_id: #{content_id}"
-          }
-        }
+    def self.call(content_id, locale, with_drafts: true, generate: false)
+      self.new(content_id, locale, with_drafts, generate).call
+    end
 
-        raise CommandError.new(code: 404, error_details: error_details)
+    def call
+      pre_computed = generate ? nil : pre_computed_link_set
```
steventux: Would it be simpler to use a naming convention like `generate_links`, `generate_links_response` and `generated_links`, `generated_links_response` (or even `pre_generated_links` etc.)? Not intended as a PR blocker.

app/workers/downstream_draft_worker.rb
```diff
@@ -75,4 +77,28 @@ def enqueue_dependencies
...
+  def update_expanded_links(downstream_payload)
+    ExpandedLinks.locked_update(
+      content_id: content_id,
+      locale: locale,
+      with_drafts: true,
+      payload_version: payload_version,
+      expanded_links: downstream_payload.expanded_links,
+    )
+
+    # we also want to update the live links since you live links can be accessed
```
steventux: `s/you/your/`

spec/queries/get_expanded_links_spec.rb
```diff
@@ -2,84 +2,73 @@
 
 RSpec.describe Queries::GetExpandedLinks do
   let(:content_id) { SecureRandom.uuid }
-  let(:live_child_taxon_content_id) { SecureRandom.uuid }
-  let(:draft_child_taxon_content_id) { SecureRandom.uuid }
+  let(:locale) { "en" }
+  let(:with_drafts) { true }
+  let(:generate) { false }
+  # let(:live_child_taxon_content_id) { SecureRandom.uuid }
```
steventux: Should these be removed?

app/controllers/v2/link_sets_controller.rb
```diff
@@ -5,13 +5,12 @@ def get_links
...
-          content_id,
-          params[:locale],
-          with_drafts: with_drafts?,
-        )
-      end
+      json = Queries::GetExpandedLinks.call(
+        content_id,
+        locale,
+        with_drafts: with_drafts?,
+        generate: params.fetch(:generate, false),
```
thomasleese: If the query is `?generate=false` the link expansion will be regenerated since `"false"` is truthy, might be worth clarifying this in the documentation or changing the logic here.

app/models/expanded_links.rb
```diff
@@ -0,0 +1,27 @@
+class ExpandedLinks < ApplicationRecord
+  self.table_name = "expanded_links"
```
thomasleese: Do we need this?

app/queries/get_expanded_links.rb
```diff
@@ -1,42 +1,75 @@
...
+      return if Document.exists?(content_id: content_id) || LinkSet.exists?(content_id: content_id)
+
+      error_details = {
+        error: {
+          code: 404,
+          message: "Could not find links for content_id: #{content_id}"
+        }
       }
+
+      raise CommandError.new(code: 404, error_details: error_details)
```
thomasleese: I think you can pass `message` here directly instead of passing the `error_details` hash.

doc/admin-tasks.md
```diff
@@ -52,3 +52,23 @@ bundle exec represent_downstream:content_id['some-content-id']
...
+It will also be rebuilt any time a piece of content is represented downstream.
+
+* To populate every document (this will take a long time - hours)
+```
+bundle exec expanded_links:populate
+```
+
+* To populate every document of a document_type
+```
+bundle exec expanded_links:populate_document_type['document-type']
```
thomasleese: I might name this `populate_by_document_type`, but that's just personal preference.

lib/tasks/expanded_links.rake
```diff
@@ -0,0 +1,73 @@
...
+    tuples.each_with_index do |(content_id, locale), index|
+      update_draft_links(content_id, locale, payload_version)
+      update_live_links(content_id, locale, payload_version)
+
+      progress(index, tuples.count, every: 100)
+    end
+  end
+
+
+  def update_draft_links(content_id, locale, payload_version)
```
thomasleese: I feel like these two methods could be simplified to reduce duplication by having a single `update_links(content_id, locale, payload_version, with_drafts:)` method.

spec/requests/expanded_links_endpoint_spec.rb
```diff
@@ -1,125 +1,150 @@
...
-      "version" => 0,
-      "content_id" => edition.document.content_id,
-      "expanded_links" => {
-        "available_translations" => translations,
-      },
-    )
+      expect(parsed_response).to eql(
+        "generated" => updated_at.utc.iso8601,
+        "expanded_links" => expanded_links.as_json,
+      )
```
emmabeynon: Just wondering how we know that the links in the reponse are 'de' links? Similarly for the previous test returning 'en' links.

kevindew:
Thanks for the reviews @thomasleese, @emmabeynon & @steventux 👍

spec/requests/expanded_links_endpoint_spec.rb
kevindew: We set the expanded_links in this context block so if it doesn't match that one the test will fail.

app/controllers/v2/link_sets_controller.rb
kevindew: Ah yeah I could use that ugly ActiveModel cast - that'll probably sort it. Boolean params are so annoying in rails

app/queries/get_expanded_links.rb
kevindew: yeah potentially, I'll have a quick think. `pre_computed` was my attempt at avoiding the word `cached` since it's not really a cache.

app/models/expanded_links.rb
kevindew: I thought I did for Rails pluralising the class name (which seemed rubbish as ExpandedLink) but seems it works fine without this so I'll drop it.

kevindew:
All comments should be addressed now @thomasleese, @emmabeynon, @steventux if you don't mind giving it another look over.

PR #982, opened 2017-07-28
Add Pact test for gds-api-adapter pagination

alecgibson:
This change adds a `provider_state` for the Pact tests for
`gds-api-adapters` to page through. For this to work, we need multiple
live editions with fixed `updated_at` times, so we have a predictable
`updated_at` time to set as the last viewed.

Corresponding `gds-api-adapters` Pull Request: https://github.com/alphagov/gds-api-adapters/pull/728

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -450,4 +450,18 @@ Pact.provider_states_for "GDS API Adapters" do
       FactoryGirl.create(:live_edition, base_path: '/bar', document: document_2)
     end
   end
+
+  provider_state "there are 4 live content items with fixed updated timestamps" do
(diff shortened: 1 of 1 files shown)


kevindew:
Do you need explicit ids set for the editions? I'd have thought they're used in the pagination key

alecgibson:
@kevindew the pagination is done on the `updated_at` date (hence the need for fixing that property).

alecgibson:
@kevindew as discussed - `content_id` is used to break ties where `updated_at` dates are equal. However, in these Pact tests, I've deliberately picked `updated_at` dates that are far away from each other to avoid this collision issue, so it should all be fine.

kevindew:
super 👍

PR #983, opened 2017-07-28
Encode request url correctly

thomasleese:
This means that the query parameters are correctly encoded in the output rather than relying on the input to be. This is mainly a problem in the pact tests where the self URL would be formatted as `fields[]=` and the next/previous URLs would be formatted as `fields%5B%5D=`. 

We don't encounter this discrepancy in the offset/limit pagination endpoints as they use regular expressions to modify the `page=` parameter and leave all the others exactly as they came in.

### DIFF
--- a/app/presenters/keyset_pagination_presenter.rb
+++ b/app/presenters/keyset_pagination_presenter.rb
@@ -36,7 +36,7 @@ module Presenters
     end
 
     def self_link
-      { href: request_url, rel: "self" }
+      { href: self_url, rel: "self" }
     end
 
     def previous_link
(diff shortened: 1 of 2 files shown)


app/presenters/keyset_pagination_presenter.rb
```diff
@@ -36,27 +36,36 @@ def next_link
...
+      page_href(except: %w(before after), before: pagination_query.next_before_key.join(","))
     end
 
-    def page_href(params)
+    def self_url
+      page_href
+    end
+
+    def page_href(params = {})
+      except = params.delete(:except)
```
kevindew: Rather than needing to pass an except in should we just remove them if there is a before or after in the params?

thomasleese: We can't do that as the `self` URL needs to leave them in, I couldn't think of a better way of doing it.

kevindew: self doesn't use the params argument right?

you could maybe except a hash of `{ before:, after: }` as the params and then have `except = before || after ? [:before, :after] : []`

thomasleese: Yep, I like that, I've pushed a fix to do it like that.

PR #984, opened 2017-07-28
Store edition diff on action

emmabeynon:
Before creating a PutContent Action we now create a diff between the previous
draft edition and the newly updated draft edition and store this on the new
Action.

https://trello.com/c/8gah2r5G/988-5-store-a-diff-of-an-edition-in-the-action-when-it-changes

I noticed that, when updating a change note for an updated draft edition, the updated draft still showed an old change note.  For now we used `edition.reload` in the PutContent command to overcome this, but we need to look at how change notes are created in the [ChangeNote model](https://github.com/alphagov/publishing-api/blob/master/app/models/change_note.rb) to find a better solution.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -3,6 +3,9 @@ module Commands
     class PutContent < BaseCommand
       def call
         PutContentValidator.new(payload, self).validate
+
+        edition_diff.previous_item = previous_drafted_edition
+
         prepare_content_with_base_path
 
         edition = create_or_update_edition
@@ -35,6 +38,10 @@ module Commands
 
     private
 
+      def edition_diff
+        @edition_diff ||= HashDiffBuilder.new(Presenters::EditionDiffPresenter)
+      end
+
       def link_diff_between(old_links, new_links)
         old_links - new_links
       end
@@ -54,8 +61,10 @@ module Commands
         access_limit(edition)
         update_last_edited_at(edition, payload[:last_edited_at])
         ChangeNote.create_from_edition(payload, edition)
-        Action.create_put_content_action(edition, document.locale, event)
+        edition.reload
         create_links(edition)
+        edition_diff.current_item = edition
+        Action.create_put_content_action(edition, document.locale, event, edition_diff.diff)
       end
 
       def create_links(edition)
--- a/app/models/action.rb
+++ b/app/models/action.rb
@@ -1,4 +1,5 @@
 class Action < ActiveRecord::Base
+  serialize :edition_diff
   belongs_to :edition, optional: true
   belongs_to :link_set, optional: true
   belongs_to :event
@@ -6,8 +7,8 @@ class Action < ActiveRecord::Base
   validate :one_of_edition_link_set
   validates :action, presence: true
 
-  def self.create_put_content_action(edition, locale, event)
-    create_publishing_action("PutContent", edition, locale, event)
+  def self.create_put_content_action(updated_draft, locale, event, edition_diff)
+    create_publishing_action("PutContent", updated_draft, locale, event, edition_diff: edition_diff)
   end
 
   def self.create_publish_action(edition, locale, event)
@@ -23,7 +24,7 @@ class Action < ActiveRecord::Base
     create_publishing_action("DiscardDraft", edition, locale, event)
   end
 
-  def self.create_publishing_action(action, edition, locale, event)
+  def self.create_publishing_action(action, edition, locale, event, edition_diff: nil)
     create!(
       content_id: edition.document.content_id,
       locale: locale,
@@ -31,6 +32,7 @@ class Action < ActiveRecord::Base
       user_uid: event.user_uid,
       edition: edition,
       event: event,
+      edition_diff: edition_diff
     )
   end
 
--- a/app/presenters/edition_diff_presenter.rb
+++ b/app/presenters/edition_diff_presenter.rb
@@ -1,0 +1,23 @@
+module Presenters
+  class EditionDiffPresenter
+    def self.call(edition)
+      attributes = {}
+      return attributes unless edition.present?
+
+      Edition::TOP_LEVEL_FIELDS.each do |field|
+        attributes[field.to_s] = edition.public_send(field)
+      end
+
+      attributes["links"] = {}
+
+      edition.links.each do |link|
+        attributes["links"][link.link_type] ||= []
+        attributes["links"][link.link_type] << link.target_content_id
+      end
+
+      attributes["change_note"] = edition.change_note&.note || {}
+
+      attributes
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170726114449) do
+ActiveRecord::Schema.define(version: 20170728101003) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -34,6 +34,7 @@ ActiveRecord::Schema.define(version: 20170726114449) do
     t.integer "event_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.text "edition_diff"
     t.index ["edition_id"], name: "index_actions_on_edition_id"
     t.index ["event_id"], name: "index_actions_on_event_id"
     t.index ["link_set_id"], name: "index_actions_on_link_set_id"
--- a/spec/presenters/edition_diff_presenter_spec.rb
+++ b/spec/presenters/edition_diff_presenter_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe Presenters::EditionDiffPresenter do
+  let(:content_id) { SecureRandom.uuid }
+  let(:edition) do
+    FactoryGirl.create(:edition,
+                        update_type: "major",
+                        links_hash: { "organisations" => [content_id] },
+                        change_note: "a note"
+                      )
+  end
+
+  let(:edition_without_links) do
+    FactoryGirl.create(:edition)
+  end
+
+  let(:edition_without_change_note) do
+    FactoryGirl.create(:edition,
+                        update_type: "minor",
+                        links_hash: { "organisations" => [content_id] }
+                      )
+  end
+
+  EXCLUDED_ATTRIBUTES = %w(updated_at created_at id publishing_request_id document_id).freeze
+
+  describe "#call" do
+    subject { described_class }
+
+    context "with links and change_note" do
+      it "returns an edition hash presented for diffing" do
+        expect(subject.call(edition)).to match a_hash_including(
+          edition.as_json.except(*EXCLUDED_ATTRIBUTES),
+          "links" => { "organisations" => [content_id] },
+          "change_note" => edition.change_note.note
+        )
+      end
+    end
+
+    context "without links" do
+      it "returns an edition hash presented for diffing" do
+        expect(subject.call(edition_without_links)).
+          to match a_hash_including("links" => {})
+      end
+    end
+
+    context "without change_note" do
+      it "returns an edition hash presented for diffing" do
+        expect(subject.call(edition_without_change_note)).
+          to match a_hash_including("change_note" => {})
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem 'bunny', '~> 2.6'
 gem 'whenever', '0.9.4', require: false
 gem "govuk_sidekiq", "~> 1.0.3"
 gem "json-schema", require: false
-gem "hashdiff", require: false
+gem "hashdiff"
 gem "sidekiq-unique-jobs", "~> 5.0", require: false
 gem "govspeak", "~> 5.0.2"
 gem "diffy", "~> 3.1", require: false
--- a/db/migrate/20170728101003_add_edition_diff_to_action.rb
+++ b/db/migrate/20170728101003_add_edition_diff_to_action.rb
@@ -1,0 +1,5 @@
+class AddEditionDiffToAction < ActiveRecord::Migration[5.1]
+  def change
+    add_column :actions, :edition_diff, :text
+  end
+end
--- a/lib/hash_diff_builder.rb
+++ b/lib/hash_diff_builder.rb
@@ -1,0 +1,29 @@
+class HashDiffBuilder
+  def initialize(presenter)
+    @presenter = presenter
+  end
+
+  def previous_item=(previous_item)
+    @previous_item = presenter ? presenter.call(previous_item) : previous_item
+  end
+
+  def current_item=(current_item)
+    @current_item = presenter ? presenter.call(current_item) : current_item
+  end
+
+  def diff
+    raise MissingItemError, "No previous item provided" if previous_item.nil?
+    raise MissingItemError, "No current item provided" if current_item.nil?
+    create_diff(previous_item, current_item)
+  end
+
+private
+
+  attr_reader :presenter, :previous_item, :current_item
+
+  def create_diff(previous_item, current_item)
+    HashDiff.diff(previous_item, current_item)
+  end
+
+  class MissingItemError < RuntimeError; end
+end
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -8,10 +8,15 @@ RSpec.describe Commands::V2::PutContent do
     end
 
     let(:content_id) { SecureRandom.uuid }
+    let(:new_content_id) { SecureRandom.uuid }
     let(:base_path) { "/vat-rates" }
     let(:locale) { "en" }
 
     let(:change_note) { { note: "Info", public_timestamp: Time.now.utc.to_s } }
+    let(:new_change_note) { { note: "Changed Info", public_timestamp: Time.now.utc.to_s } }
+
+    let(:links_1) { { organisations: [content_id] } }
+    let(:links_2) { { policy_areas: [new_content_id] } }
 
     let(:payload) do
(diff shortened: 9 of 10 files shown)


app/models/action.rb
```diff
@@ -23,14 +24,15 @@ def self.create_discard_draft_action(edition, locale, event)
     create_publishing_action("DiscardDraft", edition, locale, event)
   end
 
-  def self.create_publishing_action(action, edition, locale, event)
+  def self.create_publishing_action(action, edition, locale, event, edition_diff = nil)
```
thomasleese: I wonder if this should be a keyword argument for clarity?

lib/edition_diff.rb
```diff
@@ -12,6 +12,14 @@ def field_diff
       diff.map(&:first)
   end
 
+  def full_diff
+    if previous_edition.present?
+      HashDiff.diff(current_edition.attributes, previous_edition.attributes)
```
thomasleese: Can we do `previous_edition.try(:attributes)` here to avoid the if statement?

spec/lib/edition_diff_spec.rb
```diff
@@ -23,77 +23,94 @@
...
 
-  context "diff in title, document_type and base_path" do
-    let!(:current_edition) do
-      FactoryGirl.create(
-        :live_edition,
+    context "diff in title, document_type and base_path" do
+      let!(:current_edition) do
+        FactoryGirl.create(
+          :live_edition,
         document: document,
```
thomasleese: I think these lines need to be indented as well.

spec/lib/edition_diff_spec.rb
```diff
@@ -23,77 +23,94 @@
...
 
-  context "diff in details when a finder" do
-    let!(:current_edition) do
-      FactoryGirl.create(
-        :live_edition,
+    context "diff in details when a finder" do
+      let!(:current_edition) do
+        FactoryGirl.create(
+          :live_edition,
         document: document,
```
thomasleese: Same as above about indentation. :)

spec/lib/edition_diff_spec.rb
```diff
@@ -23,77 +23,94 @@
...
-    it "compares the two given editions" do
-      expect(
-        EditionDiff.new(new_draft_edition,
-                        previous_edition: {}).field_diff
-      ).to include(:document_type, :title)
+    context "provide the edition to compare" do
+      it "compares the two given editions" do
+        expect(
+          EditionDiff.new(new_draft_edition,
+          previous_edition: {}).field_diff
```
thomasleese: I think this line should line up with `new_draft_edition`.

spec/lib/edition_diff_spec.rb
```diff
@@ -23,77 +23,94 @@
...
+      end
+
+      let(:presented_item) do
+        new_draft_edition.to_h.deep_stringify_keys
+      end
+
+      it "compares the two given editions" do
+        expect(
+          EditionDiff.new(new_draft_edition,
+          previous_edition: presented_item).field_diff
```
thomasleese: Same as above.

spec/lib/edition_diff_spec.rb
emmabeynon: Serves me right for relying on linter auto-correct 🙄

lib/edition_diff.rb
emmabeynon: Fab suggestion, wasn't very happy with that duplication 👍

lib/edition_diff.rb
```diff
@@ -12,6 +12,14 @@ def field_diff
...
+    else
+      HashDiff.diff(current_edition.attributes, previous_edition)
+    end
```
kevindew: To safe a conditional you could use the [safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) 

I think we might need to use something more than attributes though so that we include things like links - not sure what else is part of an edition that isn't associated.

app/commands/v2/put_content.rb
```diff
@@ -3,6 +3,9 @@ module V2
     class PutContent < BaseCommand
       def call
         PutContentValidator.new(payload, self).validate
+        @edition_diff ||= Action::EditionDiffFactory.new
```
kevindew: Be nice to have this handled in a method

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
+class Action::EditionDiffFactory
+  attr_reader :previous_draft_edition, :updated_draft_edition
```
kevindew: I don't think we need to refer to these as "draft" - I also wonder if theses need to be public?

kevindew: Thinking more, it could be that these are public writer methods, and have private reader methods? not 100% sure on this though

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def add_previous_draft_edition(previous_draft_edition)
+    @previous_draft_edition = get_edition_attributes(previous_draft_edition)
+  end
+
+  def add_updated_draft_edition(updated_draft_edition)
+    @updated_draft_edition = get_edition_attributes(updated_draft_edition)
+  end
+
+  def create_draft_edition_diff
```
kevindew: I think this could have a simpler name, maybe just `diff` ?

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+    # we don't raise an error if there is no @previous_draft_edition present,
+    # as it's possible for no previous draft edition to exist
+    raise Action::NoUpdatedDraftEditionProvidedError unless @updated_draft_edition.present?
```
kevindew: Might be that we require both methods to have been called but nil results in an empty object, so if either is nil it means it wasn't called and then can throw exception - could probably avoid the need for a comment then.

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+private
+
+  def create_edition_diff(edition_a, edition_b)
+    HashDiff.diff(edition_a, edition_b)
+  end
+
+  def get_edition_attributes(edition)
+    attributes = edition.try(:attributes)
```
kevindew: be good to not include timestamps here. Rather than using attributes could loop through [TOP_LEVEL_FIELDS](https://github.com/alphagov/publishing-api/blob/master/app/models/edition.rb#L11) to call the methods for each - as there's some subtle differences between the attributes and method versions (SymbolizeJSON)

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+    return unless attributes
+
+    if edition.links.present?
+      attributes["links"] = []
+      edition.links.each_with_object({}) do |link, hash|
+        hash["link_type"] = link["link_type"]
+        hash["target_content_id"] = link["target_content_id"]
```
kevindew: We might be able to more succinctly represent these as 
`"link_type" => [content_ids]`

eg. `"organisations" => [id_a, id_b]`

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    if edition.change_note.present?
+      attributes["change_note"] = edition.change_note.note
+    end
+    attributes
+  end
+end
+
+class Action::NoUpdatedDraftEditionProvidedError < StandardError; end
```
kevindew: Be good to move this within the class, then it's within autoload scope. Might also be good to have it inherit from RuntimeError to be more semantic (http://blog.honeybadger.io/ruby-exception-vs-standarderror-whats-the-difference/#the-exception-tree)

emmabeynon:
@kevindew I've addressed all your comments 👍

kevindew:
So I pondered this a bit more, and thought that since we no longer actually create an action the name `EditionDiffFactory` no longer really makes sense as it's not a factory.

So the logical next name would be EditionDiff, but then we already have a class of EditionDiff (albeit without a namespace) which seemed to smell a little to me. So then I wondered if it needs to care about an edition, and maybe it's actually just a combination of a progressive hash diff and a presenter.

So I wondered how something like this felt to you? (excuse on the fly code that probably doesn't work)

in put content something like this
```ruby
def edition_diff
  @edition_diff ||= ProgressiveHashDiff.new(Presenters::EditionDiffPresenter)
end

# wherever it's used
edition_diff.previous = edition
# ...
edition_diff.current = edition
Action.create_...(..., edition_diff.diff)
```

And then for say lib/progressive_hash_diff.rb you have
```
class ProgressiveHashDiff
  # ...

  def previous=(item)
     @previous_hash = presenter ? presenter.call(item) : item
  end

end
```

Which I think has the nice benefit of decoupling the diff building process from the presenter concerns. What do you think?

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    if edition.links.present?
+      attributes["links"] = {}
+      edition.links.each do |link|
+        attributes["links"][link["link_type"]] = []
+      end
+      edition.links.each do |link|
+        attributes["links"][link["link_type"]] << link["target_content_id"]
+      end
```
kevindew: I'd probably go with an `each_with_object` here, you can then have that result of that be set to attributes["links"] and can avoid iterating the collection twice. 

Something like
```ruby
attributes["links"] = edition.links.each_with_object({}) do |link, memo|
  memo[link.link_type] ||= []
  memo[link.link_type] << link.target_content_id
end
```

You also don't need to check if links are present, it's better we have them as an empty hash than missing from the collection

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    if edition.change_note.present?
+      attributes["change_note"] = edition.change_note.note
```
kevindew: similar comment here, we probably want this to be nil when there isn't a change_note.

app/models/action/edition_diff_factory.rb
```diff
@@ -0,0 +1,48 @@
...
+  def create_edition_diff(edition_a, edition_b)
+    HashDiff.diff(edition_a, edition_b)
+  end
+
+  def get_edition_attributes(edition)
+    attributes = {}
+    return attributes unless edition.present?
+
+    Edition::TOP_LEVEL_FIELDS.each do |field|
+      attributes[field.to_s] = edition.send(field)
```
kevindew: Minor thing, but you're probably better to use `public_send` here as that won't accidentally send a private method. It's not really likely in this scenario, but good to indicate to future readers that you're only using public methods.

emmabeynon:
@kevindew I've addressed your new comments - let me know what you think!

spec/presenters/edition_diff_presenter_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      state: "draft",
+      title: "A title",
+      user_facing_version: 1,
+      update_type: "major",
+      links: organisation_links,
+      change_note: change_note
+    )
+  end
+
+  describe "#call" do
```
kevindew: I imagine here it'd be nice to have one test to check core values and then just need to check small sections in the different contexts.

I also wonder if rather than needing to the type out the whole thing you could use as_json on the edition (which thinking about it might be a nice alternative to using `public_send` above

spec/presenters/edition_diff_presenter_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+          "rendering_app" => "government-frontend",
+          "routes" => [{ "path" => "/path", "type" => "prefix" }],
+          "schema_name" => "answer",
+          "state" => "draft",
+          "title" => "A title",
+          "user_facing_version" => 1,
+          "update_type" => "major",
+          "links" => {},
+          "change_note" => change_note.note
+        )
```
kevindew: so you probably only need to test that this is `a_hash_including("links" => {})` for this context

app/presenters/edition_diff_presenter.rb
```diff
@@ -0,0 +1,24 @@
...
+      attributes = {}
+      return attributes unless edition.present?
+
+      Edition::TOP_LEVEL_FIELDS.each do |field|
+        attributes[field.to_s] = edition.public_send(field)
+      end
+
+      attributes["links"] = {}
+
+      edition&.links&.each do |link|
```
emmabeynon: Ah yes, I've realised in one of the tests an edition with no links was stubbed to return nil instead of an empty array, which was the reason I needed the safe navigation operator to make it work.  That's been change to return an empty array now 👍

lib/progressive_hash_diff.rb
```diff
@@ -0,0 +1,29 @@
+class ProgressiveHashDiff
+  attr_reader :presenter, :previous_item, :current_item
```
emmabeynon: I've found that by moving this into private I'm unable to test the `previous_item=` and `current_item=` methods.  Is that ok?

lib/progressive_hash_diff.rb
```diff
@@ -0,0 +1,29 @@
...
+    @previous_item = presenter ? presenter.call(previous_item) : previous_item
+  end
+
+  def current_item=(current_item)
+    @current_item = presenter ? presenter.call(current_item) : current_item
+  end
+
+  def diff
+    raise MissingItemError, 'No previous item provided' if @previous_item.nil?
+    raise MissingItemError, 'No current item provided' unless @current_item.present?
```
emmabeynon: I've switched l18 to use `if item.nil?` -- `present?` won't work for l17 as is needs to be at least an empty hash, and definitely not nil.

kevindew: cool cool 👍

lib/progressive_hash_diff.rb
kevindew: Yeah I think it's fine, we care about the outcome of `diff`matters.

spec/lib/progressive_hash_diff_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      subject.current_item = updated_edition
+
+      expect(subject.diff).to include(
+        ["~", "title", "A title", "A new title"],
+        ["-", "links.organisations", [content_id]],
+        ["+", "links.policy_areas", [content_id]]
+      )
+    end
+
+    it "raises a 'MissingItemError' error if there is no previous item or empty hash" do
```
emmabeynon: As discussed, I've implemented the one-line expectation where it's possible

emmabeynon:
@kevindew latest comments addressed 🎉

db/schema.rb
```diff
@@ -34,6 +34,7 @@
     t.integer "event_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.text "edition_diff"
```
thomasleese: It feels to me like this would be better stored as JSON?

lib/progressive_hash_diff.rb
```diff
@@ -0,0 +1,30 @@
...
+  def previous_item=(previous_item)
+    @previous_item = presenter ? presenter.call(previous_item) : previous_item
+  end
+
+  def current_item=(current_item)
+    @current_item = presenter ? presenter.call(current_item) : current_item
+  end
+
+  def diff
+    raise MissingItemError, 'No previous item provided' if previous_item.nil?
```
thomasleese: I think these should be double quotes rather than single quotes.

lib/progressive_hash_diff.rb
```diff
@@ -0,0 +1,30 @@
+class ProgressiveHashDiff
```
thomasleese: I find the naming of this class a bit confusing, not sure what `Progressive` means in this context?

kevindew: As I suggested this name, it's probably appropriate I reply.

The purpose of the class is to build up the components of a diff gradually, I.e. in a progressive manner. So at the start of a block of code you can set the first item and later can set the other one (which could be same object in a different state - which is what we do). So this class is almost used to gradually build up what will be diffed, as opposed to us needing to store variables of the components.

Did that explanation make sense?

db/schema.rb
kevindew: Another one I suggested. 

If the data is stored as JSON you lose the type of objects when converting. Eg if you have a Date object in the diff, a JSON representation would only tell you it's a string timestamp. Whereas text serialised stores the diff accurately.

lib/progressive_hash_diff.rb
thomasleese: Hmm, yeah, this makes more sense.

I think I would've gone for `HashDiffBuilder` with a more fluent interface like:

```ruby
HashDiffBuilder.new
  .previous_item(a)
  .current_item(b)
  .diff
```

db/schema.rb
thomasleese: Oh yeah, sad times 😢

lib/progressive_hash_diff.rb
kevindew: We did have actually have a EditionActionFactory for a few days

Fluid interface doesn't work unless you save the variables, which is what we worked at avoiding.

eg this crude pseudo code
```
# start of command
edition_before_command = turn_edition_into_comparable_hash(edition)
# change edition in lots of ways over many lines and methods
edition_after_command = turn_edition_into_comparable_hash(edition)
# create and save diff
```

whereas this class encapsulates the representation of the item as a hash and stores the variables for you

thomasleese: Yeah, I can see that. I think if `ProgressiveHashDiff` were renamed to `HashDiffBuilder` it would make more sense to me.

kevindew: Ah sure, yeah - that is probably a good shout. We'll let Emma make the call 👍

emmabeynon: I liked `HashDiffBuilder` 👍

emmabeynon:
@thomasleese HashDiffBuilder is go 🚀

PR #987, opened 2017-08-02
Add payload_version to message queue messages

MatMoore:
This is the same field passed to content store.

This allows the search indexing consumer to end up in the right state when
publish/unpublish messages are processed out of order, because we can ignore
version numbers that older than our existing copy for a given content id.

This means for the requeue task, we'll have to query the events from the
database and pick the most recent one for each document. This could probably be
written more efficiently but I'm expecting the bottleneck to be on the consumer
side.

Trello: https://trello.com/c/kkRWlPKy/207-spike-ignore-messages-from-old-versions

### DIFF
--- a/lib/requeue_content.rb
+++ b/lib/requeue_content.rb
@@ -20,9 +20,11 @@ class RequeueContent
 private
 
   def publish_to_queue(edition)
+    version = Event.maximum(:id)
+
     queue_payload = Presenters::EditionPresenter.new(
       edition, draft: false,
-    ).for_message_queue
+    ).for_message_queue(version)
 
     # FIXME: Rummager currently only listens to the message queue for the
     # event type 'links'. This behaviour will eventually be updated so that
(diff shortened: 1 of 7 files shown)


lib/requeue_content.rb
```diff
@@ -20,9 +21,11 @@ def call
 private
 
   def publish_to_queue(edition)
+    version = edition.document.events.last.id
```
kevindew: From a quick glance at this, you shouldn't need to look up the last event of a document, just getting the last event should be sufficient: `Event.maximum(:id)` - and you should look this up before you load the edition as if it's lazy loaded you could be missing an event that just happened.

brenetic: We've made this change. 👍 Could you take another look?

dwhenry:
@thomasleese unfortunately the `RedirectPresenter` has a `for_redirect_helper` method that does take a version so trying to make this a default field is not very clean. as such I think leaving as is may be the best option for now.

The alternative might be adding the payload version to the redirect helper schema, but this feels out of scope for now.

thomasleese:
@dwhenry Yeah, okay, `for_redirect_helper` is a bit of a weird one. Happy to merge then!

PR #991, opened 2017-08-09
Create change note model on an edition object

kevindew:
This means that the edition will have up to date information and not
need to be reloaded.

### DIFF
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -26,37 +26,35 @@ private
 
   def create_from_top_level_change_note
     return unless change_note
-    ChangeNote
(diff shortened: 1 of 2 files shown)


app/models/change_note.rb
```diff
@@ -26,39 +26,37 @@ def build
...
+      public_timestamp: history_element.fetch(:public_timestamp),
+      note: history_element.fetch(:note),
+    )
   end
 
   def update_type
     @update_type ||= payload[:update_type] || edition.update_type
   end
 
+  def change_note_model
```
thomasleese: `change_note_instance` feels better to me than `change_note_model`.

PR #992, opened 2017-08-09
[WIP] Add migration to redirect base_path_less world locations

davidbasalla:
We recently created taxon pages for all WorldLocations (eg /world/germany). The publishing api and content store still have data for the old WorldLocation pages (eg /government/world/germany). We already redirect to the new taxon pages via Whitehall's routing, so this migration cleans up the state in the PublishingAPI to match.

When attempting to run the UnpublishService, I ran into errors from the `RoutesAndRedirectsValidator` class which expects an base_path to be set. Therefore this migration adds the base_paths back in before running the UnpublishService.


davidbasalla:
Closing in favour of https://github.com/alphagov/whitehall/pull/3391

PR #995, opened 2017-08-10
Remove documents from change_notes

steventux:
https://trello.com/c/Ui6ln1cw/975-5-remove-association-between-change-notes-and-documents

Removes `document_id` column and fk from `change_notes`

Depends on https://github.com/alphagov/publishing-api/pull/994

### DIFF
--- a/app/models/change_note.rb
+++ b/app/models/change_note.rb
@@ -1,5 +1,4 @@
 class ChangeNote < ActiveRecord::Base
-  belongs_to :document
   belongs_to :edition, optional: true
 
   def self.create_from_edition(payload, edition)
@@ -54,7 +53,7 @@ private
   end
 
   def change_note_instance
-    edition.change_note || edition.create_change_note!(document: document)
+    edition.change_note || edition.create_change_note!
   end
 
   def change_note
@@ -67,9 +66,5 @@ private
 
   def change_history
     edition.details[:change_history]
-  end
-
-  def document
-    edition.document
   end
 end
--- a/app/presenters/change_history_presenter.rb
+++ b/app/presenters/change_history_presenter.rb
@@ -27,16 +27,10 @@ module Presenters
 
     def change_notes
       ChangeNote
-        .where(document: document)
-        .where("edition_id IS NULL OR edition_id IN (?)", edition_ids)
-        .order(:public_timestamp)
-    end
-
-    def edition_ids
-      Edition
-        .where(document: document)
+        .joins(:edition)
+        .where(editions: { document: document })
         .where("user_facing_version <= ?", version_number)
-        .pluck(:id)
+        .order(:public_timestamp)
     end
 
     def version_number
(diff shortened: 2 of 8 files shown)


app/models/change_note.rb
```diff
@@ -54,7 +53,7 @@ def update_type
   end
 
   def change_note_instance
-    edition.change_note || edition.create_change_note!(document: document)
+    edition.change_note || edition.create_change_note!(edition_id: edition.id)
```
thomasleese: Might be more succinct to do `edition: edition`.

app/presenters/change_history_presenter.rb
```diff
@@ -27,7 +27,6 @@ def presented_change_notes
 
     def change_notes
       ChangeNote
-        .where(document: document)
         .where("edition_id IS NULL OR edition_id IN (?)", edition_ids)
```
thomasleese: I think we can get rid of the `edition_id IS NULL` part and just do `.where(edition_id: edition_ids)` now.

kevindew: I assume we could also do this through one query with joins and skip the edition_ids query?

app/models/change_note.rb
thomasleese: In fact, do we even need to pass `edition` here? Feels like that would be handled by `create_change_note!`.

emmabeynon:
Thanks @kevindew @thomasleese I've made a couple of changes 🌵

app/presenters/change_history_presenter.rb
```diff
@@ -27,16 +27,10 @@ def presented_change_notes
...
-        .where(document: document)
-        .where("edition_id IS NULL OR edition_id IN (?)", edition_ids)
-        .order(:public_timestamp)
-    end
-
-    def edition_ids
-      Edition
-        .where(document: document)
+        .joins(:edition)
+        .where('editions.document' => document)
```
thomasleese: I think you can do `editions: { document: document }` which might be nicer.

app/models/change_note.rb
```diff
@@ -54,7 +53,7 @@ def update_type
   end
 
   def change_note_instance
-    edition.change_note || edition.create_change_note!(document: document)
+    edition.change_note || edition.create_change_note!(edition: edition)
```
thomasleese: Did you work out if it was necessary to include the `edition: edition` here?

emmabeynon:
> I wonder if we need to split this up into a PR for the code update, and then a PR for the migration and deploy those separately in that respective order? Alternatively we could deploy without migrations and then run the migrations manually.

For general ease, I could make sure to run the migrations separately.

app/models/change_note.rb
emmabeynon: Missed that earlier comment, but yes it seems to work without passing edition 👍

PR #996, opened 2017-08-10
Revert edition diff code

emmabeynon:
We need to revert edition_diff on Action and all code related to creating the diff.

https://trello.com/c/MnNWuuIj/1016-revert-the-diff-work-temporarily


kevindew:
I think there might be something that has broke in the world of end to end (yay)

PR #1002, opened 2017-08-16
Add an alert for when put_content is called without an update_type

thomasleese:
This will help us to know when it's safe to make this field mandatory.

[Trello Card](https://trello.com/c/2wWArZR3/1019-05-add-an-alert-for-when-put-content-happens-without-an-updatetype)

### DIFF
--- a/spec/commands/v2/put_content_spec.rb
+++ b/spec/commands/v2/put_content_spec.rb
@@ -221,5 +221,35 @@ RSpec.describe Commands::V2::PutContent do
         end
       end
     end
+
+    context "when no update_type is provided" do
+      before do
+        payload.delete(:update_type)
+      end
+
+      it "should send an alert to Airbrake" do
+        # swallow error about missing schema
+        expect(Airbrake).to receive(:notify)
+          .with(anything, parameters: a_hash_including(schema_path: anything))
+
+        expect(Airbrake).to receive(:notify)
+          .with(anything, parameters: a_hash_including(content_id: content_id))
+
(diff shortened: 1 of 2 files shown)


spec/commands/v2/put_content_spec.rb
```diff
@@ -221,5 +221,32 @@
...
+          .with(anything, parameters: a_hash_including(content_id: content_id))
+
+        described_class.call(payload)
+      end
+    end
+
+    context "when an update type is provided" do
+      it "should not send an alert to Airbrake" do
+        # swallow error about missing schema
+        expect(Airbrake).to receive(:notify)
```
emmabeynon: Should this be `.not to`?

thomasleese: No, it's confusing, but I couldn't think of a better way of doing this.

Since we're using a schema name of `nonexistent-schema`, we already get an Airbrake error about how the schema doesn't exist. This catches that error, and will subsequently fail if `Airbrake.notify` gets called again with different arguments (i.e. the alert about update type) so we know it's not being sent if no subsequent failure is sent out.

thomasleese: I will add an explicit `.not to` for the update type `Airbrake.notify` though which should make things clearer.

emmabeynon: Cool 👍

thomasleese: Done!

PR #1005, opened 2017-08-17
Add new date fields to Edition

emmabeynon:
We add new date fields to Edition to distinguish between dates provided by publishing apps, and those generated by the Publishing API.
This PR covers the migrations and will be followed up by another to populate the fields.


db/migrate/20170817095425_add_temporary_timestamps_to_edition.rb
```diff
@@ -0,0 +1,15 @@
+class AddTemporaryTimestampsToEdition < ActiveRecord::Migration[5.1]
+  def up
```
thomasleese: I think actually you only need a `change` method which can be the same as the `up` method here, and the `down` method can be determined automagically.

PR #1006, opened 2017-08-17
Add a rake task to clean up a corporate info locale problem

tuzz:
This problem was caused by a bug in Whitehall that has
since been fixed. The script is designed to error out
if unexpected data is encountered.

### DIFF
--- a/lib/tasks/corporate_info_fix.rake
+++ b/lib/tasks/corporate_info_fix.rake
@@ -1,0 +1,47 @@
+desc "Fix a base_path issue with some corporate info pages"
+task corporate_info_fix: :environment do
+  require_relative "../../db/migrate/helpers/delete_content"
+
+  Edition.where(schema_name: "corporate_information_page").find_each do |e|
+    next unless e.locale == "cy"
+    next if e.base_path.end_with?(".cy")
+
+    if ENV["DRY_RUN"]
+      puts [e.base_path, e.locale].inspect
+      next
+    end
+
+    #this edition has an en base_path but a cy locale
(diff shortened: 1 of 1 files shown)


lib/tasks/corporate_info_fix.rake
```diff
@@ -0,0 +1,47 @@
...
+
+      raise "expected redirect" unless edition.schema_name == "redirect"
+      raise "expected en" unless edition.locale == "en"
+
+      if edition.base_path != edition_with_incorrect_base_path.base_path
+        raise "expected matching base_path"
+      end
+
+      puts "deleting #{content_id}"
+      Helpers::DeleteContent.destroy_documents_with_links(content_id)
```
gpeng: Looks like these are showing up as duplicates in rummager too going by https://govuk.zendesk.com/agent/tickets/2304321. We'll need to explicitly remove them from there too although the rummager api removes stuff by `base_path` and the duplicates have the same `base_path` in the search results so I'm not sure how we go about that.

gpeng: Might be easier to do that in https://github.com/alphagov/whitehall/pull/3410 using https://github.com/alphagov/whitehall/blob/master/lib/whitehall/search_index.rb#L6-L18

tuzz:
@thomasalanlee yep, we'll remove it

@gpeng I've just spoken to the Search team and I'm going to run this rake task with those content_ids: https://github.com/alphagov/rummager/blob/master/lib/tasks/delete.rake#L12

PR #1007, opened 2017-08-17
Populate new date fields on Edition

emmabeynon:
We added some new date fields to Edition (see #1005).  This PR populates those fields.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -28,7 +28,6 @@ module Commands
         end
 
         set_public_updated_at
-        set_first_published_at
         set_publishing_request_id
         set_update_type
         set_timestamps
@@ -166,11 +165,6 @@ module Commands
 
       def set_timestamps
         Edition::Timestamps.live_transition(edition, previous_item)
-      end
-
-      def set_first_published_at
-        return if edition.first_published_at.present?
-        edition.update_attributes!(first_published_at: default_datetime)
       end
 
       def default_datetime
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -31,6 +31,7 @@ module Commands
         set_first_published_at
         set_publishing_request_id
         set_update_type
+        set_timestamps
         edition.publish
         remove_access_limit
         create_publish_action
@@ -161,6 +162,10 @@ module Commands
         else
           edition.update_attributes!(public_updated_at: previous_item.public_updated_at)
         end
+      end
+
+      def set_timestamps
+        Edition::Timestamps.live_transition(edition, previous_item)
       end
 
       def set_first_published_at
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -153,7 +153,7 @@ module Commands
       end
 
       def set_timestamps
-        Edition::Timestamps.live_transition(edition, previous_item)
+        Edition::Timestamps.live_transition(edition, update_type, previous_item)
       end
 
       def default_datetime
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -27,7 +27,6 @@ module Commands
           clear_published_items_of_same_locale_and_base_path
         end
 
-        set_public_updated_at
         set_publishing_request_id
         set_update_type
         set_timestamps
@@ -151,16 +150,6 @@ module Commands
           callbacks: callbacks,
           nested: true,
         )
-      end
-
-      def set_public_updated_at
-        return if edition.public_updated_at.present?
-
-        if update_type == "major" || previous_item.blank?
-          edition.update_attributes!(public_updated_at: default_datetime)
-        else
-          edition.update_attributes!(public_updated_at: previous_item.public_updated_at)
-        end
       end
 
       def set_timestamps
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -7,6 +7,7 @@ module Commands
         check_update_type
 
         edition = create_or_update_edition
+        set_timestamps(edition)
 
         update_content_dependencies(edition)
 
@@ -57,6 +58,10 @@ module Commands
         ChangeNote.create_from_edition(payload, edition)
         create_links(edition)
         Action.create_put_content_action(edition, document.locale, event)
+      end
+
+      def set_timestamps(edition)
+        Edition::Timestamps.edited(edition, payload, previously_published_edition)
       end
 
       def check_update_type
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -54,7 +54,6 @@ module Commands
       def update_content_dependencies(edition)
         create_redirect
         access_limit(edition)
-        update_last_edited_at(edition, payload[:last_edited_at])
         ChangeNote.create_from_edition(payload, edition)
         create_links(edition)
         Action.create_put_content_action(edition, document.locale, event)
@@ -148,14 +147,6 @@ module Commands
           callbacks: callbacks,
           nested: true,
         )
-      end
-
-      def update_last_edited_at(edition, last_edited_at = nil)
-        if last_edited_at.nil? && %w(major minor).include?(payload[:update_type])
-          last_edited_at = Time.zone.now
-        end
-
-        edition.update_attributes(last_edited_at: last_edited_at) if last_edited_at
       end
 
       def bulk_publishing?
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -83,7 +83,10 @@ module Commands
       end
 
       def unpublish
-        Edition::Timestamps.live_transition(edition, previous) if edition.draft?
+        if edition.draft?
+          Edition::Timestamps.live_transition(edition, edition.update_type, previous)
+        end
+
         edition.unpublish(
           payload
             .slice(:type, :explanation, :alternative_path, :unpublished_at)
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -83,6 +83,7 @@ module Commands
       end
 
       def unpublish
+        Edition::Timestamps.live_transition(edition, previous) if edition.draft?
         edition.unpublish(
           payload
             .slice(:type, :explanation, :alternative_path, :unpublished_at)
--- a/app/models/create_draft_edition.rb
+++ b/app/models/create_draft_edition.rb
@@ -39,7 +39,6 @@ private
   def fill_out_new_edition
     document.increment!(:stale_lock_version)
     set_first_published_at
-    set_last_edited_at
     set_document_owner
   end
 
@@ -48,14 +47,6 @@ private
     return if edition.first_published_at
     edition.update_attributes(
       first_published_at: previously_published_item.first_published_at,
-    )
-  end
-
-  def set_last_edited_at
-    return unless previously_published_item.has_last_edited_at?
-    return if edition.last_edited_at
-    edition.update_attributes(
-      last_edited_at: previously_published_item.last_edited_at,
     )
   end
 
--- a/app/models/create_draft_edition.rb
+++ b/app/models/create_draft_edition.rb
@@ -38,16 +38,7 @@ private
 
   def fill_out_new_edition
     document.increment!(:stale_lock_version)
-    set_first_published_at
     set_document_owner
-  end
-
-  def set_first_published_at
-    return unless previously_published_item.has_first_published_at?
-    return if edition.first_published_at
-    edition.update_attributes(
-      first_published_at: previously_published_item.first_published_at,
-    )
   end
 
   def set_document_owner
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -20,6 +20,11 @@ class Edition::Timestamps
       edition.major_published_at = previous_live_version&.major_published_at
     end
 
+    # In time we'd like to rename public_updated_at to major_published_at as
+    # part of the payload, but in the meantime we are populating the field
+    # with the optionally provided public_updated_at field.
+    edition.publisher_major_published_at = payload[:public_updated_at]
+
     edition.save!
   end
 
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -7,6 +7,7 @@ class Edition::Timestamps
   )
     edition.temporary_last_edited_at = now
     edition.publisher_last_edited_at = payload[:last_edited_at]
+    edition.last_edited_at = payload[:last_edited_at] || now
 
     # first_published_at should eventually be associated with a document model,
     # therefore avoiding the need to copy between editions as this is fragile
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -13,6 +13,8 @@ class Edition::Timestamps
     # therefore avoiding the need to copy between editions as this is fragile
     edition.temporary_first_published_at = previous_live_version&.temporary_first_published_at
     edition.publisher_first_published_at = payload[:first_published_at]
+    edition.first_published_at = payload[:first_published_at] ||
+      previous_live_version&.first_published_at
 
     # We set this on non-major updates so that editions can maintain their value
     # from previous items. In future we'd like to avoid this date being copied
@@ -43,6 +45,10 @@ class Edition::Timestamps
       edition.temporary_first_published_at = now
     end
 
+    unless edition.first_published_at.present?
+      edition.first_published_at = now
+    end
+
     edition.published_at = now
 
     if edition.update_type == "major"
@@ -54,7 +60,6 @@ class Edition::Timestamps
       # sent in put content.
       edition.major_published_at = previous_live_version&.major_published_at
     end
-
 
     edition.save!
   end
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -39,6 +39,7 @@ class Edition::Timestamps
 
   def self.live_transition(
     edition,
+    update_type,
     previous_live_version = nil,
     now = Time.zone.now
   )
@@ -54,7 +55,7 @@ class Edition::Timestamps
 
     edition.published_at = now
 
-    if edition.update_type == "major"
+    if update_type == "major"
       edition.major_published_at = now
       edition.public_updated_at = now unless edition.public_updated_at.present?
     else
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -5,6 +5,8 @@ class Edition::Timestamps
     previous_live_version = nil,
     now = Time.zone.now
   )
+    edition.temporary_last_edited_at = now
+    
     # first_published_at should eventually be associated with a document model,
     # therefore avoiding the need to copy between editions as this is fragile
     edition.temporary_first_published_at = previous_live_version&.temporary_first_published_at
@@ -24,7 +26,7 @@ class Edition::Timestamps
     # part of the payload, but in the meantime we are populating the field
     # with the optionally provided public_updated_at field.
     edition.publisher_major_published_at = payload[:public_updated_at]
-    
+
     edition.save!
   end
 
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -1,0 +1,28 @@
+class Edition::Timestamps
+  def self.edited(
+    edition,
+    payload,
+    previous_live_version = nil,
+    now = Time.zone.now
+  )
+    # first_published_at should eventually be associated with a document model,
+    # therefore avoiding the need to copy between editions as this is fragile
+    edition.temporary_first_published_at = previous_live_version&.temporary_first_published_at
+
+    edition.save!
+  end
+
+  def self.live_transition(
+    edition,
+    previous_live_version = nil,
+    now = Time.zone.now
+  )
+    # We set this because a lack of temporary_first_published_at indicates the
+    # edition has not been published
+    unless edition.temporary_first_published_at.present?
+      edition.temporary_first_published_at = now
+    end
+
+    edition.save!
+  end
+end
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -8,6 +8,7 @@ class Edition::Timestamps
     # first_published_at should eventually be associated with a document model,
     # therefore avoiding the need to copy between editions as this is fragile
     edition.temporary_first_published_at = previous_live_version&.temporary_first_published_at
+    edition.publisher_first_published_at = payload[:first_published_at]
 
     edition.save!
   end
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -6,7 +6,8 @@ class Edition::Timestamps
     now = Time.zone.now
   )
     edition.temporary_last_edited_at = now
-    
+    edition.publisher_last_edited_at = payload[:last_edited_at]
+
     # first_published_at should eventually be associated with a document model,
     # therefore avoiding the need to copy between editions as this is fragile
     edition.temporary_first_published_at = previous_live_version&.temporary_first_published_at
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -39,6 +39,8 @@ class Edition::Timestamps
       edition.temporary_first_published_at = now
     end
 
+    edition.published_at = now
+
     if edition.update_type == "major"
       edition.major_published_at = now
     else
@@ -48,6 +50,7 @@ class Edition::Timestamps
       # sent in put content.
       edition.major_published_at = previous_live_version&.major_published_at
     end
+
 
     edition.save!
   end
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -31,6 +31,9 @@ class Edition::Timestamps
     # with the optionally provided public_updated_at field.
     edition.publisher_major_published_at = payload[:public_updated_at]
 
+    # If the payload value is nil, we rely on publish to populate public_updated_at
+    edition.public_updated_at = payload[:public_updated_at]
+
     edition.save!
   end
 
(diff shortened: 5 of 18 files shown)


app/models/update_existing_draft_edition.rb
```diff
@@ -60,6 +60,7 @@ def add_extra_attributes(attributes)
     attributes.tap do |x|
```
thomasleese: I wonder if we can simplify this and do something like:

```ruby
%w(publisher_first_published_at publisher_major_published_at publisher_minor_published_at).each do |attribute|
  attributes[attribute] = payload[attribute] if payload[attribute].present?
end
```

app/models/create_draft_edition.rb
```diff
@@ -39,6 +39,10 @@ def document
   def fill_out_new_edition
     document.increment!(:stale_lock_version)
     set_first_published_at
+    set_publisher_first_published_at
```
thomasleese: Similar to in `update_existing_draft_edition.rb` I think we can simplify this with a loop over the attributes and just take them from the payload.

app/models/update_existing_draft_edition.rb
```diff
@@ -79,4 +95,20 @@ def attributes
   def edition_attributes_from_payload
     payload.slice(*Edition::TOP_LEVEL_FIELDS)
   end
+
+  def publisher_first_published_at
```
thomasleese: I don't think we need these methods if we're only using them in one place, I would just take the values straight from the payload.

app/models/create_draft_edition.rb
```diff
@@ -39,6 +39,10 @@ def document
...
+    set_publisher_major_published_at
+    set_publisher_minor_published_at
+    set_publisher_last_edited_at
     set_last_edited_at
     set_document_owner
```
thomasleese: I think we want the `temporary_{major,minor}_published_at` date fields here to be set using the value in the `previously_published_item`. In this way, an edition with a major `update_type` still has the minor date set for the last time that document had a minor update.

app/models/update_existing_draft_edition.rb
```diff
@@ -52,8 +59,15 @@ def assign_attributes_with_defaults
 
   def remove_first_published_at_if_not_explicitly_set(attributes)
     attributes.tap do |x|
-      x.delete(:first_published_at) unless payload.include?(:first_published_at)
+      x.delete(:first_published_at)
+    end
+  end
+
+  def add_extra_attributes(attributes)
```
thomasleese: I would be tempted to name this `add_publisher_date_attributes`.

app/models/update_existing_draft_edition.rb
```diff
@@ -79,4 +95,20 @@ def attributes
   def edition_attributes_from_payload
     payload.slice(*Edition::TOP_LEVEL_FIELDS)
```
thomasleese: Actually, I wonder if we can just combined `TOP_LEVEL_FIELDS` with `ATTRIBUTES_TO_BE_ADDED` here to get a slice including those `publisher_` date fields?

app/models/update_existing_draft_edition.rb
emmabeynon: The problem I have here is that the date field names don't match up with the payload attribute names e.g. the date field is `publisher_first_published_at` and the payload attribute is `first_published_at`.  I think this is correct because the caller shouldn't need to know that the publishing API is treating publisher dates differently to its own dates.  However it does mean that we can't cleanly assign the payload values to the date fields without either removing or adding `publisher_` at some point.  While the methods aren't being re-used, I think at least it does make the relationship clear between the payload attributes and the date fields.  Do you know of a way to better handle the the difference in field and payload names?

thomasleese: Oh, of course, I see what you mean... sorry, I completely missed that distinction.

I wonder if we can have some sort of mapping with something like (untested code):

```ruby
DATE_FIELDS = {
  publisher_major_updated_at: :major_updated_at,
  publisher_minor_updated_at: :minor_updated_at,
  ...
}

attributes = DATE_FIELDS.each_with_object({}) do |hsh, (internal_key, external_key)|
  hsh[internal_key] = payload[external_key] if payload[external_key].present?
end

edition.update_attributes!(attributes)
```

emmabeynon: Ok that sounds good, will give it a shot 👍

app/models/update_existing_draft_edition.rb
emmabeynon: `ATTRIBUTES_TO_BE_ADDED` is now a hash called `DATE_FIELDS` so I won't be able to combine it with `TOP_LEVEL_FIELDS` anymore.

thomasleese: Yep, that wouldn't have worked anyway as you pointed out with the distinction between the payload fields and the database field names.

emmabeynon:
@thomasleese your comments should be addressed now 🚀

app/commands/v2/publish.rb
```diff
@@ -163,6 +167,40 @@ def set_public_updated_at
...
+        end
+
+        if previous_item.present?
+          edition.update_attributes!(
+            temporary_last_edited_at: previous_item.temporary_last_edited_at
+          )
+        end
+      end
+
+      def set_temporary_first_published_at
```
thomasleese: Do we not need to take this from the previous item if it's available?

app/commands/v2/publish.rb
```diff
@@ -163,6 +167,40 @@ def set_public_updated_at
...
+        if update_type.present?
+          edition.update_attributes!(temporary_last_edited_at: default_datetime)
+          return
+        end
+
+        if previous_item.present?
+          edition.update_attributes!(
+            temporary_last_edited_at: previous_item.temporary_last_edited_at
+          )
+        end
```
thomasleese: I wonder if we need to do something here if the `previous_item` is not present, i.e. if it was the first publish? Since the state is being edited, do we need to update that here?

app/models/create_draft_edition.rb
```diff
@@ -13,6 +13,13 @@ def call
 
 private
 
+  DATE_FIELDS = {
```
thomasleese: It's annoying we have this duplication between the create and the update class, I wonder if we can create a subclass or a mixin?

app/commands/v2/publish.rb
```diff
@@ -163,6 +167,40 @@ def set_public_updated_at
         end
       end
 
+      def set_temporary_major_published_at
```
thomasleese: Up to you on this one, but I would maybe reduce this into one method, feels a bit like overkill to me at the moment.

```ruby
def set_temporary_published_at_dates
  edition.update_attributes!(temporary_major_published_at: default_datetime) if update_type == "major"
  edition.update_attributes!(temporary_minor_published_at: default_datetime) if update_type == "minor"
end
```

app/commands/v2/publish.rb
```diff
@@ -163,6 +167,40 @@ def set_public_updated_at
...
+      def set_temporary_minor_published_at
+        return unless update_type == "minor"
+
+        edition.update_attributes!(
+          temporary_minor_published_at: default_datetime
+        )
+      end
+
+      def set_temporary_last_edited_at
+        if update_type.present?
```
thomasleese: I don't think we need this, if the `update_type` isn't present then the publish will fail anyway.

thomasleese:
Sorry, I have more questions! Dates are annoying...

app/commands/v2/publish.rb
emmabeynon: Makes sense 👍

app/commands/v2/publish.rb
emmabeynon: If the `previous_item` isn't present, set it to the default datetime?

thomasleese: Actually I'm wondering if `last_edited_at` should even be modified during a publish.

emmabeynon: As discussed, moving this into PUT content.

app/models/update_existing_draft_edition.rb
```diff
@@ -56,16 +58,27 @@ def remove_first_published_at_if_not_explicitly_set(attributes)
     end
   end
 
+  def add_publisher_date_attributes(attributes)
```
thomasleese: I reckon this method can go in the common module as well.

app/models/update_existing_draft_edition.rb
```diff
@@ -56,16 +58,27 @@ def remove_first_published_at_if_not_explicitly_set(attributes)
...
+    DATE_FIELDS.each do |internal, external|
+      attributes[internal] = payload[external] if payload.include?(external)
+    end
+    attributes
+  end
+
+  def temporary_last_edited_at
```
thomasleese: I would be tempted to name this `temporary_last_edited_at_attribute`.

doc/api.md
```diff
@@ -120,6 +122,10 @@ presented edition and [warnings](#warnings).
 - `locale` *(optional, default: "en")*
   - Accepts: An available locale from the [Rails I18n gem][i18n-gem]
   - Specifies the locale of the edition.
+- `major_published_at` *(optional)*
+  - Specifies when a major update was published.
+- `minor_published_at` *(optional)*
```
kevindew: I don't think we need to add these in yet - these would also require a schema change. I think it's fine for this step in the process to fill these in based on `public_updated_at` 

Good to add `first_published_at` in though since that should have been there before

app/models/publisher_date_fields.rb
```diff
@@ -0,0 +1,17 @@
+module PublisherDateFields
+  DATE_FIELDS = {
+    publisher_first_published_at: :first_published_at,
+    publisher_last_edited_at: :last_edited_at,
+    publisher_major_published_at: :major_published_at,
+    publisher_published_at: :minor_published_at,
```
kevindew: I think this would be just `published_at` still

spec/commands/v2/publish_spec.rb
```diff
@@ -71,6 +71,25 @@
       end
     end
 
+    context "with major update_type sent in the payload" do
+      it "updates the temporary_published_at time to now" do
+        described_class.call(payload)
+
+        expect(draft_item.reload.temporary_published_at)
+          .to be_within(1.second).of(Time.zone.now)
+      end
```
kevindew: to save having to do the within(1.second) you could wrap these in Timecop.freeze blocks
eg: https://github.com/alphagov/publishing-api/blob/be9d4a2bfa70402196480f6573965ffac680fa7a/spec/queries/get_expanded_links_spec.rb#L66-L73

Or if you're doing loads you can do them in before and ends: https://github.com/alphagov/publishing-api/blob/faf9038a2a3ecef67e8b6f66d6539ee1f537c40a/spec/lib/tasks/taxonomy_event_log_spec.rb#L3-L10

app/models/temporary_date_fields.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def set_temporary_date_fields
+    DATE_FIELDS.each_with_object({}) do |attribute, hash|
+      hash[attribute] = previously_published_item.send(attribute)
+    end
+  end
+
+  def set_default_temporary_date_fields
+    DATE_FIELDS.each_with_object({}) do |attribute, hash|
+      hash[attribute] = Time.zone.now
```
kevindew: We'd probably want to avoid using Time.zone.now and instead have a date injected. Otherwise these dates which are basically the same will end up with microsecond differences which have caught people out before.

app/models/publisher_date_fields.rb
emmabeynon: Do you mean `publisher_published_at` should be renamed `published_at`?  Or that the param being sent should be called `published_at`?

kevindew: Sorry: `publisher_published_at: :minor_published_at` => `publisher_published_at: :published_at`

emmabeynon: Cool 👍

kevindew:
This is what I came up with trying to do some consolidating. It's pretty quick and scrappy: https://github.com/alphagov/publishing-api/commit/6f2a119ebe2984b41c7180ff5751f715aa5cc9ca 

Do you think something like this makes dates easier to understand? 

@thomasleese this strikes me as something you might have opinions on too

thomasleese:
@kevindew Hmm... yeah, I do quite like that. I think it makes it easier to follow through the logic of when dates are changed.

kevindew:
@emmabeynon Sorry you always seem to be the victim of me changing my mind....

I've been having a think about the setting of dates in Put Content and think the suggestion I've had where we set dates for draft is all a bit weird now and am less convinced by it. I think @steventux suggestion to handle this when presenting to draft content store is a better approach.

I imagine it's good to make this non ambiguous, so my thoughts are:

## When Put Content runs:

- Set `last_edited_at` to `payload[:last_edited_at]` or current time
- Set `temporary_last_edited_at` to current time
- Set `publisher_last_edited_at` to `payload[:last_edited_at]`
- Set `first_published_at`to `payload[:first_published_at]` or `previous_live_version.first_published_at`
- Set `temporary_first_published_at`to `previous_live_version.first_published_at`
- Set `publisher_first_published_at` to `payload[:first_published_at]`
- Set `public_updated_at` to `payload[:public_updated_at]` or `previous_live_version.public_updated_at`
- Set `temporary_major_published_at` to `previous_live_version.major_published_at` if update type is not "major"
- Set `publisher_major_published_at` to `payload[:public_updated_at]` 
- Don't set `temporary_published_at`
- Set `publisher_published_at` to nil (eventually this would be allowed via input)

## When we publish or unpublish a draft:

- Set `first_published_at`to current time unless it's already set
- Set `temporary_first_published_at`to current time unless it's already set
- Set `public_updated_at` to current time unless it's already set
- Set `temporary_major_published_at` to current time if the update type is "major"
- Set `temporary_published_at`to current time

## WTFs saved

- No longer set published_at dates before item has a state of `published` in Publishing API
- Don't keep updating `temporary_first_published_at` each put content which was a bit weird

## WTFs remaining

- Copying the dates so much seems error prone and complicated
- By allowing a document to bypass major publishing that can have a nil value

## How this could hopefully get simpler later

- Store `first_published_at` on document so it's not passed around each edition
- Associate an edition with a major edition which is used for accessing `major_published_at`

app/commands/v2/publish.rb
```diff
@@ -153,19 +152,10 @@ def clear_published_items_of_same_locale_and_base_path
...
-          edition.update_attributes!(public_updated_at: default_datetime)
-        else
-          edition.update_attributes!(public_updated_at: previous_item.public_updated_at)
-        end
-      end
-
-      def set_first_published_at
-        return if edition.first_published_at.present?
-        edition.update_attributes!(first_published_at: default_datetime)
+      def set_timestamps
```
thomasleese: If this were to happen after `set_update_type` it would mean we can avoid having to pass in the `update_type`, which I think would be nice for when we get rid of it.

spec/integration/put_content/content_with_a_previously_published_edition_spec.rb
```diff
@@ -4,10 +4,21 @@
...
+    Timecop.freeze("2017-01-02 12:23")
+  end
+
+  after do
+    Timecop.return
+  end
+
+  before do
     stub_request(:put, %r{.*content-store.*/content/.*})
+    Timecop.freeze(Time.local(2017, 9, 1, 12, 0, 0))
```
thomasleese: There seems to already be a `Timecop.freeze` above, do we need both?

emmabeynon: We don't need both - will remove one 👍

app/commands/v2/publish.rb
emmabeynon: Good shout 👍

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,72 @@
+class Edition::Timestamps
+  def self.edited(
+    edition,
+    payload,
+    previous_live_version = nil,
+    now = Time.zone.now
+  )
+
```
thomasleese: Rogue empty line!

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,72 @@
...
+    if previous_live_version
```
thomasleese: Can we do `edition.temporary_first_published_at = previous_live_version&.first_published_at` here to avoid the if?

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+    edition.temporary_last_edited_at = now
+
+    edition.publisher_first_published_at = payload[:first_published_at]
+
+    unless edition.update_type == "major"
+      edition.major_published_at = previous_live_version&.major_published_at
+    end
+
+    if payload[:last_edited_at].present?
```
thomasleese: Is it possible to do `edition.last_edited_at = payload[:last_edited_at] || now`?

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,72 @@
...
+      edition.last_edited_at = payload[:last_edited_at]
+    else
+      edition.last_edited_at = now
+    end
+
+    if payload[:first_published_at].present?
```
thomasleese: Similar to above, can we use `||` here?

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,72 @@
...
+      edition.first_published_at = payload[:first_published_at]
+    else
+      edition.first_published_at = previous_live_version&.first_published_at
+    end
+
+    if payload[:public_updated_at].present?
```
thomasleese: And similar to above.

emmabeynon:
@thomasleese your comments should be addressed now ✅

app/commands/v2/put_content.rb
```diff
@@ -53,12 +54,19 @@ def prepare_content_with_base_path
...
         create_redirect
         access_limit(edition)
-        update_last_edited_at(edition, payload[:last_edited_at])
         ChangeNote.create_from_edition(payload, edition)
         create_links(edition)
         Action.create_put_content_action(edition, document.locale, event)
       end
 
+      def set_timestamps(edition)
+        if previously_published_edition.is_a?(PreviouslyPublishedItem)
```
kevindew: ooh this looks like a confusing gotcha! - not your fault of course.

app/commands/v2/unpublish.rb
```diff
@@ -83,6 +83,10 @@ def validate
       end
 
       def unpublish
+        if edition.draft?
+          Edition::Timestamps.live_transition(edition, previous)
```
kevindew: do we have space on the line for this to be `Edition::Timestamps.live_transition(edition, previous) if edition.draft?`

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
+class Edition::Timestamps
+  def self.edited(
+    edition,
+    payload,
+    previous_live_version = nil,
+    now = Time.zone.now
+  )
+    edition.temporary_first_published_at = previous_live_version&.first_published_at
```
kevindew: Since we have that `NoPreviousPublishedItem` class we could in theory accept that into this class and have that return nil.

As I'm typing this I'm wondering if that's a bad idea as it probably creates indirection - be harder still to work out why it's nil - but thought I'd leave this comment to see your thoughts. We end up a bit caught between fighting against a pattern in https://github.com/alphagov/publishing-api/pull/1007/files#diff-d07c485229519b45d396e1797345df7cR63 and having opportunity to continue it here

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+    edition.temporary_last_edited_at = now
+    edition.publisher_first_published_at = payload[:first_published_at]
+
+    unless edition.update_type == "major"
```
kevindew: Might be worth adding a quick comment here as to why we set this when not major as the code reads a little confusing (as the situation is a bit confusing)

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+    edition.first_published_at = payload[:first_published_at] ||
+      previous_live_version&.first_published_at
+
+    edition.last_edited_at = payload[:last_edited_at] || now
+
+    edition.public_updated_at = payload[:public_updated_at] ||
+      previous_live_version&.public_updated_at
+
+    edition.publisher_major_published_at = payload[:public_updated_at]
+    edition.publisher_published_at = nil
```
kevindew: I'm guessing we might not need this set to nil as it will default to that?

kevindew: I know it was probably me who did that in first place too heh

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    edition.save!
+  end
+
+  def self.live_transition(
+    edition,
+    previous_live_version = nil,
+    now = Time.zone.now
+  )
+    unless edition.temporary_first_published_at.present?
```
kevindew: I think this would also be a candidate for a comment to explain why always set even if it's not the first item.

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+      edition.temporary_first_published_at = now
+    end
+
+    if edition.update_type == "major"
+      edition.major_published_at = now
+    else
+      edition.major_published_at = previous_live_version&.major_published_at
```
kevindew: Might be worth adding a comment or something here to explain that the else condition of this can be removed once update_type is no longer part of publish.

app/models/edition/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    unless edition.first_published_at.present?
+      edition.first_published_at = now
+    end
+
+    edition.published_at = now
+
+    unless edition.public_updated_at.present?
```
kevindew: I'm wondering if we've not quite got our logic matching for public_updated_at

compared to https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/publish.rb#L156-L164

I think we may only want to set it in [put content](https://github.com/alphagov/publishing-api/pull/1007/files#diff-e4282e225c34744d59e36393ea78c0feR21) when a value is provided and not use the `or` condition.

I feel that currently we'd have a situation where this is no longer set automatically.

Consider:

I have document 1 with edition A
I create edition B with an update type of major
And expect when I publish it to to have a public_updated_at of now

What I think happens is:
edition B has a put content without a public_updated_at so it is given the one of edition A
edition B is then published and public_updted_at is not set to now

Does that sound like a problem we have?

app/commands/v2/put_content.rb
emmabeynon: Yeah - I really don't like how PreviouslyPublishedItem works :(

app/commands/v2/unpublish.rb
emmabeynon: Yep just about :)

app/commands/v2/put_content.rb
kevindew: Yeah it's quite confusing as we don't make use of that pattern much so you'd think by the naming it'd be a (item || nil) value. 

If you've not come across this approach before this is a write up of the pattern: https://medium.com/@franzejr/null-object-pattern-in-ruby-12a809bc75cc

app/models/edition/timestamps.rb
emmabeynon: I did wonder, and thought maybe it's worth making it explicit in the code that we do want it to be nil.  Equally, I think the tests should make that clear, so... *shrugs*

app/commands/v2/put_content.rb
emmabeynon: This article makes it clearer to me 👍

app/models/edition/timestamps.rb
emmabeynon: At first, I felt like Edition::Timestamps shouldn't need to know about `NoPreviousPublishedItem` because it can't do much with it anyway e.g. `NoPreviousPublishedItem` will never have a `first_published_at` so what's the point in trying to call that on it.
Then, after reading the article about the Null Object Pattern, I can see the value in that.  I suppose my concern is having a bunch of methods that just return nil is a bit annoying.  We could alias them perhaps like this https://stackoverflow.com/questions/19979100/alias-a-method-multiple-times ... or is that really gross?

app/models/edition/timestamps.rb
emmabeynon: Just trying to remember the reasoning - was it for data completeness?

app/models/edition/timestamps.rb
emmabeynon: Ok so we need to update `public_updated_at` to current time if `update_type == major` or if it's not already present?

app/models/edition/timestamps.rb
kevindew: I wouldn't be too worried about the bunch of methods. The whole object is a bunch of methods that return effectively null or default values. 

As they return nil I think you could define them as simply as `def first_published_at; end`

app/models/edition/timestamps.rb
kevindew: Yeah I think best to just remove it. Leaving it in only makes you wonder what it might have been.

app/models/edition/timestamps.rb
kevindew: It's because we always to have a first_published_at value when something has been published. 
We assume lack of one means that something has not been published.
In future we'd like to move this onto the document model so it doesn't have to be passed around so there is less scope for error here.

app/models/edition/timestamps.rb
kevindew: there's two parts to this (and hopefully I'm not contradicting something I said before :pray:)

In put content we should only set public_updated_at if a user has given it to us. 

When going live we should check is there a public_updated_at. If there is not it should be set to now if update_type is major otherwise previous items.

emmabeynon: Ok I think I've got it - I've pushed up some changes

emmabeynon:
@kevindew All of your comments should be addressed now 👍

spec/integration/put_content/content_with_a_previously_published_edition_spec.rb
```diff
@@ -42,14 +51,58 @@
...
+
+  it "sets temporary_first_published_at to the previously published version's value" do
+    put "/v2/content/#{content_id}", params: payload.to_json
+
+    edition = Edition.last
+    expect(edition.temporary_first_published_at)
+      .to eq(temporary_first_published_at)
+  end
+
+  context "and update_type is minor" do
```
kevindew: I think this should probably be a "when update_type is minor" as it doesn't appear to already be nested in a context block.

spec/integration/put_content/content_with_a_previously_published_edition_spec.rb
```diff
@@ -42,14 +51,58 @@
...
+    it "sets major_published_at to previous published edition's value" do
+      put "/v2/content/#{content_id}", params: payload.to_json
+
+      edition = Edition.last
+      expect(edition.major_published_at.iso8601)
+        .to eq(major_published_at.iso8601)
+    end
+  end
+
+  context "and first_published_at has changed in the payload" do
```
kevindew: same comment about s/and/when here

spec/requests/unpublishing_spec.rb
```diff
@@ -107,7 +107,6 @@
           schema_name: "redirect",
           base_path: base_path,
           publishing_app: edition.publishing_app,
-          public_updated_at: Time.zone.now.iso8601,
```
kevindew: I'm not sure this is related to the commit that changes it. I'm also guessing it has no effect since no tests needed changing - but could be wrong there?

PR #1009, opened 2017-08-21
Update requeue task to requeue by document type

MatMoore:
This task was added when Finding Things team migrated all the links to the
publishing api, to sync everything back to Rummager.

We now need something similar to synchronise all existing content (not just links), so that
Rummager can be a direct downstream consumer of publishing api.

This PR includes a few implementation changes:

1. Use non persistent messages, because we want requeues to be as fast and low
   overhead as possible. Requeued messages are less important than normal
   publishing messages and should be abandoned if RabbitMQ goes down.

2. Use a separate routing key - `*.bulk.reindex` - to indicate the intent (see
   removed comment). This allows us to define a separate non-durable queue
   for these messages in Rummager; currently the `rummager_govuk_index`
   receives everything and is set up as `durable` to survive server restarts.

3. Filter on `content_store` instead of `state`, because we care about
   unpublished (especially withdrawn) editions as well as published.

4. Filter by document type for now, as we don't need the whole lot yet. We may
   add a 'requeue everything' task back in later.

Part of https://trello.com/c/5d4Qkt31/244-bulk-requeue-content-by-format / https://trello.com/c/VQsLP52d/246-create-rake-task-to-bulk-republish-by-format

### DIFF
--- a/lib/requeue_content.rb
+++ b/lib/requeue_content.rb
@@ -1,39 +1,33 @@
 class RequeueContent
-  def initialize(number_of_items: nil)
-    @number_of_items = number_of_items
+  def initialize(scope)
+    # Restrict scope to stuff that's live (published or unpublished)
+    # Unpublished content without a content store representation won't
+    # be returned, but we're not interested in this content.
+    @scope = scope.where(content_store: :live)
+    @version = Event.maximum(:id)
   end
 
-  attr_accessor :number_of_items
-
   def call
-    if number_of_items.present?
-      Edition.where(state: :published).limit(number_of_items).each do |edition|
-        publish_to_queue(edition)
-      end
-    else
-      Edition.where(state: :published).find_each do |edition|
-        publish_to_queue(edition)
-      end
+    scope.each do |edition|
+      publish_to_queue(edition)
     end
   end
 
 private
 
+  attr_reader :scope, :version
+
   def publish_to_queue(edition)
-    version = Event.maximum(:id)
+    presenter = DownstreamPayload.new(edition, version, draft: false)
+    queue_payload = presenter.message_queue_payload
+    service = PublishingAPI.service(:queue_publisher)
 
-    queue_payload = Presenters::EditionPresenter.new(
-      edition, draft: false,
-    ).for_message_queue(version)
-
-    # FIXME: Rummager currently only listens to the message queue for the
-    # event type 'links'. This behaviour will eventually be updated so that
-    # it listens to other update types as well. This will happen as part of
-    # ongoing architectural work to make the message queue the sole source of
-    # search index updates. When that happens, the event_type below should
-    # be changed - perhaps to a newly introduced, more-appropriately named
-    # one. Maybe something like 'reindex'.
-
-    PublishingAPI.service(:queue_publisher).send_message(queue_payload, event_type: "links")
+    # Requeue is considered a different event_type to major, minor etc
+    # because we don't want to send additional email alerts to users.
+    service.send_message(
+      queue_payload,
+      routing_key: "#{edition.schema_name}.bulk.reindex",
+      persistent: false
+    )
   end
 end
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -28,8 +28,16 @@ namespace :queue do
     end
   end
 
-  desc "Add published editions to the message queue, optionally specifying a limit on the number of items"
-  task :requeue_content, [:number_of_items] => :environment do |_, args|
-    RequeueContent.new(number_of_items: args[:number_of_items]).call
+  desc "Add published editions to the message queue by document type"
+  task :requeue_document_type, [:document_type] => :environment do |_, args|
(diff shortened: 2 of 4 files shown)


lib/tasks/queue.rake
```diff
@@ -28,8 +28,16 @@ namespace :queue do
     end
   end
 
-  desc "Add published editions to the message queue, optionally specifying a limit on the number of items"
-  task :requeue_content, [:number_of_items] => :environment do |_, args|
-    RequeueContent.new(number_of_items: args[:number_of_items]).call
+  desc "Add published editions to the message queue by document type"
+  task :requeue_document_type, [:document_type] => :environment do |_, args|
+    document_type = args[:document_type]
+    raise ValueError("expecting document_type") if document_type.nil?
```
thomasleese: I would do `unless document_type.present?` because it catches empty strings too.

MatMoore:
I've only tried it on dev so far, but will test the branch on integration before merging as I also want to add in a new queue on the consumer side.

lib/requeue_content.rb
```diff
@@ -1,39 +1,30 @@
...
 
 private
 
-  def publish_to_queue(edition)
-    version = Event.maximum(:id)
-
-    queue_payload = Presenters::EditionPresenter.new(
-      edition, draft: false,
-    ).for_message_queue(version)
+  attr_reader :scope
```
thomasleese: This can be one line, `attr_reader :scope, :version`.

MatMoore:
@thomasleese I think there might be another problem, which is that when binding queues to the exchange, we can't easily distinguish between a routing key of `document_type.requeue (these messages)` and `document_type.major/minor/links/unpublish` - at least without setting up multiple bindings. And our puppet module [doesn't seem to work with multiple bindings per queue (yet)](https://github.com/voxpupuli/puppet-rabbitmq/pull/504).

I'd like to route these messages to a separate transient queue, rather than having our existing queue handle them, and this would get in the way of that.

My suggestion is to change the routing key to 3 words instead of 2:
`document_type.requeue.bulk`. Then we could bind the main rummager queue to `*.*`. Do you see any problem with adding a routing key of this format?

thomasleese:
@MatMoore I would say there's nothing wrong with a three word routing key _for the moment_.

In the future we're planning to do some clean up on the naming of the routing keys so they're more consistent (https://trello.com/c/OqyPl7pc/534-stop-messagequeueeventtype-from-using-update-type), do you think there is anything we should be aware of now in preparation for that? Perhaps when we get round to that, the puppet module will support multiple bindings per queue.

MatMoore:
@thomasleese Cool, I'll proceed with the 3 word key then for now.

Looking at the proposal in the card though - should I be distinguishing between `bulk_republish` and `bulk_reindex`? I've gone with `*.requeue.bulk` here but it could also be `*.reindex.bulk` or `*.bulk.reindex`.

A couple of thoughts on the proposed renames:

Rummager now needs to subscribe to the set of {`major_update`, `minor_update`, `links_update`, `unpublish`} - basically all publishing actions - but I'm assuming it shouldn't subscribe to {`bulk_reindex`, `bulk_republish`}. I think if we were to ever hook up the content store to the message queue it would also subscribe to the same things as well, so I think if we're renaming, it would help a lot to have something common in the routing key for all of those actions.

So if we have `guide.action.major_update`, `guide.action.links_update`, vs `guide.bulk.reindex` we'd bind to `*.action.*` and `*.bulk.reindex`.

Multiple bindings would also do the job, if supported, but wildcard matching is simpler and we can definitely do it.

thomasleese:
@MatMoore I would go for `reindex` actually, since that's what we called it in the proposal. I don't think it matters at the moment whether it's `reindex.bulk` or `bulk.index`.

I'll copy your notes into the card on renaming the routing keys as that's useful information to have, thanks!

MatMoore:
@thomasleese Cool, I've made that change.

lib/requeue_content.rb
```diff
@@ -1,39 +1,33 @@
...
+    presenter = DownstreamPayload.new(edition, version, draft: false)
+    queue_payload = presenter.message_queue_payload
+    service = PublishingAPI.service(:queue_publisher)
 
-    PublishingAPI.service(:queue_publisher).send_message(queue_payload, event_type: "links")
+    # Requeue is considered a different event_type to major, minor etc
+    # because we don't want to send additional email alerts to users.
+    service.send_message(
+      queue_payload,
+      routing_key: "#{edition.schema_name}.bulk.reindex",
```
thomasleese: Do you know if it's possible to do `event_type: "bulk.reindex"` here? AFAIK it's just string interpolation.

MatMoore: I guess so but I thought this was less confusing as you can see what the key looks like all in one place.

thomasleese: Yeah, I see what you mean. It's annoying having to include the duplication of `edition.schema_name` though. It's up to you, I'll approve the PR since it's not major.

PR #1010, opened 2017-08-22
[WIP] Use Sentry for error reporting

tijmenb:
https://trello.com/c/2AP3O9dn


thomasleese:
🤖 🤖 🤖 ✅

thomasleese:
I realise this is WIP so perhaps it's not a concern, but is there a suitable replacement for `Airbrake.notify_sync`? We use it in the consistency checker to ensure the error makes it through to Errbit at the moment (as it's the last thing the process does so we found the process was terminating before sending the error). Perhaps this isn't a problem with `raven-ruby`.

thomasleese:
Ignore that... I can see there is. Not sure how I missed it the first time.

tijmenb:
Hmm, that's interesting. Raven does sync by default as far as I know, so we wouldn't need this? 

Btw `GOVUK::Error.notify_sync` doesn't actually exist, that's sloppy review on my part.

thomasleese:
Ah! Yeah, if it's synchronous by default then we wouldn't need it, but then I wonder if we would want that behaviour by default since it's that's different to how `Airbrake.notify` currently works?

tijmenb:
I'm tempted to go with Sentry's default behaviour until we see a problem. Does that sounds allright?

thomasleese:
Yep, fine by me 👍

PR #1015, opened 2017-08-25
Set first_published_at from edition history

steventux:
https://trello.com/c/BHzWqzPF/1043-3-sort-out-the-same-firstpublishedat-date-for-1-4-of-the-database

We have 800000 records with a `first_published_at` on 2016-02-29,
a quarter of these are placeholders, just over 200000 have a
'First published' change history item, so use the date from this
to set the correct `first_published_at` value.

### DIFF
--- a/db/migrate/20170901095633_repopulate_first_published_at_from_change_history.rb
+++ b/db/migrate/20170901095633_repopulate_first_published_at_from_change_history.rb
@@ -1,0 +1,42 @@
+class RepopulateFirstPublishedAtFromChangeHistory < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    min_date = '2016-02-29T09:24:09'
+    max_date = '2016-02-29T09:24:11'
+    document_ids = []
+    statements_ary = []
+
+    sql = <<-SQL.strip_heredoc
+      SELECT e.id,
+             e.document_id,
+             hist->>'public_timestamp'
+      FROM   editions e,
+             json_array_elements(details->'change_history') hist
+      WHERE  e.document_type != 'placeholder'
+      AND    e.first_published_at BETWEEN '#{min_date}' AND '#{max_date}'
+      AND    hist IS NOT NULL
+      AND    hist->>'note' = 'First published.'
+      AND    hist->>'public_timestamp' NOT BETWEEN '#{min_date}' AND '#{max_date}'
+    SQL
+
+    # ~224000 records
+    results = ActiveRecord::Base.connection.execute(sql).values
+    puts "#{results.size} records found with change history and first_published_at on 2016-02-29."
+
+    results.each do |edition_id, document_id, timestamp|
+      statements_ary << "UPDATE editions SET first_published_at = '#{timestamp}' WHERE id = #{edition_id};"
+      document_ids << document_id
+    end
+
+    statements_ary.each_slice(1000).each do |statements|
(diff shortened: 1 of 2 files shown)


db/migrate/20170825144733_repopulate_first_published_at_from_change_history.rb
```diff
@@ -0,0 +1,39 @@
+class RepopulateFirstPublishedAtFromChangeHistory < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    min_date = '2016-02-29'
+    max_date = '2016-02-29T23:59:59'
```
kevindew: I'm not sure you have to be dealing with a 24 hour window here. I think it's all the same second.

```
[3] pry(main)> Edition.where("first_published_at > ? AND first_published_at < ?", "2016-02-29", "2016-02-29T23:59:59").count
   (1055.6ms)  SELECT COUNT(*) FROM "editions" WHERE (first_published_at > '2016-02-29' AND first_published_at < '2016-02-29T23:59:59')
=> 789244
```

```
[12] pry(main)> Edition.where("first_published_at > ? AND first_published_at <= ?", "2016-02-29T09:24:09", "2016-02-29T09:24:11").count
   (1056.2ms)  SELECT COUNT(*) FROM "editions" WHERE (first_published_at > '2016-02-29T09:24:09' AND first_published_at <= '2016-02-29T09:24:11')
=> 788272
```

You might know better than me at this point though whether the whole day is a problem or just that second.

steventux: I think you're right, somewhere in analysing this I got the idea there were 2 different bulk update times that day, evidently not though:
```
SELECT COUNT(*) AS count_all, "editions"."first_published_at" AS editions_first_published_at FROM "editions" WHERE (first_published_at BETWEEN '2016-02-29' AND '2016-02-29T23:59:59') GROUP BY "editions"."first_published_at" HAVING (count(*) > 1000)
=> {Mon, 29 Feb 2016 09:24:10 UTC +00:00=>809051}
```

db/migrate/20170825144733_repopulate_first_published_at_from_change_history.rb
```diff
@@ -0,0 +1,39 @@
...
+      AND    hist->>'public_timestamp' NOT BETWEEN '#{min_date}' AND '#{max_date}'
+    SQL
+
+    # ~224000 records
+    results = ActiveRecord::Base.connection.execute(sql).values
+    puts "#{results.size} records found with change history and first_published_at on 2016-02-29."
+
+    # ~395 seconds
+    results.each do |edition_id, document_id, timestamp|
+      sql = "UPDATE editions SET first_published_at = '#{timestamp}' WHERE id = #{edition_id};"
```
thomasleese: If the slowness here is problematic, I wonder if we can make this faster by grouping lots of queries together with semi-colons into batches and then calling `execute` only a few times.

steventux: :+1: I'll batch them up, this migration isn't blocking but I'm sure it can be run quicker than 6 minutes.

steventux:
@thomasleese It was just over 2 minutes quicker, the initial find is slow because of the json column lookups. I've rebased this (and moved the migration timestamp on because of conflicts) so could you re-review?

PR #1017, opened 2017-08-30
Make the Bulk Links endpoint ready for production

whoojemaflip:
This is to productionise the [bulk links endpoint](https://github.com/alphagov/publishing-api/pull/1008).

This endpoint has changed to respond to POST requests, to accommodate larger payloads.  For example:

```
curl -d "content_ids[]=e1067450-7d13-45ff-ada4-5e3dd4025fb7&content_ids[]=72ed754c-4c82-415f-914a-ab6760454cb4" localhost:3093/v2/bulk-links
```

In order to aid the efficient parsing of responses, the output has changed to a hash of content_ids => links. For example:

```
{:"e1067450-7d13-45ff-ada4-5e3dd4025fb7"=>
  {:links=>
    {:mainstream_browse_pages=>["749762c5-0e80-4d04-b766-9850192764d4"],
     :meets_user_needs=>
      ["8042330b-d973-41c1-991c-dffada2950a3",
       "436205fc-1848-41c3-b0df-359ecfbb78b2"],
     :ordered_related_items=>
      ["124edf01-5637-43d1-9c05-fe22530d54b4",
       "b589f602-7427-4ac0-b220-ee35f711548b",
       "f9d0c33a-5b1f-4ec3-8226-5eb665f33b54"],
     :organisations=>["04148522-b0c1-4137-b687-5f3c3bdd561a"],
     :parent=>["749762c5-0e80-4d04-b766-9850192764d4"],
     :taxons=>["13bba81c-b2b1-4b13-a3de-b24748977198"]},
   :version=>10},
 :"72ed754c-4c82-415f-914a-ab6760454cb4"=>
  {:links=>
    {:mainstream_browse_pages=>["26c60c2f-52f7-4da6-98fc-0183776f7fdc"],
     :ordered_related_items=>
      ["1e333395-5dd5-4452-96a3-fbe939928761",
       "e2044da3-5e2f-4fdf-bfb0-4aa5b4357c02"],
     :organisations=>["04148522-b0c1-4137-b687-5f3c3bdd561a"],
     :parent=>["26c60c2f-52f7-4da6-98fc-0183776f7fdc"],
     :taxons=>["13bba81c-b2b1-4b13-a3de-b24748977198"]},
   :version=>15}}
```

[Trello](https://trello.com/c/H3BqIPWc/190-current-tags-are-visible-in-the-bulk-tagging-tool)

### DIFF
--- a/app/queries/get_bulk_links.rb
+++ b/app/queries/get_bulk_links.rb
@@ -1,18 +1,40 @@
 module Queries
-  module GetBulkLinks
-    def self.call(content_ids = [])
-      content_ids.each_with_object({}) do |content_id, hsh|
-        hsh[content_id] = link_set(content_id)
+  class GetBulkLinks
+    def self.call(content_ids)
+      new(content_ids).call
+    end
+
+    def initialize(content_ids)
+      @content_ids = content_ids
+    end
+
+    def call
+      content_ids.each_with_object({}) do |content_id, obj|
+        obj[content_id] = link_set(content_id)
       end
     end
 
-    def self.link_set(content_id)
-      link_set = LinkSet.find_by(content_id: content_id)
+  private
+
+    attr_reader :content_ids
+
+    def link_set(content_id)
+      link_set = link_sets[content_id]
       return { links: {}, version: 0 } unless link_set
 
       Presenters::Queries::LinkSetPresenter
         .present(link_set)
         .slice(:links, :version)
+    end
+
+    def link_sets
+      @_link_sets ||= begin
+        LinkSet
+          .includes(:links) # avoid an N+1 in the presenter class
+          .joins(:links) # combines with the `include` to make a single SQL statement
+          .where("content_id IN (?)", content_ids)
+          .index_by(&:content_id) # reform the Relation into a hash, keyed by content_id
+      end
     end
   end
 end
--- a/app/queries/get_bulk_links.rb
+++ b/app/queries/get_bulk_links.rb
@@ -1,0 +1,18 @@
+module Queries
+  module GetBulkLinks
+    def self.call(content_ids = [])
+      content_ids.each_with_object({}) do |content_id, hsh|
+        hsh[content_id] = link_set(content_id)
+      end
+    end
+
+    def self.link_set(content_id)
+      link_set = LinkSet.find_by(content_id: content_id)
+      return { links: {}, version: 0 } unless link_set
+
+      Presenters::Queries::LinkSetPresenter
+        .present(link_set)
+        .slice(:links, :version)
+    end
+  end
+end
--- a/app/controllers/v2/link_sets_controller.rb
+++ b/app/controllers/v2/link_sets_controller.rb
@@ -9,10 +9,7 @@ module V2
         raise "This experimental endpoint has been disabled, please ask Taxonomy team to remove it."
       end
 
-      json = params.fetch(:content_ids).map do |content_id|
-        Queries::GetLinkSet.call(content_id)
-      end
-
+      json = Queries::GetBulkLinks.call(content_ids)
       render json: json
     end
 
@@ -50,6 +47,10 @@ module V2
 
     def generate?
       ActiveModel::Type::Boolean.new.cast(params.fetch(:generate, false))
+    end
+
+    def content_ids
+      params.fetch(:content_ids)
     end
 
     def links_params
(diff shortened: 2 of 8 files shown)


app/queries/get_bulk_links.rb
```diff
@@ -0,0 +1,22 @@
+module Queries
+  module GetBulkLinks
+    def self.call(content_ids = [])
+      content_ids.reduce({}) do |acc, content_id|
```
thomasleese: I would use `.each_with_object` here and then you shouldn't need the `acc.tap`.

app/queries/get_bulk_links.rb
```diff
@@ -0,0 +1,22 @@
...
+        acc.tap { acc[content_id] = link_set(content_id) }
+      end
+    end
+
+    def self.link_set(content_id)
+      if (link_set = LinkSet.find_by(content_id: content_id))
+        Presenters::Queries::LinkSetPresenter
+          .present(link_set)
+          .slice(:links, :version)
+      else
```
thomasleese: I think I would structure this as an early return rather than an `else`.

```ruby
return { links: {}, version: 0 } unless link_set
```

whoojemaflip:
@thomasleese All good points & addressed.

whoojemaflip:
@thomasleese I concur.

kevindew:
Hey there, code looks good 👍 

Biggest concern I have is that the unlimited input amount means effectively an unlimited output (unlimited data outputs are the bane of our lives on Pub API). Is there scope to put a limit on the number of content_ids in? We might be able to work that out by the amount of queries we consider reasonable for a request, as I think this takes 2 queries per item so if limit was 1000 we’d have a rather painful 2000 queries running.

Other things could be that this is something calculated asynchronously but that feels like overkill.

I’m also wondering about the naming too. We do have a similar endpoint [/lookup-by-base-path](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-lookup-by-base-path) though I don’t think that’s particularly well named either so hesitant to suggest consistency with that.

I kinda like the idea that it’s handled as a nested route within the `/v2/links` segment, that feels more RESTful to me? In many ways this endpoint is very similar to index endpoint for `/v2/links` and just setting content_ids as a filter, but I guess there’s not the need to create that for this.

So some options within `/v2/links` could be:
- `/v2/links/by-content-id` - quite descriptive, limits expansion
- `/v2/links/bulk-lookup` - annoyingly generic, keeps door open for expansion

What do you think? I was going to suggest looking at what we have to see what seems consistent - but they seem a bit of a mess unfortunately.

tijmenb:
I went with the `/v2/bulk` prefix because I had a vague idea that that could be a pattern to provide consistent bulk-versions of endpoints.

So `GET /v2/bulk/links?content_ids[]=content-id-1` returns the same thing as `GET /v2/links/content-id-1`, but in an array (or hash).
 
The rule would apply for other `/:content_id` endpoints (https://github.com/alphagov/publishing-api/blob/master/config/routes.rb#L18-L30): 

`PUT /bulk/content` can be used to update multiple content items,  `PATCH /bulk/links` to update links, `POST /bulk/content/publish` to publish multiple items, etc etc.

But I don't feel strongly about this, haven't considered any other alternatives.

kevindew:
@tijmenb that makes sense - thanks for explaining.

I'd be of the opinion that we'd want to exhaust doing things in a more RESTful `resource/action` style before exploring that. Feels more like something we should consider to group a bunch of similar items than something to introduce in isolation.

whoojemaflip:
@kevindew I've refactored the Query object so that it makes a max of two DB queries. They might be massive queries, but at least it's only two. Both are SELECT lookups against indexed columns, so I would expect the additional load on the service to be minimal.  

If it's OK with you I'd prefer to take a rain-check on limiting the size of payload for this endpoint until we have a better idea of how it will be used. It could potentially create a bunch of annoying client-side workarounds that probably won't be needed once we completely understand the user needs for the consuming service. If we notice any problems caused by excessive payloads or memory usage, it should be easy enough to drop in a limit at that point. 

What do you think?

kevindew:
@whoojemaflip great thanks for refactoring - looks like it'd be n + 1 queries though right rather than 2? due to this https://github.com/alphagov/publishing-api/blob/master/app/presenters/queries/link_set_presenter.rb#L22 unless I'm missing something.

Sorry to be a pain but would prefer we started with a limit as normally once things become a problem they're far too baked in that they're easy to solve. Is 1000 reasonable? can always amend it as your knowledge of needs grows.

whoojemaflip:
@kevindew Your feedback has been addressed, sir.

PR #1018, opened 2017-08-30
Migration to remove orphaned links

kevindew:
This should not be merged until https://github.com/alphagov/publishing-api/pull/1016 has been deployed

This removes Links which have an edition_id of nil and a link_set_id of
nil. Which were created by us not understanding the default behaviour of
`edition.links.delete_all`.

This deletes approximately 3 million rows, I did consider running it
through a find_in_batches but it only took 5 seconds to run on my
machine and as these are orphaned I don't think this will lock anything
that's used.


kevindew:
Sorry @thomasleese going to need another tick on this

PR #1020, opened 2017-08-31
Add major_published_at to Edition

emmabeynon:

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -8,6 +8,7 @@ module Presenters
       :document_id,
       :id,
       :last_edited_at,
+      :major_published_at,
       :publisher_first_published_at,
       :publisher_major_published_at,
       :publisher_published_at,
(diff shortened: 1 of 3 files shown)


kevindew:
How does this fit in with temporary_major_published_at ? I assume we need one of them rather than both - but I might be missing something

emmabeynon:
> How does this fit in with temporary_major_published_at ? I assume we need one of them rather than both - but I might be missing something

My understanding from our chat earlier was that `temporary_major_published_at` would be set to the previous item’s `major_published_at`, which we don’t have at the moment.

kevindew:
> My understanding from our chat earlier was that temporary_major_published_at would be set to the previous item’s major_published_at, which we don’t have at the moment.

Gotcha... Yeah that would be the same value as temporary_major_published_at, so I guess either we'd want to alias the `major_published_at` method to point to `temporary_major_published_at` or to replace `temporary_major_published_at` with `major_published_at`.

We have same situation with `published_at` as well I guess, where it has a `temporary_` prefix in the db but doesn't conflict with another name so may not actually need it.

emmabeynon:
> Gotcha... Yeah that would be the same value as temporary_major_published_at, so I guess either we'd want to alias the major_published_at method to point to temporary_major_published_at or to replace temporary_major_published_at with major_published_at.

I think it's probably easiest to rename `temporary_major_published_at` to `major_published_at`.  What do you think?

emmabeynon:
Closed in favour of https://github.com/alphagov/publishing-api/pull/1022

PR #1024, opened 2017-09-01
Record changes in links through patch_links

koetsier:
After each call to Patch Links, the Links added and Removed are recorded in a new LinkChange model. The source- and target_content_id are recorded, the link type and if the change was an addition (change=1) or a removal (change=-1), the index of the record and the related Action

The LinkChanges are exposed through a new API call (/v2/link-changes?start=<id>) which outputs the LinkChanges in JSON. The number of records can be limited through the start parameter which takes the ID of the first LinkChange to fetch

Trello: https://trello.com/c/162iqEul/185-update-change-history-every-minute-from-publishing-api

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -9,7 +9,7 @@ module Commands
 
         link_set.increment!(:stale_lock_version)
 
-        links_before_patch = link_set.links.map(&:target_content_id)
+        before_links = link_set.links.to_a
 
         grouped_links.each do |group, payload_content_ids|
           # For each set of links in a LinkSet scoped by link_type, this iterator
@@ -25,13 +25,16 @@ module Commands
         # we need to reload the link_set as the links association will be stale
         link_set.reload
 
-        orphaned_content_ids = link_diff_between(links_before_patch, link_set.links.map(&:target_content_id))
+        orphaned_content_ids = link_diff_between(before_links.map(&:target_content_id), link_set.links.map(&:target_content_id))
 
         after_transaction_commit do
           send_downstream(orphaned_content_ids)
         end
 
-        Action.create_patch_link_set_action(link_set, event)
+        action = Action.create_patch_link_set_action(link_set, event)
+
+        after_links = link_set.links.to_a
+        LinkChangeService.new(action, before_links, after_links).record
 
         presented = Presenters::Queries::LinkSetPresenter.present(link_set)
         Success.new(presented)
--- a/app/controllers/v2/link_changes_controller.rb
+++ b/app/controllers/v2/link_changes_controller.rb
@@ -1,0 +1,7 @@
+module V2
+  class LinkChangesController < ApplicationController
+    def index
+      render json: Queries::GetLinkChanges.new(params).as_json
+    end
+  end
+end
--- a/app/controllers/v2/link_changes_controller.rb
+++ b/app/controllers/v2/link_changes_controller.rb
@@ -1,7 +1,7 @@
 module V2
   class LinkChangesController < ApplicationController
     def index
-      render json: Queries::GetLinkChanges.new(params).as_json
+      render json: Queries::GetLinkChanges.new(params, request.original_url).as_json
     end
   end
 end
--- a/db/migrate/20170905123649_create_link_changes.rb
+++ b/db/migrate/20170905123649_create_link_changes.rb
@@ -1,0 +1,13 @@
+class CreateLinkChanges < ActiveRecord::Migration[5.1]
+  def change
+    create_table :link_changes do |t|
+      t.uuid :source_content_id, null: false
+      t.uuid :target_content_id, null: false
+      t.string :link_type, null: false
+      t.integer :change, null: false
+      t.references :action, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170901095633) do
+ActiveRecord::Schema.define(version: 20170905123649) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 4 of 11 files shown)


app/controllers/v2/link_changes_controller.rb
```diff
@@ -0,0 +1,9 @@
+module V2
+  class LinkChangesController < ApplicationController
+    def search
+      result = LinkChange
+      result = result.where("ID > ?", params[:start]) unless params[:start].nil?
```
tijmenb: To make Kevin happy, can we limit the number of results returned here?

app/commands/v2/patch_link_set.rb
```diff
@@ -25,19 +25,38 @@ def call
...
+        record_link_changes(before_links, link_set.links.to_a)
+
         presented = Presenters::Queries::LinkSetPresenter.present(link_set)
         Success.new(presented)
       end
 
-    private
+      private
+
+      def record_link_changes(before_links, after_links)
```
tijmenb: I think this method is sufficiently complex to pull out into a class. That may make it easier to refactor too.

app/commands/v2/patch_link_set.rb
```diff
@@ -25,20 +25,39 @@ def call
...
 
+        record_link_changes(before_links, link_set.links.to_a)
+
         presented = Presenters::Queries::LinkSetPresenter.present(link_set)
         Success.new(presented)
       end
 
     private
 
+      def record_link_changes(before_links, after_links)
```
tijmenb: It may make sense to pull this method out into a class, it's pretty complex.

app/commands/v2/patch_link_set.rb
```diff
@@ -25,20 +25,39 @@ def call
...
+          create_link_change(link: link, change: 1) unless before_set.include?([link.target_content_id, link.link_type])
+        end
+
+        before_links.each do |link|
+          create_link_change(link: link, change: -1) unless after_set.include?([link.target_content_id, link.link_type])
+        end
+      end
+
+      def create_link_change(link:, change:)
+        LinkChange.create!(source_content_id: content_id, target_content_id: link.target_content_id, link_type: link.link_type, change: change)
```
tijmenb: I think we're going to need to associate the `LinkChange` with the related `Action` so we know the user that performed the action.

app/controllers/v2/link_changes_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+      render json: result.all.as_json(only: %w[id source_content_id target_content_id link_type change])
```
tijmenb: Can we expose the `created_at` here too? We'll need that for the interface.

app/controllers/v2/link_changes_controller.rb
```diff
@@ -0,0 +1,9 @@
+module V2
+  class LinkChangesController < ApplicationController
+    def search
+      result = LinkChange
```
tijmenb: Can we add a filter for link_type here too? That would prevent us from having to deal with organisations/topic/policies tagging elsewhere.

app/commands/v2/patch_link_set.rb
```diff
@@ -25,13 +25,16 @@ def call
...
 
         after_transaction_commit do
           send_downstream(orphaned_content_ids)
         end
 
-        Action.create_patch_link_set_action(link_set, event)
+        action = Action.create_patch_link_set_action(link_set, event)
+
+        after_links = link_set.links.to_a
+        LinkChangeService.new(action, before_links, after_links).record
```
kevindew: I feel like this LinkChange code belongs inside the `create_patch_link_set_action` method. Logic being that these aren't things that live independently and you wouldn't want to be creating patch_link_set actions without this data.

app/controllers/v2/link_changes_controller.rb
```diff
@@ -1,7 +1,7 @@
 module V2
   class LinkChangesController < ApplicationController
     def index
-      render json: Queries::GetLinkChanges.new(params).as_json
+      render json: Queries::GetLinkChanges.new(params, request.original_url).as_json
```
kevindew: I wonder if we have scope to re-use the keyset pagination code that was wrote for the editions endpoint? https://github.com/alphagov/publishing-api/blob/master/app/controllers/v2/editions_controller.rb#L3-L30

I know @thomasleese intended for it to be reusable and if this endpoint is quite consistent with that this could be a good chance for consistent approaches.

db/schema.rb
```diff
@@ -130,6 +130,17 @@
     t.index ["content_id", "locale", "with_drafts"], name: "expanded_links_content_id_locale_with_drafts_index", unique: true
   end
 
+  create_table "link_changes", force: :cascade do |t|
+    t.uuid "source_content_id", null: false
+    t.uuid "target_content_id", null: false
+    t.string "link_type", null: false
+    t.integer "change", null: false
```
kevindew: I'm not sure an integer is the right field here since we only care about two values. Maybe it's better to have an enum (well the rails pseudo enum) assuming bool is a bit too crude.

db/migrate/20170905123649_create_link_changes.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLinkChanges < ActiveRecord::Migration[5.1]
+  def change
+    create_table :link_changes do |t|
+      t.uuid :source_content_id, null: false
+      t.uuid :target_content_id, null: false
+      t.string :link_type, null: false
+      t.integer :change, null: false
+      t.references :action, null: false
```
kevindew: Be good to get a foreign key and index here, also nice if we had a on_delete behaviour - which I guess would be cascade unless you want to go for a more paranoid restrict - example: https://github.com/alphagov/publishing-api/blob/5b385fe1d79687a646f8600030964dc3cc878a06/db/migrate/20170131143927_add_content_item_fk_to_links.rb#L3

app/controllers/v2/link_changes_controller.rb
thomasleese: Yeah, I agree it would be good to use the keyset pagination code here. Probably easier to explain off GitHub though, so let me know if you want more background on how you could do this.

app/commands/v2/patch_link_set.rb
tijmenb: Excellent point

tijmenb:
I'm splitting this PR into 2 for easier review. https://github.com/alphagov/publishing-api/pull/1026 implements the change history, with @kevindew's comments about `Action`, enums and indexes addressed.

PR #1025, opened 2017-09-05
Make change notes deletable

tuzz:
We have found a case where an Edition originally has an `update_type` of
"major" and a change note, and subsequently has its `update_type` changed to
"minor", the change note persists. This ensures that in these cases the change
note is deleted.

This is to fix an issue relating to this Zendesk ticket:
https://govuk.zendesk.com/agent/tickets/2393067

For [Trello](https://trello.com/c/GbvuESJT/1112-remove-change-note-when-switching-update-type-from-major-to-minor)

### DIFF
--- a/app/models/update_existing_draft_edition.rb
+++ b/app/models/update_existing_draft_edition.rb
@@ -42,6 +42,10 @@ private
     # including them in the payload.
     edition.links.delete_all
 
+    if edition.update_type == "minor" && edition.change_note.present?
+      edition.change_note.delete
+    end
+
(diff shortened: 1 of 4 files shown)


kevindew:
Oh wow I was just writing this up as a card for our backlog - fast work @tuzz

tuzz:
Ahh, well it's not finished yet– I need to move on to other things now so hopefully someone can finish this up.

kevindew:
Sure, I'll continue with the card then. Thanks for sharing.

PR #1026, opened 2017-09-06
Keep track of change history for links

tijmenb:
This is the first PR to implement a change tracking feature for links. It's half of https://github.com/alphagov/publishing-api/pull/1024.

We'll create a separate PR to expose the change history via the API. That information will be used in content tagger by users who want to see when a page was tagged to a certain taxon and who tagged what to which taxon. This enables them to monitor tagging behaviour.

Initial code by @koetsier.

https://trello.com/c/KDFBohiV

### DIFF
--- a/spec/factories/link_change.rb
+++ b/spec/factories/link_change.rb
@@ -1,0 +1,9 @@
+FactoryGirl.define do
+  factory :link_change do
+    source_content_id { SecureRandom.uuid }
+    target_content_id { SecureRandom.uuid }
+    action
+    link_type "taxons"
+    change "add"
+  end
+end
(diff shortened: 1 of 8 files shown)


spec/factories/link_change.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :link_change do
+    source_content_id { SecureRandom.uuid }
+    target_content_id { SecureRandom.uuid }
+    action
+    link_type "taxons"
+    change 1
```
kevindew: I think this should be your enum rather than an int

tijmenb: Amended.

PR #1027, opened 2017-09-06
Ensure that content_id and locale are populated

steventux:
https://trello.com/c/AJxouyp6/1044-2-fix-issue-with-contentid-locale-sometimes-not-being-sent-to-content-store-in-expanded-links

These were not being populated correctly when `LinkExpansion` was initialised by edition, because
they are not  included in `Edition#attributes`.
Make sure these are correctly populated so that calling `LinkExpansion.by_edition` produces the same output as `LinkExpansion.by_content_id`.

See https://github.com/alphagov/collections/pull/365/commits/9bdf07195c72b001a3c7ad3c32a9fa93d1d2b4f7 for workaround in collections.

@MatMoore - Hopefully this means you can remove the defensive code in the commit above.

### DIFF
--- a/lib/link_expansion/edition_hash.rb
+++ b/lib/link_expansion/edition_hash.rb
@@ -2,7 +2,7 @@ class LinkExpansion::EditionHash
   class << self
     def from(values)
       return nil unless values.present?
-      hash = values.is_a?(Array) ? hash_for(values) : values
+      hash = hash_for(values)
       hash = SymbolizeJSON.symbolize(hash)
       hash = hash.slice(*edition_fields)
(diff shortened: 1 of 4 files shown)


lib/link_expansion/edition_hash.rb
```diff
@@ -20,7 +20,16 @@ def edition_fields
...
-      Hash[edition_fields.zip(values)]
+      case values
+      when Array
+        Hash[edition_fields.zip(values)]
+      when Edition
+        values.attributes.merge(
+          content_id: values.content_id,
+          locale: values.locale
+        )
+      else values
```
thomasleese: Maybe we should only do this `when Hash` and then raise an exception otherwise?

PR #1028, opened 2017-09-06
Add endpoint for tagging history

tijmenb:
This is the follow up to https://github.com/alphagov/publishing-api/pull/1026, which implemented tracking changes to the links. This PR exposes the data via the API. It shouldn't be merged until we have [used it in content-tagger][ct], because that will be the primary user of the endpoint.

Things that we need to do before merging:

- [x] Use it in content tagger
- [x] Add documentation

https://trello.com/c/Ays5PBSH

[ct]: https://trello.com/c/woqZQI8n

### DIFF
--- a/app/queries/get_link_changes.rb
+++ b/app/queries/get_link_changes.rb
@@ -1,0 +1,82 @@
+module Queries
+  class GetLinkChanges
+    # Current maximum number of results. This is an arbitrary number to prevent
+    # too much data being returned, and to hold off from implementing pagination.
+    MAXIMUM_NUMBER_OF_RESULTS = 250
+
+    attr_reader :params
+
+    def initialize(params)
+      @params = params
+    end
+
+    def as_hash
+      results = link_changes.map do |link_change|
+        {
+          source: expand_edition(link_change.source_content_id),
+          target: expand_edition(link_change.target_content_id),
+          link_type: link_change.link_type,
+          change: link_change.change,
+          user_uid: link_change.action.user_uid,
+          created_at: link_change.created_at
+        }
+      end
+
+      {
+        link_changes: results,
+      }
+    end
+
+  private
+
+    def expand_edition(content_id)
+      editions = all_related_editions[content_id]
+      editions && editions.first.slice(%w[title base_path content_id])
+    end
+
+    def link_changes
+      @link_changes ||= LinkChange
+                          .order(created_at: :desc)
+                          .where(where_query_from_params)
+                          .limit(MAXIMUM_NUMBER_OF_RESULTS)
+                          .includes(:action)
+    end
+
+    def where_query_from_params
+      # Link type is required. Calling `fetch` on a Rails parameters object
+      # will bubble up to the controller to return a 4XX response.
+      query = { link_type: params.fetch(:link_types) }
+
+      if params[:source_content_ids]
+        query[:source_content_id] = params[:source_content_ids]
+      end
+
+      if params[:target_content_ids]
+        query[:target_content_id] = params[:target_content_ids]
+      end
+
+      if params[:users]
+        query['actions.user_uid'] = params[:users]
+      end
+
+      query
+    end
+
+    # Returns all editions (with their document) that are either the source or
+    # the target of the link change.
+    def all_related_editions
+      @all_related_editions ||= begin
+        content_ids_relevant = (
+          link_changes.map(&:source_content_id) +
+          link_changes.map(&:target_content_id)
+        ).uniq
+
+        Queries::GetLatest.call(
+          Edition
+            .eager_load(:document)
+            .where('documents.content_id IN (?)', content_ids_relevant)
+        ).group_by(&:content_id)
+      end
+    end
+  end
+end
--- a/doc/api.md
+++ b/doc/api.md
@@ -23,6 +23,7 @@ message queue for other apps (e.g. `email-alert-service`) to consume.
 - [`GET /v2/expanded-links/:content_id`](#get-v2expanded-linkscontent_id)
 - [`GET /v2/linked/:content_id`](#get-v2linkedcontent_id)
 - [`GET /v2/linkables`](#get-v2linkables)
+- [`GET /v2/links/changes`](#get-v2linkschanges)
 - [`GET /v2/editions`](#get-v2editions)
 - [`POST /v2/links/by-content-id`](#post-v2linksby-content-id)
 - [`POST /lookup-by-base-path`](#post-lookup-by-base-path)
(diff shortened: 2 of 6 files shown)


app/queries/get_link_changes.rb
```diff
@@ -0,0 +1,79 @@
...
+      {
+        link_changes: results,
+      }
+    end
+
+  private
+
+    def expand_edition(content_id)
+      all_related_editions
+        .find { |edition| edition.content_id == content_id }
```
kevindew: I don't think you have any guarantee that this will return something. 

You might do better grouping editions by content_id to avoid iterating?

app/queries/get_link_changes.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    # Returns all editions (with their document) that are either the source or
+    # the target of the link change.
+    def all_related_editions
+      @all_related_editions ||= begin
+        content_ids_relevant = (link_changes.map(&:source_content_id) + link_changes.map(&:target_content_id)).uniq
+
+        Queries::GetLatest.call(
+          Edition.joins(:document).where('documents.content_id IN (?)', content_ids_relevant)
+        )
```
kevindew: Do you mind that drafts are included in this?

app/queries/get_link_changes.rb
```diff
@@ -0,0 +1,79 @@
...
+        .order(:id)
+        .where(where_query_from_params)
+        .limit(MAXIMUM_NUMBER_OF_RESULTS)
+        .includes(:action)
+    end
+
+    def where_query_from_params
+      # Link type is required. Calling `fetch` on a Rails parameters object
+      # will bubble up to the controller to return a 4XX response.
+      query = { link_type: params.fetch(:link_type) }
```
kevindew: Could make this a path parameter? unless you're planning to allow an array of course.

tijmenb: Perhaps an array is a good idea (also for all other params).

app/queries/get_link_changes.rb
tijmenb: No, drafts should be included. Drafts can be tagged as well.

app/queries/get_link_changes.rb
cbaines: I think Tijmen or Jos have changed the implementation now.

cbaines:
I've spent some time looking at this and testing it.

I've made a few changes:
 - I've changed the ordering to be descending, rather than ascending, as this means that if there are more than 250 results, the latest 250 changes will be returned, rather than the oldest.
 - I've added the link_type to the fields returned, as even though you have to filter by link_type, you can specify more than one, so you need the link_type per change in the results to tell which is which.
 - I've added some documentation to doc/api.md.

The changes to get content-tagger using this are on the `add_analytics` branch for now [1]. I'm happy with the functionality this provides, so I've ticked the "Use it in Content Tagger" box, even though this isn't yet deployed.

1: https://github.com/alphagov/content-tagger/tree/add_analytics

spec/queries/get_link_changes_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe Queries::GetLinkChanges do
+  describe '#as_json' do
```
tijmenb: I think this needs to change to `as_hash` as well.

cbaines: Good spot, I've updated :)

doc/api.md
```diff
@@ -567,6 +568,28 @@ which is `details.internal_name` and falls back to `title`.
...
+### Query string parameters
+
+- `link_types[]` *(required)*
+  - Filter the changes by link type.
+- `source_content_ids[]`
+  - Filter the changes by source content id.
+- `target_content_ids[]`
+  - Filter the changes by source content id.
+- `users[]`
+  - Filter the changes by user UIDs.
```
kevindew: As these don't have to be arrays it might be more appropriate to specify them as non arrays and suggest they can optionally be arrays.

Though I'm not sure how this has been handled in other parts of this doc and it's probably best to just be consistent.

doc/api.md
```diff
@@ -567,6 +568,28 @@ which is `details.internal_name` and falls back to `title`.
...
   - The `document_type` value that returned editions has.
 
+## `GET /v2/links/changes`
+
+Returns an array of changes to links, giving the information on what
+changed (source and target content ids, and the link type), when the
+change happened, what user was associated with the action, and if the
+link was created or deleted.
+
+The results will be in decending order by link change id, which
```
kevindew: I wonder if created_at might be a more intuitive field to order by? I have a comment to make about link change id that is related that I'll add in a different spot.

app/queries/get_link_changes.rb
```diff
@@ -0,0 +1,83 @@
...
+    attr_reader :params
+
+    def initialize(params)
+      @params = params
+    end
+
+    def as_hash
+      results = link_changes.map do |link_change|
+        {
+          id: link_change.id,
```
kevindew: Is this field made use of it? Exposing it makes me feel like you should be able to do something with it - like look up the change directly, which we don't allow, so just wanted to check that we expose this for a reason and it's used.

app/queries/get_link_changes.rb
```diff
@@ -0,0 +1,83 @@
...
+    def all_related_editions
+      @all_related_editions ||= begin
+        content_ids_relevant = (
+          link_changes.map(&:source_content_id) +
+          link_changes.map(&:target_content_id)
+        ).uniq
+
+        Queries::GetLatest.call(
+          Edition.joins(:document)
+            .eager_load(:document)
```
kevindew: Do we need both of these?

doc/api.md
tijmenb: Not sure, I think not using arrays here would be confusing. Does it matter that you technically can send single values?

app/queries/get_link_changes.rb
tijmenb: Removed.

app/queries/get_link_changes.rb
tijmenb: Removed.

doc/api.md
tijmenb: Done.

tijmenb:
@kevindew updated & force-pushed. Added more to the commit, including a note about the pagination.

doc/api.md
kevindew: Nah, that's not a problem.

kevindew:
Cool thanks @tijmenb, looks like you've just got a test to fix and it's good

PR #1035, opened 2017-09-12
Re-enable Edition Diffing

thomasleese:
We had to disable this as it was found to be too slow, but since there is a new `use_lcs` option in HashDiff, we can re-enable it.

[Trello Card](https://trello.com/c/nxfFJGNe/1048-1-reimplement-edition-diffing)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ gem 'bunny', '~> 2.6'
 gem "govuk_sidekiq", "~> 2.0"
 gem 'whenever', '0.9.4', require: false
 gem "json-schema", require: false
-gem "hashdiff"
+gem "hashdiff", "~> 0.3.6"
 gem "sidekiq-unique-jobs", "~> 5.0", require: false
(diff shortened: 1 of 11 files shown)


thomasleese:
Benchmarks of https://www.gov.uk/government/statistical-data-sets/unclaimed-estates-list suggest no statistical difference in speed for saving a draft, both take about 6 seconds.

```ruby
> e = Edition.find_by(base_path: '/government/statistical-data-sets/unclaimed-estates-list')
> Action.find_by(edition_id: e.id).edition_diff.length
22
```

kevindew:
@thomasleese 6 seconds sounds an awfully long time. Is that compared to without the diffing or the previous diffing implementation?

thomasleese:
@kevindew Sorry, that's how long it takes currently with or without diffing, i.e. the diffing makes no difference, it still takes 6 seconds.

kevindew:
@thomasleese oh good! my heart sank a bit at first hehe.

PR #1036, opened 2017-09-15
Link expansion bug

kevindew:
This fixes something that was overlooked when the link caching was
introduced. When a reverse has direct links, the resultant SQL query was
using the reverse link name and not finding results.

eg `WHERE link_type = 'child_taxons' AND other_link_type IN
('associated_taxons')`

Which would yield no results as child_taxons isn't in the database

This corrects this to be

`WHERE link_type = 'parent_taxons' AND other_link_type IN
('associated_taxons')`

### DIFF
--- a/spec/integration/dependency_resolution_spec.rb
+++ b/spec/integration/dependency_resolution_spec.rb
@@ -142,6 +142,24 @@ RSpec.describe "Dependency Resolution" do
     end
   end
 
+  context "where there are reverse links: child_taxons and direct links: associated_taxons" do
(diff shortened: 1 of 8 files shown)


spec/integration/dependency_resolution_spec.rb
```diff
@@ -61,6 +61,8 @@
     end
   end
 
+
```
emmabeynon: I think there are a couple of errant blank lines here.

kevindew: Thanks! fixed

PR #1038, opened 2017-09-20
Add update_type to redirect payload

emmabeynon:
We found a bug where redirect payloads don't include `update_type` when sent
with PUT content, and `update_type` is soon to be a required field for PUT
content requests. This ensures that `update_type` is now sent in this
situation.

For [Trello](https://trello.com/c/UaxgGUDI/1115-require-updatetype-in-put-content)

### DIFF
--- a/app/presenters/redirect_presenter.rb
+++ b/app/presenters/redirect_presenter.rb
@@ -27,7 +27,7 @@ class RedirectPresenter
   end
 
   def for_redirect_helper(content_id)
-    present.merge(content_id: content_id)
+    present.merge(content_id: content_id, update_type: "major")
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/presenters/redirect_presenter.rb
```diff
@@ -26,8 +26,8 @@ def for_message_queue(payload_version)
     )
   end
 
-  def for_redirect_helper(content_id)
-    present.merge(content_id: content_id)
+  def for_redirect_helper(content_id, update_type)
+    present.merge(content_id: content_id, update_type: update_type)
```
kevindew: I'd probably just include a hard coded update_type of major here. 
I think having a variety of update_types on redirects is a bit confusing (well the whole concept of an update_type on a redirect is confusing).

This would also be consistent with how they are currently published: https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/publish.rb#L186

And I also wonder once we add this in if we can remove the line I linked to above?

emmabeynon: This has now been updated 🌞

PR #1039, opened 2017-09-20
Add provider state for LinkChanges pact 

koetsier:
trello: https://trello.com/c/nYBkNBDc/294-switch-to-using-the-link-changes-endpoint-through-the-gds-api-adapters-gem

PR #1040, opened 2017-09-22
reformat pact_helper to make it more consistent

koetsier:
Addresses a missed comment in this https://github.com/alphagov/publishing-api/pull/1039 pull request

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -324,34 +324,42 @@ Pact.provider_states_for "GDS API Adapters" do
         document_b1 = FactoryGirl.build(:document, content_id: 'bbbbbbbb-bbbb-1bbb-bbbb-bbbbbbbbbbbb')
         document_b2 = FactoryGirl.build(:document, content_id: 'bbbbbbbb-bbbb-2bbb-bbbb-bbbbbbbbbbbb')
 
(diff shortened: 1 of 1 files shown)


spec/service_consumers/pact_helper.rb
```diff
@@ -344,13 +347,13 @@ def url_encode(str)
...
-                           action: FactoryGirl.create(:action, user_uid: '11111111-1111-1111-1111-111111111111'),
+                           action: action1,
                            change: 'add')
 
         FactoryGirl.create(:link_change,
                            source_content_id: document_a2.content_id,
                            target_content_id: document_b2.content_id,
-                           action: FactoryGirl.create(:action, user_uid: '22222222-2222-2222-2222-222222222222'),
+                           action: action2,
                            change: 'remove')
```
cbaines: I'm guessing that what @kevindew meant by [1] was make the indentation like [2], where the symbol is on the same line, and the hash keys are two spaces in from `Factory...`.

1: https://github.com/alphagov/publishing-api/pull/1039#discussion_r140194927
2: https://github.com/alphagov/publishing-api/blob/master/spec/service_consumers/pact_helper.rb#L77-L93

koetsier: 👍

kevindew:
Awesome thank you @koetsier 👍

PR #1042, opened 2017-09-27
Rake task to set new Publishing API dates

kevindew:
I've put this PR up before I go away as it'd be great to get some feedback and sanity checks as to whether this seems accurate, and whether I've missed anything (or got bad logic). If all looks good I'll plan to run once I'm back from holidays.

This is used to populate the date fields of:

- temporary_first_published_at
- publisher_first_published_at
- major_published_at
- publisher_major_published_at
- published_at
- publisher_published_at
- temporary_last_edited_at
- publisher_last_edited_at

To be consistent with the new date fields introduced by:
https://github.com/alphagov/publishing-api/pull/1007

This is a slow process, I think the script probably takes ~6 hours to
run full. It also has a few limitations:

- Only able to use events when there is a single locale
- Unable to work out major_publishing when the update_type was set in
publish (lost in events at that point)
- Makes some approximations based on nano-second value

My feeling is that although this isn't completely accurate, it's
probably accurate enough to lay some decent foundation. It introduces
a Edition#new_dates_valid? method which can be used to check that the
dates haven't deviated from what will be presented to Content Store and
other services.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -231,6 +231,38 @@ class Edition < ApplicationRecord
     Plek.current.website_root + base_path
   end
 
+  # @FIXME this is only supposed to be in for temporarily changing of date fields
+  def new_dates_valid?
+    within_1_sec = ->(date_a, date_b) do
+      diff = (date_b.to_f * 1000).to_i - (date_a.to_f * 1000).to_i
+      diff.abs < 1000
+    end
+    valid = ->(old, publisher_new, api_new) do
+      if old
+        within_1_sec.(old, (publisher_new || api_new))
+      else
+        old == publisher_new
+      end
+    end
+
+    first_published_at_valid = valid.(
+      first_published_at,
+      publisher_first_published_at,
+      temporary_first_published_at
+    )
+    public_updated_at_valid = valid.(
+      public_updated_at,
+      publisher_major_published_at,
+      major_published_at
+    )
+    last_edited_at_valid = valid.(
+      last_edited_at,
+      publisher_last_edited_at,
+      temporary_last_edited_at
+    )
+    first_published_at_valid && public_updated_at_valid && last_edited_at_valid
+  end
+
 private
 
   def renderable_content?
--- a/lib/tasks/split_dates.rake
+++ b/lib/tasks/split_dates.rake
@@ -1,0 +1,13 @@
+  # @TODO Remove this file once the split dates task has run
+namespace :split_dates do
+  desc "Populate the split dates introduced September 2017"
+  task :populate, [:threads] => :environment do |_, args|
+    threads = Integer(args.fetch(:threads, 5))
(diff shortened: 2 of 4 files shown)


app/models/edition.rb
```diff
@@ -230,6 +230,20 @@ def web_url
     Plek.current.website_root + base_path
   end
 
+  # @FIXME this is only supposed to be in for temporarily changing of date fields
+  def new_dates_valid?
+    within_1_sec = -> (date_a, date_b) do
```
thomasleese: This doesn't actually do what is implied by `within_1_sec`, although I don't know if this is an issue:

```
[1] pry(main)> a = DateTime.parse('2017-01-01 10:00:00.999')
=> Sun, 01 Jan 2017 10:00:00 +0000
[2] pry(main)> b = DateTime.parse('2017-01-01 10:00:01.000')
=> Sun, 01 Jan 2017 10:00:01 +0000
[3] pry(main)> a.to_i == b.to_i
=> false
```

When the two dates are clearly within 1 second of each other.

lib/tasks/split_dates/populate_edition.rb
```diff
@@ -0,0 +1,265 @@
...
+
+  def has_last_edited_dates?
+    edition.temporary_last_edited_at || edition.publisher_last_edited_at
+  end
+
+  def can_use_events?
+    @can_use_events ||= Document.where(content_id: document.content_id).count == 1
+  end
+
+  def within_1_sec(date_a, date_b)
```
thomasleese: Same issue as above.

app/models/edition.rb
kevindew: @thomasleese you're right, I was just being lazy as I figured the odds of it slipping over a second would generally be low. But I should and will sort it out.

kevindew: Should be fixed now.

app/models/edition.rb
```diff
@@ -230,6 +230,21 @@ def web_url
...
+      return true if date_a == date_b
+      return false if [date_a, date_b].include?(nil)
+      diff = (date_b.to_f * 1000).to_i - (date_a.to_f * 1000).to_i
+      diff.abs > 1000
```
thomasleese: I think this should be `< 1000`?

kevindew:
Marked this DNM as should probably [fix some dates](https://trello.com/c/SSbguUQA/1172-fix-whitehall-firstpublishedat-dates-in-publishing-api) before this goes in

adrianclay:
Hey @kevindew ,

Just to let you know, the publishing-e2e-tests are failing for this commit because it branches off an old version of master which has an old Dockerfile.  If you branch off a more recent version of master the publishing-e2e-tests should pass.

kevindew:
@adrianclay Thanks 👍  - I'm actually doing that on purpose to test the slack message, sorry should have given a heads up.

PR #1043, opened 2017-09-28
Update Whitehall about page first_published_at

steventux:
https://trello.com/c/SSbguUQA/558-fix-whitehall-firstpublishedat-dates-in-publishing-api

Around 80000 Editions relating various Whitehall formats have the same `first_published_at` of `2016-02-29 09:24:10`.
Using an export of data from Whitehall update these Editions to use the correct `first_published_at` value.

Dev VM run:

```
== 20170928103153 FixWhitehallFirstPublishedAtForMultipleDocumentTypes: migrating 
Updated 28526 editions with document_type: about
Updated 24514 editions with document_type: press_release
Updated 21858 editions with document_type: news_story
74898 editions updated.
== 20170928103153 FixWhitehallFirstPublishedAtForMultipleDocumentTypes: migrated (187.5310s) 
```


db/migrate/20170928103153_fix_whitehall_about_page_first_published_at.rb
```diff
@@ -0,0 +1,22 @@
...
+class FixWhitehallAboutPageFirstPublishedAt < ActiveRecord::Migration[5.1]
+  def up
+    update_count = 0
+    data = CSV.read(Rails.root.join("db", "migrate", "data", "about-first-published-at.csv"))
+    start_datetime = "2016-02-29 09:24:09"
+    end_datetime = "2016-02-29 09:24:11"
+
+    data.each do |content_id, first_published_at|
+      count = Edition
+        .with_document
```
thomasleese: I don't think we need this line since we never access the document as part of a model instance.

db/migrate/20170928103153_fix_whitehall_about_page_first_published_at.rb
```diff
@@ -0,0 +1,22 @@
...
+        .where("first_published_at BETWEEN '#{start_datetime}' AND '#{end_datetime}'")
+        .where(publishing_app: "whitehall", document_type: "about")
+        .where(:"documents.content_id" => content_id)
```
thomasleese: Might be nicer to do `documents: { content_id: content_id }`.

bevanloon:
I notice the CSV has a mixture of summer and winter times (+0100 and +0000) - is that going to cause any problems given the data is coming from mysql and going into postgres (so I guess, do pg and mysql handle date formats the same way)?

steventux:
@thomasleese refactored accordingly (also added a guard clause for nil `first_published_at` data from the CSV)
@bevanloon looks like postgres imports the dates and converts to UTC:
eg. the date for `454efa5c-c09f-4e4e-9f14-2f4fad4e328d` is `2016-08-31 09:24:51 +0100` in the CSV but migrates to `Wed, 31 Aug 2016 08:24:51 UTC +00:00`

steventux:
@thomasleese I've reworked this as I was gonna repeat the same migration for several formats.
So now there are 3 document types to be updated: `about`, `press_release` and `news_story`
This fixes about 75000 records.

PR #1045, opened 2017-10-02
Update first_published_at for Dfid Research Output editions

steventux:
Part of https://trello.com/c/1gZgk0HF/10-switch-specialist-documents-to-use-firstpublishedat

There are 63000 `dfid_research_output` editions which have a `first_published_at` date
after their `public_updated_at` date. The correct date is stored in the edition metadata
so update these records with the correct `first_published_at` value.

```
== 20171002152414 FixDfidResearchOutputFirstPublishedAt: migrating ============
== 20171002152414 FixDfidResearchOutputFirstPublishedAt: migrated (41.4836s) ==
```

### DIFF
--- a/db/migrate/20171005150944_fix_dfid_research_output_first_published_at.rb
+++ b/db/migrate/20171005150944_fix_dfid_research_output_first_published_at.rb
@@ -1,0 +1,31 @@
+class FixDfidResearchOutputFirstPublishedAt < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  # There are ~63000 editions which have a first_published_at timestamp
+  # later than their public_updated_at timestamp. So use legacy data from
+  # metadata to fix this.
+  #
+  def up
+    content_ids = Edition.joins(:document)
+                         .where(publishing_app: "specialist_publisher",
+                                document_type: "dfid_research_output",
+                                state: %w(draft published unpublished))
+                         .where("first_published_at > (public_updated_at + interval '1 second')")
+                         .pluck(:"documents.content_id")
+
+    sql = <<-SQL.strip_heredoc
+      UPDATE editions
+      SET    first_published_at = (details->'metadata'->>'first_published_at')::timestamp
+      WHERE  publishing_app = 'specialist-publisher'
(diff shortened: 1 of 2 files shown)


db/migrate/20171002152414_fix_dfid_research_output_first_published_at.rb
```diff
@@ -0,0 +1,18 @@
...
+  # later than their public_updated_at timestamp. So use legacy data from
+  # metadata to fix this.
+  #
+  def up
+    sql = <<-SQL.strip_heredoc
+      UPDATE editions
+      SET    first_published_at = (details->'metadata'->>'first_published_at')::date
+      WHERE  publishing_app = 'specialist-publisher'
+      AND    document_type  = 'dfid_research_output'
+      AND    first_published_at > (public_updated_at + interval '1 second')
```
thomasleese: Why do we need the 1 second interval here?

steventux: I found that an edition could have the same datetime to the second or even millisecond for both fields but the storage to microsecond precision meant that they were not deemed equal when executing the query.
So we'd get a lot of legitimately near-equal datetimes being caught by what is essentially a search for nonsensical date combinations. In other words rounding.

thomasleese: Ah yep, makes sense.

On a side note, I'll try and figure out why the e2e tests are failing...

db/migrate/20171002152414_fix_dfid_research_output_first_published_at.rb
```diff
@@ -0,0 +1,29 @@
...
+      AND    (details->'metadata'->>'first_published_at')::timestamp < (public_updated_at + interval '1 second')
+    SQL
+
+    ActiveRecord::Base.connection.execute(sql)
+
+    if Rails.env.production?
+      Commands::V2::RepresentDownstream.new.call(content_ids)
```
thomasleese: You'll need to include `disable_ddl_transaction!` to do this.

PR #1046, opened 2017-10-02
Fix cma_case first_published_at

steventux:
Part of https://trello.com/c/1gZgk0HF/10-switch-specialist-documents-to-use-firstpublishedat

There are 2700 cma cases with a first published at date later
than their public updated at date. The correct date may be stored
in the metadata as the opened date, so update these editions to use
this date where possible.

```
== 20171004172111 FixCmaCaseFirstPublishedAt: migrating =======================
== 20171004172111 FixCmaCaseFirstPublishedAt: migrated (3.6276s) ==============
```

This PR should be merged after https://github.com/alphagov/publishing-api/pull/1049 as it is preferable we use the change history dates where possible.


db/migrate/20171002155811_fix_cma_case_first_published_at.rb
```diff
@@ -0,0 +1,15 @@
+class FixCmaCaseFirstPublishedAt < ActiveRecord::Migration[5.1]
+  def change
+    sql = <<-SQL.strip_heredoc
+      UPDATE editions
+      SET    first_published_at = (details->'metadata'->>'opened_date')::date
+      WHERE  publishing_app = 'specialist-publisher'
+      AND    document_type  = 'cma_case'
+      AND    first_published_at > (public_updated_at + interval '1 second')
```
thomasleese: Similar to the previous PR, why do we need the 1 second interval here?

db/migrate/20171002155811_fix_cma_case_first_published_at.rb
```diff
@@ -0,0 +1,15 @@
+class FixCmaCaseFirstPublishedAt < ActiveRecord::Migration[5.1]
+  def change
+    sql = <<-SQL.strip_heredoc
+      UPDATE editions
+      SET    first_published_at = (details->'metadata'->>'opened_date')::date
```
steventux: I think this needs to cast to a `timestamp` and not `date` to preserve the time. Will update.

PR #1047, opened 2017-10-04
Preview of rabbit MQ message

dwhenry:
This is helpful when building consumers as otehrwise
it is hard to reason about the actual message that will be received
and setting the system up to capture the message is time consuming.

This functionality has been added to the documentation.

https://github.com/alphagov/govuk-developer-docs/pull/488
https://trello.com/c/KDfyOeS1/331-specialist-publisher-implementation

### DIFF
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -40,4 +40,24 @@ namespace :queue do
 
     RequeueContent.new(scope).call
   end
+
+  desc "Preview of the message published onto rabbit MQ"
+  task :preview_recent_message, [:document_type] => :environment do |_, args|
+    document_type = args[:document_type]
+    raise ValueError("expecting document_type") unless document_type.present?
+
+    edition = Edition
+              .with_document
+              .order(public_updated_at: :desc)
+              .find_by!(
+                editions: { state: 'published' },
+                document_type: document_type
+              )
+    version = Event.maximum(:id)
+
+    presenter = DownstreamPayload.new(edition, version, draft: false)
+    payload = presenter.message_queue_payload
+
+    pp payload
+  end
 end
--- a/spec/lib/tasks/queue_spec.rb
+++ b/spec/lib/tasks/queue_spec.rb
@@ -1,0 +1,17 @@
+require "rails_helper"
+
+RSpec.describe "Queue rake task" do
+  before do
+    Rake::Task['queue:preview_recent_message'].reenable
+  end
+
+  it "previews the rabbit MQ message for a document type" do
+    edition = FactoryGirl.create(:live_edition, base_path: '/ci1')
+    FactoryGirl.create(:event, id: 12)
+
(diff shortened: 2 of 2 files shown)


lib/tasks/queue.rake
```diff
@@ -40,4 +40,20 @@ namespace :queue do
 
     RequeueContent.new(scope).call
   end
+
+  desc "Preview of the message published onto rabbit MQ"
+  task :preview_last_message, [:document_type] => :environment do |_, args|
+    document_type = args[:document_type]
+    raise ValueError("expecting content_id") unless document_type.present?
```
thomasleese: Should be "expecting document_type".

lib/tasks/queue.rake
```diff
@@ -40,4 +40,20 @@ namespace :queue do
...
+
+    scope = Edition
```
thomasleese: I don't know if this is the correct way of working out what was last put onto the queue, since unpublished editions are also put onto the queue now. Perhaps a better way might be to use the last `Action` to determine the edition?

lib/tasks/queue.rake
```diff
@@ -40,4 +40,20 @@ namespace :queue do
...
+              .with_document
+              .where(editions: { state: 'published' })
+              .where(document_type: document_type)
+              .order(public_updated_at: :desc)
+              .limit(1)
+
+    pp RequeueContent.new(scope).preview
+    nil
```
thomasleese: Don't think we need to have this here.

lib/tasks/queue.rake
dwhenry: corrected

lib/tasks/queue.rake
dwhenry: I have renamed the message to "preview_recent_message" as discussed

lib/tasks/queue.rake
dwhenry: removed

lib/tasks/queue.rake
```diff
@@ -40,4 +40,19 @@ namespace :queue do
...
+    raise ValueError("expecting document_type") unless document_type.present?
+
+    scope = Edition
+              .with_document
+              .where(editions: { state: 'published' })
+              .where(document_type: document_type)
+              .order(public_updated_at: :desc)
+              .limit(1)
+
+    pp RequeueContent.new(scope).preview
```
thomasleese: I don't think the preview functionality should be the responsibility of the `RequeueContent` class since that's just concerned about requeueing all the content, not everything related to the queue. I would be tempted to include the lines which call the DownstreamPayload into this task directly and leave `RequeueContent` as it was.

dwhenry: @thomasleese Have moved all the code to the rake task and added a test to ensure it is updated if the code changes

dwhenry:
@thomasleese can I get an approval on this one if you are happy now. :)

PR #1051, opened 2017-10-09
Fix specialist-publisher editions with that first_published_at date

steventux:
There are around 17000 specialist publisher editions which have
a `first_published_at` date of `2016-02-29 09:24:10` and a `public_updated_at`
timestamp earlier than this. So update these records to the `public_updated_at`
timestamp.


steventux:
Not sure where the branch build went for this...

PR #1052, opened 2017-10-10
Update specialist publisher first_published_at from legacy dates

steventux:
https://trello.com/c/L7IfKMRs/13-sanitise-specialist-content-firstpublishedat-dates
More `first_published_at` fun.

There are a few hundred editions which match legacy specialist publisher
db records having meaningful first published data. (The older db shows all the signs of a mass import on `2014-09-24` so I've omitted these records).
Use the legacy data to update specialist-publisher editions where possible, ie. don't attempt to update records where the public_updated_at date is earlier than the legacy date.

```
== 20171010151218 ImportLegacySpecialistPublisherFirstPublishedAtDates: migrating 
182 specialist-publisher editions updated.
== 20171010151218 ImportLegacySpecialistPublisherFirstPublishedAtDates: migrated (15.3880s)
```


db/migrate/20171010151218_import_legacy_specialist_publisher_first_published_at_dates.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  # ~4832 records match
+  def up
+    count = 0
+    content_ids = []
+    csv_data = CSV.read(Rails.root.join("db", "migrate", "data", "specialist-publisher-legacy-first-published-at.csv"))
+    csv_data.each do |content_id, slug, first_published_at|
+      updated = Edition.joins(:document)
+                       .where(publishing_app: "specialist-publisher",
+                              state: %w(draft published unpublished),
```
kevindew: can we also do this for superseded records since they share the same first_published_at data? (and we really want to rid this rod)

kevindew:
I admire your persistence with these dates Steve 👍  the database is going to be gleaming

steventux:
@kevindew updated.

PR #1053, opened 2017-10-12
Update first_published_at for remaining specialist publisher editions

steventux:
https://trello.com/c/L7IfKMRs/13-sanitise-specialist-content-firstpublishedat-dates

There are ~165 remaining specialist publisher editions with a first published at
after the public updated at value. There's no accurate data available as
to when these were first published, so make the first published at date
the same as the current public updated at so the dates appear logical.

Publishing API date whack-a-mole level complete.

### DIFF
--- a/db/migrate/20171011154900_fix_remaining_specialist_publisher_first_published_at_dates.rb
+++ b/db/migrate/20171011154900_fix_remaining_specialist_publisher_first_published_at_dates.rb
@@ -1,0 +1,19 @@
+class FixRemainingSpecialistPublisherFirstPublishedAtDates < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher")
+                   .where.not(document_type: %w(finder finder_email_signup gone redirect))
+                   .where("first_published_at > (public_updated_at + interval '1 second')")
+
+    content_ids = scope.joins(:document).pluck(:content_id)
+
+    count = scope.update_all("first_published_at = public_updated_at")
+
+    if Rails.env.production?
+      Commands::V2::RepresentDownstream.new.call(content_ids)
+    end
+
+    puts "Updated #{count} specialist-publisher editions."
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20171011154900_fix_remaining_specialist_publisher_first_published_at_dates.rb
```diff
@@ -0,0 +1,19 @@
+class FixRemainingSpecialistPublisherFirstPublishedAtDates < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    scope = Edition.where(publishing_app: "specialist-publisher")
+                   .where.not(document_type: %w(finder finder_email_signup gone redirect))
```
thomasleese: Would it be better to say `where(schema_name: 'specialist_document')`?

steventux: I get slightly different results using that clause rather than the exclusive one. eg.

```
irb(main):001:0> Edition.where(publishing_app: "specialist-publisher").where.not(document_type: %w(finder finder_email_signup gone redirect)).where("first_published_at > (public_updated_at + interval '1 second')").group(:document_type).count
=> {"cma_case"=>31, "drug_safety_update"=>6, "aaib_report"=>8, "raib_report"=>8, "dfid_research_output"=>95, "esi_fund"=>11, "medical_safety_alert"=>5, "specialist_document"=>1}
irb(main):002:0> Edition.where(publishing_app: "specialist-publisher").where(schema_name: "specialist_document").where("first_published_at > (public_updated_at + interval '1 second')").group(:document_type).count
=> {"cma_case"=>26, "drug_safety_update"=>6, "aaib_report"=>7, "raib_report"=>8, "dfid_research_output"=>95, "esi_fund"=>10, "medical_safety_alert"=>5, "specialist_document"=>1}
```

thomasleese: Hmm, strange... looks like a few more editions will be fixed by your query so I'm happy to go with yours.

PR #1061, opened 2017-11-13
[WIP] Add rake task to export content with taxon links

davidbasalla:
We require a dump of this for the data science team to analyse the current state of the munged legacy taxonomy.

Currently only outputs a sample of 10000 items, will be updated to loop through all content eventually.


davidbasalla:
This is not needed anymore

PR #1062, opened 2017-11-14
Only represent one copy of each document when republishing

suzannehamilton:
Fix the `represent_downstream` tasks so that they represent unique documents rather than unique editions.

In both cases, only the latest edition would be republished, so this fix does not affect what ends up in the content store and search index. It only reduces the number of republishing tasks.

https://trello.com/c/64Nv2XP8/433-migrate-hmrc-manual-formats-to-new-index

### DIFF
--- a/lib/tasks/represent_downstream.rake
+++ b/lib/tasks/represent_downstream.rake
@@ -10,7 +10,7 @@ namespace :represent_downstream do
   desc "Represent all editions downstream"
   task all: :environment do
     represent_downstream(
-      Edition.with_document.where.not(content_store: nil)
+      Document.joins(:editions).where.not(editions: { content_store: nil })
     )
   end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/represent_downstream.rake
```diff
@@ -10,7 +10,7 @@ namespace :represent_downstream do
   desc "Represent all editions downstream"
   task all: :environment do
     represent_downstream(
-      Edition.with_document.where.not(content_store: nil)
+      Document.joins(:editions).where(editions: { content_store: nil })
```
MatMoore: This has lost the `not`.

My understanding is that this `where` is just an optimisation and there will always be one edition with a value of `draft` and/or one with `live`, and anything with a nil value is in a superseded state. I could be wrong though - I'm not sure why we have this here and not in the other tasks.

PR #1063, opened 2017-11-15
Add migration to fix taxons visible_to_department_editors

davidbasalla:
For https://trello.com/c/b3TTdin2/157-prevent-publishers-seeing-draft-taxons-in-government-frontend

This value should be `true` for all published editions, since this
will now control whether taxons are visible in the taxonomy sidebar.

We are doing work in ContentTagger to set this value to true once
a taxon is published so that the state is maintained properly going
forward.

See https://github.com/alphagov/content-tagger/pull/516

### DIFF
--- a/db/migrate/20171114152513_fix_visible_to_departmental_editors_field_for_taxons.rb
+++ b/db/migrate/20171114152513_fix_visible_to_departmental_editors_field_for_taxons.rb
@@ -1,0 +1,19 @@
+class FixVisibleToDepartmentalEditorsFieldForTaxons < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    published_taxons = Edition.where(document_type: "taxon").where(state: "published")
+    content_ids = published_taxons.joins(:document).pluck(:content_id)
+
+    published_taxons.each do |t|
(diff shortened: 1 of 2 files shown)


db/migrate/20171114152513_fix_visible_to_departmental_editors_field_for_taxons.rb
```diff
@@ -0,0 +1,17 @@
...
+    published_taxons = Edition.where(document_type: "taxon").where(state: "published")
+    content_ids = published_taxons.joins(:document).pluck(:content_id)
+
+    published_taxons.each do |t|
+      updated_details = t.details.merge(visible_to_departmental_editors: true)
+      t.update(details: updated_details)
+    end
+
+    if Rails.env.production?
+      Commands::V2::RepresentDownstream.new.call(content_ids)
```
thomasleese: You'll need to include the line `disable_ddl_transaction!` at the top of the class to do this.

davidbasalla: 👌 thanks, added it!

davidbasalla:
Closing in favour of alphagov/govuk_navigation_helpers#82

PR #1065, opened 2017-11-16
Use standard buildProject in Jenkinsfile

h-lame:
For: https://trello.com/c/QtpGwa2A/288-improve-feedback-on-github-for-changes-to-govuk-content-schemas

Using the standard buildProject defined in our groovy library means we get
consistent behaviour across all applications, and allows us to improve our
CI process in one place.

We do need to take advantage of some of the customisation hooks to keep
the build behaviour the same though.

1. Add extra parameters to the build for configuring which branches to use
   in the the content-store and publishing end-to-end test jobs
2. use beforeTest to set some extra env vars for coverage and pact tests
3. set publishingE2ETests to true so we always run the publishing
   end-to-end tests (our custom jenkinsfile always ran these but
   buildProject only does it if you tell it to)
4. use afterTest to publish the output of rcov
5. use afterTest to run pact verification and publish the output
6. use afterTest to run pact verification against content-store

The test task in our custom script was to run `rspec` whereas the default
in buildProject is to run `rake default`.  We could use overrideTestTask
here, but it turns out that our `rake default` behaviour is to run rspec
anyway so while it would be marginally faster to run `rspec` directly
rather than go via `rake`, consistency wins out.

There was also some difference between how our custom script ran the
publishing end-to-end tests and how buildProject does it, but as far
as I can tell the eventual outcome is the same.  The major difference
was that we would always run it whereas buildProject is conditional, but
all the parameters come from the same place (or resolve to the same
values).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,63 +17,81 @@ node {
       )
     ],
     beforeTest: {
-      govuk.setEnvar("RCOV", "1")
-      govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+      setExtraEnvVars(govuk);
     },
     publishingE2ETests: true,
     afterTest: {
-      stage("Publish coverage") {
-        publishHTML(target: [
-          allowMissing: false,
-          alwaysLinkToLastBuild: false,
-          keepAll: true,
-          reportDir: "coverage/rcov",
-          reportFiles: "index.html",
-          reportName: "RCov Report"
-        ])
-      }
+      publishCoverage(govuk);
 
-      stage("Verify pact") {
-        sh "bundle exec rake pact:verify"
-      }
+      runPublishingApiPactTests(govuk);
 
-      stage("Publish pacts") {
-        withCredentials([[$class: "UsernamePasswordMultiBinding", credentialsId: "pact-broker-ci-dev",
-          usernameVariable: "PACT_BROKER_USERNAME", passwordVariable: "PACT_BROKER_PASSWORD"]]) {
-          withEnv(["PACT_TARGET_BRANCH=branch-${env.BRANCH_NAME}"]) {
-            sshagent(["govuk-ci-ssh-key"]) {
-              sh "bundle exec rake pact:publish:branch"
-            }
-          }
-        }
-      }
+      runContentStorePactTests(govuk);
+    }
+  )
+}
 
-      stage("Checkout content store") {
-        sh("rm -rf tmp/content-store")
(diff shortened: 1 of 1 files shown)


kevindew:
Cool! Thanks for doing this @h-lame 😎 🎉 

As the blocks are quite big can we put them in variables or functions? as we've ended up with a huge amount of code lines going into the arguments for the function

h-lame:
@kevindew - good idea.  I'll see what my groovy skills let me do.  🤞

h-lame:
@kevindew - took a few attempts, but how does 3bbc6d78748d6e806e1a01da30a746f12fb81c73 grab you?

Jenkinsfile
```diff
@@ -1,182 +1,97 @@
...
         name: "PUBLISHING_E2E_TESTS_BRANCH",
-        defaultValue: DEFAULT_PUBLISHING_E2E_TESTS_BRANCH,
+        defaultValue: "test-against",
         description: "The branch of publishing-e2e-tests to test against"
-      ),
-    ])
-  ])
-
-  try {
-    govuk.initializeParameters([
```
thomasleese: I think we have an issue here, now that this isn't called, the parameters don't get set as environment variables the first time as per: https://github.com/alphagov/govuk-puppet/blob/5ad9ad4b597a76a3011ccc8ff9cdc03f81dd4384/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L378

So this means below where they are used `sh("git checkout ${env.CONTENT_STORE_BRANCH}")`, it's coming up as `null`, as can be seen in: https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/3/console

h-lame: Ah, interesting.  Do you think this this something we should add back into this Jenkinsfile, or add to `buildProject` 'cos it's a more widespread problem?

thomasleese: I'm surprised this hasn't come up before, so I'm wondering how other teams are mitigating the problem. Adding it to `buildProject` feels like the right thing to do though.

h-lame: Yeah, I'll take a look - I think we can iterate over the existing params definitions and use their default values so we should be able to add it without extra config from the calling Jenkinsfile.

h-lame: @thomasleese I've got a PR in puppet here: https://github.com/alphagov/govuk-puppet/pull/6826 and I've used the jenkins replay feature to try it out on your failing build - https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/use_rabbitmq_envvar/15/console is a passing one (it took me a while).

PR #1067, opened 2017-11-21
Bump govspeak gem to include new button component

nickcolley:
I've tested this previously on integration

https://github.com/alphagov/govspeak/pull/118#issuecomment-346089074

https://trello.com/c/ppqXaGPk/122-create-a-govspeak-component-for-buttons


nickcolley:
Will hold off merging until after the freeze, so if someone has an emergency deployment they don't have to figure out if this change is going to block them 😄

PR #1068, opened 2017-11-22
Use rabbbitmq hosts defined in an environment variable.

deanwilson:
We now access these via different naming schemes in different environments
so accept the values provided by puppet rather than the hard coded
ones. CC /@kevindew 

This a subset of the changes made here - https://github.com/alphagov/email-alert-service/pull/80/files#diff-64f27c9e1694187fb14d1429cd7d409fR21

### DIFF
--- a/config/rabbitmq.yml
+++ b/config/rabbitmq.yml
@@ -16,9 +16,10 @@ test:
 
 production:
   hosts:
-    - rabbitmq-1.backend
-    - rabbitmq-2.backend
-    - rabbitmq-3.backend
+    <% hosts = ENV['RABBITMQ_HOSTS'] || 'localhost' %>
+    <% hosts.split(",").each do |host| %>
+    - <% host %>
+    <% end %>
   user: publishing_api
   pass: <%= ENV['RABBITMQ_PASSWORD'] %>
   port: 5672


kevindew:
This looks good to me. I am just wondering/investigating if we can strip this down to have an env_var of like:

```
production:
  url: <%= ENV['RABBITMQ_URL'] %>
  exchange: published_documents
  recover_from_connection_close: true
```
and an env var of: `ampq://user:pass@rabbitmq-1:5672.backend,rabbitmq-2.backend:5672,rabbitmq-3.backend:5672/`

Just to save us having extra env vars and making it more configurable for e2e tests (and other non puppet environments)

deanwilson:
We could possibly change the format but it'd be nice to get this in to fix AWS and then roll the further change across all the apps in a batch.

kevindew:
@deanwilson yup sure, it doesn't support the multi-host name syntax anyway 😢 - have you already got the env vars in puppet? or should the definition be `ENV.fetch('RABBITMQ_HOSTS'], 'rabbitmq-1.backend,rabbitmq-2.backend,rabbitmq-3.backend')` ?

surminus:
Can confirm we have the hosts in Puppet:

https://github.com/alphagov/govuk-puppet/blob/a863202a2af7169456cc4cc69da3b66be01c607a/hieradata/common.yaml#L479-L482

When we move to AWS, the way that we cluster RabbitMQ in an autoscaling group means that adding the host will be much nicer anyway:

https://github.com/alphagov/govuk-puppet/blob/a863202a2af7169456cc4cc69da3b66be01c607a/hieradata_aws/common.yaml#L484-L485

PR #1069, opened 2017-11-24
Fix Mainstream Publisher first published at dates

thomasleese:
I took the dates by looking at the earliest `updated_at` value for the archived or published editions of any document. We can't use `created_at` as that is when the document is drafted. There doesn't seem to be a better date field to pick though.

```
editions = Edition.archived_or_published.group_by { |e| e.content_id }.map { |k, v| [k, v.map(&:updated_at).sort.first] }
File.open("/tmp/mainstream-publisher-first-published-at.csv", "w") { |file| file.write(editions.compact.map(&:to_csv).join("")) }
```

[Trello Card](https://trello.com/c/m4oNTUum/390-fix-remaining-firstpublishedat-dates)


kevindew:
@thomasleese Isn't this date when it was first created as a draft rather than when it was published?

thomasleese:
@kevindew Yep... you're right... knew it was somehow too simple

thomasleese:
@kevindew This should have the right dates now.

kevindew:
@thomasleese Sorry I totally missed your notification all that time ago!

Only thing I'm concerned about is that could be other things (maybe unpublishing) ends up tweaking that updated at date. We do have actions available e.g: `Edition.first.actions.where(request_type: "publish").first.created_at`

PR #1076, opened 2017-12-07
 Add 'with_draft' option to '/lookup-by-base-path' endpoint

tomsabin:
During the development of the Topic Taxonomy in Content Tagger, a lot of
taxons are still in their draft states. Recently, we've had the
requirement to lookup taxons by their base path to provide an
informative error message when attempting to create a taxon with a
conflicting base path.

By introducing the 'with_draft' parameter, we can optionally include
draft states and content from both the live and draft content stores.

Trello: https://trello.com/c/WVTtiLie/150-provide-a-useful-error-message-in-content-tagger-when-you-try-to-create-or-move-a-taxon-where-one-already-exists

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -1,9 +1,16 @@
 class LookupsController < ApplicationController
   def by_base_path
-    # return content_ids for content that is visible on the live site
-    # withdrawn items are still visible
-    states = %w(published unpublished)
     base_paths = params.fetch(:base_paths)
+
+    if with_drafts
+      states = %w(draft published unpublished)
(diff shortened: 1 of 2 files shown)


kevindew:
Only had a quick look but this looks good to me. I'm just wondering if it ties into this work: https://github.com/alphagov/publishing-api/issues/812 at all? It looks a bit like that one was paused indefinitely but I'm not sure of full context

tomsabin:
Thanks for sharing the issue, I didn't know there had been prior work on a v2 lookup endpoint. Because of that, I'm also unsure of the full context, but I'll read through that issue and the related Trello cards, and catch up with the relevant people to find out more.

PR #1089, opened 2017-12-12
Upgrade gds-sso to latest version

adrianclay:
This didn't get picked up by dependabot and is needed by the E2E tests to run signon in mock mode while the applications run in production.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ else
   gem 'gds-api-adapters', "~> 50.5.0"
 end
 
-gem "gds-sso", "13.0.0"
+gem "gds-sso", "~> 13.4"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -18,7 +18,7 @@ else
   gem 'gds-api-adapters', "~> 47.9.1"
 end
 
-gem "gds-sso", "13.0.0"
+gem "gds-sso"
```
kevindew: Should this be tagged to `~> 13.4` so that it doesn't get bumped into breaking changes unexpectantly?

adrianclay:
Closing as has merge conflicts and dependabot has raised a PR at #1096

PR #1099, opened 2017-12-14
Add level_one_taxon<->root_taxon reverse links

koetsier:
Level one taxons are now linked to the home page via the 'root_taxon' link
The homepage in turn has a reverse link back to the level one taxons in a 'level_one_taxons' link.

See also: https://github.com/alphagov/govuk-content-schemas/pull/691


cbaines:
This is quite separated from the taxonomy stuff, so I think it would be really good to put lots of detail about what was previously being done, and the new situation this change is moving us towards.

PR #1104, opened 2017-12-18
Requeue by document type using sidekiq

dwhenry:
But moving the JSON generation to run as a sidekiq process this should
significantly improve the throughput when we need to requeue documents
on rabbit MQ for indexing in rummager.

https://trello.com/c/OBhgB1Lw/531-code-tidy-up


spec/workers/requeue_content_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe RequeueContent do
+  let(:edition) { FactoryGirl.create(:live_edition, base_path: '/ci1') }
+
+  describe "#call" do
```
brenetic: Might be good to change this from `#call`

dwhenry: good call..

PR #1107, opened 2017-12-20
Support enqueuing editions for multiple document types

Rosa-Fox:
[Trello](https://trello.com/c/2zUNyO8g/536-check-the-data-in-mainstream-is-consistent-with-the-data-in-govuk)

Paired with @JonathanHallam and @brenetic

### DIFF
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -29,21 +29,19 @@ namespace :queue do
   end
 
   desc "Add published editions to the message queue by document type"
-  task :requeue_document_type, [:document_type] => :environment do |_, args|
-    document_type = args[:document_type]
-    raise ValueError("expecting document_type") unless document_type.present?
+  task :requeue_document_type, [] => :environment do |_, args|
+    document_types = args.extras
 
-    # Restrict scope to stuff that's live (published or unpublished)
-    # Unpublished content without a content store representation won't
-    # be returned, but we're not interested in this content.
-    scope = Edition
-      .with_document
-      .with_unpublishing
-      .where(document_type: document_type)
-      .where(content_store: :live)
-      .select(:id)
+    raise(StandardError, "expecting document_type") unless document_types.present?
 
     version = Event.maximum(:id)
+
+    # Restrict scope to stuff that's live (published or withdrawn)
+    scope = Edition
+      .with_document
+      .where(document_type: document_types)
+      .where(content_store: :live)
+      .select(:id)
 
     scope.find_each do |edition|
       RequeueContent.perform_async(edition.id, version)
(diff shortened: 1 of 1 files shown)


lib/tasks/queue.rake
```diff
@@ -1,3 +1,4 @@
+class ValueError < RuntimeError; end
```
suzannehamilton: Where was this defined before, when it was called from line 34?

Rosa-Fox: It wasn't defined anywhere.

MatMoore: This is probably my fault as I wrote the original. I must have been thinking of ArgumentError and didn't test this case. I don't think there's any need to introduce a new class, just raise something sensible instead. 🙂

lib/tasks/queue.rake
```diff
@@ -29,31 +30,35 @@ namespace :queue do
...
+      # Unpublished content without a content store representation won't
+      # be returned, but we're not interested in this content.
+      scope = Edition
+        .with_document
+        .with_unpublishing
+        .where(document_type: document_type)
+        .where(content_store: :live)
+        .select(:id)
+
+      version = Event.maximum(:id)
```
MatMoore: you can take this outside of the loop. The purpose of the version is just to make sure that any later publishing events will always have a later version, so that if the represent downstream takes long a time to process, it can't overwrite updates that happened since it was triggered.

cbaines:
As far as I can tell "allow multiple editions to be sent to the message queue" was something that was previously supported. Would "Support enqueuing editions for multiple document types" better describe what is changing here?

lib/tasks/queue.rake
```diff
@@ -29,31 +30,35 @@ namespace :queue do
...
 
-    version = Event.maximum(:id)
+    document_types.each do |document_type|
+
+      # Restrict scope to stuff that's live (published or unpublished)
+      # Unpublished content without a content store representation won't
+      # be returned, but we're not interested in this content.
+      scope = Edition
+        .with_document
+        .with_unpublishing
```
cbaines: I know that you're just moving this code, but as far as I can see, the join on the unpublishings table isn't used.

MatMoore: Yep I think you're right. It was originally there as a way of filtering published+withdrawn content, but the `content_store: live` filter handles this now.

Rosa-Fox: @cbaines why wouldn't it be used?

lib/tasks/queue.rake
Rosa-Fox: @MatMoore have moved

lib/tasks/queue.rake
Rosa-Fox: @MatMoore have changed

lib/tasks/queue.rake
MatMoore: It's an outer join but I don't think is anything using values from `unpublishings` (correct me if I'm wrong)

cbaines: Indeed, I was also assuming that `with_unpublishing` was doing an outer join on the unpublishings table, and if you're just selecting the edition id, then doing (or not doing) that outer join shouldn't make a difference to the intentional behaviour of the code.

Rosa-Fox: 👍

PR #1115, opened 2017-12-28
Fix Collections Publisher first_published_at dates

kevindew:
Trello: https://trello.com/c/m4oNTUum/390-fix-remaining-firstpublishedat-dates

Publish dates are not stored within collections-publisher so the
approach I took to determining these was to use the public_updated_at
value for the first version of each document from collections-publisher.

Data was generated via the Publishing API using:

```
irb(main):004:0> first_published = Edition.with_document.where("first_published_at BETWEEN '2016-02-29 09:24:09' AND '2016-02-29 09:24:11'").where(publishing_app: "collections-publisher", user_facing_version: 1).pluck(:content_id, :public_updated_at)
irb(main):005:0> File.open("/tmp/collections-publisher-first-published-at.csv", "w") { |file| file.write(first_published.map(&:to_csv).join("")) }
```

On my local VM:

Before:
```
[1] pry(main)> Edition.with_document.where("first_published_at BETWEEN '2016-02-29 09:24:09' AND '2016-02-29 09:24:11'").where(publishing_app: "collections-publisher").count
=> 11773
```

After:
```
[3] pry(main)> Edition.with_document.where("first_published_at BETWEEN '2016-02-29 09:24:09' AND '2016-02-29 09:24:11'").where(publishing_app: "collections-publisher").count
=> 0
```


kevindew:
These dates will get resolved as part of: https://github.com/alphagov/publishing-api/pull/1117 so I'm going to close this.

PR #1129, opened 2018-01-10
Log invalid publishing queue messages

suzannehamilton:
Log messages which don't match the relevant notification schema to help us find inconsistencies between the publishing API and the content schemas.

Apps which consume these messages rely on the schemas to know what fields to expect.

Just log to a file for now. We may log to Sentry once we have an idea of the error volume and we have fixed the most common errors.

It would be good to get these logs into Kibana too, but I couldn't figure out how to do it. Does anyone know? I tried formatting them as JSON, but it didn't help. It might be because they end up in the Sidekiq lots in procfile_worker.err.log rather than app.err.log. Do we send Sidekiq logs to Kibana at all?

https://trello.com/c/uV5zDvzu/97-add-schema-tests-to-the-publishing-api

cc @kevindew

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -122,8 +122,8 @@ module Commands
       def schema_validator
         @schema_validator ||= SchemaValidator.new(
           payload: payload[:links],
-          links: true,
-          schema_name: schema_name
+          schema_name: schema_name,
+          schema_type: :links,
         )
       end
 
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -4,11 +4,11 @@ require 'govuk_schemas'
 class SchemaValidator
   attr_reader :errors
 
-  def initialize(payload:, links: false, schema: nil, schema_name: nil)
+  def initialize(payload:, schema: nil, schema_name: nil, schema_type: :publisher)
     @payload = payload
-    @links = links
     @schema = schema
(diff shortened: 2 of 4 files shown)


kevindew:
@suzannehamilton thanks for looking into this.

Sidekiq should be logged into kibana, and it did used to be - I'll have a look see if I can see any evidence that this has changed.

kevindew:
@suzannehamilton I had a look and they are still sent to kibana only they're now part of the same application name now rather than being publishing-api-sidekiq as they used to be: https://github.com/alphagov/govuk-puppet/commit/fb5a9fc6a5c15dae962cddda92692cad564bcc16 

I think the logging isn't working correctly on integration at the moment as I can't find anything under "application: publishing-api" there. 

I'm going to put in a PR to tag sidekiq log files with sidekiq so it's easier to find them. I'm not quite sure what something logged via the logger would appear as in them. By the way I'd probably suggest having this as a warning rather than an info logging. Also I think it'd be fine if they went into sentry tagged as a warning (but I'm not fully up to date on the guidance surrounding that).

suzannehamilton:
@kevindew Aha, I only checked on integration which is why I couldn't find any Sidekiq messages. I've changed it to `warn` and formatted the message as JSON. I'll check if it appears in Kibana on staging and production.

I'll add the logs to Sentry if they're not too noisy. I've found some validation failures for unpublishing messages, but I haven't seen any others yet.

PR #1141, opened 2018-01-18
Add task to setup exchange for RabbitMQ

RyanMacG:
We need the exchange and it's queues in E2E testing this rake task
allows for them to be setup using the existing config in publishing-api.


RyanMacG:
@kevindew yeah we just based it off the existing config bits that were used in other tasks, previously we hardcoded them to the values we expected but were concerned about that not being maintainable

PR #1157, opened 2018-02-08
Removes a change note

elenatanasoiu:
This commit removes a change note for an edition. 


elenatanasoiu:
Closing this PR as it's not needed. The change note has not been propagated from whitehall into publishing-api.

PR #1158, opened 2018-02-13
Add a rake task to generate a CSV of live documents

h-lame:
For: https://trello.com/c/Q4Ml1UnP/68-2-uprating-list-of-all-live-pages-that-came-through-publisher-or-are-smart-answers

This takes a list of publishing apps and extracts all live editions
published by those apps.  The editions are written to a CSV including
URL, title, document type, and first published at.

This report is intended to give a list of documents to the content
team to help them work out what content might need uprating.

### DIFF
--- a/lib/live_content_report_exporter.rb
+++ b/lib/live_content_report_exporter.rb
@@ -1,0 +1,56 @@
+require 'csv'
+
+class LiveContentReportExporter
+  def initialize(publishing_apps)
+    @publishing_apps = publishing_apps
+  end
+
+  def total
+    editions.count
+  end
+
+  def file_path
+    @file_path ||= Rails.root.join("live_content_report_#{publishing_apps_for_file_path}_#{time_for_file_path}.csv")
+  end
+
+  def export(progress: nil)
+    CSV.open(file_path, 'w') do |csv_out|
+      csv_out << csv_headers
+      editions.find_each.with_index do |edition, i|
+        progress.call(i, total) unless progress.nil?
+        csv_out << csv_row(edition)
+      end
+    end
+  end
+
+private
+
+  attr_reader :publishing_apps
+
+  def publishing_apps_for_file_path
+    @publishing_apps_for_file_path ||= publishing_apps.join('_')
+  end
+
+  def time_for_file_path
+    @time_for_file_path ||= Time.zone.now.strftime('%Y%m%d%H%M%S%N')
+  end
+
+  def editions
+    @editions ||= Edition.
+      renderable_content.
+      where(phase: 'live', state: 'published', publishing_app: publishing_apps)
+  end
+
+  def csv_headers
+    @csv_headers ||= ['URL', 'Page title', 'Format', 'First published at']
+  end
+
+  def csv_row(edition)
+    [
+      edition.web_url,
+      edition.title,
+      edition.document_type,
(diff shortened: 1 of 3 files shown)


lib/tasks/live_content_report.rake
```diff
@@ -0,0 +1,38 @@
+require 'csv'
+require 'tempfile'
+
+desc 'Generates a CSV report of all live documents for the supplied publishing app(s)'
+task :live_content_report, [] => :environment do |_, args|
```
thomasleese: Given the size of this Rake task, I would be tempted to move it into a `ContentReportExporter` class in `lib` which does all this and then it can be easily tested.

h-lame: I thought it was roughly equivalent to existing rake tasks, but rubocop seems to agree with you `lib/tasks/live_content_report.rake:5:1: C: Metrics/BlockLength: Block has too many lines. [28/25]`, so I'll do a thing.

h-lame: Did a rebase where I extracted code from the rake file to an object in lib that can be tested.

lib/live_content_report_exporter.rb
```diff
@@ -0,0 +1,40 @@
+require 'csv'
+
+class LiveContentReportExporter
+  attr_reader :publishing_apps
```
thomasleese: I would make this private if possible.

lib/live_content_report_exporter.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(publishing_apps)
+    @publishing_apps = publishing_apps
+  end
+
+  def total
+    editions.count
+  end
+
+  def file_path
+    @file_path ||= Rails.root.join("live_content_report_#{publishing_apps.join('_')}_#{Time.zone.now.strftime('%Y%m%d%H%M%S%N')}.csv")
```
thomasleese: Might be clearer to extract the `publishing_apps.join('_')` and `Time.zone.now.strftime('%Y%m%d%H%M%S%N')` into variables and then interpolate them in.

lib/live_content_report_exporter.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def export(progress: nil)
+    CSV.open(file_path, 'w') do |csv_out|
+      csv_out << ['URL', 'Page title', 'Format', 'First published at']
+      editions.find_each.with_index do |edition, i|
+        progress.call(i, total) unless progress.nil?
+        csv_out << [
```
thomasleese: I might extract this into a `csv_columns_for_row(edition)` method and then have a separate `csv_headers` method which returns the first row headers.

lib/live_content_report_exporter.rb
h-lame: :+1:

lib/live_content_report_exporter.rb
h-lame: ugh, yeah, I should have spotted that (I used methods, not variables, because I prefer memoized methods to be oneliners).

lib/live_content_report_exporter.rb
h-lame: yeah, that gives us an extension point if we want other types of reports.  :+1:

PR #1160, opened 2018-02-15
Removes change note

elenatanasoiu:
This will remove a change note added in error on a publication:

https://www.gov.uk/government/publications/sen-and-disability-reform-s31-grant-determinations#history

The change was requested in this zendesk ticket: https://govuk.zendesk.com/agent/tickets/2566949

### DIFF
--- a/db/migrate/20180215112954_fix_whitehall_remove_change_note.rb
+++ b/db/migrate/20180215112954_fix_whitehall_remove_change_note.rb
@@ -1,0 +1,27 @@
+class FixWhitehallRemoveChangeNote < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  
+  def up
+    # Find change notes
+    d = Document.where(content_id: "b3b96cb9-e7fa-493c-87a5-31d71d79c7a7").first
+    if d.present?
+      change_notes = d.editions.map(&:change_note).compact
+      change_notes.select{ |change_note| change_note.note == "Added funding allocations for individual LAs for 2018 to 2019."}
+  
+      # Get rid of change notes
+      change_notes.map(&:destroy)
+  
+      # Re-present editions to content store
+      content_ids = change_notes.map(&:edition_id)
+      puts "The editions that need to be represented downstream are: #{content_ids}"
+  
+      if Rails.env.production?
+        Commands::V2::RepresentDownstream.new.call(content_ids)
+      end
+    end
+  end
+  
+  def down
+    # This migration is not reversible
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


db/migrate/20180215112954_fix_whitehall_remove_change_note.rb
```diff
@@ -0,0 +1,15 @@
+class FixWhitehallRemoveChangeNote < ActiveRecord::Migration[5.1]
+  def up
+    d = Document.where(content_id: "b3b96cb9-e7fa-493c-87a5-31d71d79c7a7").first
+    change_notes = d.editions.map(&:change_note).compact
+    change_notes.select{ |change_note| change_note.note == "Added funding allocations for individual LAs for 2018 to 2019."}
+    # Re-present editions to content store
+    puts "The editions that need to be represented downstream are: #{change_notes.map(&:edition_id)}"
```
thomasleese: It is actually actually possible to represent stuff from the migration, see https://github.com/alphagov/publishing-api/blob/fab6a131d52b9759160b4004f9e11953dad50ba7/db/migrate/20171011154900_fix_remaining_specialist_publisher_first_published_at_dates.rb#L13 for some prior art here. An important thing to note is the `disable_ddl_transaction!` at the top.

elenatanasoiu: Thanks!

elenatanasoiu:
I've added the `schema.rb`.

I've also added a link to the Zendesk ticket requiring the change. In a nutshell an HTML attachment was added to a different document by mistake and we've had to make a redirect for it. The person who added the original attachment also added a change note which is visible in the UI and is no longer accurate. Initially we were provided with the whitehall-admin link of the document and asked to remove the latest change note. This was actually not the same note as the latest one visible in the UI, which is the note they were actually referring to, but hadn't specified. 

So now we're removing the visible UI change note as requested.

db/migrate/20180215112954_fix_whitehall_remove_change_note.rb
```diff
@@ -0,0 +1,27 @@
...
+  
+  def up
+    # Find change notes
+    d = Document.where(content_id: "b3b96cb9-e7fa-493c-87a5-31d71d79c7a7").first
+    if d.present?
+      change_notes = d.editions.map(&:change_note).compact
+      change_notes.select{ |change_note| change_note.note == "Added funding allocations for individual LAs for 2018 to 2019."}
+  
+      # Get rid of change notes
+      change_notes.map(&:destroy)
```
thomasleese: I wonder if this might delete all the change notes because the `select` call above doesn't get reassigned to the `change_notes` variable?

elenatanasoiu: Thanks! Corrected here: https://github.com/alphagov/publishing-api/pull/1166

PR #1167, opened 2018-02-16
Clears change note history together with deletion

elenatanasoiu:
Change note history does not get regenerated when you delete a change note and represent it downstream to content store.

This has to be cleared manually.

### DIFF
--- a/db/migrate/20180216150252_remove_change_history_for_edition.rb
+++ b/db/migrate/20180216150252_remove_change_history_for_edition.rb
@@ -1,0 +1,21 @@
+class RemoveChangeHistoryForEdition < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  
+  def change
+    begin
+      edition = Edition.find(2975826)
+  
+      if edition.present?
+        edition_details = edition.details
+        edition_details.delete(:change_history)
+        edition.update!(details: edition_details)
+    
+        if Rails.env.production?
+          Commands::V2::RepresentDownstream.new.call(edition.content_id)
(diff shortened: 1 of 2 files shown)


db/migrate/20180216150252_remove_change_history_for_edition.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveChangeHistoryForEdition < ActiveRecord::Migration[5.1]
+  def change
+    edition = Edition.find(2975826)
+
+    edition_details = edition.details
+    edition_details[:change_history].reject!{|n| n[:note] == "Added funding allocations for individual LAs for 2018 to 2019."}
+    edition.update!(details: edition_details)
+
+    if Rails.env.production?
+      Commands::V2::RepresentDownstream.new.call(edition.content_id)
```
thomasleese: You'll need to use `disable_ddl_transaction!` to use this.

db/migrate/20180216150252_remove_change_history_for_edition.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveChangeHistoryForEdition < ActiveRecord::Migration[5.1]
+  def change
+    edition = Edition.find(2975826)
```
thomasleese: Would be good to quit early if the `edition` doesn't exist.

elenatanasoiu:
Thanks @thomasleese
I've gone with your suggestion and removed change notes entirely. And added a check for edition. You caught me before I could add it. :)

PR #1179, opened 2018-02-27
Bump govuk_document_type gem version

fmlharrison:
The new version of the gem has the newly added
content_purpose_supergroup and content_purpose_subgroup content item fields

https://trello.com/c/IsIQwZzn/93-republish-content-items-to-pickup-subgroup-and-supergroup


leenagupte:
@fmlharrison I think it's this test that's failing: https://github.com/alphagov/publishing-api/blob/master/spec/pacts/content_store/put_endpoint_spec.rb#L87

PR #1184, opened 2018-03-01
Change a represent_downstream:document_types task

fmlharrison:
An update to the rake task so that it accepts a list of multiple
document types as a parameter. This change was made in order to
republish content items in batches to include the new supergroup and subgroup
fields.

https://trello.com/c/IsIQwZzn/93-republish-content-items-to-pickup-subgroup-and-supergroup

### DIFF
--- a/lib/tasks/represent_downstream.rake
+++ b/lib/tasks/represent_downstream.rake
@@ -15,13 +15,14 @@ namespace :represent_downstream do
   end
 
   desc "
-  Represent downstream for a specific document_type
+  Represent downstream for individual or multiple document_types
   Usage
-  rake 'represent_downstream:document_type[:document_type]'
+  rake 'represent_downstream:document_type[:document_types]'
   "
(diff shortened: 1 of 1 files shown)


lib/tasks/represent_downstream.rake
```diff
@@ -15,13 +15,14 @@ namespace :represent_downstream do
...
 
   desc "
-  Represent downstream for a specific document_type
+  Represent downstream for individual or multiple document_types
   Usage
-  rake 'represent_downstream:document_type[:document_type]'
+  rake 'represent_downstream:document_type[:document_types]'
   "
-  task :document_type, [:document_type] => :environment do |_t, args|
+  task :document_types, [:document_types] => :environment do |_t, args|
```
thomasleese: I think we should keep the task name as `document_type` as that might break some other users of this task.

fmlharrison: great point

PR #1189, opened 2018-03-12
Determine payload_version number in worker

kevindew:
This is to address an issue in the Publishing API where the content
store is not updated due to conflicts.

What can happen is that an item can be in the queue for downstreaming
with an older payload_version than the current one. Any updates to the
item will trigger a new queuing with a newer payload version - however
because of sidekiq jobs these will be discarded. Thus when the job
eventually runs it may present an up to date representation of a
document but it will still contain an old payload_version id which
causes the content store to ignore it with a 409.

The approach to fix this is to calculate the payload version at the time
we generate the document for downstreaming and we can do this by pulling
out the last id in the event table for this content id (it could even be
the last id in the entire table though as it doesn't matter if this is
associated with the edition or not).

The passing of payload_version as part of downstreaming made more sense
when we were constructing the downstream representation at runtime. This
is now no longer the case and thus risks problems.

The only thing I think we lose by doing this is having the payload
version associated with the sidekiq job as a means to trace back to
the event. I'm not aware that anyone used this for tracing though.

### DIFF
--- a/app/workers/downstream_draft_worker.rb
+++ b/app/workers/downstream_draft_worker.rb
@@ -54,7 +54,7 @@ private
     @content_id = attributes.fetch(:content_id)
     @locale = attributes.fetch(:locale)
     @edition = Queries::GetEditionForContentStore.(content_id, locale, true)
-    @payload_version = attributes.fetch(:payload_version)
+    @payload_version = Event.payload_version(content_id)
     @orphaned_content_ids = attributes.fetch(:orphaned_content_ids, [])
     @update_dependencies = attributes.fetch(:update_dependencies, true)
     @dependency_resolution_source_content_id = attributes.fetch(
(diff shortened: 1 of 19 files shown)


app/workers/downstream_draft_worker.rb
```diff
@@ -54,7 +54,7 @@ def assign_attributes(attributes)
     @content_id = attributes.fetch(:content_id)
     @locale = attributes.fetch(:locale)
     @edition = Queries::GetEditionForContentStore.(content_id, locale, true)
-    @payload_version = attributes.fetch(:payload_version)
+    @payload_version = Event.where(content_id: content_id).maximum(:id) || 0
```
thomasleese: Can we use the `Event.payload_version` method here?

kevindew: Whoops - thanks!

PR #1194, opened 2018-03-13
Adds the root-taxon link to the expansion rules

koetsier:
Level one taxons do not have any links to parent taxons but do have a link to the root taxon. This change adds a

`links->taxons->links->root_taxon`

link to the expanded links of all content that is directly linked to a level one taxon.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,3 +7,7 @@ AllCops:
 
 Style/ConditionalAssignment:
   Enabled: false
+
+Style/SymbolArray:
+  Exclude:
+    - lib/expansion_rules.rb
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -18,6 +18,7 @@ module ExpansionRules
     [:parent.recurring],
     [:parent_taxons.recurring],
     [:parent_taxons.recurring, :root_taxon],
+    [:taxons, :root_taxon],
     [:taxons, :parent_taxons.recurring],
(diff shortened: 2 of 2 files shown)


lib/expansion_rules.rb
```diff
@@ -18,6 +18,7 @@ def recurring
     [:parent.recurring],
     [:parent_taxons.recurring],
     [:parent_taxons.recurring, :root_taxon],
+    %i[taxons root_taxon],
```
cbaines: To improve readability, I'd prefer to always use the `[:taxons, :root_taxon]` representation.

koetsier: rubocop says different :(

cbaines: Maybe disable that check for this file? I think you can add a comment of this form `# rubocop:disable ...` [1].

1: https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md

koetsier: Good idea! Added that in new commit. 👍

.rubocop.yml
```diff
@@ -7,3 +7,7 @@ AllCops:
 
 Style/ConditionalAssignment:
   Enabled: false
+
+Style/SymbolArray:
+  Exclude:
+    - lib/expansion_rules.rb
```
cbaines: I'd have gone for the comment in the file, as I think that is a little easier to understand compared to this, but I don't have a strong preference.

PR #1211, opened 2018-03-26
Send payload version without querying on the content_id

thomasleese:
In 11178765670842cca8124a79dc8509dd6b322518 we made to change to take the `payload_version` from the maximum ID of all events associated with the `content_id`. Previously the `payload_version` was coming from the ID of the event which triggered the main publishing event.

This PR slightly reverts the behaviour to instead use the `payload_version` taken from the maximum ID of any event rather than an event associated with the `content_id`.

This fixes a problem we have where publishing a document which has links to other documents would cause those linked documents to be republished, but using their last event ID as the payload version rather than the event ID of the original publish, meaning that it will arrive at the content store with a `payload_version` that isn't greater than the version stored in the content store and therefore end with a 409 conflict error.

An alternative to this PR would be to create a new event for each linked document when publishing a document.


kevindew:
Only had a quick glance over this - will look more soon - but would an alternative be to just use the maximum event id?

thomasleese:
@kevindew the title of this PR might be badly written... but that is basically what we're doing in this PR

kevindew:
ah ha sorry - saw revert and thought it took it back to how it used to be.

I couldn't think of a problem with this approach back when I did the last tweaks but was just a bit nervous about it incase I'd overlooked something. Seemingly I overlooked a different thing!

I feel it might be prudent to drop the payload_version method on Event and replace it with a maximum_id one or just call that directly, feels like it maybe adds an unnecessary bit of misdirection.

PR #1224, opened 2018-04-17
Pagination can use multiple sort fields.

koetsier:
Pagination now allows multiple fields to be specified to order results. They should be separated by
commas; i.e. "-public_updated_at, id".

Previously pagination used just public_updated_at as a default field to sort results. As this field is not guaranteed to be unique, this can (and does) cause unexpected results where duplicates are returned.

This PR adds 'id' as second default sort field.

### DIFF
--- a/app/pagination.rb
+++ b/app/pagination.rb
@@ -28,19 +28,21 @@ private
   end
 
   def order_from_options(options)
-    order = options.fetch(:order, "-public_updated_at")
+    order_string = options.fetch(:order, '-public_updated_at')
 
-    if order.start_with?("-")
-      field = order[1..-1].to_sym
-      direction = :desc
-    else
-      field = order.to_sym
-      direction = :asc
+    orders = order_string.split(',').map(&:strip)
+
+    orders.map do |order|
+      if order.start_with?("-")
+        field = order[1..-1].to_sym
+        direction = :desc
+      else
+        field = order.to_sym
+        direction = :asc
+      end
+      raise_unless_valid_order_field(field)
+      [field, direction]
     end
-
-    raise_unless_valid_order_field(field)
-
-    { field => direction }
   end
 
   def raise_unless_valid_order_field(field)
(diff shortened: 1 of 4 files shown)


tuzz:
If the desired result is the results come back in a deterministic order, would it make sense to always tie-break based on `content_id`. That way, if a caller overrides this sort option, it would still produce deterministic results?

thomasleese:
I agree with @tuzz here, and I wonder if we should just use `id` rather than `content_id` as it might make it clearer that it's only to sort out tie breaks rather than to get the results back in a specific order.

app/pagination.rb
```diff
@@ -28,19 +28,22 @@ def offset_from_page
...
-      field = order[1..-1].to_sym
-      direction = :desc
-    else
-      field = order.to_sym
-      direction = :asc
-    end
-
-    raise_unless_valid_order_field(field)
+    orders = order_string.split(',').map(&:strip)
+    orders << 'id' if orders.none? { |o| o.match('[-]?id') }
```
cbaines: Do you need the `[ ]`'s here around `-`, as it's only one character?

app/pagination.rb
```diff
@@ -28,19 +28,22 @@ def offset_from_page
...
+    orders.each_with_object([]) do |order, result|
+      if order.start_with?("-")
+        field = order[1..-1].to_sym
+        direction = :desc
+      else
+        field = order.to_sym
+        direction = :asc
+      end
+      raise_unless_valid_order_field(field)
+      result << [field, direction]
```
cbaines: Could this be a `map`, rather than `each_with_object`, as you are constructing another array?

app/pagination.rb
koetsier: fixed

app/pagination.rb
koetsier: Good point 😀 - fixed

PR #1233, opened 2018-04-30
Stop presenting need_ids

cbaines:
Don't include `need_ids` in the payloads for the Content Store or
message queue. This information is now represented through the
`meets_user_needs` link type.

Turns out the previous Pull Request [1] didn't quite make that happen.

1: https://github.com/alphagov/publishing-api/pull/1230


cbaines:
The first step, is to make them optional in the `govuk-content-schemas`, which this PR does https://github.com/alphagov/govuk-content-schemas/pull/754

PR #1241, opened 2018-05-22
Add a legacy_taxons expansion rule

cbaines:
Whitehall loads the Topic Taxonomy through expanded links for each
level 1 taxon, so this means it will be included in the response.

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -14,6 +14,7 @@ module ExpansionRules
   MULTI_LEVEL_LINK_PATHS = [
     [:associated_taxons.recurring],
     [:child_taxons, :associated_taxons.recurring],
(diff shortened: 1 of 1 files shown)


cbaines:
> Why is this needed?

It makes it possible to easily load the legacy taxons in to the Topic Taxonomy data in Whitehall. The Whitehall part is here https://github.com/alphagov/whitehall/pull/4049/commits/7c818e826aab9d9be8682cdcafeb56c20d7ee0ed#diff-879c9da7af56ccdb0248979fd73a5cc2R33

I know we've discussed putting the data in Whitehall only, but using the existing support in Content Tagger and loading the mapping in to Whitehall through the Publishing API is the way I think I can get this done with the least amount of effort on my part.

PR #1242, opened 2018-05-25
Task to queue messages for all content

MatMoore:
Paired with @chao-xian 

Example usage: `rake queue:requeue_all_the_things[bulk.data-warehouse]`

Currently in content performance manager, the way we initially
populate our content data is very different to how we incrementally
update it as content changes. This is because the API returns us
data in a different format to the message queue.

In particular, it doesn't send us links or the details hash,
so we have to make additional calls to the content store to
fully populate the content. These calls are architecturally
undesirable because the content store is supposed to be used
for serving the frontend of the site.

We are thinking that it would be simpler
if we could drive everything from the message queue (this
is the solution previously adopted by search). To do this
we need some way of sending all the live content items through
the system on demand. This is expected to be run only as a developer
task, so we would pause the normal consumer in content performance
manager, run this rake task, and wait for everything to be
processed.

# Next steps
- Update the new CPM consumer to process messages instead of discarding them
- We need to measure how long this takes end to end
- We should write documentation for it

# Trello
- https://trello.com/c/bIKQuMU3/75-content-performance-manager-is-talking-to-content-store
- https://trello.com/c/m3vzLnR5/372-use-only-publishing-api-to-preload-items

### DIFF
--- a/app/workers/requeue_content.rb
+++ b/app/workers/requeue_content.rb
@@ -4,7 +4,7 @@ class RequeueContent
 
   sidekiq_options queue: :import
 
-  def perform(edition_id, version)
+  def perform(edition_id, version, action = "bulk.reindex")
     edition = Edition.find(edition_id)
     presenter = DownstreamPayload.new(edition, version, draft: false)
     queue_payload = presenter.message_queue_payload
@@ -14,7 +14,7 @@ class RequeueContent
     # because we don't want to send additional email alerts to users.
     service.send_message(
       queue_payload,
-      routing_key: "#{edition.schema_name}.bulk.reindex",
+      routing_key: "#{edition.schema_name}.#{action}",
       persistent: false
     )
   end
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -1,3 +1,5 @@
+MAX_IMPORT_QUEUE_SIZE = 1000
+
 namespace :queue do
   desc "Watch the queue, and print messages on the console"
   task watcher: :environment do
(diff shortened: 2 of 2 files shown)


chao-xian:
[(3) [Tech debt] Use only publishing api to preload items](https://trello.com/c/m3vzLnR5/372-3-tech-debt-use-only-publishing-api-to-preload-items)

kevindew:
Hello :wave:, sorry to wade in without maybe understanding everything.

This strikes me as a little strange as we're using an API app to tell it to send all it's data to a different app. Whereas a more logical approach would strike me as we traverse the data from the API endpoints. However I do understand you may not have everything as you'd expect which causes problems.

I wonder though if efforts would be better used to making the API endpoints provide this functionality - as it just seems a little odd and architectually undesirable to utilise message queues for this type of communication.

If I recall the /editions endpoint was created to ease content-performance-manager with traversing lots of Publishing API data. Might this be something that can be iterated to achieve the same goals and make it easier for subsequent applications to do similar tasks.

Apologies if I've overlooked something large in my comment here.

MatMoore:
Not at all, this is worth discussing. Me and @chao-xian are both off most of next week so we're not planning on merging this right away.

We discussed a couple of different options as a team but haven't considered changing the API. My only concern is that making the API work the way we'd like it to is not going to be straightforward.

Ideally I'd like the basic structure of the content items we process to be consistent, regardless of how we get them (message queue or API). We don't expect to preload items very often, so we don't want to have to change and test two different processes every time something changes upstream.

I have previously worked on a [bulk read endpoint](https://github.com/alphagov/publishing-api/pull/305) for something else, but it wasn't done in a way that was consistent with the rest of the API and we removed it later.

@pmanrubia what do you think?

kevindew:
Sorry for the delay getting back to you. 

> We only need to run this code once in the lifetime of the product, so from an architectural perspective, doesn't feel like a big deal. It is a one-off thing.

If this is just a one off and it's due to be removed then sure it's fine to do whatever is easiest.

> Publishing-API raises timeout errors from time to time. It seems that it relies on the database Cache to be able to serve requests that demand complex results.

Do you know if this is with the edition endpoint or the get content endpoint? Doesn't seem good

I think in general we should be avoiding using message queue to populate things that feels much more what the API should do but I appreciate the API doesn't have parity there. Hypothetically if the API did return responses consistent with message queue responses would that be a cleaner approach for you? well if we got past the timeouts and such (not that we'll have any time to work on this soon but architectural goals and that)

pmanrubia:
> Do you know if this is with the edition endpoint or the get content endpoint? Doesn't seem good

It happens in the editions endpoint, but also with single requests to retrieve a content item from the publishing api. It randomly returns 504, but retrying  use to do the trick. This is why we used Sidekiq and [report to Sentry only the latest retry fails] (https://github.com/alphagov/content-performance-manager/blob/9a26dde218d34791fb0236340069e966cb32addb/app/etl/items/jobs/application_job.rb#L12-L12). 

> I think in general we should be avoiding using message queue to populate things that feels much more what the API should do but I appreciate the API doesn't have parity there. Hypothetically if the API did return responses consistent with message queue responses would that be a cleaner approach for you? well if we got past the timeouts and such (not that we'll have any time to work on this soon but architectural goals and that)

Yes, I totally agree @kevindew; we started using the API but couldn't make it work consistently. I will raise this issue (about the Publishing-API timeouts) so we can put a solution for good, as this is a pattern that is likely to be repeated.

We really appreciate your help, @kevindew, as always 👍

MatMoore:
Hi all, 

Any objections to me merging this as is? I'm going to flesh out the consumer process separately.

PR #1252, opened 2018-06-12
EU Exit Data Export

thomasleese:
This adds a new Rake task which exports data from the `publishing-api` as requested by the EU exit team. The task takes a few hours to run.

It's an export of content items from Whitehall for each organisation, with information about what taxons that content is tagged to, a CSV file with these headings `title,url,last updated,document type,the taxons it is tagged to,"Is this Brexit-related?"`.

I've intentionally not added tests to this class because I don't think they would add any value for something we will run once, but happy to include some if people think it is worth it.

Paired with @binaryberry and @jonodrew.

This is the first part of [Trello Card](https://trello.com/c/1ykB19S8/209-data-exports-for-eu-exit).

### DIFF
--- a/lib/tasks/whitehall_eu_exit_report.rake
+++ b/lib/tasks/whitehall_eu_exit_report.rake
@@ -1,0 +1,4 @@
+desc "Generate a report of all Whitehall content per organisation."
+task :whitehall_eu_exit_report, [:path] => :environment do |_, args|
+  WhitehallEuExitReport.call(path: args[:path])
+end
--- a/lib/whitehall_eu_exit_report.rb
+++ b/lib/whitehall_eu_exit_report.rb
@@ -1,0 +1,177 @@
+require "fileutils"
+
+class WhitehallEuExitReport
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def initialize(path:)
+    @path = path
+  end
+
+  def call
+    FileUtils::mkdir_p(path)
+
+    organisations.find_each do |organisation|
+      content = content_for(organisation: organisation)
+      next if content.empty?
+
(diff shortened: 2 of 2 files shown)


lib/brexit_data_exporter.rb
```diff
@@ -0,0 +1,176 @@
...
+    "c64eb759-93ff-4bfd-86f2-16753fc1e062", # entering and staying in the uk / immigration rules
+    "29480b00-dc4d-49a0-b48c-25dda8569325", # entering and staying in the uk / visas and entry clearance
+    "262e9ea7-90fc-4ac8-8aa7-fc294282a1d4", # entering and staying in the uk / border control
+    "fa13521f-9285-45b0-bd65-4a472a8037e7", # entering and staying in the uk / immigration offences
+    "fef7e737-6f1a-4ef4-b844-aa24b630ad03", # entering and staying in the uk / permanant stay in the uk
+    "08a8a69f-2825-4fe2-a4cf-c83458a5629e", # entering and staying in the uk / refugee asylum and human rights
+
+    "b14a6778-60a2-483a-924c-774d1040cbf4", # environment / commercial fishing and fisheries
+    "52ff5c99-a17b-42c4-a9d7-2cc92cccca39", # environment / food and farming
+    "42fdd045-cadf-474b-b8a6-ca20a77c43bb", # environment / marine
```
binaryberry: You're missing environment - Farming and food grants and payments

thomasleese: Good catch, thanks, I've added that in now!

thomasleese:
Replaced by https://github.com/alphagov/publishing-api/pull/1253

PR #1253, opened 2018-06-13
[WIP] Whitehall EU Exit Report

thomasleese:


lib/whitehall_eu_exit_report.rb
```diff
@@ -0,0 +1,202 @@
...
+
+  def initialize(path:)
+    @path = path
+  end
+
+  def call
+    organisations.find_each do |organisation|
+      puts organisation.title, organisation.content_id
+
+      slug = organisation.base_path[26..-1] # /government/organisations/
```
kevindew: Are we confident these don't contain any nasty characters?

thomasleese: I've ran it locally with a full replication without any weird characters!

PR #1254, opened 2018-06-14
Rake tasks to create and publish special routes

kevindew:
Trello: https://trello.com/c/j3ZOI1fQ/227-add-route-to-the-design-system-at-design-system

This introduces two rake tasks that allow special routes to be created and published.

It is intended to be used for publishing the GOV.UK design system by running the task as `special_route:draft[/design-system, Design System, design-system]` and then a subsequent publish task.

There is a content schemas PR, https://github.com/alphagov/govuk-content-schemas/pull/775, which adds the necessary app names so this can be used.

### DIFF
--- a/doc/admin-tasks.md
+++ b/doc/admin-tasks.md
@@ -26,38 +26,38 @@ The following tasks will allow you to specify which content items/editions to ad
 
 * Represent all editions downstream
 ```
-bundle exec represent_downstream:all
+bundle exec rake represent_downstream:all
 ```
 N.B. This task will take several hours and should be used with caution.
 
 * Represent downstream for a specific document_type
 ```
-bundle exec represent_downstream:document_type['a-document-type']
+bundle exec rake represent_downstream:document_type['a-document-type']
 ```
 
 * Represent downstream for a rendering application
 ```
-bundle exec represent_downstream:rendering_app['application-name']
+bundle exec rake represent_downstream:rendering_app['application-name']
 ```
 
 * Represent downstream for a publishing application
 ```
-bundle exec represent_downstream:publishing_app['application-name']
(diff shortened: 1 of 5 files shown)


MatMoore:
FYI centralising special routes came up before in https://github.com/alphagov/govuk-rfcs/pull/85

I'm a bit uneasy about this change because it feels like special routes are not very well defined - some have very minimal metadata whereas others seem to add HTML or other data to the special route content item. 

It's clear how they are rendered, but how should other parts of the system handle them?

When we were working on search we had to add a [blacklist](https://github.com/alphagov/rummager/blob/master/config/govuk_index/migrated_formats.yaml#L290) because it wasn't clear which things should be searchable. And if there is no text in the content item then all search has to go on is the title.

The same thing applies to the data warehouse - is this something we want to analyse in the same as way as every other content item? If so we need a way of extracting the text of the page.

kevindew:
@MatMoore Yeah they're an opaque part of the stack.

I think with their "special" nature there isn't a guarantee of consistency between them. I think of them as basically means to mount applications on GOV.UK, by forgoing the standard content item approach they can't be searched, they can't be email alerts, they're an outlier - but that's just my opinion.

I think though that this PR is just there to add ease of how to add ease and consistency in adding them to the system rather than whether they should exist as a concept. It feels like that has already past and they're now a part of the stack whether or not they should be.

tijmenb:
Since the design system will live on a service domain for now, maybe we can take the opportunity to rethink this? I think it would be better to either do this in service manual (because it's related), or start doing these special routes from short-url-manager. I think it would be a good principle to keep that all publishing of content should go via publishing-api's HTTP API, for consistency with the rest of the platform.

bevanloon:
Closing as the team are going with a service domain

PR #1258, opened 2018-06-18
Check for invalid links in the patch links command

MatMoore:
We shouldn't be allowing publishing apps to send any old rubbish to the publishing API.
This code is supposed to log when links are invalid, but it didn't actually do anything.
For now, we just want to log to sentry. Once sentry is reporting no new errors coming in,
we can make this a 400 error (similar to put content).

That said, the request can currently accept links for content that doesn't
exist; i.e. the publishing app calls patch links *before* creating a draft.
I'm not sure if we actually need to support this use case?

As long as we do support this there will always been an edge case where invalid links
can sneak into the publishing API. But checking what we can is better than not doing
the check at all.

<strike>I've also added a rake task to revalidate the existing link sets.</strike>

See also: https://github.com/alphagov/govuk-content-schemas/pull/781


app/commands/v2/patch_link_set.rb
```diff
@@ -109,17 +109,27 @@ def downstream_live(content_id, locale, orphaned_content_ids)
...
         return true unless schema_name
 
-        # Do not raise anything yet
-        # Only send errbit notification
-        schema_validator.valid?
+        unless schema_validator.valid?
+          # Do not raise anything yet, only report a warning.
+          # This should be changed to an error once we are confident nothing is sending
+          # broken links.
+          GovukError.notify(
```
tijmenb: Bit late to this party, but this is contrary to the approach described in https://docs.publishing.service.gov.uk/manual/errors.html#manually-logged-errors. Would it be better if this just logs to Kibana and/or statsd?

MatMoore: Ah. Yeah, this should log to kibana instead. I can raise another PR but probably won't get to it until Friday or next week. If anyone wants to raise this now let me know and I'll review.

PR #1275, opened 2018-07-03
GET /v2/editions can now filter on document_types

koetsier:
- Allow GET /v2/editions to return empty sets
- GET /v2/editions can now filter on document_types

### DIFF
--- a/spec/controllers/v2/editions_controller_spec.rb
+++ b/spec/controllers/v2/editions_controller_spec.rb
@@ -124,6 +124,16 @@ RSpec.describe V2::EditionsController do
       end
     end
 
+    context "filtered so that the result set is empty" do
+      it "returns an empty set" do
+        get :index, params: { locale: "nowhere" }
+        expect(parsed_response["results"].count).to eq(0)
+        expect(parsed_response["links"]).to eq([
+          { "href" => "http://test.host/v2/editions?locale=nowhere", "rel" => "self" },
+        ])
+      end
+    end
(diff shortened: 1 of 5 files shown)


spec/controllers/v2/editions_controller_spec.rb
```diff
@@ -124,6 +124,27 @@ def u(string)
       end
     end
 
+    context "filtered so that the result set is empty" do
+      it "returns an empty set" do
+        get :index, params: { locale: "nowhere" }
+        expect(parsed_response["results"].count).to eq(0)
+        expect(parsed_response["links"]).to eq([
+                                                 { "href" => "http://test.host/v2/editions?locale=nowhere", "rel" => "self" },
```
cbaines: This line is a bit long.

spec/controllers/v2/editions_controller_spec.rb
```diff
@@ -124,6 +124,27 @@ def u(string)
...
+                                               ])
+      end
+    end
+
+    context "filtered by document type" do
+      it "returns only the document type specified" do
+        create(:draft_edition, document_type: 'taxon', id: 10000)
```
cbaines: It might be good here to create something which isn't a taxon, just to make sure that the filtering is actually happening.

koetsier: There are 150 other documents which have already been created in the before(:context) block

PR #1284, opened 2018-07-11
Decrease Sidekiq concurrency from 10 to 6

cbaines:
Currently the load on the PostgreSQL primary machine is often in the
mid 20s. This is a bit too high for a machine with now 8 cores, and it
looks like the culprit is the Publishing API Sidekiq workers.

Drop the connections back down to 6, to hopefully reduce the load on
the machine.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,6 +1,6 @@
 ---
 :verbose: false
-:concurrency: 10
+:concurrency: 6
 :logfile: ./log/sidekiq.json.log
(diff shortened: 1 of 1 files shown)


cbaines:
> Seems reasonable but you might want to check with @sihugh and/or @barrucadu before merging, as I think there might have been problems with some of the Whitehall/Asset Manager stuff when this was previously lowered.

Sure, I'll leave this until tomorrow when hopefully they can have a look.

PR #1287, opened 2018-07-16
Change RabbitMQ config

cbaines:
A few minor refactorings, plus disable RabbitMQ automatic recovery.

The Publishing API uses Sidekiq for retrying jobs, and there has
potentially been issues with mixing the indefinite retrying from
Bunny, with Sidekiq workers [1]. As Sidekiq handles retrying, stop
RabbitMQ from doing so, and instead use the Sidekiq retrying methods.

1: 2018-07-05:
There was a incident involving Sidekiq using up too many threads,
potentially because the indefinite retrying for connection issues was
causing issues with Sidekiq workers.

### DIFF
--- a/config/rabbitmq.yml
+++ b/config/rabbitmq.yml
@@ -5,6 +5,7 @@ defaults: &defaults
   pass: publishing_api
   exchange: published_documents
   recover_from_connection_close: true
+  automatically_recover: false
 
 development:
   <<: *defaults
--- a/config/rabbitmq.yml
+++ b/config/rabbitmq.yml
@@ -1,5 +1,4 @@
 defaults: &defaults
-  host: localhost
   port: 5672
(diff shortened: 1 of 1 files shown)


config/rabbitmq.yml
```diff
@@ -15,14 +16,11 @@ test:
   exchange: published_documents_test
 
 production:
+  <<: *defaults
```
thomasleese: This object will have both as `host` key (set to `localhost`) and a `hosts` key set to the RabbitMQ hosts - will that cause any problems?

cbaines: Good question. I was hoping at the time that it would just use `hosts` if specified. I've checked though, and it at least checks if both have been given, and warns against this, so I've moved it out of the defaults now.

PR #1293, opened 2018-07-24
Remove change note 157789

alex-ju:
This PR remove a public change note (dated 20 July 2018) for [VAT Notice 708: buildings and construction](https://www.gov.uk/government/publications/vat-notice-708-buildings-and-construction)

This request comes from the following ticket: https://govuk.zendesk.com/agent/tickets/2927623

### DIFF
--- a/db/migrate/20180724140319_remove_change_note.rb
+++ b/db/migrate/20180724140319_remove_change_note.rb
@@ -1,0 +1,33 @@
+# Remove a public change note (dated 20 July 2018) for "VAT Notice 708: buildings and construction"
+# Ticket: https://govuk.zendesk.com/agent/tickets/2927623
+# Page: https://www.gov.uk/government/publications/vat-notice-708-buildings-and-construction
+#
+# Prior steps:
+# Queried `Document` for the `content_id: "5f623c6e-7631-11e4-a3cb-005056011aef"`
+
+class RemoveChangeNote < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    # Find document
+    document = Document.where(content_id: "5f623c6e-7631-11e4-a3cb-005056011aef").first
+
+    if document.present?
+      live_edition = document.live
+      details = live_edition.details
+      change_history = details[:change_history]
+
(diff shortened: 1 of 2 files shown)


db/migrate/20180724140319_remove_change_note.rb
```diff
@@ -0,0 +1,17 @@
...
+# Ticket: https://govuk.zendesk.com/agent/tickets/2927623
+# Page: https://www.gov.uk/government/publications/vat-notice-708-buildings-and-construction
+#
+# Prior steps:
+# Queried `Document` for the `content_id`, then grabbed the `id` of the `ChangeNote` from the result
+
+class RemoveChangeNote < ActiveRecord::Migration[5.1]
+  def up
+    # Get rid of change note
+    ChangeNote.delete(157789)
```
thomasleese: Once you've got the `ChangeNote`, you will need to do [something like this](https://github.com/alphagov/publishing-api/blob/8c43bf55d0c58b3da9650560c882f3819c79fcff/db/migrate/20180215112954_fix_whitehall_remove_change_note.rb#L18-L20) to represent the content item to the content store. Similarly, you'll need to include `disable_ddl_transaction!` at the top for that to work.

I also wonder, for such a simple change, it might be easier to just run the command directly on the machine rather than creating a migration.

cbaines: > I also wonder, for such a simple change, it might be easier to just run the command directly on the machine rather than creating a migration.

I think so, but I also like the idea of keeping some kind of record of this change, so I like having it as a migration.

alex-ju: @thomasleese added `disable_ddl_transaction!` and the `RepresentDownstream` for production.

The PR fails and asks for a `rails db:migrate` which I can't successfully run as I don't have the DB setup yet 😞.

thomasleese: If you wanted to skip that, since you don't make any schema changes, you could [manually update the date in the schema file](https://github.com/alphagov/publishing-api/blob/master/db/schema.rb#L13) to the date of your migration.

PR #1298, opened 2018-07-30
Add back in empty migration methods

vanitabarrett:
We have replaced this migration, but need the empty methods there otherwise we get errors


kevindew:
You could delete the whole file rather than adding back the methods in.

PR #1324, opened 2018-09-04
Rake task to use high priority queue for represent_downstream

kevindew:
This task allows us to resend items to the high priority queue.

I went for a syntax of `represent_downstream:high_priority:content_id`
as we couldn't use a second option in the content_id method and we don't
have a precedent in this app for using env vars for rake tasks. If
this needs extending to other represent_downstream tasks then it
probably makes sense to have some sort of global option for them, but I
don't really anticipate it as we really shouldn't be putting anything
more than the odd thing through the high priority queue.

This feature has been added in response to it being needed during a
period of publishing sensitive documents and it being done via the
console. This will instead allow it to be run through Jenkins.

### DIFF
--- a/doc/admin-tasks.md
+++ b/doc/admin-tasks.md
@@ -47,10 +47,21 @@ bundle exec represent_downstream:tagged_to_taxon
 bundle exec represent_downstream:content_id['some-content-id']
 ```
 
+* Represent an individual edition downstream via the high priority queue
+```
+bundle exec represent_downstream:high_priority:content_id['some-content-id']
+```
+
 * Represent multiple editions downstream
 ```
 bundle exec represent_downstream:content_id['some-content-id some-other-content-id']
 ```
+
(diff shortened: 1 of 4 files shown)


doc/admin-tasks.md
```diff
@@ -47,10 +47,21 @@ bundle exec represent_downstream:tagged_to_taxon
 bundle exec represent_downstream:content_id['some-content-id']
 ```
 
+* Represent an individual edition downstream via the high priority queue
+```
+bundle exec represent_downstream:high_priority:content_id['some-content-id']
```
steventux: Are these tasks called via `rake`? If so they should read:

```
bundle exec rake represent_downstream:high_priority:content_id['some-content-id']
```

kevindew:
@steventux Thanks - oops that was missing on nearly all the tasks

PR #1327, opened 2018-09-06
Use a LATERAL JOIN on aggregates

thomasleese:
This should mean that documents with large numbers of editions get a significant performance increase as the aggregate subquery will only run once the most relevant edition has been found rather than for each of them.

For those not familiar with the [`LATERAL JOIN`](https://heapanalytics.com/blog/engineering/postgresqls-powerful-new-join-type-lateral) it allows a query to join on multiple subqueries where subsequent subqueries can use rows from the previous ones.

Comparing [an example query which currently takes around 50s to run](https://gist.github.com/thomasleese/effcf02e9c740e75fa36d9187622e044) with [a query that utilises the `LATERAL JOIN`](https://gist.github.com/thomasleese/71300c106cbc97c7691ec7906c30d3ce) we see the new one takes just `159.812 ms` to run with the following query plan:

```
Sort  (cost=2810.88..2810.90 rows=6 width=686) (actual time=158.837..158.838 rows=1 loops=1)
  Sort Key: editions.id
  Sort Method: quicksort  Memory: 176kB
  ->  WindowAgg  (cost=2374.00..2810.80 rows=6 width=686) (actual time=158.764..158.781 rows=1 loops=1)
        ->  Nested Loop Left Join  (cost=2374.00..2810.73 rows=6 width=686) (actual time=149.269..158.443 rows=1 loops=1)
              ->  Unique  (cost=2373.57..2373.60 rows=6 width=860) (actual time=134.887..144.061 rows=1 loops=1)
                    ->  Sort  (cost=2373.57..2373.58 rows=6 width=860) (actual time=134.886..143.753 rows=4701 loops=1)
                          Sort Key: editions.document_id, (CASE editions.state WHEN 'draft'::text THEN 0 WHEN 'published'::text THEN 1 WHEN 'unpublished'::text THEN 1 WHEN 'superseded'::text THEN 2 ELSE NULL::integer END), editions.user_facing_version
                          Sort Method: external merge  Disk: 2552kB
                          ->  Nested Loop Left Join  (cost=5.83..2373.49 rows=6 width=860) (actual time=1.878..82.571 rows=4701 loops=1)
                                ->  Nested Loop Left Join  (cost=5.41..83.83 rows=6 width=518) (actual time=1.666..35.585 rows=4701 loops=1)
                                      ->  Nested Loop  (cost=4.99..81.13 rows=6 width=485) (actual time=1.638..20.794 rows=4701 loops=1)
                                            ->  Index Scan using index_documents_on_content_id_and_locale on documents  (cost=0.42..8.45 rows=1 width=27) (actual time=0.062..0.064 rows=1 loops=1)
                                                  Index Cond: ((content_id = '08d48cdd-6b50-43ff-a53b-beab47f4aab0'::uuid) AND ((locale)::text = 'en'::text))
                                            ->  Bitmap Heap Scan on editions  (cost=4.56..72.51 rows=17 width=462) (actual time=1.568..19.362 rows=4701 loops=1)
                                                  Recheck Cond: (document_id = documents.id)
                                                  Filter: ((state)::text = ANY ('{draft,published,unpublished,superseded}'::text[]))
                                                  ->  Bitmap Index Scan on index_editions_on_document_id_and_user_facing_version  (cost=0.00..4.56 rows=17 width=0) (actual time=1.024..1.024 rows=4701 loops=1)
                                                        Index Cond: (document_id = documents.id)
                                      ->  Index Scan using index_change_notes_on_edition_id on change_notes  (cost=0.42..0.44 rows=1 width=37) (actual time=0.003..0.003 rows=0 loops=4701)
                                            Index Cond: (edition_id = editions.id)
                                ->  Index Scan using index_unpublishings_on_edition_id on unpublishings  (cost=0.42..0.45 rows=1 width=342) (actual time=0.002..0.002 rows=0 loops=4701)
                                      Index Cond: (edition_id = editions.id)
                                SubPlan 1
                                  ->  Values Scan on "*VALUES*"  (cost=0.00..0.02 rows=1 width=136) (actual time=0.001..0.001 rows=1 loops=4701)
                                SubPlan 2
                                  ->  Aggregate  (cost=381.11..381.12 rows=1 width=32) (actual time=0.003..0.003 rows=1 loops=4701)
                                        ->  Index Scan using index_links_on_edition_id on links  (cost=0.43..380.84 rows=108 width=32) (actual time=0.002..0.002 rows=0 loops=4701)
                                              Index Cond: (edition_id = editions.id)
              ->  GroupAggregate  (cost=0.43..72.82 rows=1 width=14) (actual time=14.378..14.378 rows=1 loops=1)
                    ->  Index Scan using index_editions_on_document_id on editions e  (cost=0.43..72.73 rows=17 width=14) (actual time=0.036..6.156 rows=4701 loops=1)
                          Index Cond: (document_id = editions.document_id)
Total runtime: 159.812 ms
```

### DIFF
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -60,6 +60,7 @@ module Presenters
 
       def query
         ordering_query = Edition.select("*, COUNT(*) OVER () as total").from(fetch_items_query)
+        ordering_query = join_lateral_aggregates(ordering_query)
         ordering_query = ordering_query.order(order.map { |o| o.join(' ') }.join(', ')) if order
         ordering_query = ordering_query.limit(limit) if limit
         ordering_query = ordering_query.offset(offset) if offset
@@ -114,12 +115,8 @@ module Presenters
             "to_char(public_updated_at, '#{ISO8601_SQL}') as public_updated_at"
           when :first_published_at
             "to_char(first_published_at, '#{ISO8601_SQL}') as first_published_at"
-          when :state_history
-            "#{STATE_HISTORY_SQL} AS state_history"
           when :unpublishing
             "#{UNPUBLISHING_SQL} AS unpublishing"
-          when :links
-            "#{LINKS_SQL} AS links"
           when :change_note
             "change_notes.note AS change_note"
           when :base_path
@@ -128,6 +125,10 @@ module Presenters
             "documents.locale as locale"
           when :content_id
             "documents.content_id as content_id"
+          when :state_history
+            nil
+          when :links
+            nil
           when :total
             nil
           else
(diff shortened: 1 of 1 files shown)


kevindew:
Wow - this sounds amazing. I'm so impressed with the amount of juice you've got out of that query, I had long thought we'd exhausted that. Excellent work.

I had no idea about lateral joins - will have to brush up. 

It's a bit of a shame that the thing that needed this optimisation is state history since that's such a weird thing for us to return anyway - since it's mostly predictable. Does dropping that altogether have any further impact?

thomasleese:
Yeah, my first instinct was to just remove `state_history` from the output which would have had the same effect on the speed of the query (I moved `links` as well because it's an aggregate function but it wasn't really causing any problems). However, when I had a quick search through GitHub I saw numerous places where the `state_history` is being used, specifically in specialist-publisher.

kevindew:
Has it spread any wider than Specialist Publisher? I remember what must now be a couple of years ago lamenting that Specialist Publisher needed it. I wonder if we can find a way of removing it there so we can eventually kill it - as I guess editions are only going to increase still.

But anyway that seems like a later problem.

thomasleese:
I found examples of it being used in content-tagger as well :sad:

PR #1338, opened 2018-09-24
Enable redirects to the homepage.

koetsier:
The validation that an internal redirect is not allowed to end
with a '/' makes it hard to redirect content to the Gov.uk
homepage.

This change puts an exception to this rule if redirecting to '/'

Related to trello: https://trello.com/c/VERGP1FP/215-remove-and-redirect-policy-area-finder and https://trello.com/c/I3WFXlpg/219-remove-and-redirect-policy-finder

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -173,7 +173,7 @@ private
 
     def validate_internal_redirect(destination)
       errors << "internal redirects cannot end with /" if
-        destination.end_with? "/"
+        destination != "/" && (destination.end_with? "/")
     end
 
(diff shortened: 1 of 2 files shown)


app/validators/routes_and_redirects_validator.rb
```diff
@@ -173,7 +173,7 @@ def invalid_destination?(destination)
 
     def validate_internal_redirect(destination)
       errors << "internal redirects cannot end with /" if
-        destination.end_with? "/"
+        destination!="/" && (destination.end_with? "/")
```
cbaines: Rubocop probably objects to the spacing around `!=`.

PR #1339, opened 2018-09-24
Update govuk_app_config to 1.9.1 and oj to 3.6.10

barrucadu:
Dependabot can't automatically update govuk_app_config because the
version constraint on oj is too restrictive.

---

[Trello card](https://trello.com/c/hMj68z21/509-update-oj-gem-in-publishing-api)

### DIFF
--- a/spec/controllers/v2/content_items_controller_spec.rb
+++ b/spec/controllers/v2/content_items_controller_spec.rb
@@ -447,18 +447,6 @@ RSpec.describe V2::ContentItemsController do
         expect(parsed_response_body["content_id"]).to eq(content_id)
       end
     end
-
-    context "with invalid request body" do
-      before do
-        request.env["CONTENT_TYPE"] = "application/json"
-        request.env["RAW_POST_DATA"] = ""
-        put :put_content, params: { content_id: content_id }
-      end
-
-      it "responds with 400" do
-        expect(response.status).to eq(400)
-      end
-    end
   end
 
   describe "publish" do
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ else
 end
 
 gem "gds-sso", "~> 13.6"
-gem "govuk_app_config", "~> 1.8"
+gem "govuk_app_config", "~> 1.9"
 gem "govuk_document_types", "~> 0.8.0"
 gem "govuk_schemas", "~> 3.2"
 gem "govuk_sidekiq", "~> 3.0"
@@ -40,10 +40,7 @@ group :development do
   gem "web-console", "~> 3"
 end
 
-# Lock to 2.18.3 because later patch versions are not listed in the oj changelog
-# and cause test failures.
-gem "oj", "2.18.3"
-gem "oj_mimic_json", "~> 1.0.1"
+gem "oj", "~> 3.0"
 
 group :development, :test do
   gem "database_cleaner"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -588,6 +588,8 @@ GEM
       aws-sigv4 (~> 1.0)
     aws-sigv2 (1.0.1)
     aws-sigv4 (1.0.2)
+    aws-xray-sdk (0.10.2)
+      oj (~> 3.0)
     bindex (0.5.0)
     builder (3.2.3)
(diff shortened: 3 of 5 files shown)


spec/controllers/v2/content_items_controller_spec.rb
```diff
@@ -449,14 +449,10 @@
...
-      before do
+      it "throws a JSON decoding error" do
         request.env["CONTENT_TYPE"] = "application/json"
         request.env["RAW_POST_DATA"] = ""
-        put :put_content, params: { content_id: content_id }
-      end
-
-      it "responds with 400" do
-        expect(response.status).to eq(400)
+        expect { put :put_content, params: { content_id: content_id } }.to raise_error
```
thomasleese: I think we'd still want to test that we get a 400 response here rather than raising an exception.

Do you know if the exception which gets raised has changed from https://github.com/alphagov/publishing-api/blob/master/app/controllers/application_controller.rb#L5 ?

barrucadu: It's an `ActionDispatch::Http::Parameters::ParseError`, and it seems to get thrown before the controller is involved at all as this doesn't work:
```diff
diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
index 81664f4..eecb6b3 100644
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::API
 
   rescue_from ActionController::ParameterMissing, with: :parameter_missing_error
   rescue_from JSON::ParserError, with: :json_parse_error
+  rescue_from ActionDispatch::Http::Parameters::ParseError, with: :json_parse_error
   rescue_from CommandError, with: :respond_with_command_error
 
   before_action :authenticate_user!
```

thomasleese: Hmm, that sounds like Rails is doing the JSON parsing not `oj`.

I think maybe we have to use `Oj.load` instead of `JSON.load` here: https://github.com/alphagov/publishing-api/blob/master/app/controllers/application_controller.rb#L48

Or maybe call `Oj.optimize_rails` somewhere? https://github.com/ohler55/oj/blob/master/pages/Rails.md

thomasleese: Or perhaps `Oj.mimic_JSON` so we can stick to using `JSON.load`? https://github.com/ohler55/oj/blob/master/pages/JsonGem.md

barrucadu: I think oj is being involved somehow, because otherwise upgrading it wouldn't make this error appear.

barrucadu: I've managed to get Rails using oj,  but I can't get `JSON.parse` using it:

```
> tracer = TracePoint.new(:c_call) do |tp| p [tp.lineno, tp.event, tp.defined_class, tp.method_id] end

> tracer.enable { Time.now.to_json }
[103, :c_call, Array, :last]
[8, :c_call, NilClass, :nil?]
[14, :c_call, #<Class:Time>, :now]
[14, :c_call, Time, :initialize]
[14, :c_call, Integer, :+]
[34, :c_call, Kernel, :is_a?]
[20, :c_call, #<Class:Oj::Rails::Encoder>, :new]
[20, :c_call, Oj::Rails::Encoder, :encode]
[20, :c_call, Time, :utc_offset]
[20, :c_call, Time, :utc?]
=> "\"2018-09-25T09:25:45.790+00:00\""

> tracer.enable { nil.to_json }
[34, :c_call, Kernel, :is_a?]
[20, :c_call, #<Class:Oj::Rails::Encoder>, :new]
[20, :c_call, Oj::Rails::Encoder, :encode]
=> "null"

> tracer.enable { JSON.parse(nil) }
[10, :c_call, #<Class:JSON>, :parse]
[10, :c_call, Class, :new]
[10, :c_call, Exception, :initialize]
[10, :c_call, Exception, :exception]
[10, :c_call, Exception, :backtrace]
TypeError: Nil is not a valid JSON source.
from (pry):10:in `parse'
```

PR #1340, opened 2018-09-25
Require a schema to exist for put content and use that for base_path presence

kevindew:
Loosely related to: https://trello.com/c/3GT17tkH/211-associate-content-with-ministers-l

Do not merge until this schema is deployed: https://github.com/alphagov/govuk-content-schemas/pull/815

This change causes a validation error if PutContent is called for when a schema does not exist. This should really have been applied a long time ago since I can't find any references in Sentry to schema not existing warnings and we really shouldn't be accepting any content that doesn't pass a schema (though we do still have placeholder_* ones)

By doing this we can then rely on schema to decide whether a base_path is required or not and save ourselves from having to repeat this logic in Publishing API.

The prompt for doing this card was discovering that items of the role schema were all failing to enter the Publishing API when they don't have a base_path (which means we only have ministers in there as they have base_paths).

As the tests and schema behaviour are quite coupled this required quite a substantial amount of changes within them - it also revealed a couple of bugs where we were testing situations which the schemas would not permit (hence https://github.com/alphagov/govuk-content-schemas/pull/815 to fix bulk_publishing option)


spec/models/edition_spec.rb
```diff
@@ -254,53 +254,6 @@
     end
   end
 
-  context "EMPTY_BASE_PATH_FORMATS" do
```
benthorner: Could this be in a subsequent commit?

spec/commands/v2/put_content_spec.rb
```diff
@@ -218,10 +212,6 @@
       end
 
       it "should send an alert to GovukError" do
-        # swallow error about missing schema
```
benthorner: Does this belong in this commit?

spec/validators/schema_validator_spec.rb
```diff
@@ -16,13 +16,13 @@
   describe "#validate" do
     context "unknown schema name" do
       let(:schema) { nil }
-      let(:payload) { { schema_name: "test" } }
+      let(:payload) { { schema_name: "nonexistent-schema" } }
 
-      it "logs to airbrake with an unknown schema_name" do
-        expect(GovukError)
```
benthorner: Does this belong in this commit?

spec/integration/put_content/new_edition_spec.rb
```diff
@@ -100,20 +100,19 @@
...
+      # This needs to be a schema that supports change_history hence news_story
+      payload.merge!(
+        document_type: "news_story",
+        schema_name: "news_article",
+        details: {
+          body: "News",
+          government: { title: "Test", slug: "test", current: true },
+          political: false,
+          change_history: [
+            { note: change_note, public_timestamp: Time.now.utc.rfc3339 },
```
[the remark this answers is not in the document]
kevindew: Yup - we're missing an integration test for that. We could do one but it'd be verbose as it needs to create a document that supports that form of schema.

It is covered by this unit test: https://github.com/alphagov/publishing-api/blob/master/spec/models/change_note_spec.rb#L47-L53

kevindew:
Thanks for your review @benthorner - I've done some more organising of tests into the previous commits when they're more related to them (which trims some of the fat on the big commit)

spec/integration/put_content/new_edition_spec.rb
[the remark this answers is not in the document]
benthorner: 👍

PR #1341, opened 2018-09-25
Do not parse payload

barrucadu:
As long as an appropriate content-type is sent, Rails will decode the
request body for us.  So by decoding it ourselves, we're doing this
twice!


kevindew:
You _might_ be able to use something like this: https://github.com/alphagov/email-alert-api/blob/master/config/routes.rb#L2 to tell Rails to default to JSON requests

barrucadu:
Might pick this up later, might not.

PR #1346, opened 2018-09-28
Use Oj.load instead of JSON.parse

barrucadu:
It seems like oj, after the upgrasde, is no longer monkey-patching
JSON.parse.  Rather than try to get that working again, just be
explicit and call Oj.load directly.

### DIFF
--- a/app/controllers/debug_controller.rb
+++ b/app/controllers/debug_controller.rb
@@ -11,7 +11,7 @@ class DebugController < ApplicationController
     }
 
     @mismatched_responses = mismatched_responses.map { |json|
-      parsed = JSON.parse(json)
+      parsed = Oj.load(json)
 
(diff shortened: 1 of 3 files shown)


kevindew:
Change seems good, I'm just wondering do we know why we use OJ over ruby JSON? It seems to have caused us quite a few problems recently and isn't a library (as far as I'm aware) we use in other parts of GOV.UK

barrucadu:
The publishing-api memory usage has been much higher since yesterday's deploy, which we suspect is because oj is no longer completely replacing the JSON gem.  So we can either make it use oj again for parsing JSON, or bump the memory we give to publishing-api.

barrucadu:
I'll keep an eye on integration and see if this has the desired effect.

PR #1347, opened 2018-10-02
[WIP] Handle person and role related link expansion

cbaines:
This should be helpful with getting person pages out of Whitehall
frontend.

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -23,7 +23,10 @@ module ExpansionRules
     [:taxons, :parent_taxons.recurring],
     [:taxons, :parent_taxons.recurring, :root_taxon],
     [:ordered_related_items, :mainstream_browse_pages, :parent.recurring],
-    [:ordered_related_items_overrides, :taxons]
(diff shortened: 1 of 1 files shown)


lib/expansion_rules.rb
```diff
@@ -65,6 +68,7 @@ def recurring
     { document_type: :need,                       fields: DEFAULT_FIELDS_WITH_DETAILS },
     { document_type: :finder, link_type: :finder, fields: DEFAULT_FIELDS_WITH_DETAILS },
     { document_type: :step_by_step_nav,           fields: DEFAULT_FIELDS_WITH_DETAILS },
+    { document_type: :role,                       fields: DEFAULT_FIELDS_WITH_DETAILS },
```
thomasleese: Is there a way we can be more explicit in which fields we expand from the `details` hash here? Similar to how we define the fields for `:travel_advice`?

thomasleese:
Done in https://github.com/alphagov/publishing-api/pull/1628.

PR #1349, opened 2018-10-03
Don't expand all organisation details

thomasleese:
I've identified that these fields are the only ones being used, so we don't need to present the entire details hash.

We should limit the number of fields being included in link expansion to reduce the size of content items being presented.

If we were to make dependency resolution work on second level fields, this would also help to reduce the amount of activity that happens when organisations change as there would be a limited number of fields which when changed cause updates to the dependencies.


kevindew:
Nice one Tom 👍

PR #1352, opened 2018-10-04
 Compare link expansion version of editions during edition diffing 

thomasleese:
Edition diffing is used to work out whether an update to an item should trigger dependency resolution. At the moment it works by diffing the entire hash of the edition and and then comparing whether the fields that changed might be exposed during link expansion. However, it only works on the first level. This means that if any field deeply nested in the `details` hash has changed regardless of whether it would ever get exposed in link expansion (for example in the case of organisations where [we only expose `details.logo` and `details.brand`](https://github.com/alphagov/publishing-api/blob/master/lib/expansion_rules.rb#L63)), we would detect that as a change to the edition and perform dependency resolution.

This PR changes that logic by first presenting the edition into what it would look like as part of link expansion before doing the process of diffing the hashes.

This should mean that we avoid performing dependency resolution on editions where no relevant fields have changed. I think it will make the queue sizes lower and content will move through the system faster.

This builds on the work in #1351 and combined with https://github.com/alphagov/publishing-api/pull/1349 I'm hopeful that it will reduce the problem we have at the moment related to organisation links.

[Trello Card](https://trello.com/c/EPXyTyeP/520-reduce-number-of-fields-in-organisation-links)

### DIFF
--- a/lib/link_expansion/edition_diff.rb
+++ b/lib/link_expansion/edition_diff.rb
@@ -8,8 +8,7 @@ class LinkExpansion::EditionDiff
   end
 
   def field_diff
-    ExpansionRules.potential_expansion_fields(current_edition.document_type) &
-      diff.map(&:first)
+    link_expansion_fields & diff.map(&:first)
   end
 
   def should_update_dependencies?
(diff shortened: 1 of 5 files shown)


lib/link_expansion/edition_diff.rb
```diff
@@ -8,21 +8,37 @@ def initialize(current_edition, version: nil, previous_edition: nil)
   end
 
   def field_diff
```
barrucadu: Do we actually need a diff here, or do we just need to know that the old and new are unequal?

thomasleese: We do, because we then compare the diff (which is just a list of fields which have changed) with "potential link expansion" fields - which is basically the top level list of fields which might change.

Only if there are fields that are present in both of those does it proceed with dependency resolution.

thomasleese: Actually having thought about this, I think you're right. We can remove the `field_diff` method.

thomasleese: @barrucadu Thanks for your review! I've added a commit to now just check if the hashes are different.

spec/lib/link_expansion/edition_diff_spec.rb
```diff
@@ -63,6 +63,52 @@
...
+          :live_edition,
+          document: document,
+          user_facing_version: 2,
+          document_type: "travel_advice",
+          base_path: "/foo",
+          details: { country: "en", unrelated_field: "en" },
+        )
+      end
+
+      it "includes the subfield in the diff" do
```
barrucadu: `it "does not include the subfield in the diff"`

barrucadu: Ah, the problems of reviewing one commit at a time.

PR #1360, opened 2018-10-19
Remove change note

MuriloDalRi:


db/migrate/20181019160402_remove_change_note_improvement_fund.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# Prior steps:
+# Queried `Document` for the `content_id: "6022c077-7631-11e4-a3cb-005056011aef"`
+
+class RemoveChangeNoteImprovementFund < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    # Find document
+    document = Document.where(content_id: "6022c077-7631-11e4-a3cb-005056011aef").first
```
thomasleese: You could do `Document.find_by(content_id: "...")` here.

thomasleese: You could do `Document.find_by(content_id: "...")` here.

thomasleese: You could do `Document.find_by(content_id: "...")` here.

db/migrate/20181019160402_remove_change_note_improvement_fund.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    if document.present?
+      live_edition = document.live
+      details = live_edition.details
+      change_history = details[:change_history]
+
+      details[:change_history] = change_history.drop(1)
```
thomasleese: In the Publishing API change notes are stored in a `ChangeNote` model and then presented to the Content Store in the `details` hash later on. Instead of modifying `details[:change_history]` directly, you'll want to find the appropriate `ChangeNote` object here and delete that and then call `RepresentDownstream`.

PR #1378, opened 2018-11-08
Don't include public_updated_at in organisation links

thomasleese:
When organisations are published in Whitehall, they are always updated as a major update. This means that the `public_updated_at` field gets updated every time meaning dependency resolution and link expansion is always happening.

I don't think this field is being used anywhere so I think we can probably drop it.


thomasleese:
Represented everything on staging, and I can't see any related errors.

thomasleese:
Will deploy next week when I'm back from annual leave.

sihugh:
_or so he thought_

PR #1380, opened 2018-11-09
Don't expand all finder details

deborahchua:
It appears that only one field is being used from the `details` hash for `finder` links.

We are aiming to reduce the number of fields included in link expansion so that the payload size is smaller, and we are only including the ones we actually need.

Trello card: https://trello.com/c/HHBxFrJr/612-do-not-expand-unnecessary-links-for-finder


deborahchua:
Currently testing the branch on `staging`

PR #1392, opened 2018-11-19
Add locale to vanish messages

theseanything:
Extract the locale from the edition and add to vanish message for the message queue. Locales are needed by the content performance manager to properly track editions.


theseanything:
[(3) Investigate: Page data pages (detailed metrics pages) don't exist for some live content items](https://trello.com/c/E1ZAulUd/831-3-investigate-page-data-pages-detailed-metrics-pages-dont-exist-for-some-live-content-items)

PR #1419, opened 2019-01-04
Update rails and arel

cbaines:


cbaines:
The pact tests with the GDS API Adapters gem are failing. This should resolve the failure: https://github.com/alphagov/gds-api-adapters/pull/876

PR #1426, opened 2019-01-22
Add a Rake task for removing change notes

thomasleese:
While on 2nd line, it's often the case that you may be asked to remove a change note from a document. This is poorly documented and can take a long time to figure out what to do. It also requires writing migrations which aren't easy to test and often don't work the first time.

Instead, we can provide a Rake task which enables developers to complete this task easily.

It provides two interfaces and uses fuzzy searching to match a users query with a specific change note:

```
data_hygiene:remove_change_note:dry[content_id,locale,query]
data_hygiene:remove_change_note:real[content_id,locale,query]
```

It doesn't change the update type of an edition from minor to major since we want to preserve history as best we can, plus email subscribers will have already received an email about content with the change note.

[Trello Card](https://trello.com/c/eE4pp5Zm/721-rake-task-for-removing-changenotes-in-whitehall)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem "govuk_sidekiq", "~> 3.0"
 gem "aws-sdk", "~> 3"
 gem 'bunny', '~> 2.13'
 gem "diffy", "~> 3.3", require: false
+gem "fuzzy_match", "~> 2.1"
 gem "govspeak", "~> 5.9.0"
 gem "hashdiff", "~> 0.3.8"
 gem "json-schema", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -629,6 +629,7 @@ GEM
     ffi (1.9.25)
     filelock (1.1.1)
     find_a_port (1.0.1)
+    fuzzy_match (2.1.0)
     gds-api-adapters (57.0.0)
       addressable
       link_header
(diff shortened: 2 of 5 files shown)


cbaines:
Which publishing apps is this intended to work for, as change note handling is a bit varied?

thomasleese:
@cbaines It's designed to be generic enough that it works with all publishing apps, but specifically this was built with Whitehall in mind, turning https://docs.publishing.service.gov.uk/manual/howto-remove-change-note-from-whitehall.html into a Rake task.

If there are little extra bits required for it to work with other publishing apps, we could update that later on as we discover them.

cbaines:
> @cbaines It's designed to be generic enough that it works with all publishing apps, but specifically this was built with Whitehall in mind, turning https://docs.publishing.service.gov.uk/manual/howto-remove-change-note-from-whitehall.html into a Rake task.

I can't remember exactly, but I thought Whitehall still sent the entire `change_history` through to the Publishing API for each edition. So changing it in the Publishing API will only apply until Whitehall sends another edition?

kevindew:
That's correct @cbaines - to fix a Whitehall one it's easier to change Whitehall and re-present. I worked with @deborahchua on a 2nd line issue of this recently.

thomasleese:
Yeah, looks like it does, perhaps those docs aren't correct... the most recent data migration I can find for removing a change note in Whitehall is https://github.com/alphagov/whitehall/blob/master/db/data_migration/20181023123751_remove_change_note_improvement_fund.rb though.

kevindew:
@thomasleese yeah - the most recent one was dev-ops'd I'm afraid.

thomasleese:
Okay, so it looks like we need a Rake task in Whitehall as well. I think this one still has a place though since it won't always be Whitehall change notes that need to be removed, we've had requests for Specialist Publisher before which will need to be done via the Publishing API.

PR #1432, opened 2019-01-29
Document status checker

thomasleese:
This is a Rake task that will check the status of a document and tell you if it has made it to the Publishing API, Content Store and Router.

This is intentionally different from the existing "Content Consistency Checker" because it does a much simpler check, but also because this Rake task is a bit of an experiment. The idea is that the we would encourage 2nd line to run this task when dealing with publishing problems to get some data about where documents get stuck. Using the results from the "Content Consistency Checker" can be incredibly unwieldy as it's complicated to run and there are so many inconsistencies that might not be relevant to "the page isn't visible on the website and I just pressed publish" type of query that often arrives on 2nd line.

Depends on https://github.com/alphagov/govuk-puppet/pull/8597 https://github.com/alphagov/gds-api-adapters/pull/896 https://github.com/alphagov/gds-api-adapters/pull/897

[Trello Card](https://trello.com/c/dskGorBB/737-rake-task-to-check-where-content-has-ended-up)

### DIFF
--- a/lib/data_hygiene/document_status_checker.rb
+++ b/lib/data_hygiene/document_status_checker.rb
@@ -1,0 +1,40 @@
+class DataHygiene::DocumentStatusChecker
+  def initialize(document)
+    @document = document
+  end
+
+  def content_store?
+    routes.each do |route|
+      content_item = GdsApi.content_store.content_item(route[:path])
+      updated_at = DateTime.parse(content_item["updated_at"])
+      return false unless updated_at >= edition.published_at # content-store must be later due to latency
+    end
+
+    true
+  rescue GdsApi::ContentStore::ItemNotFound
+    false
+  end
+
+  def router?
+    routes.each do |route|
+      route = GdsApi.router.get_route(route[:path])
+      return false unless route["backend_id"] == edition.rendering_app
+    end
+
+    true
+  rescue GdsApi::HTTPNotFound
+    false
+  end
+
+private
+
+  attr_reader :document
+
+  def edition
+    @edition ||= document.live
+  end
+
+  def routes
+    @routes ||= edition.routes
+  end
+end
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -23,4 +23,23 @@ namespace :data_hygiene do
       call_change_note_remover(args[:content_id], args[:locale], args[:query], dry_run: false)
     end
   end
+
+  desc "Check the status of a document whether it's in Content Store or Router."
+  task :document_status_check, %i[content_id locale] => :environment do |_, args|
+    document = Document.find_by!(args)
+    status = DataHygiene::DocumentStatusChecker.new(document)
+
+    content_store = status.content_store?
+    router = status.router?
+
+    puts "Has the document made it to:"
+    puts "Content Store? #{content_store ? 'Yes' : 'No'}"
+    puts "Router? #{router ? 'Yes' : 'No'}"
+
+    unless content_store && router
+      puts ""
+      puts "You could try running:"
+      puts "rake 'represent_downstream:content_id[#{args[:content_id]}]'"
+    end
+  end
 end
--- a/spec/lib/data_hygiene/document_status_checker_spec.rb
+++ b/spec/lib/data_hygiene/document_status_checker_spec.rb
@@ -1,0 +1,78 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/content_store'
+require 'gds_api/test_helpers/router'
+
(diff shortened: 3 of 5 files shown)


lib/data_hygiene/document_status_checker.rb
```diff
@@ -0,0 +1,40 @@
+class DataHygiene::DocumentStatusChecker
+  def initialize(document)
+    @document = document
+  end
+
+  def content_store?
+    routes.each do |route|
+      content_item = GdsApi.content_store.content_item(route[:path])
+      updated_at = DateTime.parse(content_item["updated_at"])
+      return false unless updated_at > edition.published_at # content-store must be later due to latency
```
elliotcm: What if it makes it there in under a second?  Should this be `>=` instead?

lib/data_hygiene/document_status_checker.rb
```diff
@@ -0,0 +1,40 @@
...
+      return false
+    end
+
+    true
+  end
+
+  def router?
+    routes.each do |route|
+      route = GdsApi.router.get_route(base_path)
+      return false unless route["backend_id"] == edition.rendering_app
```
elliotcm: This might just be some divergent terminology, but is it correct that `backend_id` would contain the `rendering_app` and not the publishing app?

lib/data_hygiene/document_status_checker.rb
```diff
@@ -0,0 +1,40 @@
...
+    rescue GdsApi::ContentStore::ItemNotFound
```
elliotcm: I'm not sure if it makes any real difference but because you're `return`ing out of this loop you could pull the exception handling up to the level of the `router?` method rather than having it defined within the loop.

lib/tasks/data_hygiene.rake
```diff
@@ -23,4 +23,13 @@ namespace :data_hygiene do
       call_change_note_remover(args[:content_id], args[:locale], args[:query], dry_run: false)
     end
   end
+
+  desc "Check the status of a document whether it's in Content Store or Router."
+  task :document_status_check, %i[content_id locale] => :environment do |_, args|
+    document = Document.find_by!(content_id: args[:content_id], locale: args[:locale])
```
elliotcm: Could just pass `args` in here?

lib/tasks/data_hygiene.rake
```diff
@@ -23,4 +23,13 @@ namespace :data_hygiene do
...
+    status = DataHygiene::DocumentStatusChecker.new(document)
+
+    puts "Content Store? #{status.content_store?}"
+    puts "Router? #{status.router?}"
```
elliotcm: Is it possible to have more descriptive output here?  I'm not sure to someone approaching it for the first time (e.g. me) that e.g. "Content Store? true" means that it's in the content store, rather than actually *being* the content store, or published by the content store, or whatever.

spec/factories/edition.rb
```diff
@@ -41,6 +41,10 @@
         )
       end
 
+      if item.state == "published"
+        item.update!(published_at: item.created_at)
+      end
```
elliotcm: Whitehall uses state machine code for this, so it could be rewritten as
```
item.update!(published_at: item.created_at) if item.published?
```

spec/lib/data_hygiene/document_status_checker_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/content_store'
+
+RSpec.describe DataHygiene::DocumentStatusChecker do
+  include GdsApi::TestHelpers::ContentStore
+  # include GdsApi::TestHelpers::Router
```
elliotcm: Can be removed?

spec/factories/edition.rb
thomasleese: Yeah, I'm not 100% happy with this being here anyway, because editions in a `superseded` or `withdrawn` state should also have a `published_at` date set. I might move it into the `live_edition` factory.

lib/data_hygiene/document_status_checker.rb
thomasleese: Yeah, basically `backend_id` is the name of the app that router will send requests to, so it has to be set to the frontend app that will handle the request.

lib/data_hygiene/document_status_checker.rb
thomasleese: I figured `>` would check milliseconds, but on second thought I think it's better to use `>=`.

thomasleese:
Thanks @elliotcm, I think I've addressed everything you raised now!

PR #1433, opened 2019-01-30
Add Rake task to update related links

karlbaker02:
This PR adds a new Rake task which reads in a JSON file and uses it to update each content item it finds to a set of associated related links.

For example:

```
{
  "a118443c-0a43-48c0-a4b6-80e3da714d9a": [
    "94a2c30c-a477-49dd-a7f4-0fa29e72f32b"
  ],
  "5c71dc4b-7631-11e4-a3cb-005056011aef": [
        "94a2c30c-a477-49dd-a7f4-0fa29e72f32b", "1602726f-05d2-44fd-b465-ace19524da56"
  ],
  "1602726f-05d2-44fd-b465-ace19524da56": []
}

```

Trello card: https://trello.com/c/oxyJRTGK

Here, the three content items specified by their content id as the key are each mapped to the set of content which is deemed to be related.

Solo: @karlbaker02

### DIFF
--- a/lib/tasks/update_related_links_from_json.rake
+++ b/lib/tasks/update_related_links_from_json.rake
@@ -1,0 +1,48 @@
+namespace :content do
+  desc "Updates suggested related links for content from a JSON file"
+  task :update_related_links_from_json, [:json_path] => :environment do |_, args|
+    puts 'Reading and parsing data...'
+
+    url = URI.parse(args[:json_path]) rescue false
+    file =
+      if url.is_a?(URI::HTTP) || url.is_a?(URI::HTTPS)
+        url.open(&:read)
+      else
+        File.read(args[:json_path])
+      end
+
+    json = JSON.parse(file)
+
+    @failed_content_ids = []
+
+    start_time = Time.now
+    puts "Start updating content items, at #{start_time}"
+
+    json.each_pair do |source_content_id, related_content_ids|
+      update_content(source_content_id, related_content_ids)
+    end
+
+    end_time = Time.now
+    elapsed_time = end_time - start_time
+
+    puts "Total elapsed time: #{elapsed_time}s"
+    puts "Failed content ids: #{@failed_content_ids}"
+  end
+
+  def update_content(source_content_id, related_content_ids)
+    response = Commands::V2::PatchLinkSet.call(
+      content_id: source_content_id,
+      links: {
+        suggested_ordered_related_items: related_content_ids
+      },
+      bulk_publishing: true
+    )
+
+    if response.code == 200
+      puts "Updated related links for content #{source_content_id}"
+    else
+      @failed_content_ids << source_content_id
+      STDERR.puts "Failed to update content id #{source_content_id} - response status #{response.code}"
+    end
+  end
+end


lib/tasks/update_related_links_from_json.rake
```diff
@@ -0,0 +1,48 @@
+namespace :content do
```
koetsier: Might be a good idea to put all this code in a service class so it can be tested

karlbaker02: It's possible. Originally I wrote this script was in content-tagger and followed the existing pattern there when talking to our APIs (i.e. https://github.com/alphagov/content-tagger/blob/master/lib/tasks/load_legacy_mappings.rake).

lib/tasks/update_related_links_from_json.rake
```diff
@@ -0,0 +1,48 @@
...
+  desc "Updates suggested related links for content from a JSON file"
+  task :update_related_links_from_json, [:json_path] => :environment do |_, args|
+    puts 'Reading and parsing data...'
+
+    url = URI.parse(args[:json_path]) rescue false
```
bilbof: What is the intent of the `rescue false`?

lib/tasks/update_related_links_from_json.rake
bilbof: I agree with moving some of the logic out of the rake task - it'll make it easier down the line to turn this into an automated process.

lib/tasks/update_related_links_from_json.rake
```diff
@@ -0,0 +1,48 @@
...
+    response = Commands::V2::PatchLinkSet.call(
+      content_id: source_content_id,
+      links: {
+        suggested_ordered_related_items: related_content_ids
+      },
+      bulk_publishing: true
+    )
+
+    if response.code == 200
+      puts "Updated related links for content #{source_content_id}"
```
bilbof: 500k lines from this puts will result in quite a big page in Jenkins :D

lib/tasks/update_related_links_from_json.rake
```diff
@@ -0,0 +1,48 @@
...
+    file =
+      if url.is_a?(URI::HTTP) || url.is_a?(URI::HTTPS)
+        url.open(&:read)
+      else
+        File.read(args[:json_path])
+      end
+
+    json = JSON.parse(file)
+
+    @failed_content_ids = []
```
bilbof: If something fails, this could be filled with 500k content ids. Do you think this is an issue?

lib/tasks/update_related_links_from_json.rake
karlbaker02: If we can't parse the path as a URI then we just want to assign some default non-true value to `url`. It could as easily be assigned `nil`.

lib/tasks/update_related_links_from_json.rake
karlbaker02: This could be an issue, especially if it were automated. It will be run manually to begin with, and it would be useful to record all of the failed content ids in a central place so that only those content items can be retried later on.

lib/tasks/update_related_links_from_json.rake
karlbaker02: That's true. We can split up the ingestion process to work on smaller batch sizes, which might be advisable to begin with anyway to ensure that the generated links look good.

lib/tasks/update_related_links_from_json.rake
karlbaker02: If it's not essential for the present, then it will be good to see if the related links test performs well. Definitely agree though.

PR #1439, opened 2019-02-06
Create a republish endpoint

kevindew:
Trello: https://trello.com/c/yhcKNEWr/587-allow-users-undo-withdraw

This endpoint is to fulfill a need Content Publisher has in the ability
to transition an unpublished item back to being published. At the time
of writing this is specifically for the ability to revert a withdrawal.

This provides an endpoint called republish that can be used by any live
edition and will update the content store and message queue with the
updated document using the existing update_type of "republish".

Editions that are published won't have their state changed, just the
publishing_request_id updated. Whereas unpublished editions will have
their state set back to published.

Initially this endpoint was going to be part of the publish endpoint to
work in a similar way to the unpublish one where it can be called
multiple times to change states. However this proved to just make
something complex more complicated and raised questions about how to
indicate whether you're dealing with draft or live editions.

### DIFF
--- a/app/commands/v2/republish.rb
+++ b/app/commands/v2/republish.rb
@@ -1,0 +1,75 @@
+module Commands
+  module V2
+    class Republish < BaseCommand
+      delegate :content_id, :locale, to: :document
+
+      def call
+        validate
+        republish_edition
+        after_transaction_commit { send_downstream } if downstream
+
+        Success.new(content_id: content_id)
+      end
+
+    private
+
+      def document
+        @document ||= Document.find_or_create_locked(
+          content_id: payload[:content_id],
+          locale: payload.fetch(:locale, Edition::DEFAULT_LOCALE),
+        )
+      end
+
+      def edition
+        document.live
+      end
+
+      def validate
+        no_republishable_item_exists unless edition
+        previous_version_number = payload[:previous_version].to_i if payload[:previous_version]
+        check_version_and_raise_if_conflicting(document, previous_version_number)
+      end
+
+      def no_republishable_item_exists
+        message = "A live item with content_id #{content_id} and locale #{locale} does not exist"
+        raise_command_error(404, message, fields: {})
+      end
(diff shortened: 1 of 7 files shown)


app/commands/v2/republish.rb
```diff
@@ -0,0 +1,74 @@
...
+        validate
+        republish_edition
+        after_transaction_commit { send_downstream } if downstream
+
+        Success.new(content_id: content_id)
+      end
+
+    private
+
+      def document
```
thomasleese: Not related to this PR, but I feel like we could move this into `BaseCommand`.

doc/api.md
```diff
@@ -225,6 +226,36 @@ will be presented in the live content store. Uses
...
+
+- [`content_id`](model.md#content_id)
+  - Identifies the document which has an edition to republish.
+
+### JSON attributes
+
+- `locale` *(optional, default: "en")*
+  - Accepts: An available locale from the [Rails I18n gem][i18n-gem]
+  - Specifies the locale of the document.
+- `previous_version` *(optional, recommended)*
```
thomasleese: I realise it goes against the other endpoints, but since this is a new endpoint which will presumably only be used bye one app at the moment could we consider making this a required field?

It feels like it should always have been a required field to me.

app/commands/v2/republish.rb
kevindew: I don't think commands are specific to documents unfortunately and similarly it's a bit of "where did that come from inheritance issue" :-/

doc/api.md
kevindew: The problem is that no-one uses it as far as I'm aware, we're not using it on Content Publisher. I think it makes more sense if Publishing API is the source of truth but when you're syncing state to Publishing API then I guess we care more that we're up to date.

kevindew:
Thanks for looking over this @thomasleese 👍

PR #1468, opened 2019-03-15
Add a new endpoint to unreserve a path

benthorner:
https://trello.com/c/6s9Bhzyj/632-delete-publishing-api-path-reservations-when-we-discard-drafts

This adds a new command to unreserve (destroy) a PathReservation,
provided the reservation exists and matches the given publishing app.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -29,6 +29,7 @@ message queue for other apps (e.g. `email-alert-service`) to consume.
 - [`POST /v2/links/by-content-id`](#post-v2linksby-content-id)
 - [`POST /lookup-by-base-path`](#post-lookup-by-base-path)
 - [`PUT /paths/:base_path`](#put-pathsbase_path)
+- [`DELETE /paths/:base_path`](#delete-pathsbase_path)
 - [`GET /debug/:content_id`](#get-debugcontent_id)
 
 ### Optimistic locking (`previous_version`)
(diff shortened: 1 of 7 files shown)


spec/requests/unreserve_path_requests_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe "DELETE /paths", type: :request do
+  let(:request_body) { payload.to_json }
+
+  def do_request(body: request_body, headers: {})
+    delete request_path, params: body, headers: headers
```
kevindew: ooh TIL - did not realise request_path could be in scope here

doc/api.md
```diff
@@ -718,6 +719,29 @@ Reserves a path for a publishing application. Returns success or failure only.
...
+- `publishing_app` *(required)*
+  - The name of the application making this request, words separated with hyphens.
+
+### State changes
+
+- If no path reservation for the supplied base_path is present, the command will
+  fail.
+- If a path reservation exists for the supplied base_path but a different
+  publishing_app, the command will fail.
+
```
kevindew: Nice one doing this 👍

spec/commands/unreserve_path_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        payload = { publishing_app: "foo", base_path: "/bar" }
+        create(:path_reservation, payload)
+        described_class.call(payload)
+        expect(PathReservation.count).to be_zero
+      end
+    end
+
+    context "when the path is not owned by the app" do
+      it "returns an error" do
+        payload = { base_path: "/bar" }
```
kevindew: This test looks a bit like it's testing what happens when a required parameter is missing rather than owned by a different app.

benthorner: Fair enough - I was being minimal. Amended :-).

cbaines:
What's the motivation for this change? It would be good if you could add that to the commit message.

benthorner:
@cbaines sure, I've updated the commit - the Trello card on this PR should also give more context :-).

cbaines:
> @cbaines sure, I've updated the commit - the Trello card on this PR should also give more context :-).

Great, thanks. Scary, but still good to know.

PR #1470, opened 2019-03-19
Adds some documenation around publishing intents

1pretz1:
Up to now there is little to no documentation around publishing intents
in the publising-api. We used publishing intents in content-publisher
when working on the scheduling workflow and had to trawl through code to
find out how they worked and how to interact with the endpoint.

This adds some guidance around using publishing intents.

Trello:
https://trello.com/c/fXtvPpIO/696-create-publish-intent

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -30,6 +30,8 @@ message queue for other apps (e.g. `email-alert-service`) to consume.
 - [`POST /lookup-by-base-path`](#post-lookup-by-base-path)
 - [`PUT /paths/:base_path`](#put-pathsbase_path)
 - [`DELETE /paths/:base_path`](#delete-pathsbase_path)
+- [`PUT /publish-intent(/*base_path)`](#put-publish-intentbase_path)
+- [`DELETE /publish-intent(/*base_path)`](#delete-publish-intentbase_path)
 - [`GET /debug/:content_id`](#get-debugcontent_id)
 
 ### Optimistic locking (`previous_version`)
@@ -742,6 +744,58 @@ Unreserves a path for a publishing application. Returns success or failure only.
 - If a path reservation exists for the supplied base_path but a different
   publishing_app, the command will fail.
 
+## `PUT /publish-intent(/*base_path)`
+
+[Request/Response detail][put-intent-pact]
+
+Used to create or update a publishing intent for a particular base_path and
+reserve the base path in the publishing-api. A publishing intent sets when the
+cache should expire for a piece of content.
+
+### Path parameters
+
+- `base_path`
+  - Identifies the path to create the publishing intent for
+
+### JSON parameters:
+
+- `publishing_app` *(required)*
+  - The name of the application making this request.
+- `rendering_app` *(required)*
+  - The name of the application rendering the content the base path refers to.
+- `publish_time` *(required)*
+  - The `DateTime` the content is set to be published.
+- `routes` *(optional)*
+  - `Array` - Fields supported are path and type, the exact requirements can be
+    found [here][routes-content-schema]. This is optional as it is only
+    required if there needs to be a route created in router-api for the
+    `base_path` at the same time the publishing intent is made.
+
+### State changes
+
+- `base_path` is reserved for use of the given `publishing_app`.
+- Creates a `PublishingIntent` in content-store to record when the cache should
+  expire for a `base_path`.
+- If `routes` are given, adds a route set in router-api.
+
+## `DELETE /publish-intent(/*base_path)`
+
+Used to delete the publishing intent for a particular `base_path`. Removing a
+publishing intent means that the content will continue to be cached as
+normal.
+
+[Request/Response detail][delete-intent-pact]
+
+### Path parameters
+
+- `base_path`
+  - Identifies the path to delete the publishing intent for
+
+### State changes
+
+- The `PublishingIntent` for the passed in base_path is removed from the
+  content-store.
+
 ## `GET /debug/:content_id`
 
 Displays debug information for `content_id`.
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
 - If a path reservation exists for the supplied base_path but a different
   publishing_app, the command will fail.
 
+## `PUT /publish-intent(/*base_path)`
+
+[Request/Response detail][put-intent-pact]
+
+Used to create or update a publishing intent for a particular base_path and
+reserve the base path in the publishing-api. A publishing intent sets when the TTL
+cache should be cleared.
```
benthorner: The terminology isn't quite right here - the goal of the intent is to reduce caching of *content* at the specified time. It's actually a bug that the TTL itself is also cached by Fastly :-|.

doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+  - Identifies the path to create the publishing intent for
+
+### JSON parameters:
+
+- `publishing_app` *(required)*
+  - The name of the application making this request, words separated with
+    hyphens.
+- `rendering_app` *(required)*
+  - The name of the application rendering the content the base path refers to,
+    words separated with hyphens.
```
benthorner: This looks helpful, but I'm curious where the hyphen rule is coming from?

doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+- `publish_time` *(required)*
+  - The `DateTime` the content is set to be published.
+- `routes` *(optional)*
+  - `Array` - Fields supported are path and type. `path` can either be an exact
```
benthorner: This also looks helpful, but I notice the other docs don't go into this much detail. It might be more future-proof to just link to the [definition in the content schemas](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/routes_redirects.jsonnet), which I'm guessing is what it's validating against.

doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+    base_path or a subpath. If `path` is an exact base_path then `type` must be
+    `exact`, and if `path` is a subpath then `type` must have value `prefix`.
+    Required if you need a route to be added to the router-api for the given
+    base_path.
+
+### State changes
+
+- `base_path` is reserved for use of the given `publishing_app`.
+- Creates `PublishingIntent` in content-store to record when TTL cache should
```
benthorner: Similar terminology clash here :-).

doc/api.md
benthorner: Perhaps it would be better to indicate why this is optional.

doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+  expire.
+- If `routes` are given, adds a route set in router-api.
+
+## `DELETE /publish-intent(/*base_path)`
+
+Used to delete the publishing intent for a particular base_path. A publishing
+intent sets when the TTL cache should be cleared.
```
benthorner: Since we're deleting, rather than creating, should we indicate why deletion is supported instead of what creation does?

doc/api.md
```diff
@@ -742,6 +744,60 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+[Request/Response detail][delete-intent-pact]
+
+### Path parameters
+
+- `base_path`
+  - Identifies the path to delete the publishing intent for
+
+### State changes
+
+- The `PublishingIntent` for the passed in base_path is removed from the
```
benthorner: Does this have any affect on the router?

doc/api.md
1pretz1: I'm not sure exactly where it originates from, probably in the router so it can find which app to send the requests to when they come in.

`rendering_app` is one of the fields used to populate the PublishingIntent record in content-store which then populates the `backend_id` field if a new route is created (depending on presence of routes argument) in the router-api.

1pretz1: Sorry if this is really small, you may have to get your specs on! 👁 

This shows the pattern of using hyphen separated words for the frontend app names...

Screen Shot 2019-03-20 at 16 13 48

doc/api.md
1pretz1: Nope not that i'm aware of!

doc/api.md
1pretz1: Well technically it's not reducing the cache on content as it's just reducing the cache for a certain base_path as the content isn't there yet, but i see your point and i will make it clearer! :)

doc/api.md
1pretz1: Yep I can do both of those things, was unsure about how much detail to add here!

doc/api.md
1pretz1: Noted

doc/api.md
1pretz1: Yep sure

doc/api.md
1pretz1: As per our discussion the rendering_app is validated against a regex which only allows lowercase, numbers and hyphens. So the names don't necessarily need to be seperated by hyphens so i will remove the rule! :)

doc/api.md
```diff
@@ -742,6 +744,58 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+
+- `base_path` is reserved for use of the given `publishing_app`.
+- Creates a `PublishingIntent` in content-store to record when the cache should
+  expire for a `base_path`.
+- If `routes` are given, adds a route set in router-api.
+
+## `DELETE /publish-intent(/*base_path)`
+
+Used to delete the publishing intent for a particular `base_path`. Removing a
+publishing intent would mean that the cache for the `base_path` would no longer
```
benthorner: The phrasing is a little strange here - the use of 'would' makes it sound like it's conditional on something. Also, the use of 'the cache' is a little misleading, as there many caches (known and unknown); it would be correct to say 'means the content will continue to be cached as normal'.

doc/api.md
```diff
@@ -742,6 +744,58 @@ Unreserves a path for a publishing application. Returns success or failure only.
...
+- `publishing_app` *(required)*
+  - The name of the application making this request.
+- `rendering_app` *(required)*
+  - The name of the application rendering the content the base path refers to.
+- `publish_time` *(required)*
+  - The `DateTime` the content is set to be published.
+- `routes` *(optional)*
+  - `Array` - Fields supported are path and type, the exact requirements can be
+    found [here][routes-content-schema]. This is optional as it is only
+    required if there needs to be a route created in router-api for the
```
benthorner: Minor one, but would be helpful to indicate this is done immediately ;-).

doc/api.md
1pretz1: Yep i tried to convey that in this sentence 'would no longer have a specific time in which it should be cleared by.` but if that's not clear enough I'll amend it to your suggestion :)

doc/api.md
1pretz1: Yep sure, i've reworded the sentence to make it clearer that this is done when the publishing intent is made. I steered clear of using the word `immediately` as to me this denotes that that if it isn't done immediately, it will be done asynchronously.

PR #1472, opened 2019-03-20
Only run dependency resolution for PatchLinkSet when links have changed

karlbaker02:
This PR updates the `PatchLinkSet` to only trigger dependency resolution when links within a `LinkSet` change. Previously, dependency resolution would be triggered regardless of whether there were any changes.

This brings us closer in line to what already happens with the `Publish` and `PutContent` commands.

Trello: https://trello.com/c/0CnfwxR6

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -26,9 +26,10 @@ module Commands
         link_set.reload
 
         orphaned_content_ids = link_diff_between(before_links.map(&:target_content_id), link_set.links.map(&:target_content_id))
+        added_content_ids = link_diff_between(link_set.links.map(&:target_content_id), before_links.map(&:target_content_id))
 
         after_transaction_commit do
-          send_downstream(orphaned_content_ids)
+          send_downstream(orphaned_content_ids, added_content_ids)
         end
 
         Action.create_patch_link_set_action(link_set, before_links, event)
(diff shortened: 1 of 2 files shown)


karlbaker02:
Initially I was going to check for changes to the latest `Edition` for the given `content_id` and `locale`, but unless we update the `links` with those given to the `PatchLinkSet`, we won't be comparing anything. `links` also do not form part of an `Edition`.

thomasleese:
Adding our conversation from Slack so we have a record:

> [...concluded that the solution is this PR is unlikely to work because adding a new link will always trigger dependency resolution...]

> I see the problem with using `LinkExpansion::EditionDiff.new(edition).should_update_dependencies?` is that a new edition isn't being created so it can't compare with anything
> but I feel like that would be the ideal solution

> I think it might need some refactoring on how `LinkExpansion::EditionDiff.new(edition).should_update_dependencies?` works to allow it to work for link set links, the process would be something like:

> - generate an expanded links representation of the edition as it already is in the database
> - add/delete the new links
> - generate an expanded links representation of the same edition, but now with the new links
> - compare the expanded links representations to decide whether to perform dependency resolution 

> compared to the process now where it just compares an expanded links representation of the current edition and the previous edition as they are in the database
> I think that's the only way to know for sure whether we need to perform dependency resolution

PR #1482, opened 2019-04-02
Add notify endpoint

steventux:
https://trello.com/c/0wiO10tC/327-email-trigger-mechanism-for-new-content-tagger-process

[RFC relating to this change](https://github.com/alphagov/govuk-rfcs/blob/2cc32ae534d0f30f9702401a31ccd69f73bc4706/rfc-102-add-notification-workflow.md)

#### Use case for this endpoint

[We currently send emails when published content items are first tagged to the business readiness finder](https://github.com/alphagov/rummager/blob/master/lib/indexer/metadata_tagger.rb#L36).

When we start using the new tagging interface in Content Tagger the same email trigger won't exist.
We need a similar mechanism in the publishing pipeline as [tagging content will be handled by patching links in the Publishing API](https://github.com/alphagov/content-tagger/pull/884/files).

This PR adds a `notify` endpoint which can be used to send a message to the message queue (using a distinct message type). This will be used by email alert service to trigger a notification in the email alert api.

The optional key `workflow_message` is merged into the downstream payload, this can be used as a custom transient change note for the notification.

### DIFF
--- a/app/presenters/message_queue_presenter.rb
+++ b/app/presenters/message_queue_presenter.rb
@@ -1,25 +1,32 @@
 module Presenters
   class MessageQueuePresenter
-    attr_reader :edition, :draft, :workflow_message, :edition_presenter
+    attr_reader :edition, :draft, :notification_attributes, :edition_presenter
 
-    def initialize(edition, draft: false, workflow_message: nil)
+    def initialize(edition, draft: false, notification_attributes: {})
       @edition = edition
       @draft = draft
-      @workflow_message = workflow_message
+      @notification_attributes = notification_attributes
       @edition_presenter = EditionPresenter.new(edition, draft: draft)
     end
 
     def for_message_queue(payload_version)
       edition_presenter.for_message_queue(payload_version)
+        .merge(presented_publishing_app)
         .merge(presented_workflow_message)
     end
 
   private
 
     def presented_workflow_message
-      return {} unless workflow_message
+      return {} unless notification_attributes[:workflow_message].present?
 
-      { workflow_message: workflow_message }
+      { workflow_message: notification_attributes[:workflow_message] }
+    end
+
+    def presented_publishing_app
+      return {} unless notification_attributes[:publishing_app].present?
+
+      { publishing_app: notification_attributes[:publishing_app] }
     end
   end
 end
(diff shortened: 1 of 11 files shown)


steventux:
@peteglondon would you mind re-reviewing? I've had to modify this PR to allow the endpoint to send a custom message to the message queue.

app/presenters/message_queue_presenter.rb
```diff
@@ -0,0 +1,25 @@
+module Presenters
+  class MessageQueuePresenter
+    attr_reader :edition, :draft, :workflow_message, :edition_presenter
+
+    def initialize(edition, draft: false, workflow_message: nil)
+      @edition = edition
+      @draft = draft
+      @workflow_message = workflow_message
+      @edition_presenter = EditionPresenter.new(edition, draft: draft)
```
leenagupte: Does `edition_presenter` need to be an attribute? Can it just be a method?

lib/expansion_rules.rb
```diff
@@ -65,9 +65,9 @@ def details_fields(*fields)
   STEP_BY_STEP_FIELDS = (DEFAULT_FIELDS + [%i(details step_by_step_nav title), %i(details step_by_step_nav steps)]).freeze
   TRAVEL_ADVICE_FIELDS = (DEFAULT_FIELDS + details_fields(:country, :change_description)).freeze
   WORLD_LOCATION_FIELDS = [:content_id, :title, :schema_name, :locale, :analytics_identifier].freeze
-  FACET_GROUP_FIELDS = (%i[content_id title schema_name] + details_fields(:name, :description)).freeze
+  FACET_GROUP_FIELDS = (%i[content_id title locale schema_name] + details_fields(:name, :description)).freeze
```
leenagupte: Is this a bug fix?

steventux: Yeah, this is necessary to comply with the notification schema links definition.

leenagupte: Ah. Could you add that to the commit note? I don't think locale is required in the [links schema](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/facet_group/publisher_v2/links.json).

steventux: It's not required in the publisher_v2 links schema, it's required in the [frontend links schema](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/frontend_links.jsonnet#L9) I think the downstream payloads are validated against this.

steventux:
This endpoint isn't needed at the moment. We've decided to use a direct route to email-alert-api for the time being.

PR #1532, opened 2019-07-04
Try resolve the temporary dates debt

kevindew:
Trello: https://trello.com/c/colkZywn/43-complete-publishing-api-date-migration

This is an attempt to try wrap up a bunch of the work started in 2017 to migrate Publishing API to change the timestamps that are stored.

Currently Publishing API stores the timestamps for when a document is first published, the time it had a major publishing (public updated at) and when it was last edited. Unfortunately, as publishing applications can override these values, we often have no easily available data on when things occurred in the Publishing API. The hypothesis back in 2017 was that we would split these fields into two: one for publisher provided dates, the other for Publishing API ones. Work began to implement this, until work on Publishing API was paused.

This work left an element of debt where we had lots of timestamp fields on an Edition and it was unclear which ones of these were used. 2 years later it's unclear when or where any work on this could fit into any teams roadmap.

Given this, I've decided to try wrap up this work on timestamps with something of a compromise approach. I chose to give up on the idea of having two sets of date and instead continue with the existing behaviours for `first_published_at`, `public_updated_at` and `last_edited_at`. To store when editions were first published and last edited in Publishing API (regardless of user input) we have fields: `publishing_api_first_published_at` and `publishing_api_last_edited_at` and to capture some of the other missing times this introduces :`published_at` and `major_published_at`.

This is a draft PR, I don't expect it to be merged directly. It is more used to illustrate to tie up the date work and likely will go through in multiple PR's that cherry pick bits which will allow us to apply this with zero downtime.

A couple of things this also does is fix the way `public_updated_at` is set, these will use `first_published_at` rather than current time, as I can't think why they'd need this not to match.

It also changes the draft edition factory to not set public_updated_at and first_published_at since first drafts would not have this (which is what breaks the GDS API adapters pact test)


kevindew:
Now fully implemented by separate PRs

PR #1539, opened 2019-07-12
Add deferred priority Sidekiq queue

karlbaker02:
This PR adds a new `deferred` priority Sidekiq queue. Currently, the majority of publishing requests to publishing-api are placed on the `low` queue, which includes both departmental and automated, bulk publishing requests. The result is that departmental edits can get "stuck" behind less urgent updates - a problem exacerbated through dependency resolution of bulk publishing updates added back to the `low` queue.

Recently, we have introduced more load to this queue in the form of publishing related links (albeit on an infrequent basis). This adds yet more load to the `low queue`.

To start and abate some of these problems, we have decided to add a new, lower-priority `deferred` queue to the publishing-api. This queue will be primarily used by related links updates, but can and should be used for any publishing which can operate whenever there is free capacity within the publishing-api to do so (think of it like an AWS EC2 spot instance).

As related links is only concerned with patching link sets, this is the only place where the new queue is being used at present.

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -87,8 +87,12 @@ module Commands
         payload.fetch(:bulk_publishing, false)
       end
 
+      def deferred_publishing?
+        payload.fetch(:deferred_publishing, false)
+      end
+
       def downstream_draft(content_id, locale, orphaned_content_ids)
-        queue = bulk_publishing? ? DownstreamDraftWorker::LOW_QUEUE : DownstreamDraftWorker::HIGH_QUEUE
+        queue = get_publishing_queue
         DownstreamDraftWorker.perform_async_in_queue(
           queue,
           content_id: content_id,
@@ -98,7 +102,7 @@ module Commands
       end
 
       def downstream_live(content_id, locale, orphaned_content_ids)
-        queue = bulk_publishing? ? DownstreamLiveWorker::LOW_QUEUE : DownstreamLiveWorker::HIGH_QUEUE
+        queue = get_publishing_queue
         DownstreamLiveWorker.perform_async_in_queue(
           queue,
           content_id: content_id,
@@ -133,6 +137,14 @@ module Commands
         @schema_name ||= Queries::GetLatest.(
           Edition.with_document.where("documents.content_id": content_id)
         ).pluck(:schema_name).first
+      end
+
+      def get_publishing_queue
+        if deferred_publishing?
+          DownstreamLiveWorker::DEFERRED_QUEUE
+        else
+          bulk_publishing? ? DownstreamLiveWorker::LOW_QUEUE : DownstreamLiveWorker::HIGH_QUEUE
+        end
       end
     end
   end
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -6,6 +6,7 @@
   - downstream_high
   - dependency_resolution
   - downstream_low
+  - downstream_deferred
   - experiments
   - default
   - import
--- a/lib/downstream_queue.rb
+++ b/lib/downstream_queue.rb
@@ -1,4 +1,5 @@
 module DownstreamQueue
   HIGH_QUEUE = "downstream_high".freeze
   LOW_QUEUE = "downstream_low".freeze
+  DEFERRED_QUEUE = "downstream_deferred".freeze
 end
--- a/spec/commands/v2/patch_link_set_spec.rb
+++ b/spec/commands/v2/patch_link_set_spec.rb
@@ -231,6 +231,13 @@ RSpec.describe Commands::V2::PatchLinkSet do
       described_class.call(payload.merge(bulk_publishing: true))
     end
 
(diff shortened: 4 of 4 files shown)


cbaines:
> This PR adds a new `deferred` priority Sidekiq queue. Currently, the majority of publishing requests to publishing-api are placed on the `low` queue, which includes both departmental and automated, bulk publishing requests. The result is that departmental edits can get "stuck" behind less urgent updates - a problem exacerbated through dependency resolution of bulk publishing updates added back to the `low` queue.

What's a departmental edit? Anything that's directly related to an action in a publishing application should be on the high queue (unless that publishing application has specifically chosen otherwise).

karlbaker02:
When we ran A/B tests a few quarters ago, some edits that departments were making _were_ ending up on the low queue - whether this was dependency resolution updates or not I'm unsure, but this was the case.

cbaines:
> When we ran A/B tests a few quarters ago, some edits that departments were making _were_ ending up on the low queue - whether this was dependency resolution updates or not I'm unsure, but this was the case.

Do you think there was a relationship between the A/B tests on the related links, and content changes being processed in the low queue?

Also, do you know what publishing applications were making the affected content changes?

I know that there's a known issue with changing the base path of a page, as the prioritisation gets a bit muddled, but if there are additional issues which are delaying changes getting out, then it would be good to identify them.

karlbaker02:
I don't think there's any relationship between the A/B tests we ran and content changes being processed in the low queue; simply, a big backlog of updates in the low queue, after which someone decided to publish some updates, highlighted an existing problem. 

Afraid that I don't know which publishing app(s) made the affected content updates. @sihugh was the person I believe who originally saw this, not sure if he has any more insight?

kevindew:
My understanding is consistent with Chris'. The only known issue of things affected is when things go through with a "republish" update type and end up on the low queue. I guess there is always the problem that any imported content doesn't make it through until dependency resolution is through.

If we're allowing the ability to specify which queue to use by a payload parameter could we enable this for all queues rather than just this deferred one? It looks like we're building up different ways to set different queues. This one introduces a deferred flag, we have a [different flag on put content](https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/put_content.rb#L153). It might be better to just call a spade a spade and let users specify, which could be used to fix the redirect problem.

A consequence this change is that the initial document that has changed may be put into the deferred queue but any dependency resolution that results from this will be [put into the low queue](https://github.com/alphagov/publishing-api/blob/master/app/workers/dependency_resolution_worker.rb#L48-L68) meaning they'll probably still flood the low queue.

Also I'm a bit concerned about the wording as I think low and deferred effectively mean the same thing. It might be worth just going with high, medium, low or having a re-think of all of them.

karlbaker02:
We can also use the `bulk_publishing` on patch content, so that flag is somewhat generic.

Agreed that it could be the way forwards to let users specify which queue. This PR is an initial foray into splitting up the queues a bit and hopefully something we can build on. I'm aware that any updates triggered by dependency resolution from this new `deferred` queue may end up on the `low` queue, however it's still going to save a lot of updates from ever needing to go into the `low` queue in the first place.

In terms of naming, we could go for high, medium and low, and then rename everything that's currently on the low queue to go to the medium? I was talking to @rubenarakelyan about this and even though we only intend to use this queue initially for related links, it's definitely something which has to be generic enough for others to use it in the future. Given that it only operates when both the high and low queues have less updates to process then the total number of workers configured, I was trying to capture this in the name. We might also ask a content designer for their opinion.

karlbaker02:
To summarise the chat I had with @kevindew yesterday, the best approach might be that instead of creating a `deferred` queue with yet more logic to determine where we push updates to, we allow the caller to specify the queue as part of making the update; this would require updating everything that pushes Sidekiq to allow the new param to be passed.

leenagupte:
@karlbaker02 We're looking at a related problem for step by step pages. When step by steps are published, the step by step itself is added to the high priority queue in publishing-api, so that the changes to the overview page appear almost immediately. 

However publishers want to review the step by step in the sidebar, but updates to the side bar are triggered by dependency resolution and can take a long time to appear.

We'd like to be able to send all updates triggered by dependency resolution for a step by step to a higher priority queue.

sihugh:
@leenagupte anecdotally, I think link updates have been getting through pretty promptly recently.  Perhaps due to Breville's crew?

thomasleese:
Certainly since we deployed https://github.com/alphagov/publishing-api/pull/1581 the latency has been significantly lower than before https://grafana.publishing.service.gov.uk/dashboard/file/publishing_api_metrics.json?refresh=1m&orgId=1&var-queue=downstream_low&var-aggregation=max.

barrucadu:
@karlbaker02 Is this still happening, or can this PR be closed?

karlbaker02:
@kevindew @leenagupte I think we should have a catch-up regarding this next week and whether it's still necessary.

kevindew:
> @kevindew @leenagupte I think we should have a catch-up regarding this next week and whether it's still necessary.

@leenagupte is off for the next few weeks - I'm happy to have a catch up though

PR #1547, opened 2019-07-17
Remove edition diff code

kevindew:
A long time ago efforts were made by the then Content History team to store
the difference between changes of content to make way for allowing
published content to be patched. This was built on the expectation that
we would intend to slowly expose more parts of GOV.UK history (as
stored by the Publishing API) and thus had reasons to keep a cleaner
history of editions.

2 years later the Publishing API is no nearer to allowing the patching
of published content and the drive to expose a public history of the
Publishing API is no longer present. Thus the work that was
done to store this data is likely not presenting us with any value and
is instead incurring quite a large cost given the additional size this
adds to the database (extra 10GB and growing).

The simplest thing to do is to remove this, we can always dig out this
code again if it's something that's later useful.


kevindew:
> We can remove `Hashdiff` as a dependency now I believe.

That's still used by a [govspeak compare class](https://github.com/alphagov/publishing-api/blob/793fb5888694117e9a7ee0c23d174d40d3a3d07d/lib/data_hygiene/govspeak_compare.rb) which I expect is also a candidate for removing since that won't have been used since Specialist Publisher launch.

thomasleese:
Oh yes... somehow I'd forgotten that even though I just did this... https://github.com/alphagov/publishing-api/pull/1544/commits/5a405ef8a311a3f325e421fa112fcc1b33a975f0

PR #1552, opened 2019-07-19
Add support for organisation access limiting

theseanything:
This adds support access limiting by organisation. Changes in this PR include:
- Adding organisations attribute to the access limits model
- Updating access limited organisations via the PutContent endpoint
- Forwarding access limited organisations downstream to the content store

### DIFF
--- a/app/models/access_limit.rb
+++ b/app/models/access_limit.rb
@@ -3,6 +3,7 @@ class AccessLimit < ApplicationRecord
 
   validate :user_uids_are_strings
   validate :auth_bypass_ids_are_uuids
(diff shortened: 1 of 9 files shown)


app/models/access_limit.rb
```diff
@@ -3,6 +3,7 @@ class AccessLimit < ApplicationRecord
 
   validate :user_uids_are_strings
   validate :auth_bypass_ids_are_uuids
+  validate :user_organisations_are_strings
```
kevindew: This should be uuids btw

PR #1555, opened 2019-07-22
Update pact configuration to avoid multiple runs

kevindew:
Previously the pact would run multiple times for a Publishing API this
seemed to be a result of additional configuration that may have lost
it's original intent.

To update this code I've tried to break this down to be less custom so
to have less challenges. The setup we use for this is a bit weird still
as it only seems to be configured for a single apps pact tests - luckily
this doesn't seem to be the case on any GOV.UK apps currently.

### DIFF
--- a/doc/pact_testing.md
+++ b/doc/pact_testing.md
@@ -42,19 +42,15 @@ file to the broker as `branch-<branch-name>` using the `pact:publish:branch`
 rake task. It then checks out the master branch of publishing-api and verifies
 the pact against that branch using the `pact:verify:branch` rake task.
 
-When a new version of gds-api-adapters is released by bumping the version, the
-job will also publish the pact as the released version, by running the
-`pact:publish:released_version` rake task.
-
 In publishing-api itself, the Jenkins job runs the `pact:verify` rake task to
-verify the current branch against both master and that released version pact.
+verify the current branch against the master branch.
 
 
 ## Running the pacts in development
 
 You can use `pact:verify` locally to run the current branch against the master
-and released version pacts stored in Pact Broker. If you need to run them
-against a local version of gds-api-adapters, run the tests in that directory
+branch of pacts stored in Pact Broker. If you need to run them against a local
+version of gds-api-adapters, run the tests in that directory
 and then set the `USE_LOCAL_PACT` env variable:
 
     USE_LOCAL_PACT=1 bundle exec rake pact:verify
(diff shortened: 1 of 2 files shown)


lib/tasks/pact.rake
```diff
@@ -1,62 +1,30 @@
...
+require 'pact/tasks'
+require 'pact_broker/client/tasks'
+require 'pact/tasks/task_helper'
 
-    Pact::TaskHelper.handle_verification_failure do
-      Pact::TaskHelper.execute_pact_verify
-    end
-
-    unless ENV['USE_LOCAL_PACT'] # avoid running twice against the same pact file.
-      with_temporary_env('GDS_API_PACT_VERSION', "master") do
```
theseanything: Is it the case the we want to test against master _and_ the released version for `gds-api-adapters`? If not we should update the [docs](https://github.com/alphagov/publishing-api/blob/master/doc/pact_testing.md#how-the-tests-run-in-ci).

lib/tasks/pact.rake
```diff
@@ -1,62 +1,30 @@
...
+  task.pact_broker_base_url = ENV.fetch("PACT_BROKER_BASE_URL")
 
-  PactBroker::Client::PublicationTask.new("branch") do |task|
-    task.consumer_version = ENV.fetch("PACT_TARGET_BRANCH")
-    configure_pact_broker_location(task)
+  if ENV['PACT_BROKER_USERNAME']
+    task.pact_broker_basic_auth = {
+      username: ENV['PACT_BROKER_USERNAME'],
+      password: ENV['PACT_BROKER_PASSWORD']
+    }
```
theseanything: Would be nice if these changes were in a separate commit as they don't seem related to reducing repetitive pact tests?

lib/tasks/pact.rake
kevindew: Ah so that must be why it was set to run twice.

So going back through history. It was originally set to run the "latest" version: https://github.com/alphagov/publishing-api/commit/375233a81644151dd9f24bc41e554a0fa8ef97b8 which wasn't actually the released version but just the [most recent branch that was committed](https://github.com/alphagov/publishing-api/commit/f095211c3e3a3ac2dfbbcb059922880a1e387d66).

Hmm I don't think it makes sense to test against both master and released version - since if one fails you've got a failed build anyway. I'll see though whether I can change it to run against just released_version rather than master by default and update docs.

Thanks for catching the docs issue.

lib/tasks/pact.rake
kevindew: Sure, I can sort that

kevindew:
@theseanything I've updated this now.

It turns out that there isn't a released_version option we can run against so master is the best option we have. Instead of their being a tag of released_version they are actually tags of individual versions, and they themselves are broken 😞 - I've opened a PR here to remove them: https://github.com/alphagov/gds-api-adapters/pull/947

doc/pact_testing.md
```diff
@@ -42,19 +42,15 @@ file to the broker as `branch-<branch-name>` using the `pact:publish:branch`
...
 ## Running the pacts in development
 
 You can use `pact:verify` locally to run the current branch against the master
-and released version pacts stored in Pact Broker. If you need to run them
-against a local version of gds-api-adapters, run the tests in that directory
+branch of pacts stored in Pact Broker. If you need to run them against a local
+version of gds-api-adapters, run the tests in that directory
 and then set the `USE_LOCAL_PACT` env variable:
 
     USE_LOCAL_PACT=1 bundle exec rake pact:verify
```
theseanything: Looks like we've removed this altogether too - maybe just remove this section from the docs? I can't see where we use `USE_LOCAL_PACT` in the code.

kevindew: It's used here: https://github.com/alphagov/publishing-api/blob/4e3ef4f4e3bef61c5ecea0199e0ede5a8300d4b1/spec/service_consumers/pact_helper.rb#L21-L23

PR #1560, opened 2019-07-24
Switch to using RSpecMatches from the govuk_schemas gem

cbaines:
As it looks like these were copied there in an effort to remove the
need for the govuk-content-schemas-test-helpers gem.


cbaines:
The test failures here will probably easier to fix once https://github.com/alphagov/govuk_schemas/pull/52 is released.

ChrisBAshton:
Closing old PR.

PR #1562, opened 2019-07-30
Remove description from link expansion

thomasleese:
I can't see that it gets used anywhere on the frontend and it's often the thing about an organisation that changes (which then causes many many documents to be represented) so hopefully this change should keep the low queue length down.

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -49,7 +49,6 @@ module ExpansionRules
     :api_path,
     :base_path,
     :content_id,
-    :description,
     :document_type,
     :locale,
     :public_updated_at,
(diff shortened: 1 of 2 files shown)


cbaines:
This sounds OK, but this is affecting all document types using the default fields right? I guess this would be a little less risky just changing the organisation fields.

thomasleese:
@cbaines I agree it's more risky, but I'm reasonably confident this isn't being used anywhere.

Right now it's currently deployed to integration and [a lot of the content has been republished](https://grafana.integration.publishing.service.gov.uk/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=publishing-api&var-Queues=All&from=now-6h&to=now). I currently can't find any errors in Sentry or high 5xx rates for our apps. I also tested this yesterday and waited for all the content to be republished and didn't find anything.

cbaines:
> @cbaines I agree it's more risky, but I'm reasonably confident this isn't being used anywhere.

OK, that's fine then.

PR #1567, opened 2019-08-02
Small link and dependency resolution improvements

thomasleese:
This PR introduces a number of small improvements to dependency resolution and link expansion related services.

It refactors some of the code to make it easier to read, and it also changes the publish command to only perform dependency resolution on draft editions if necessary (currently a draft edition that needs to be represented will always perform dependency resolution).

Probably easiest to review commit by commit as they're not really related to each other.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -193,7 +193,7 @@ module Commands
       end
 
       def update_dependencies?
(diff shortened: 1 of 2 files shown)


thomasleese:
@kevindew Yeah, I was also tempted to set `update_dependencies: false` on the draft side, but since `public_updated_at` might have changed, I decided to leave it as consistent with the live side.

PR #1570, opened 2019-08-05
Add description to a possible field for expansion

thomasleese:
These fields are used to decide which fields to pluck from the database.

We need this for #1569 to work.

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -106,7 +106,7 @@ module ExpansionRules
 
   POSSIBLE_FIELDS_FOR_LINK_EXPANSION = DEFAULT_FIELDS +
     %i[details] +
-    %i[id state phase unpublishings.type] -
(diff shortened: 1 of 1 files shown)


lib/expansion_rules.rb
```diff
@@ -106,7 +106,7 @@ def details_fields(*fields)
 
   POSSIBLE_FIELDS_FOR_LINK_EXPANSION = DEFAULT_FIELDS +
     %i[details] +
-    %i[id state phase unpublishings.type] -
+    %i[id state phase description unpublishings.type] -
```
edwardkerry: tangent question: is there a particular reason why `:details` has it's own array? Just line length/readability?

thomasleese: I'm not actually sure, but I wanted to keep it as it is. I think it probably is for readability.

PR #1576, opened 2019-08-09
Track the sources of dependency resolution

thomasleese:
This adds tracking on three sources of dependency resolution (command, document type and changed fields) so that we can gather a better picture about what is triggered dependency resolution. This should help us in the quest to reduce the latency of the downstream low queue by building a dashboard telling us where to focus our efforts.

The tracking works by sending the data to Graphite via Statsd which we will use to build a Grafana dashboard in the future.

[Trello Card](https://trello.com/c/7SfWSbLs/1189-improve-tracking-of-what-triggers-dependency-resolution)

### DIFF
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -6,6 +6,7 @@ module Commands
 
         check_version_and_raise_if_conflicting(document, payload[:previous_version])
 
+        save_document_type
         delete_supporting_objects(document.draft)
         delete_draft_from_database
         increment_lock_version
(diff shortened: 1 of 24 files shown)


app/commands/v2/discard_draft.rb
```diff
@@ -6,6 +6,7 @@ def call
 
         check_version_and_raise_if_conflicting(document, payload[:previous_version])
 
+        save_document_type
```
cbaines: This is because the draft is deleted before the information is required right? It might be useful to add a comment in here to mention this if this is indeed the reason.

thomasleese: Good idea. That is the reason - I've amended this with a comment above the method.

PR #1581, opened 2019-08-15
Avoid updating dependencies if links haven't changed

thomasleese:
If the links given to the patch link set command aren't any different to the current links, we don't need to update any dependencies because the expanded links won't have changed.

A few of our publishing applications (notably [hmrc-manuals-api](https://github.com/alphagov/hmrc-manuals-api/blob/69cc927ebe786ef4bcb6ae1edb6d2bcdc09ca963/app/notifiers/publishing_api_notifier.rb#L6-L11) and whitehall) will always call patch link set when publishing a piece of content, often with the same links as before. This can lead to high numbers of items on the downstream low queue.

[Trello Card](https://trello.com/c/uhuMk5bI/1200-patch-link-set-should-check-whether-dependency-resolution-is-necessary)

### DIFF
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -26,9 +26,10 @@ module Commands
         link_set.reload
 
         orphaned_content_ids = link_diff_between(before_links.map(&:target_content_id), link_set.links.map(&:target_content_id))
+        update_dependencies = link_set.links_changed?(before_links)
 
         after_transaction_commit do
-          send_downstream(orphaned_content_ids)
+          send_downstream(orphaned_content_ids, update_dependencies)
         end
 
         Action.create_patch_link_set_action(link_set, before_links, event)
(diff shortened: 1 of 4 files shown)


cbaines:
> If the links given to the patch link set command aren't any different to the current links, we don't need to update any dependencies because the expanded links won't have changed.

Might it be sensible (at some point) to do even less if the patch links request does nothing?

With these changes, there are still jobs enqueued, which won't do anything right, as the links haven't changed?

thomasleese:
@cbaines I agree. There's certainly even more we could do in this area, and I think this is just a start.

I chose to leave that out of this PR to keep it quite small, and I'm hopeful that this change alone will make a reasonably significant dent in the average size of the downstream low queue.

cbaines:
> @cbaines I agree. There's certainly even more we could do in this area, and I think this is just a start.
> 
> I chose to leave that out of this PR to keep it quite small, and I'm hopeful that this change alone will make a reasonably significant dent in the average size of the downstream low queue.

:+1:

app/commands/v2/patch_link_set.rb
```diff
@@ -25,20 +24,24 @@ def call
...
 
         presented = Presenters::Queries::LinkSetPresenter.present(link_set)
         Success.new(presented)
       end
 
     private
 
+      def link_set
+        @link_set ||= LinkSet.find_or_create_locked(content_id: content_id)
+      end
```
cbaines: I'm reluctantly OK with the pattern of functions like this for queries, but given this can actually change the database, I'm unsure what this means for the readability of the code.

In my mind, it would be easier to see where the find or create happens if in `call` there was
```ruby
@link_set = LinkSet.find_or_create_locked(content_id: content_id)
```

thomasleese: Yeah, I think you're right. It makes it harder to know exactly at which point the potential create happens. I've amended the commit without this method.

cbaines: Great, that's looking really good. I like making `update_dependencies` an argument as well, as that makes it clearer to see where it's used also. :+1:

PR #1597, opened 2019-09-16
Present updated_at downstream

barrucadu:
This is so we can use this field in search-api when generating
sitemaps.

---

[Trello card](https://trello.com/c/fVDwVHnm/1001-use-updateddate-in-sitemapxml)

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -17,7 +17,6 @@ module Presenters
       publishing_api_last_edited_at
       state
       unpublishing_type
-      updated_at
       user_facing_version
       web_url
       withdrawn


kevindew:
> This is so we can use this field in search-api when generating
> sitemaps.
> 
> [Trello card](https://trello.com/c/fVDwVHnm/1001-use-updateddate-in-sitemapxml)

Is public_updated_at an option for this? 

updated_at is a rather delicate field to downstream for a couple of reasons: 1) content_store has it's own updated_at field which means something different 🤦‍♀ 2) it doesn't really mean anything concrete, just the time a Publishing API record was changed which has a variety of different interpretations and 3) gets bogged down in the public/private aspect of Publishing API updates.

barrucadu:
> Is public_updated_at an option for this?

We currently use that, but that means the updated timestamp in the sitemap only reflects the last major change.  As search engines use sitemaps, that means we're potentially making them check pages less frequently that they should.

> updated_at is a rather delicate field to downstream for a couple of reasons: 1) content_store has it's own updated_at field which means something different 🤦‍♀ 2) it doesn't really mean anything concrete, just the time a Publishing API record was changed which has a variety of different interpretations and 3) gets bogged down in the public/private aspect of Publishing API updates.

Yeah, as soon as I realised I'd have to change whitehall, publishing-api, *and* search-api, I started to wonder if this was the right thing to do.  I suspect the time a change got pushed to search-api is good enough, in which case I can change just that.

kevindew:
> > Is public_updated_at an option for this?
> 
> We currently use that, but that means the updated timestamp in the sitemap only reflects the last major change. As search engines use sitemaps, that means we're potentially making them check pages less frequently that they should.

Gotcha, makes sense

> > updated_at is a rather delicate field to downstream for a couple of reasons: 1) content_store has it's own updated_at field which means something different 🤦‍♀ 2) it doesn't really mean anything concrete, just the time a Publishing API record was changed which has a variety of different interpretations and 3) gets bogged down in the public/private aspect of Publishing API updates.
> 
> Yeah, as soon as I realised I'd have to change whitehall, publishing-api, _and_ search-api, I started to wonder if this was the right thing to do. I suspect the time a change got pushed to search-api is good enough, in which case I can change just that.

That sounds good - I don't think we actually know the timestamp much clearer than that anyway. We could know each time the content of a document changed, but that wouldn't factor for any automatic changes to the content caused by linking. So I think you're right to use that timestamp there even if it has the potential to be more grossly more recent than the actual value.

PR #1614, opened 2019-10-08
Add auth bypass ids to editions

1pretz1:
This is the first part of work being done to separate auth_bypass_ids from access limiting. It adds a new column to editions and copies over the auth_bypass_ids to this column every time its saved on the access limit.

We need to do this so all new editions being updated have the auth_bypass_ids column populated which allows us to run a data migration transferring the access limit records to the editions without losing any data. Without this step if someone was to update an access limit while the data migration was running it wouldn't be transferred to the edition, then when we removed the auth_bypass_ids column on the access limit table the record would be lost.

Trello:
https://trello.com/c/DnYanQdY/132-separate-auth-bypassing-from-access-limiting-in-publishing-api


app/presenters/edition_presenter.rb
```diff
@@ -29,11 +29,11 @@ def initialize(edition, draft: false)
     end
 
     def for_content_store(payload_version)
-      present.except(:update_type).merge(payload_version: payload_version)
+      present.except(:update_type, :auth_bypass_ids).merge(payload_version: payload_version)
     end
 
     def for_message_queue(payload_version)
-      present.merge(
+      present.except(:auth_bypass_ids).merge(
```
kevindew: I'd suggest using the [NON_PRESENTED_PROPERTIES](https://github.com/alphagov/publishing-api/blob/b8c428c9039d1477bf558aea02a37a20b9e76281/app/presenters/edition_presenter.rb#L3-L24) constant for this rather than adding individual excepts.

app/models/edition.rb
```diff
@@ -74,6 +76,12 @@ class Edition < ApplicationRecord
 
   delegate :content_id, :locale, to: :document
 
+  def auth_bypass_ids_are_uuids
+    unless auth_bypass_ids.all? { |id| UuidValidator.valid?(id) }
```
kevindew: Oh interesting - I didn't realise these were validated to be uuids. I've had to amend https://trello.com/c/jKRvjyAe/130-store-authbypassid-as-random-data-in-the-collection-publisher-database to reflect this.

app/presenters/edition_presenter.rb
1pretz1: Sure, although this will be removed in favour of a singular except on the `for_message_queue` method at a later date when we're ready to talk to the content-store.

app/models/edition.rb
1pretz1: Strictly UUID or no entry! ⚔️

1pretz1:
I've made the amendment you've suggested @kevindew.

As for your question around Postgres arrays, I've not hit any weirdness using them apart from when I was playing about with casting from a postgres json array to a normal posgres array, turns out you cant!

PR #1616, opened 2019-10-09
Custom expansion fields can handle multiple links

kevindew:
Trello: https://trello.com/c/jmf3okaM/133-update-publishing-api-link-expansion-field-check-to-include-all-possible-fields

This changes the `find_custom_expansion_fields` to behave differently
whether a link_type is provided or not.

When a link_type of nil is provided this will find all possible fields
that this document type can provide by collating all the options with
links in. The expectation is that the use of nil is used when the
context is link agnostic and thus needs all data, for this reason
default fields are also collated into this data if some link types may
use default fields.

When a link type is given this ensures that a match for a document type
and a link will take precedence over a document_type without a link (as
was previously the case).

This change is done for the introduction of new custom fields that use
link type and document type when there is already specific fields for
that document type (step_by_step_nav).

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -148,22 +148,41 @@ module ExpansionRules
     @dependency_resolution ||= ExpansionRules::DependencyResolution.new(self)
   end
 
-  def find_custom_expansion_fields(document_type, options = {})
-    should_check_link_type = options[:link_type]
-    link_type = options[:link_type].try(:to_sym)
-
-    condition = CUSTOM_EXPANSION_FIELDS.find do |cond|
-      next if should_check_link_type && cond.fetch(:link_type, link_type) != link_type
-
-      cond[:document_type] == document_type.to_sym
+  def expansion_fields(document_type, link_type = nil)
+    if link_type
+      expansion_fields_for_linked_document_type(document_type, link_type)
+    else
+      expansion_fields_for_document_type(document_type)
     end
-
-    condition[:fields] if condition
   end
 
-  def expansion_fields(document_type, link_type = nil)
-    find_custom_expansion_fields(document_type, link_type: link_type) ||
-      DEFAULT_FIELDS
+  def expansion_fields_for_document_type(document_type)
+    matching_document_types = CUSTOM_EXPANSION_FIELDS.select do |item|
+      item[:document_type] == document_type.to_sym
+    end
+
+    return DEFAULT_FIELDS unless matching_document_types.any?
+
+    collated_fields = matching_document_types.flat_map { |item| item[:fields] }
+    matches_any_link_type = matching_document_types.any? { |item| item[:link_type].nil? }
+
+    collated_fields += DEFAULT_FIELDS unless matches_any_link_type
+    collated_fields.uniq
+  end
+
+  def expansion_fields_for_linked_document_type(document_type, link_type)
+    matching_link = CUSTOM_EXPANSION_FIELDS.find do |item|
+      item[:document_type] == document_type.to_sym &&
+        item[:link_type] == link_type.to_sym
+    end
+    return matching_link[:fields] if matching_link
+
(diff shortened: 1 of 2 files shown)


spec/lib/expansion_rules_spec.rb
```diff
@@ -38,6 +38,70 @@
...
+
+      it "collates the fields for document_types with multiple link types" do
+        expect(rules.find_custom_expansion_fields(:news, nil))
+          .to match_array(%i[news_a news_b news_c news_d news_e])
+      end
+
+      it "includes default fields for a document type that only exists in "\
+        "custom fields with a link" do
+        expect(rules.find_custom_expansion_fields(:editorial, nil))
+          .to contain_exactly(:editorial_a, :editorial_b, *ExpansionRules::DEFAULT_FIELDS)
```
ChrisBAshton: I'm a little unsure why we're using `eq` on line 89, `match_array` on line 94 and `contain_exactly` on line 100.

lib/expansion_rules.rb
```diff
@@ -148,21 +148,30 @@ def dependency_resolution
...
 
-      cond[:document_type] == document_type.to_sym
-    end
+    if link_type.nil?
+      collated_fields = matching_document_types.inject([]) { |memo, item| memo + item[:fields] }
+      may_use_default_fields = matching_document_types.all? { |item| item[:link_type].present? }
 
-    condition[:fields] if condition
+      collated_fields += DEFAULT_FIELDS if may_use_default_fields
+      collated_fields.uniq
```
ChrisBAshton: Are we testing the `.uniq` behaviour anywhere in "expansion_rules_spec.rb"?

lib/expansion_rules.rb
```diff
@@ -148,21 +148,30 @@ def dependency_resolution
...
+    else
+      matched =
+        matching_document_types.find { |item| item[:link_type] == link_type.to_sym } ||
+        matching_document_types.find { |item| item[:link_type].nil? }
+
+      matched[:fields] if matched
+    end
```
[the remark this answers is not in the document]
kevindew: Thanks - I like this suggestion I've reworked this into two methods.

lib/expansion_rules.rb
```diff
@@ -148,21 +148,30 @@ def dependency_resolution
...
 
-    condition = CUSTOM_EXPANSION_FIELDS.find do |cond|
-      next if should_check_link_type && cond.fetch(:link_type, link_type) != link_type
+    return nil unless matching_document_types.any?
 
-      cond[:document_type] == document_type.to_sym
-    end
+    if link_type.nil?
+      collated_fields = matching_document_types.inject([]) { |memo, item| memo + item[:fields] }
```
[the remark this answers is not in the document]
ChrisBAshton: Interesting! Learned something new today 👍

lib/expansion_rules.rb
kevindew: Got that covered now, thanks

lib/expansion_rules.rb
```diff
@@ -148,22 +148,41 @@ def dependency_resolution
...
   end
 
   def expansion_fields(document_type, link_type = nil)
-    find_custom_expansion_fields(document_type, link_type: link_type) ||
-      DEFAULT_FIELDS
+    if link_type
+      expansion_fields_for_linked_document_type(document_type, link_type)
+    else
+      expansion_fields_for_document_type(document_type)
+    end
```
ChrisBAshton: I would move this method up to be above the definitions for `expansion_fields_for_linked_document_type` and `expansion_fields_for_document_type`, so that it reads top-to-bottom. Same with the accompanying tests.

spec/lib/expansion_rules_spec.rb
ChrisBAshton: Thanks for the explanation and tweak.

lib/expansion_rules.rb
kevindew: Ok done.

PR #1625, opened 2019-10-14
Use editions auth bypass

1pretz1:
Publishing API stores auth_bypass_ids inside access limit models. They should instead be stored distinctly on an edition. This makes it easier for us to share auth_bypass_id via link expansion and helps reduce an area of frequent confusion in the code base.

Tested in integration using content-publisher (sending root auth_bypass_ids and nested access_limited auth_bypass_ids). In all variations, the code worked as expected.

More information on the implementation in the commits: ->

Trello:
https://trello.com/c/DnYanQdY/132-separate-auth-bypassing-from-access-limiting-in-publishing-api

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -104,7 +104,6 @@ module Commands
             access_limit.update_attributes!(
               users: (payload[:access_limited][:users] || []),
               organisations: (payload[:access_limited][:organisations] || []),
-              auth_bypass_ids: (payload[:access_limited][:auth_bypass_ids] || []),
             )
           end
         else
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -7,6 +7,7 @@ module Commands
         prepare_content_with_base_path
         check_update_type
 
+        update_root_payload_with_auth_bypass_ids
         edition = create_or_update_edition
         set_timestamps(edition)
 
@@ -98,7 +99,7 @@ module Commands
       end
 
       def access_limit(edition)
-        if payload[:access_limited]
+        if payload[:access_limited].present?
           AccessLimit.find_or_create_by(edition: edition).tap do |access_limit|
             access_limit.update_attributes!(
               users: (payload[:access_limited][:users] || []),
@@ -108,6 +109,14 @@ module Commands
           end
         else
           AccessLimit.find_by(edition: edition).try(:destroy)
+        end
+      end
+
+      def update_root_payload_with_auth_bypass_ids
+        return if payload[:auth_bypass_ids]
+
+        if payload.dig(:access_limited, :auth_bypass_ids)
+          payload[:auth_bypass_ids] = payload[:access_limited].delete(:auth_bypass_ids)
         end
       end
 
--- a/app/models/access_limit.rb
+++ b/app/models/access_limit.rb
@@ -1,17 +1,11 @@
 class AccessLimit < ApplicationRecord
   belongs_to :edition
 
-  after_save :copy_auth_bypass_ids_to_edition
-
   validate :user_uids_are_strings
   validate :auth_bypass_ids_are_uuids
   validate :user_organisations_are_uuids
 
 private
-
-  def copy_auth_bypass_ids_to_edition
-    edition.update!(auth_bypass_ids: auth_bypass_ids)
-  end
 
   def user_uids_are_strings
     unless users.all? { |id| id.is_a?(String) }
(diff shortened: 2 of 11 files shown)


app/presenters/edition_presenter.rb
```diff
@@ -34,7 +33,7 @@ def for_content_store(payload_version)
     end
 
     def for_message_queue(payload_version)
-      present.merge(
+      present.except(:auth_bypass_ids).merge(
```
ChrisBAshton: I can't see a test for the absence of `auth_bypass_ids` in the message queue (unless I'm misunderstanding something) - can you create one?

app/presenters/edition_presenter.rb
```diff
@@ -89,9 +88,9 @@ def access_limited
       else
         {
           access_limited: {
-            users: access_limit.users,
-            organisations: access_limit.organisations,
-            auth_bypass_ids: access_limit.auth_bypass_ids,
+            users: access_limit&.users,
+            organisations: access_limit&.organisations,
```
ChrisBAshton: I'm not clear why you've added the ampersands here?

app/presenters/queries/content_item_presenter.rb
```diff
@@ -223,11 +223,14 @@ def join_lateral_aggregates(scope)
       def parse_results(results)
         json_columns = %w(details routes redirects state_history unpublishing links)
         int_columns = %w(user_facing_version lock_version)
+        psql_array_column = "auth_bypass_ids"
 
         Enumerator.new do |yielder|
           results.each do |result|
             json_columns.each { |c| parse_json_column(result, c) }
             int_columns.each { |c| parse_int_column(result, c) }
+            parse_psql_array_column(result, psql_array_column)
```
ChrisBAshton: I think the `psql_array_column` variable is unnecessary, and would just pass `"auth_bypass_ids"` directly, as it's not being reused and it's already clear from the method name that we're passing the name of a column.

app/commands/v2/put_content.rb
```diff
@@ -111,6 +112,14 @@ def access_limit(edition)
         end
       end
 
+      def update_root_payload_with_auth_bypass_ids
+        if payload[:auth_bypass_ids]
+          payload
+        elsif payload.dig(:access_limited, :auth_bypass_ids)
+          payload[:auth_bypass_ids] = payload.dig(:access_limited, :auth_bypass_ids)
+        end
+      end
```
ChrisBAshton: Nice 👍

app/models/access_limit.rb
```diff
@@ -13,12 +12,6 @@ def user_uids_are_strings
     end
   end
 
-  def auth_bypass_ids_are_uuids
-    unless auth_bypass_ids.all? { |id| UuidValidator.valid?(id) }
-      errors.add(:auth_bypass_ids, ["contains invalid UUIDs"])
-    end
-  end
```
ChrisBAshton: ~Should this validation still happen, just on the auth_bypass_ids column on editions in the previous commit?~ Ah, it happens in edition.rb. Glad you removed the duplicate code!

app/presenters/edition_presenter.rb
1pretz1: I've removed this `except` in favour of adding the field to NON_PRESENTED_PROPERTIES, the logic is still the same however so i've added a test here - https://github.com/alphagov/publishing-api/pull/1625/files#diff-8180fe08951fde62fc6dd9c41bee0d17R60-R62.

app/presenters/edition_presenter.rb
1pretz1: I've added ampersands here to signify that access_limit could be nil. Adding an `&` after access_limit means that if access_limit is `nil` then calling users or organisations on this value wont throw a method error and instead return nil.

This is needed as the check here https://github.com/alphagov/publishing-api/pull/1625/files#diff-5e3b9d2d7f141f5b5d7383dac0157593R87 means that there doesn't necessarily need to be an access_limit if an edition has `auth_bypass_ids`, the code here will be altered and removed as we only need it to keep this way to preserve the legacy way of sending bypass_ids to content-store.

app/presenters/queries/content_item_presenter.rb
1pretz1: Sure!

app/presenters/edition_presenter.rb
ChrisBAshton: Ahhhh thanks for the explanation. I'd forgotten about that.

PR #1627, opened 2019-10-15
Use govuk_test

thomasleese:
Unfortunately this does add a few dependencies we don't need for the Publishing API, but it does install Brakeman for us.

In the future `govuk_test` might do more than just configure a test environment for frontend apps, so it's still good to use this Gem.


thomasleese:
@kevindew I would be up for renaming it if we want to be clear that it should be used only for frontend testing - I think at the moment we don't have a clear vision on what belongs where.

In my mind `govuk_app_config` (which should maybe be renamed to `govuk_app`?) contains things necessary to run the app while `govuk_test` contains things necessary to test the app (in this case Brakeman is used to test the app for security vulnerabilities and runs in CI). One key distinction between them is that `govuk_test` belongs in the `development` Bundle group so the dependencies aren't necessary to run the app. For example, if we wanted to include `govuk-lint` as part of this I think it would go in `govuk_test` and not in `govuk_app_config`.

It is unfortunate that `govuk_test` currently contains only frontend related testing, but I can imagine a future where it would include more backend-y things like setting up `govuk_sidekiq` for testing.

kevindew:
> @kevindew I would be up for renaming it if we want to be clear that it should be used only for frontend testing - I think at the moment we don't have a clear vision on what belongs where.
> 
> In my mind `govuk_app_config` (which should maybe be renamed to `govuk_app`?) contains things necessary to run the app while `govuk_test` contains things necessary to test the app (in this case Brakeman is used to test the app for security vulnerabilities and runs in CI). One key distinction between them is that `govuk_test` belongs in the `development` Bundle group so the dependencies aren't necessary to run the app. For example, if we wanted to include `govuk-lint` as part of this I think it would go in `govuk_test` and not in `govuk_app_config`.
> 
> It is unfortunate that `govuk_test` currently contains only frontend related testing, but I can imagine a future where it would include more backend-y things like setting up `govuk_sidekiq` for testing.

Oh yes good point about development/test dependencies.

Hmm does feel a bit wrong to load this up with the other deps but I can't see a better route forward now.

PR #1645, opened 2019-11-19
Add reverse links for appointments

thomasleese:
This means that when you link to a person or a role from a role appointment, the person and the role will automatically get a link back to the appointment.

This is useful because it means we don't need to maintain four links for each appointment (person -> appointment, appointment -> person, role -> appointment, appointment -> role) rather than two links (appointment -> role, appointment -> person). Having four links means that it's possible for the links to become out of sync.

In practice this means that Whitehall needs to republish people and roles whenever an appointment changes using `after_save` hooks without relying entirely on the Publishing API for this kind of dependency resolution.

I've left the expansion rules for the old `ordered_current_appointments` style links, but once the frontend has been converted to read the data from the new `appointments` link, we can remove those rules.

This will end up merging the two `ordered_current_appointments` and `ordered_previous_appointments` into one link (`appointments`) which means the frontends will have to parse the appointments to work out if they are current or not. This pushes more logic into frontends, but it means there is less duplicate state to manage in the publishing apps (we don't need to patch links when an appointment transitions from previous to current).

I've chosen the name `appointed_person` and `appointed_role` for the explicit links because we already have links to `person` and `role` for attaching documents to a particular person and I didn't want them to get a reverse link.

I've chosen the name `appointments` for the reverse link since it's the same as `ordered_current_appointments` except that I don't believe we can make any guarantees on the ordering any more and because it will include both the current and the previous appointments.

[Trello Card](https://trello.com/c/roHfuPUV/1362-use-reverse-links-for-role-appointments)

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -34,6 +34,8 @@ module ExpansionRules
     [:ordered_current_appointments, :person],
     [:ordered_previous_appointments, :role, :ordered_parent_organisations],
     [:ordered_previous_appointments, :person],
+    [:role_appointments, :person],
+    [:role_appointments, :role, :ordered_parent_organisations],
   ].freeze
 
   REVERSE_LINKS = {
@@ -46,6 +48,8 @@ module ExpansionRules
     pages_related_to_step_nav: :related_to_step_navs,
     legacy_taxons: :topic_taxonomy_taxons,
     pages_secondary_to_step_nav: :secondary_to_step_navs,
+    person: :role_appointments,
+    role: :role_appointments,
   }.freeze
 
   DEFAULT_FIELDS = [


kevindew:
> This means that when you link to a person or a role from a role appointment, the person and the role will automatically get a link back to the appointment.

This is great. It's a feature we want for Content Publisher so that we can link directly to role appointments rather than having to [look up role and person](https://github.com/alphagov/content-publisher/blob/master/app/services/preview_service/payload.rb#L94-L103) - though we'll still need to enable link expansion on edition links for that, which isn't going to be the most fun afternoon (is that a problem you'll hit?)

> This is useful because it means we don't need to maintain four links for each appointment (person -> appointment, appointment -> person, role -> appointment, appointment -> role) rather than two links (appointment -> role, appointment -> person). Having four links means that it's possible for the links to become out of sync.
> 
> In practice this means that Whitehall needs to republish people and roles whenever an appointment changes using `after_save` hooks without relying entirely on the Publishing API for this kind of dependency resolution.
> 
> I've left the expansion rules for the old `ordered_current_appointments` style links, but once the frontend has been converted to read the data from the new `appointments` link, we can remove those rules.

Are there no concerns about losing the ordered aspect?

> This will end up merging the two `ordered_current_appointments` and `ordered_previous_appointments` into one link (`appointments`) which means the frontends will have to parse the appointments to work out if they are current or not. This pushes more logic into frontends, but it means there is less duplicate state to manage in the publishing apps (we don't need to patch links when an appointment transitions from previous to current).

Do we need to do any link expansion to reveal whether they're current or not?

> I've chosen the name `appointed_person` and `appointed_role` for the explicit links because we already have links to `person` and `role` for attaching documents to a particular person and I didn't want them to get a reverse link.

Hmm is that going to create a strange situation where we have person and role links on the role-appointment but then appointed_person and appointed_role when expanded? or are you planning to re-name all the links? I think I just don't understand this decision, so maybe I'm on the wrong track.

> I've chosen the name `appointments` for the reverse link since it's the same as `ordered_current_appointments` except that I don't believe we can make any guarantees on the ordering any more and because it will include both the current and the previous appointments.
> 
> [Trello Card](https://trello.com/c/roHfuPUV/1362-use-reverse-links-for-role-appointments)

Should we go with role_appointments here so it's consistent with the document type. I feel appointments makes it more ambiguous than it needs to be.

thomasleese:
Had a discussion in person, I've pushed up an amended commit which should resolve the problems raised.

PR #1647, opened 2019-11-20
Support reverse links with the same link type

thomasleese:
We want reverse links for `role_appointments` to come from direct links of both a `person` and `role` type. Currently reverse links can only be associated with one type of direct link.

The reverse links with the same type were added in: https://github.com/alphagov/publishing-api/pull/1645

The current behaviour means that we only see the `role_appointments` links in a person content item, and not in the role. The reason for this is through the use of `REVERSE_LINKS.key` which only returns the key for the first occurrence of a value, and not all the keys associated with a value. Person happens to be above role in the `REVERSE_LINKS` hash so the Publishing API thinks that only people need to get the reverse links for role appointments.

[Trello Card](https://trello.com/c/roHfuPUV/1549-8-use-reverse-links-for-role-appointments)

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -173,7 +173,7 @@ module ExpansionRules
     %i[api_path withdrawn]
 
   def reverse_links
-    REVERSE_LINKS.values
+    REVERSE_LINKS.values.uniq
   end
 
   def reverse_link_type(link_type)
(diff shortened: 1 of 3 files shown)


thomasleese:
Continuing my journey into the depth of the Publishing API...

I found understanding a lot of this code quite difficult, so hopefully I got this right. I've tested it on integration and it does solve the problem we're seeing.

thomasleese:
@kevindew Yep, that's right. We now support a 1 to many mapping for reverse links.

PR #1659, opened 2019-11-29
Enable rails-specific rules for rubocop-govuk

benjamineskola:
trello: https://trello.com/c/Lijki8aj/1584-3-enable-rails-rubocop-rules-via-rubocop-govuk-on-all-our-rails-apps


benjamineskola:
Remaining broken checks should be fixed by #1660.

spec/lib/events/s3_exporter_spec.rb
```diff
@@ -1,12 +1,12 @@
...
 
 RSpec.describe Events::S3Exporter do
-  let(:created_before) { Time.now }
+  let(:created_before) { Time.zone.now }
   let(:created_on_or_after) { nil }
   let(:client) { Aws::S3::Client.new(region: "eu-west-1", stub_responses: true) }
   let(:resource) { Aws::S3::Resource.new(client: client) }
   let(:object_exists?) { false }
-  let(:s3_key) { "events/#{created_before.to_s(:iso8601)}.csv.gz" }
+  let(:s3_key) { "events/#{created_before.strftime('%FT%T%:z')}.csv.gz" }
```
AlanGabbianelli: I'm not sure how much clearer this is... 🤔

issyl0: I think it's the polar opposite of clear.

benjamineskola: It's not clearer, but this is the format being used in the [implementation](https://github.com/alphagov/publishing-api/blob/411556fe0849c0cee3c798705f2d679e853e74ff/lib/events/s3_exporter.rb#L46).

Changing the implementation to be the clearer version would also be an option, but that would have an actual effect in production so I thought I should steer clear.

The problem is that these two formats aren't precisely identical; one shows the timezone as `Z`, and the other shows it as `+00:00`.

Actually this has made me wonder whether this would behave correctly when the timezone changes; rubocop defaults to using local time, and the tests passed that way, but local time == UTC. It's possible that it'll break in another timezone, and it's also possible that this change would be unnecessary with UTC time objects (although I think I checked that already). I'll investigate further.

benjamineskola: Double-checked and it doesn't seem to be related to the timezone issue (though I did track down an edge case where one test would fail if it was run in a timezone different to the UK's).

PR #1663, opened 2019-12-06
Add rake task to republish documents by document type

deborahchua:
We would like to prioritise republishing organisation pages if there
is a delay due to a large publishing queue.

This rake task gives us the flexibilty to pass any document type(s)
and republish all affected editions using the `high_priority`
sidekiq queue.

Trello card: https://trello.com/c/uzKUr4WT/1613-3-make-sure-organisation-pages-show-people-changes-quickly-during-reshuffle

### DIFF
--- a/lib/tasks/represent_downstream.rake
+++ b/lib/tasks/represent_downstream.rake
@@ -99,5 +99,19 @@ namespace :represent_downstream do
         queue: DownstreamQueue::HIGH_QUEUE,
       )
     end
+
+    desc "
+    Represent documents by document type(s) downstream on the high_priority sidekiq
+    queue
+    Usage
+    rake 'represent_downstream:high_priority:document_type[organisation]'
+    "
+    task :document_type, [:document_types] => :environment do |_t, args|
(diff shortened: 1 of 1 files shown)


lib/tasks/represent_downstream.rake
```diff
@@ -85,6 +85,21 @@ namespace :represent_downstream do
...
   end
 
+  desc "
+  Represent documents by document type(s) downstream on the high_priority sidekiq
+  queue
+  Usage
+  rake 'represent_downstream:high_priority_document_type[organisation]'
+  "
+
+  task :high_priority_document_type, [:document_types] => :environment do |_t, args|
```
thomasleese: The code looks good! Could we move this task into the `:high_priority` namespace below and change the name to just `document_type`? Then it'll be called with `rake represent_downstream:high_priority:document_type[...]`.

deborahchua: Thanks @thomasleese! I have updated this now 😄

PR #1682, opened 2020-01-08
Create a rake task to remove links

laurentqro:
As part of auditing the Business Finder, we discovered multiple pages with links (in the sidebar) to the Business Finder.

As we are shutting the Business Finder down, we should remove these links from content.

https://trello.com/c/iExS63UL/342-remove-any-sidebar-links-to-the-business-finder

### DIFF
--- a/lib/tasks/links.rake
+++ b/lib/tasks/links.rake
@@ -43,9 +43,17 @@ namespace :links do
 
     old_links = link_set.links.where(link_type: link_type)
 
+    puts "############ OLD LINKS ###############"
+    puts old_links.inspect
+    puts "###########################"
+
     new_links = old_links.reject do |link|
       link.target_content_id == target_content_id && link.link_type == link_type
     end
+
+    puts "############ NEW LINKS ###############"
+    puts new_links.inspect
+    puts "###########################"
 
     payload = {
       content_id: link_set.content_id,
@@ -53,6 +61,10 @@ namespace :links do
       previous_version: link_set.stale_lock_version,
       bulk_publishing: true,
     }
+
+    puts "############ PAYLOAD ###############"
+    puts payload.inspect
+    puts "###########################"
 
     Commands::V2::PatchLinkSet.call(payload)
   end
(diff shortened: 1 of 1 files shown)


lib/tasks/links.rake
```diff
@@ -0,0 +1,58 @@
...
+    old_links = link_set.links.where(link_type: link_type)
+
+    new_links = old_links.reject do |link|
+      link.target_content_id == target_content_id && link.link_type == link_type
+    end
+
+    payload = {
+      content_id: link_set.content_id,
+      links: { link_type.to_sym => new_links },
+      bulk_publishing: true,
```
koetsier: I believe this also may need the previous version of the links:

 @option params [Integer] previous_version The previous version (returned by `get_links`). If this version is not the current version, the publishing-api will reject the change and return 409 Conflict. (optional)

lib/tasks/links.rake
```diff
@@ -0,0 +1,58 @@
...
+    link_sets.each do |link_set|
+      remove_link_from_link_set(args, link_set)
+    rescue CommandError => e
+      puts e.message
+      puts "Skipping removal for document with content_id: #{link_set.content_id} ..."
+      next
+    end
+  end
+
+  desc "Remove link of given type to given target from given document"
```
koetsier: is this task needed?

lib/tasks/links.rake
```diff
@@ -0,0 +1,58 @@
...
+  task :remove_from_all_documents, %w[target_content_id link_type] => :environment do |_, args|
+    raise "Missing parameter: target_content_id" unless args.target_content_id
+
+    target_content_id = args.target_content_id
+
+    raise "Missing parameter: link_type" unless args.link_type
+
+    link_type = args.link_type
+
+    link_sets = LinkSet.joins(:links).where(links: { target_content_id: target_content_id, link_type: link_type })
```
koetsier: Actually looking at @vanitabarrett's rake task: https://github.com/alphagov/content-tagger/compare/find-business-finder-related-links, we could use ```publishing_api.get_linked_items``` which means we don't have to do raw database access and we can go through the external api which may be a bit better

laurentqro:
so it turns out this is not needed. Taking down the Business Finder has had the effect of removing all the sidebar links to the Business finder from pages that had them. So ... :tada:

PR #1706, opened 2020-02-06
Add temporary Rake task to update unpublished specialist document paths

edwardkerry:
A large number of SpecialistDocuments were urgently unpublished before redirects were created.

This PR adds a temporary Rake task to add the redirect, after running it can be reverted.


lib/tasks/update_export_health_certificate_paths.rake
```diff
@@ -0,0 +1,153 @@
...
+
+                    }
+    unpublishing.redirects << redirect_details
+    unpublishing.save!
+
+    content_ids << unpublishing.edition.content_id
+  end
+
+  represent_downstream(
+      Document.where(content_id: content_ids.split(" ")),
```
thomasleese: I don’t think you need to call split here as it’ll already be an array.

lib/tasks/update_export_health_certificate_paths.rake
```diff
@@ -0,0 +1,153 @@
...
+
+  paths.each do |path|
+    unpublishing = Edition.live.find_by(base_path: path).unpublishing
+    redirect_details = {
+                      "path"       =>  path,
+                      "type"        =>  "exact",
+                      "destination" =>  redirect_path
+
+                    }
+    unpublishing.redirects << redirect_details
```
thomasleese: I think you’ll want to assign this rather than append to the array so you can replace the existing redirect.

edwardkerry: whoops, thanks

lib/tasks/update_export_health_certificate_paths.rake
edwardkerry: 👍

lib/tasks/update_export_health_certificate_paths.rake
```diff
@@ -0,0 +1,157 @@
...
+    unpublishing.redirects = redirect_details
+    unpublishing.save!
+
+    content_ids << unpublishing.edition.content_id
+  end
+
+  represent_downstream(
+    Document.where(
+      content_id: content_ids,
+      queue: DownstreamQueue::HIGH_QUEUE,
```
thomasleese: I think this argument belongs to `represent_downstream` rather than `Document.where`.

lib/tasks/update_export_health_certificate_paths.rake
```diff
@@ -0,0 +1,153 @@
...
+    unpublishing = Edition.live.find_by(base_path: path).unpublishing
+    redirect_details = [{
+                      "path"        =>  path,
+                      "type"        =>  "exact",
+                      "destination" =>  redirect_path,
+
+                    }]
+    unpublishing.redirects = redirect_details
+    unpublishing.type = "redirect"
+    unpublishing.save!
```
cbaines: It's pretty irregular to work with the data in the Publishing API directly, rather than going through the API. Is there an option of unpublishing this content from a publishing app?

edwardkerry: As I understand it, Specialist Publisher doesn't have it's own database but also reads from the publishing-api, so having discussed with @thomasleese this was the suggested approach

edwardkerry:
Replaced by https://github.com/alphagov/specialist-publisher/pull/1580

PR #1739, opened 2020-03-09
Add JSONB columns

MuriloDalRi:
This is the first of a series of PRs to switch to using JSONB instead of the JSON data type. This PR will:

1. Add columns of the JSONB type
2. Add callbacks to save data to both the old and new columns
3. Add a rake task to backfill the new columns

The next steps in following PRs will be:
1. Switch to only using the new columns, adding non-null constraints and default values
2. Remove callbacks to save to new columns, remove the old columns and rename the new ones

[trello](https://trello.com/c/JJ194ZeT/1800-8-publishing-api-use-jsonb-instead-of-json-postgres-data-type-%F0%9F%8D%90)

### DIFF
--- a/lib/tasks/backfill_new_jsonb_columns.rake
+++ b/lib/tasks/backfill_new_jsonb_columns.rake
@@ -1,0 +1,91 @@
+desc "Migrate access_limits json columns"
+task access_limits: :environment do
+  puts "starting access limits"
+  acc_lim_sql = "UPDATE access_limits SET temp_users = users, temp_organisations = organisations;"
+  ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(acc_lim_sql) }
+end
+
+desc "Migrate editions json columns"
+task editions: :environment do
+  record = Edition.find_by(id: 704530)
+  details = record.details.to_json.gsub(/\\u0000/, "")
+  details = JSON.parse(details).symbolize_keys
+  record.update(details: details)
+
+  low = 0
+  high = 1000
+  total = Edition.maximum("id")
+
+  sql = "UPDATE editions SET temp_details = details, temp_routes = routes, " \
+        "temp_redirects = redirects WHERE id > #{low} AND id <= #{high};"
+
+  while high < total
+    ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(sql) }
+    puts "finished editions batch #{low} to #{high}"
+    low += 1000
+    high += 1000
+  end
+end
+
+desc "Migrate events json columns"
+task events: :environment do
+  event = Event.find_by(id: 31148138)
+  payload = event.payload.to_json.gsub(/\\u0000/, "")
+  payload = JSON.parse(payload).symbolize_keys
+  event.update(payload: payload)
+
+  low = 0
+  high = 1000
+  total = Event.maximum("id")
+
+  sql = "UPDATE events SET temp_payload = payload WHERE id > #{low} AND id <= #{high};"
+
+  while high < total
+    ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(sql) }
+    puts "finished events batch #{low} to #{high}"
+    low += 1000
+    high += 1000
+  end
+end
+
+desc "Migrate expanded_links json columns"
+task expanded_links: :environment do
+  low = 0
+  high = 1000
+  total = ExpandedLinks.maximum("id")
+
+  sql = "UPDATE expanded_links SET temp_expanded_links = expanded_links " \
+        "WHERE id > #{low} AND id <= #{high};"
+
+  while high < total
+    ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(sql) }
+    puts "finished expanded links batch #{low} to #{high}"
+    low += 1000
+    high += 1000
+  end
+end
+
+desc "Migrate unpublishings json columns"
+task unpublishings: :environment do
+  low = 0
+  high = 1000
+  total = Unpublishing.maximum("id")
+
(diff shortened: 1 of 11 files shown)


spec/lib/events/s3_importer_spec.rb
```diff
@@ -16,15 +16,16 @@ def gzipped_file(contents)
   let(:file) { gzipped_file(file_contents) }
   let(:file_contents) do
     <<-CSV.strip_heredoc
-      id,action,payload,user_uid,created_at,updated_at,request_id,content_id
-      10,Publish,"{""content_id"":""2cc503f1-5fef-4fac-8381-63f6689cd6a2""}",,2015-11-11 09:55:00 UTC,2015-11-11 09:55:00 UTC,,2cc503f1-5fef-4fac-8381-63f6689cd6a2
-      11,Publish,"{""content_id"":""8bc0c1dd-4842-4283-a123-5671a34b67eb""}",e676cb22-0c0c-4534-9514-a65ebcc7a9e2,2015-11-11 14:00:00 UTC,2015-11-11 14:00:00 UTC,2027-1477408502.282-80.194.77.100-1361,8bc0c1dd-4842-4283-a123-5671a34b67eb
+      id,action,temp_payload,payload,user_uid,created_at,updated_at,request_id,content_id
+      10,Publish,"{""content_id"":""2cc503f1-5fef-4fac-8381-63f6689cd6a2""}","{""content_id"":""2cc503f1-5fef-4fac-8381-63f6689cd6a2""}",,2015-11-11 09:55:00 UTC,2015-11-11 09:55:00 UTC,,2cc503f1-5fef-4fac-8381-63f6689cd6a2
+      11,Publish,"{""content_id"":""8bc0c1dd-4842-4283-a123-5671a34b67eb""}","{""content_id"":""8bc0c1dd-4842-4283-a123-5671a34b67eb""}",e676cb22-0c0c-4534-9514-a65ebcc7a9e2,2015-11-11 14:00:00 UTC,2015-11-11 14:00:00 UTC,2027-1477408502.282-80.194.77.100-1361,8bc0c1dd-4842-4283-a123-5671a34b67eb
```
thomasleese: Do we need to add `temp_payload` to the CSV file? Because that will break importing the existing CSV files while we have this field. It seems like because we do [`event.update!`](https://github.com/alphagov/publishing-api/blob/e542e87669a9ab1a71f5f45c4de002b746dbb332/lib/events/s3_importer.rb#L10) the `before_save` should meant that `temp_payload` gets assigned anyway.

MuriloDalRi: I _think_ I fixed that now!

MuriloDalRi:
> This looks good to me! Have you tried it out in integration to see how long the migration takes?

Not yet but I'll do that today. I'll keep this PR open at least until we have the second PR ready (to backfill the new columns). Adding this temporary column will probably make the performance even worse until we finish doing the migration and clean up everything, so I want to be very careful.

lib/tasks/backfill_new_jsonb_columns.rake
```diff
@@ -0,0 +1,30 @@
...
+  AccessLimit.find_each.with_index do |access_limit, index|
+    access_limit.update(temp_users: access_limit.users,
+                        temp_organisations: access_limit.organisations)
+    puts "Updated #{(index + 1)} temp_users and temp_organisations"
+  end
+
+  Edition.find_each.with_index do |edition, index|
+    edition.update(temp_details: edition.details.delete("\u0000"),
+                   temp_routes: edition.routes.delete("\u0000"),
+                   temp_redirects: edition.redirects.delete("\u0000"))
```
thomasleese: Did we ever find out where these came from? Does calling `.delete` work on the `Hash`? I imagined we might need to do it on the JSON string.

MuriloDalRi: We couldn't figure out how that got there... the .delete didn't work, we'll have to change the rake task

spec/lib/events/s3_importer_spec.rb
thomasleese: I think this might have come back in a rebase!

MuriloDalRi: Urrgh yeah I'll get rid of it

lib/tasks/backfill_new_jsonb_columns.rake
```diff
@@ -0,0 +1,112 @@
...
+  sql = "UPDATE events SET temp_payload = payload WHERE id > #{low} AND id <= #{high};"
+
+  while high < total
+    Event.find_each(start: low, finish: high) do |event|
+      next if event.payload.nil?
+
+      payload = event.payload.to_json.gsub(/\\u0000/, "")
+      payload = JSON.parse(payload).symbolize_keys
+      event.update(payload: payload)
+    end
```
thomasleese, line 45: Do we still need to do this? I thought it was only one `Edition` that had a null byte in the string?

MuriloDalRi: There's at least one event with null bytes (id 31148138)

thomasleese: Hmm... I wonder if that event is the event that caused the edition with the null byte.

I think it would be better to deal with these similar to how we dealt with the edition with a special case, because then it's easier to see exactly what data we're changing and there is less of a risk of changing something accidentally. (Plus it would make the task a slightly bit faster).

MuriloDalRi: I've changed this now so we only remove the one event and edition

thomasleese:
One thing I've just noticed is that these new columns don't have the same `default` set as the old ones. We probably want to do that in this PR as changing the default later can sometimes lock tables as well I think.

PR #1765, opened 2020-04-07
Switch columns over

MuriloDalRi:
Both the new JSONB and the old JSON columns should match now.

[Trello](https://trello.com/c/S7iyCLbj/1830-3-pub-api-jsonb-migration-pt-iii-switch-to-new-columns)

### DIFF
--- a/db/migrate/20200406130817_rename_json_columns.rb
+++ b/db/migrate/20200406130817_rename_json_columns.rb
@@ -1,0 +1,42 @@
+class RenameJsonColumns < ActiveRecord::Migration[5.2]
+  def change
+    change_table :access_limits, bulk: true do |t|
+      t.rename :users, :old_users
+      t.rename :organisations, :old_organisations
+      t.rename :temp_users, :users
+      t.rename :temp_organisations, :organisations
+      t.rename :old_users, :temp_users
+      t.rename :old_organisations, :temp_organisations
+    end
+
+    change_table :editions, bulk: true do |t|
+      t.rename :details, :old_details
+      t.rename :routes, :old_routes
+      t.rename :redirects, :old_redirects
+      t.rename :temp_details, :details
+      t.rename :temp_routes, :routes
+      t.rename :temp_redirects, :redirects
+      t.rename :old_details, :temp_details
+      t.rename :old_routes, :temp_routes
+      t.rename :old_redirects, :temp_redirects
+    end
+
+    change_table :events, bulk: true do |t|
+      t.rename :payload, :old_payload
+      t.rename :temp_payload, :payload
+      t.rename :old_payload, :temp_payload
+    end
+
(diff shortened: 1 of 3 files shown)


db/migrate/20200406130817_rename_json_columns.rb
```diff
@@ -0,0 +1,42 @@
+class RenameJsonColumns < ActiveRecord::Migration[5.2]
```
barrucadu, line 1: We'll briefly have the old app running with the new database (unless we deploy with a hard restart), will the old app using `next unless column.sql_type == "json"` mean some data won't be saved to the database between this migration running and the old app processes being killed?

thomasleese: Hmm that's a good point, perhaps we should deploy the `next unless column.sql_type == "json"` work in a separate PR first before actually switching over the columns.

barrucadu: What would happen to the old app in that case - would it save data in one column but try to read it from another?

thomasleese: I think it should be fine, that line `next unless column.sql_type == "json"` is only used to automatically symbolise the keys of the `json` (and soon the `jsonb`) columns.

MuriloDalRi: Thanks, @barrucadu, I've moved that to a separate PR: https://github.com/alphagov/publishing-api/pull/1771

spec/lib/events/s3_exporter_spec.rb
```diff
@@ -23,9 +23,10 @@
   end
 
   def build_csv(expected_events)
-    columns = "id,action,payload,user_uid,created_at,updated_at,request_id,content_id\n"
+    columns = "id,action,user_uid,created_at,updated_at,request_id,content_id,payload\n"
     expected_events.inject(columns) do |memo, event|
-      memo << CSV.generate_line([event.id, event.action, event.payload.to_json, event.user_uid, event.created_at, event.updated_at, event.request_id, event.content_id])
+      event.reload
+      memo << CSV.generate_line([event.id, event.action, event.user_uid, event.created_at, event.updated_at, event.request_id, event.content_id, event.payload.to_json])
```
cbaines, line 29: Will these changes cause issues when restoring the CSV files?

thomasleese: My understanding is that it won't cause any issues because the importer parses the fields in any order they come in: https://github.com/alphagov/publishing-api/blob/4028a4e690d1ac5b1ca0b9fce268b2f0bd892d03/lib/events/s3_importer.rb#L37-L43

cbaines: Good good :+1: I think it's rare that we import data from S3 anyway.

PR #1782, opened 2020-04-28
Fix rake task

MuriloDalRi:
The original rake task didn't copy all the data over to the new columns

[Trello](https://trello.com/c/IJ6dRZyw/1829-8-pub-api-jsonb-migration-pt-ii-backfill-new-columns)

### DIFF
--- a/lib/tasks/backfill_new_jsonb_columns.rake
+++ b/lib/tasks/backfill_new_jsonb_columns.rake
@@ -1,91 +1,26 @@
-desc "Migrate access_limits json columns"
-task access_limits: :environment do
-  puts "starting access limits"
-  acc_lim_sql = "UPDATE access_limits SET temp_users = users, temp_organisations = organisations;"
-  ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(acc_lim_sql) }
-end
-
-desc "Migrate editions json columns"
-task editions: :environment do
-  record = Edition.find_by(id: 704530)
-  details = record.details.to_json.gsub(/\\u0000/, "")
-  details = JSON.parse(details).symbolize_keys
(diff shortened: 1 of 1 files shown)


lib/tasks/backfill_new_jsonb_columns.rake
```diff
@@ -1,91 +1,26 @@
...
-  event.update(payload: payload)
-
-  low = 0
-  high = 1000
-  total = Event.maximum("id")
-
-  sql = "UPDATE events SET temp_payload = payload WHERE id > #{low} AND id <= #{high};"
+desc "Backfill new jsonb columns with data from original json columns"
+task backfill_new_jsonb_columns: :environment do
+  sql = "UPDATE editions SET temp_details = details, temp_routes = routes, temp_redirects = redirects WHERE id in (SELECT id FROM editions WHERE temp_details IS NULL LIMIT 100);"
```
barrucadu, line 3: Does the nested query need to be `where temp_details is null or temp_routes is null or temp_redirects is null`, or are they only non-null in the case `temp_details` is?

lib/tasks/backfill_new_jsonb_columns.rake
```diff
@@ -1,91 +1,26 @@
...
 
-  while high < total
+  num = (Edition.count / 100) + 1
+  num.times do |i|
     ActiveRecord::Base.connection_pool.with_connection { |con| con.exec_query(sql) }
```
barrucadu, line 7: Are newly created / updated editions correct?  Is it only old editions which have the problem?

MuriloDalRi: It seems like all the new ones are fine

PR #1827, opened 2020-07-08
Automatically discard path reservations for unused drafts

kevindew:
This introduces the automatic removal of PathReservation models when drafts are either discarded or have their base_path changed. This has been introduced to replace a manual process where developers need to remove PathReservation models manually when a publisher tries to create a draft of a document in Content Publisher that they previously discarded in Whitehall.

The approach taken here is to assume that editing the base_path of a draft edition or discarding a draft edition is an implicit giving up of a path reservation when it is not also used for a live edition. This is to reflect that almost all path reservations are created implicitly as part of a put content rather than explicitly via a call to a specific endpoint.

The motivation for adding this code to Publishing API was because I didn't want to create similar code in a number of publishing applications. We currently have [code to manage this](https://github.com/alphagov/content-publisher/blob/d41fabb5ce00f137ed6b409c130c51c499f91b9f/app/services/discard_draft_edition_service.rb#L14) in Content Publisher and would need similar in Whitehall, which I [initially coded](https://github.com/alphagov/whitehall/tree/delete-path-reservations-on-discard-draft) before deciding it would be better placed in Publishing API.

## The good

- Support should no longer receive tickets being unable to move working on content from Whitehall to Content Publisher (and vice-versa)
- Can remove code from Content Publisher
- Handling this in Publishing API can be done simpler than publishing apps as it can be done implicitly in existing requests whereas publishing apps need new requests for this
- We'll stop building up such a big pile of URLs that are reserved and never going to be used (currently have 35,000 reserved without a corresponding edition)

## The bad

- This complicates the path reservation system further. The system itself is confusing because it can be populated both implicitly (through put content) or explicitly (through path reservation endpoint) without distinguishing between them. This change thus makes an implicit assumption that discarding means no longer wanting a path which could be surprising if you're using path reservations explicitly.

## The ugly

- The tweaking the command classes and their tests was pretty hairy, it's all rather fragmented. I did what I could without wanting to do any wider rewriting.

More info in the commits.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -4,7 +4,9 @@ module Commands
       def call
         PutContentValidator.new(payload, self).validate
 
-        prepare_content_with_base_path
+        remove_previous_path_reservations
+        reserve_current_path
+        clear_draft_items_of_same_locale_and_base_path
         check_update_type
 
         update_root_payload_with_auth_bypass_ids
@@ -42,11 +44,21 @@ module Commands
         old_links - new_links
       end
 
-      def prepare_content_with_base_path
+      def remove_previous_path_reservations
+        to_discard = previous_drafted_edition&.base_path
+        return if to_discard.blank? || to_discard == payload[:base_path]
+        return if Edition.exists?(base_path: to_discard,
+                                  content_store: :live,
+                                  publishing_app: payload[:publishing_app])
+
+        PathReservation.where(base_path: to_discard, publishing_app: payload[:publishing_app])
+                       .delete_all
+      end
+
+      def reserve_current_path
         return unless payload[:base_path]
 
         PathReservation.reserve_base_path!(payload[:base_path], payload[:publishing_app])
-        clear_draft_items_of_same_locale_and_base_path
       end
 
       def update_content_dependencies(edition)
@@ -143,6 +155,8 @@ module Commands
       end
 
       def clear_draft_items_of_same_locale_and_base_path
+        return unless payload[:base_path]
+
         SubstitutionHelper.clear!(
           new_item_document_type: payload[:document_type],
           new_item_content_id: document.content_id,
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -49,8 +49,8 @@ module Commands
       end
 
       def delete_supporting_objects
-        AccessLimit.find_by(edition: draft).try(:destroy)
-        ChangeNote.where(edition: draft).destroy_all
+        AccessLimit.where(edition: draft).delete_all
+        ChangeNote.where(edition: draft).delete_all
       end
 
       def increment_lock_version
--- a/app/commands/v2/discard_draft.rb
+++ b/app/commands/v2/discard_draft.rb
@@ -7,26 +7,22 @@ module Commands
         check_version_and_raise_if_conflicting(document, payload[:previous_version])
 
         save_document_type
-        delete_supporting_objects(document.draft)
+        delete_supporting_objects
         delete_draft_from_database
         increment_lock_version
 
-        after_transaction_commit do
-          downstream_discard_draft(
-            document.draft.base_path,
-            document.content_id,
-            document.locale,
-          )
-        end
+        after_transaction_commit { downstream_discard_draft }
 
-        Action.create_discard_draft_action(document.draft, document.locale, event)
-        Success.new(content_id: document.content_id)
+        Action.create_discard_draft_action(draft, locale, event)
+        Success.new(content_id: content_id)
       end
 
     private
 
+      delegate :draft, :content_id, :locale, to: :document
+
       def raise_error_if_missing_draft!
-        return if document.draft.present?
+        return if draft.present?
 
         code = document.published_or_unpublished.present? ? 422 : 404
         message = "There is not a draft edition of this document to discard"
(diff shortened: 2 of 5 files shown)


app/commands/v2/put_content.rb
```diff
@@ -42,11 +44,19 @@ def link_diff_between(old_links, new_links)
         old_links - new_links
       end
 
-      def prepare_content_with_base_path
+      def remove_previous_path_reservations
+        to_discard = previous_drafted_edition&.base_path
+        return if to_discard.blank? || to_discard == payload[:base_path]
+        return if to_discard == previously_published_edition.previous_base_path
```
benthorner, line 50: I found this a bit hard to understand, since there may be other historical paths we care about. Are we really asking `will_redirect?`

kevindew: The question isn't whether it will redirect it's asking is the path already in use on published content. Your question has prompted me to wonder whether I've done this right though because there are scenarios where this could discard a path reservation still used by an app (for example you create a new document to replace an unpublished one) so it might be better to instead query where there is a live content item using that base_path.

kevindew:
> Nice one - I think this is a positive change, and we can always deal with the implicit/explicit ambiguity if it becomes a problem. The code looks good.

Yeah I'm expecting it's used sufficiently infrequently that it may never catch someone out.

> I'm wondering about the relationship between the path reservations and the routes. Will the discarded paths still be present in the router?

Interesting question. Having a bit of a look, it seems Content Store [clean ups routes when a draft is discarded](https://github.com/alphagov/content-store/blob/9b099ad6fcd6f366bed0ba70285a0d021dcb2c85/app/controllers/content_items_controller.rb#L54), however it looks like the process to [replace routes when content is updated](https://github.com/alphagov/content-store/blob/9b099ad6fcd6f366bed0ba70285a0d021dcb2c85/app/models/content_item.rb#L201) does no deleting (and there'd be none if a content item changed from having a base_path to not) so it seems very possible there are a quantity of ghost routes in router-api where content has been replaced on both draft/live stacks. These shouldn't cause any problems to apps that can handle the content store returning 404 responses, but is a bit of a smell - one of many in this area of the stack.

I consider this beyond the scope of this PR by the way.

kevindew:
@benthorner I've pushed a change for this regarding live editions if you wouldn't mind giving it another look.

app/commands/v2/put_content.rb
benthorner: I'm still not sure I understand. While the new code is self-explanatory (thanks!), I'm still wondering about the behaviour of this with respect to redirects on previously live (superseded) editions: this looks like [it will discard them](https://github.com/alphagov/publishing-api/blob/master/app/models/edition.rb#L165) (and therefore allow them to be overridden)? Perhaps we should increase the scope to be "NOT content_store = 'draft'".

benthorner:
@kevindew thanks for investigating. My question was broad, but really targeted towards the behaviour of previous redirects. I've posed this in my inline comment.

> the process to replace routes when content is updated does no deleting (and there'd be none if a content item changed from having a base_path to not)

Would this be partially covered [by the RedirectService](https://github.com/alphagov/publishing-api/blob/master/app/services/redirect_service.rb)? I think the only problematic scenario here is "if a content item changed from having a base_path to not", but like you say, it's unlikely to have any user impact.

app/commands/v2/put_content.rb
kevindew: Gotcha, I don't think it will prove problematic - I'll try explain my reasoning. Once any redirects related to a superseding are live, they will have the base_path and same publishing app of the content that was superseded, thus still meet the criteria in the query. In theory, and assuming everything now and previously works properly, every base_path ever publicly used on GOV.UK should have an Edition with a content_store of live in some state.

I don't think it should go to the degree of asking NOT content_store draft, as that means the past can interfere with the process in subtle, surprising ways (for example you happen to put content an old path that was once used in 2016 and has nothing else and this process fails). In general, I also think if we end up with a base_path not marked as being on live content store then it's effectively historical and fair game to be taken.

Does that help at all?

kevindew:
> @kevindew thanks for investigating. My question was broad, but really targeted towards the behaviour of previous redirects. I've posed this in my inline comment.
> 
> > the process to replace routes when content is updated does no deleting (and there'd be none if a content item changed from having a base_path to not)
> 
> Would this be partially covered [by the RedirectService](https://github.com/alphagov/publishing-api/blob/master/app/services/redirect_service.rb)? I think the only problematic scenario here is "if a content item changed from having a base_path to not", but like you say, it's unlikely to have any user impact.

Yup, that's correct. The situation I was latching on to was that a content item can have any number of routes under the base path and they aren't cleaned up.

So say you have travel advice of Poland which is published with 3 routes `/foreign-travel-advice/poland`, /foreign-travel-advice/poland.atom` and `/foreign-travel-advice/poland/print`. Then if you replaced that base_path there appears to be nothing that would ever delete those atom and print routes 😬  - but that seems like something not to worry about right now.

app/commands/v2/put_content.rb
benthorner: > they will have the base_path and same publishing app of the content that was superseded

Ah, that's good to know. Perhaps there's still a loophole here for a chain of redirects, but that also seems (a) unlikely and (b) low impact; at worst, we might lead people to mildly irrelevant content.

PR #1833, opened 2020-07-14
Add Dependabot configuration

ChrisBAshton:
In accordance with [RFC 126], we're configuring Dependabot to only
raise Pull Requests for the following three categories:

1. Security updates
2. Internal libraries
3. Framework libraries

Things have moved on slightly since the RFC; we're now using
[GitHub native] configs rather than the legacy `.dependabot`
approach. The syntax is slightly different to the example config
provided in the RFC but most of it is easily transferable.

We've handled (2) and (3) in config, and have set it to only raise
PRs for direct updates, i.e. updates to subdependencies of the
named dependencies will not raise PRs.

We've chosen a `govuk*` wildcard to cover things like
`gds-api-adapters` and `govuk_app_config` and other unknown future
dependencies. Having `govuk` as a [suffix is not encouraged][wildcards]
so instead of a `*govuk` wildcard we've named `rubocop-govuk`
explicitly. We explicitly name dependencies beginning with `gds`
[for the same reason][discussion].

Trello: https://trello.com/c/uPoriyfJ/2049-add-dependabot-configuration-to-each-repo-blitz-pair

[discussion]: https://github.com/alphagov/content-store/pull/750#discussion_r454208436
[GitHub native]: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates
[security]: https://github.community/t/how-to-get-dependabot-to-trigger-for-security-updates-only/117257/5
[RFC 126]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md
[wildcards]: https://github.com/alphagov/content-store/pull/750#discussion_r453693082

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,44 @@
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: /
+    schedule:
+      interval: daily
+    allow:
+      # Security updates
+      - dependency-name: brakeman
+        dependency-type: direct
+      # Internal gems
+      - dependency-name: "govuk*"
+        dependency-type: direct
+      - dependency-name: gds-sso
+        dependency-type: direct
+      - dependency-name: govspeak
+        dependency-type: direct
+      - dependency-name: plek
+        dependency-type: direct
+      - dependency-name: rubocop-govuk
+        dependency-type: direct
+      # Framework gems
+      - dependency-name: factory_bot_rails
+        dependency-type: direct
+      - dependency-name: pg
+        dependency-type: direct
+      - dependency-name: rails
+        dependency-type: direct
+      - dependency-name: rspec
+        dependency-type: direct
+      - dependency-name: rspec-rails
+        dependency-type: direct
+
+  # Ruby needs to be upgraded manually in multiple places, so cannot
+  # be upgraded by Dependabot. That effectively makes the below
+  # config redundant, as ruby is the only updatable thing in the
+  # Dockerfile, although this may change in the future. We hope this
+  # config will save a dev from trying to upgrade ruby via Dependabot.
+  - package-ecosystem: docker
+    ignore:
+      - dependency-name: ruby
+    directory: /
+    schedule:
+      interval: weekly


.github/dependabot.yml
```diff
@@ -0,0 +1,38 @@
...
+      - dependency-name: "rubocop-govuk"
+        dependency-type: direct
+      - dependency-name: "gds-sso"
+        dependency-type: direct
+      - dependency-name: govspeak
+        dependency-type: direct
+      - dependency-name: plek
+        dependency-type: direct
+      - dependency-name: pg
+        dependency-type: direct
```
kevindew, line 23: I don't think you meant postgres to be an internal gem.

.github/dependabot.yml
```diff
@@ -0,0 +1,38 @@
...
+      # Framework gems
+      - dependency-name: factory_bot_rails
+        dependency-type: direct
+      - dependency-name: rails
+        dependency-type: direct
+      - dependency-name: rspec
+        dependency-type: direct
+      - dependency-name: rspec-rails
+        dependency-type: direct
+  - package-ecosystem: docker
```
kevindew, line 33: What does this do? I hadn't noticed it before.

.github/dependabot.yml
ChrisBAshton: I just realised this and was going back to change it - good spot! Will move the comment.

kevindew: Ah ha, sorry - you nearly got away with it.

.github/dependabot.yml
ChrisBAshton: This has Dependabot checking Dockerfile for dependencies. There's not a lot of info about it on the [GitHub native Dependabot docs](https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates), but the [release notes from 2017](https://dependabot.com/blog/dependabot-now-supports-docker/) suggest that it looks at the `FROM` field in the Dockerfile and checks for newer versions.

I've been going on the basis that if a repo has a Dockerfile, it can't hurt to have a corresponding check in Dependabot. However, we're explicitly ignoring updates to `ruby` as upgrading ruby requires manual intervention, so it's quite possible that this extra config is giving us nothing. I could omit the Docker config altogether...?

thomasleese: I preferred the idea of leaving it in to make it more explicit (if it's not there, someone might think one day we could enable Docker upgrades only to discover that it doesn't really work as we have the Ruby version stored in multiple places).

I was also thinking that in the future there might be some other way we use Docker and it might be nice having a pre-configured configuration file... but that's just anticipation...

kevindew: Oh I see, so we're actively turning this on in order to turn it off? ie, we expect the only dependency it'll manage is ruby (as it's the only thing with a FROM) and so we're telling it not to do that?

Am i understanding correctly there? if so do we need the config for directory and schedule?

ChrisBAshton: That's correct. And yes, I think [directory and schedule are required fields](https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates).

kevindew: Sure, could we add a comment to explain at all? it feels like rather niche information.

ChrisBAshton: Done 👍 could you check it makes sense?

PR #1853, opened 2020-09-21
Better facilitate changing the locale of published content

cbaines:
Similar to 8471eba4553564a489f074b1b87fb39534003486, this commit aims
to make it possible to take some content, and change it's
locale. Because the locale is part of the document in the Publishing
API, this it turns out was rather difficult.

With this commit, it should be possible to publish a draft, where it
has a different locale to some content which shares the content id and
base path.

This is motivated by some changes in Publisher to allow changing the
locale, but should help other publishing apps as well.

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/publishing-api), after merging.

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -18,6 +18,7 @@ module Commands
         unless edition.pathless?
           redirect_old_base_path
           clear_published_items_of_same_locale_and_base_path
+          clear_published_item_of_different_locale_but_matching_base_path
         end
 
         set_publishing_request_id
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Makes sense to me

PR #1855, opened 2020-09-22
Add rake task to allow updating of country names in specialist documents

Tetrino:
Turns out when the work was put in to allow updating of country names in
our database, we left a large blindspot in specialist publications. This
was found when it was reported that North Macedonia had not been set as
the correct name for the country for more than two years after the name
change.

This rake task allows the relevant documents to be updated when needed,
and should be done after changes made to relevant files in
govuk-content-schemas and specialist-publisher as seen in these two
pull requests:

https://github.com/alphagov/govuk-content-schemas/pull/1011
https://github.com/alphagov/specialist-publisher/pull/1714

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/publishing-api), after merging.

### DIFF
--- a/lib/tasks/update_specialist_publisher_countries.rake
+++ b/lib/tasks/update_specialist_publisher_countries.rake
@@ -1,0 +1,35 @@
+namespace :content do
+  #See git commit log on this comment for details
+  desc "Updates a country name for specialist publications"
+  task :update_specialist_publication_countries, %i[old_country new_country] => :environment do |_, args|
+    if args.old_country.blank? || args.new_country.blank?
+      raise ArgumentError, "This task takes two arguments: the current country name and the new country name."
+    end
+
+    health = Edition.where(document_type: 'export_health_certificate')
+    puts "Attempting to update type: export_health_certificate"
+    filt_health = health.select { |h| h.details[:metadata][:destination_country].present? && h.details[:metadata][:destination_country].include?(args.old_country) }
+    filt_health.each do |doc|
+      puts "Processing #{doc.title}"
+      details = doc.details
+      met = details[:metadata]
+      met[:destination_country].map! { |c| c == args.old_country ? args.new_country : c }
+      details[:metadata] = met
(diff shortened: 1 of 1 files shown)


benthorner:
This looks to be a regular kind of task that we probably want [to document](https://docs.publishing.service.gov.uk/manual/rename-a-country.html) and have some tests around. Looking at our general approach for renaming countries, it would be more consistent (and probably easier!) to implement this in Specialist Publisher itself, since it knows the most about the change. Something like

```
ResearchForDevelopmentOutput.all.each do |doc|
   next unless doc.country == "macedonia"
   doc.country = "north-macedonia"
   doc.save
   Republisher.republish_one(doc.content_id)
end
```

Tetrino:
Caught this by chance on the train! - Documentation was going to come when I got back, as for the reason I did it in publishing API, I caught somewhere (possibly the comments one of the rake tasks in specialist publisher itself) that republishing affects the update times for the documents which in turn can cause issues down the line in P-API.

benthorner:
Hmm, that's interesting. I can see the comment you're talking about [here](https://github.com/alphagov/specialist-publisher/blob/master/lib/tasks/republish.rake#L9). However, it looks like the actual code [has since been used](https://github.com/alphagov/specialist-publisher/commits/master/app/services/republish_service.rb), so it may not be so dodgy. I've added [a task to do the renaming in Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/1722), but realise I should have investigated your comment more first. I'll try republishing some documents and see if that leads to any issues. Hopefully it will be possible to locate these tasks in Specialist Publisher 🤞.

PR #1863, opened 2020-10-15
Run rubocop, spec and verify:pact as default task

kevindew:
Trello: https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

This task is used by CI to build the app. To reflect this I've removed
the pact verify stage from the Jenkins stages.

This turns on linting that had accidentally been switched off following https://github.com/alphagov/govuk-jenkinslib/pull/74 it also includes some other minor clean ups of files related to Jenkins and Rake. See commits for more details

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,12 +20,13 @@ node("postgresql-9.6") {
     beforeTest: {
       setExtraEnvVars(govuk);
     },
+
     publishingE2ETests: true,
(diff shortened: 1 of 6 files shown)


ChrisBAshton (approved):
Thanks for the detailed commits 👍

PR #1884, opened 2020-12-22
Add rake tasks to export CSVs of publishings and unpublishings 

kevindew:
Trello: https://trello.com/c/P3fpshLs/1010-add-publishing-api-rake-task-for-new-year

These tasks provide the means to create a CSV report of editions that
have been published or unpublished on GOV.UK between the dates provided.
They output CSV directly to stdout. The expectation is that a developer
will retrieve the CSV by copy and pasting from a plain text Jenkins log of
the run.

These tasks were added shortly before 1st of January 2021 with the
expectation that they would be needed to produce reports on the content
that changed for the end of the transition period. They are added
somewhat speculatively. I did wonder if Content Data could meet this need,
but I couldn't find the functionality.

The brief I had for this was to create a task that produced a report to
show which pages were created, updated or removed. However, combining
removed with created and updated is a somewhat complex process that
isn't fully logical (in some cases the content is still there) and follows
a different date system. Therefore I decided to separate this into two
tasks: publishings and unpublishings and include more specific data in each
one.

The publishings task deliberately excludes "republish" update_types to
exclude when content isn't changed by a publisher.

The unpublishings task excludes the automatic unpublishing type of
"substitute" which is used when a redirect or gone is automatically
replaced. The unpublishing modelling in Publishing API doesn't keep a
log of all changes so the output of this task isn't fully conclusive if
a user has edited or changed an unpublishing type.

Examples:

Running `bundle exec rake "date_range:publishings['2020-12-1 00:00',
'2020-12-22 00:00']"`

```
published_at,base_path,content_id,locale,title,document_type,update_type,first_publishing
2020-12-01 00:15:04 UTC,/government/news/clinical-evaluation-confirms-accuracy-of-lamp-test,ed7492fd-e438-4c71-a7a5-fd4805c432bb,en,Clinical evaluation confirms accuracy of LAMP test,press_release,major,true
2020-12-01 00:15:12 UTC,/government/publications/evaluation-of-the-early-years-local-government-programme,c9494e09-597c-4091-b235-73902669198d,en,Evaluation of the early years local government programme,research,major,true
2020-12-01 00:15:12 UTC,/government/publications/making-a-claim-to-an-employment-tribunal-t420,f1092bc4-20b7-4355-809e-e36200f4fd76,en,Making a claim to an employment tribunal (T420),guidance,major,false
```

Running `bundle exec rake "date_range:unpublishings['2020-12-1 00:00',
'2020-12-22 00:00']"`

```
unpublished_at,base_path,content_id,locale,title,document_type,unpublishing_type
2020-12-01 00:15:12 UTC,/government/publications/visitor-visa-guide-to-supporting-documents/visitor-supporting-documents-guide,db319240-9fe7-496a-83a3-9678b1cd7285,en,Visitor: supporting documents guide 访问者: 支持材料 指南 (Chinese translation),html_publication,redirect
2020-12-01 08:18:30 UTC,/government/publications/philippines-list-of-medical-practitioners,82d48f7b-7280-43be-9feb-91876b946067,en,Philippines - list of medical practitioners,guidance,redirect
2020-12-01 08:37:11 UTC,/government/publications/emissions-to-air-from-large-combustion-plant-in-the-transitional-national-plan-rps-c18,6939791f-b048-4a84-af00-b5d70dfd8204,en,Emissions to air from Large Combustion Plant in the Transitional National Plan: RPS C18,guidance,withdrawal
```


AlanGabbianelli (commented):
Hey @kevindew nicely done on this. I particularly like that you updated the docs, added clear tests for the two new tasks and wrote comprehensive commit messages. ⭐

I only have a couple of comments:

- It would be useful to explicitly state in the docs whether the two required timestamps are inclusive or exclusive to avoid needing to find that info in the code itself. Currently, the `from_time` is included and the `until_time` is excluded, maybe that could be changed to include both so in the docs we could simply have "[...] between two times (inclusive) run:"?

- I'm not convinced by the `date_range` filename and namespace. It doesn't seem to fully convey what the tasks are for. It would be useful to include terms like "csv report" to convey the reporting nature of these tasks. Something along the lines of `date_range_csv_report` or `csv_report_in_date_range` perhaps?

kevindew:
> Hey @kevindew nicely done on this. I particularly like that you updated the docs, added clear tests for the two new tasks and wrote comprehensive commit messages. ⭐
> 
> I only have a couple of comments:
> 
> * It would be useful to explicitly state in the docs whether the two required timestamps are inclusive or exclusive to avoid needing to find that info in the code itself. Currently, the `from_time` is included and the `until_time` is excluded, maybe that could be changed to include both so in the docs we could simply have "[...] between two times (inclusive) run:"?
> * I'm not convinced by the `date_range` filename and namespace. It doesn't seem to fully convey what the tasks are for. It would be useful to include terms like "csv report" to convey the reporting nature of these tasks. Something along the lines of `date_range_csv_report` or `csv_report_in_date_range` perhaps?

Thanks for the review Alan, I've updated it to address your comments. It was a good suggestion about the namespace. I've switched to using a `csv_reports` one to group reports.

AlanGabbianelli:
Good choice @kevindew, I like how both `csv_report:unpublishings_by_date_range` and `csv_report:publishings_by_date_range` read.

Do you have any thoughts about my other suggestion to be more explicit about the date range in the docs?

kevindew:
> Good choice @kevindew, I like how both `csv_report:unpublishings_by_date_range` and `csv_report:publishings_by_date_range` read.
> 
> Do you have any thoughts about my other suggestion to be more explicit about the date range in the docs?

Thanks. Yup, I amended the docs to cover that, you can see changes here: https://github.com/alphagov/publishing-api/compare/3a072d30f2d7a2a8b410342b59962f385b683c22..d09cbbe66f93b07f1d3e6bbc0b775f37b65833fb but main change was adding: "Both task require a `from` timestamp argument and a `until` timestamp argument."

AlanGabbianelli:
What do think about making both inclusive (`from_time..until_time` vs `from_time...until_time`) and stating this in the docs?

kevindew:
> What do think about making both inclusive (`from_time..until_time` vs `from_time...until_time`) and stating this in the docs?

I think it's better to be exclusive. Doing inclusive with time is quite confusing as you'd have exactly the time on the milisecond included but nothing later and it would mean that it is possible for consecutive reports to include the same data (should you be (un)lucky enough to have a event fall on the exact time at 0 milliseconds). I don't actually think it's worth documenting about inclusive/exclusive aspect as it is such a muddle to consider and this would work implicitly as you'd expect it to.

AlanGabbianelli:
I see what you mean. If you created two reports, for example, one from "2020-12-30 12:00:00" to "2020-12-30 13:00:00" and one from "2020-12-30 13:00:00" to "2020-12-30 14:00:00", you'd potentially have some repeated data using inclusive timestamps, but no repeated data using inclusive `from_time` and exclusive `until_time`. That's fair, thanks for clarifying.

docs/admin-tasks.md
```diff
@@ -95,3 +95,21 @@ bundle exec rake expanded_links:populate_by_document_type['document-type']
 ```
 bundle exec rake expanded_links:truncate
 ```
+
+## Generating CSV reports of publishings and unpublishing by date range
+
+There are two tasks provided which to create CSV reports of publishings and
```
AlanGabbianelli, line 101:
```suggestion
There are two tasks provided which create CSV reports of publishings and
```

kevindew: Thank you, I've amended.

AlanGabbianelli (approved):

PR #1891, opened 2021-01-11
Fix more Abort errors

1pretz1:
This builds upon the work done recently [1] to reduce the amount of
Abort errors [2] we've been seeing for the downstream draft and live
workers.

On further investigation we've found that other orphaned content ids can
raise the same error under different circumstances (all seem to stem
from the patch_links endpoint). Below are some scenarios where this
could happen:

1) First draft document is added and removed (whitehall) from a live
   document collection. This causes an orphan content id to be
   represented downstream live but then only to fail as the orphaned
   document only resides in the draft content store.

2) First draft edition deleted (whitehall) and then the draft is removed
   from the owning document collection. This leaves the draft document
   with no editions in Publishing API which raises the error for the
   draft & live downstream workers.

3) Foreign translation edition is unpublished (publisher) which sets the
   `content_store` field to nil for that edition in Publishing API. This
   results in errors for live and draft downstream workers.

For the errors I've investigated there doesn't seem to be actually any
bugs in the system as the code is working as intended but the lack of
filtering on the orphaned content ids is causing a lot of indirect
noise.

I'm sure there are more scenarios where this error is raised but for the
specific errors I've looked into then the simplest solution is to simply
check for an edition for the given content store which mitigates the
error for all of the above scenarios.

[1]: https://github.com/alphagov/publishing-api/pull/1890
[2]: https://github.com/alphagov/publishing-api/blob/master/app/workers/downstream_draft_worker.rb#L25-L27

Trello:
https://trello.com/c/SeUUIIVK/2259-5-investigate-and-fix-abortworkererror-sidekiq-downstreamworker

### DIFF
--- a/app/workers/dependency_resolution_worker.rb
+++ b/app/workers/dependency_resolution_worker.rb
@@ -13,11 +13,6 @@ class DependencyResolutionWorker
       send_downstream(content_id, locale)
     end
 
-    orphaned_content_ids_for_locale = Document.where(
-      content_id: orphaned_content_ids,
-      locale: locale,
-    ).pluck(:content_id)
-
     orphaned_content_ids_for_locale.each { |content_id| send_downstream(content_id, locale) }
   end
 
@@ -41,6 +36,20 @@ private
     @source_fields = args.fetch(:source_fields, [])
   end
 
+  def orphaned_content_ids_for_locale
+    Document
+      .distinct
+      .joins(:editions)
+      .where(editions: { content_store: content_stores },
+             content_id: orphaned_content_ids,
+             locale: locale)
+      .pluck(:content_id)
+  end
+
+  def content_stores
+    draft? ? %w[draft live] : %w[live]
+  end
+
   def send_source_stats
     prefix = "dependency_resolution.source"
 
@@ -62,7 +71,7 @@ private
     Queries::ContentDependencies.new(
       content_id: content_id,
       locale: locale,
-      content_stores: draft? ? %w[draft live] : %w[live],
+      content_stores: content_stores,
     ).call
   end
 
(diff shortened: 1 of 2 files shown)


thomasleese (approved):

app/workers/dependency_resolution_worker.rb
```diff
@@ -41,6 +36,20 @@ def assign_attributes(args)
     @source_fields = args.fetch(:source_fields, [])
   end
 
+  def orphaned_content_ids_for_locale
+    Document
+      .distinct
```
ChrisBAshton, line 41: Can you explain why you're using `distinct` here? I've done some searching and it looks like `.distinct` is [usually applied _after_ a query](https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/QueryMethods.html#method-i-distinct). I'm sure it's just something I've not come across before :)

ChrisBAshton (approved):
LGTM 👍 again, excellent investigative work!

I think you can definitely consider https://trello.com/c/SeUUIIVK/2259-5-investigate-and-fix-abortworkererror-sidekiq-downstreamworker closed once this is merged 🎉

app/workers/dependency_resolution_worker.rb
1pretz1: Sure thing. I'm using `distinct` so that there are no duplicates returned in the result set (`distinct` being ActiveRecord's answer to [SQL distinct](https://www.w3schools.com/sql/sql_distinct.asp)). I noticed duplicate content ids may be returned if a live and draft edition are present for a document when querying for the draft content store, e.g. ` .where(editions: { content_store: %w[draft live] }`.

As for placement of the mighty distinct. I usually place it at the start, probably just out of habit much the same as when you write a `DISTINCT` in SQL it's added at the start of the query.

Hope that answers your question 😄

ChrisBAshton: It does - thanks!

PR #1898, opened 2021-02-04
Allow redirects to etl.beis.gov.uk

brucebolt:
Allows withdrawn documents to redirect to `etl.beis.gov.uk`.  This has been approved as a domain we are permitted to redirect documents to.

Trello card: https://trello.com/c/138P3vIk

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -166,7 +166,7 @@ private
     end
 
     def government_domain?(host)
-      host.end_with?(".gov.uk") || host.end_with?(".judiciary.uk")
+      host.end_with?(".gov.uk", ".judiciary.uk", "etl.beis.gov.uk")
     end
 
     def invalid_destination?(destination)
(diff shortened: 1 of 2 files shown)


app/validators/routes_and_redirects_validator.rb
```diff
@@ -166,7 +166,7 @@ def internal?(destination)
     end
 
     def government_domain?(host)
-      host.end_with?(".gov.uk") || host.end_with?(".judiciary.uk")
+      host.end_with?(".gov.uk") || host.end_with?(".judiciary.uk") || host.end_with?("etl.beis.gov.uk")
```
1pretz1, line 169: Minor: looks like you can use `end_with?` with multiple arguments which would make this statement a bit more readable -> https://apidock.com/ruby/String/end_with%3F

1pretz1 (approved):
Looks good Bruce, just one minor, non blocking comment.

app/validators/routes_and_redirects_validator.rb
ChrisBAshton: TIL!

brucebolt: Ooh, thanks, for the pointer.  I'll update to use this.

PR #1899, opened 2021-02-09
Add locale support for Punjabi Pakistan and Punjabi India

edwardkerry:
Adds support for Punjabi Pakistan (pa-pnb)
Adds support for Punjabi India (pa-pan)

[trello](https://trello.com/c/bJZCbzaI/2352-3-add-3-new-languages-to-whitehall)

[government-frontend](https://github.com/alphagov/government-frontend/pull/2009)
[whitehall](https://github.com/alphagov/whitehall/pull/5978)
[govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/pull/1042)
[content-store](https://github.com/alphagov/content-store/pull/807)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -79,6 +79,7 @@ module PublishingAPI
       nl
       no
       pa
+      pa-pan
       pa-pnb
       pa-ur
       pl
--- a/config/application.rb
+++ b/config/application.rb
@@ -79,6 +79,7 @@ module PublishingAPI
       nl
       no
       pa
+      pa-pnb
       pa-ur
       pl
       ps


config/application.rb
```diff
@@ -79,6 +79,8 @@ class Application < Rails::Application
       nl
       no
       pa
+      pa-pan
+      pa-pnb
       pa-ur
```
thomasleese, line 84: Not related to this PR, but I think this should have been `pa-Urdu` according [to the specification](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags):

> An optional script subtag, based on a four-letter script code from ISO 15924 (usually written in Title Case);

thomasleese (commented):
Sorry, I could be wrong, but I think the codes might need changing...

config/application.rb
thomasleese: Reading more I'm a bit confused by this now because I think the introduction of `pa-ur` was supposed to be the Pakistan variant of Punjabi... reading the PR description:

> We already support `pa`, Punjabi Gurmukhi, a language used in Punjab,
> India. This is to add `pa-ur`, Punjabi Shahmukhi, a language used in
> Punjab, Pakistan, that uses the Urdu script.

And reading the section on Wikipedia about Indian/Pakistan variants:

https://en.wikipedia.org/wiki/Punjabi_language#Arabic_and_Persian_influence_on_Punjabi

thomasleese: The Zendesk ticket suggests the same https://govuk.zendesk.com/agent/tickets/4371686 but I think `pa-ur` should have been `pa-pk`.

thomasleese: Just to back up my thoughts on this, I think ideally the code would be something like `pa-Shah` [to represent the script](https://en.wikipedia.org/wiki/Shahmukhi_alphabet) as per the specification:

> An optional script subtag, based on a four-letter script code from ISO 15924 (usually written in Title Case);

However, ISO 15924 doesn't seem to contain an entry for the Shahmukhi script, so instead my feeling is the next best option is to use `pa-PK` as per the specification:

> An optional region subtag based on a two-letter country code from ISO 3166-1 alpha-2 (usually written in upper case), or a three-digit code from UN M.49 for geographical regions;

PR #1901, opened 2021-02-16
Add temporary rake task to fix 'value_of_funding' values

ChrisBAshton:
Dependent on https://github.com/alphagov/govuk-content-schemas/pull/1043

The previous "Up to £100,000" value will now become "£10,001 to
£100,000", in preparation for a new category "Up to £10,000". The
department have confirmed that no current international development
funds are below £10,001, so "10001-to-100000" should be the new
value for these.

Once the Rake task has been executed, we can revert this commit
as the code will no longer be needed.

Trello: https://trello.com/c/eDFwuCFY/2356-5-changes-to-international-development-funding-finder

### DIFF
--- a/lib/tasks/tmp_international_development_funds.rake
+++ b/lib/tasks/tmp_international_development_funds.rake
@@ -2,11 +2,13 @@ namespace :international_development_funds do
   desc "Republishes international development funds with new 'value_of_funding' values"
   task update_value_of_funding: :environment do
     Edition.where(document_type: "international_development_fund").each do |edition|
-      next unless edition.details.dig(:metadata, :value_of_funding) == %w[up-to-100000]
+      value_of_funding = edition.details.dig(:metadata, :value_of_funding) || []
+      next unless value_of_funding.include?("up-to-100000")
 
+      other_values = value_of_funding.reject { |val| val == "up-to-100000" }
       edition.details = edition.details.deep_merge({
         metadata: {
-          value_of_funding: %w[10001-to-100000],
+          value_of_funding: other_values.unshift("10001-to-100000"),
         },
       })
       edition.save!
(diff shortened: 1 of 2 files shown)


lib/tasks/tmp_international_development_funds.rake
```diff
@@ -0,0 +1,17 @@
...
+  desc "Republishes international development funds with new 'value_of_funding' values"
+  task update_value_of_funding: :environment do
+    Edition.where(document_type: "international_development_fund").each do |edition|
+      value_of_funding = edition.details.dig(:metadata, :value_of_funding) || []
+      next unless value_of_funding.include?("up-to-100000")
+
+      other_values = value_of_funding.reject { |val| val == "up-to-100000" }
+      edition.details = edition.details.deep_merge({
+        metadata: {
+          value_of_funding: other_values.unshift("10001-to-100000"),
```
thomasleese, line 11: Not related to your PR, but `unshift` seems like a very poorly named function for what this actually does!

thomasleese (approved):

lib/tasks/tmp_international_development_funds.rake
ChrisBAshton: I've never agreed with the `shift`/`unshift` naming of these methods 😁

ChrisBAshton: ...I enjoyed [this StackOverflow answer](https://stackoverflow.com/questions/6085518/what-is-the-easiest-way-to-push-an-element-to-the-beginning-of-the-array#comment7055991_6085543) too 💩

thomasleese: Haha that's amazing!

PR #1904, opened 2021-02-22
Update rubocop-govuk to 4.0.0.pre.1

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of [trialling the new rubocop-govuk release](https://github.com/alphagov/rubocop-govuk/issues/129). It seems to have gone into this app without too many problems which is good news. 

All changes here are to resolve linter issues, the pin to 4.0.0.pre.1 is intended to be removed once the final release of rubocop-govuk 4.0.0 is out.


barrucadu (approved):
I particularly like the changes to method arguments: changing positional arguments with default values to keyword arguments with default values, and putting required keyword arguments before optional ones.

ChrisBAshton (approved):
LGTM 👍

PR #1929, opened 2021-04-07
Bugfix: Enable exporting events twice in a week

bilbof:
We have a cron job set up to run the rake task
events:export_to_s3 every 7 days.

The task was run on 28 March 2021 and 04 April 2021.

The second run failed with the error:
Events::S3Exporter::EventsExportExistsError: S3 already
has an export for events/2021-02-28T00:00:00+00:00.csv.gz

This is because the default time to fetch events before is
`1.month.ago.beginning_of_week(:sunday)`, which produces the
same date on both March 28 and April 04:

```ruby
(DateTime.new(2021, 03, 28, 04, 38) - 1.month).beginning_of_week(:sunday)
(DateTime.new(2021, 04, 04, 04, 38) - 1.month).beginning_of_week(:sunday)
```

Both of the above commands give the same date: Feb 28 2021.

Changing the period to 30.days (without specifying the
weekday to start from) fixes the issue by ensuring that
running the same command on any day will give a unique
date.

This allows us to run the task multiple times per calendar week.

### DIFF
--- a/lib/tasks/events.rake
+++ b/lib/tasks/events.rake
@@ -1,7 +1,7 @@
 namespace :events do
   desc "Export events to S3. Defaults to the beginning of week 2 months ago, otherwise dates can be specified in a form that Date.parse understands."
   task :export_to_s3, %i[created_before created_on_or_after] => :environment do |_, args|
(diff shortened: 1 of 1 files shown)


brucebolt (approved):

kevindew:
Ha wow great catch, aren't calendars tricky?

PR #1936, opened 2021-04-15
Remove startup.sh and make README consistent with current conventions

ChrisBAshton:
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This removes the startup.sh script as per the conventions outlined in https://github.com/alphagov/govuk-developer-docs/pull/2897 and updates the README to reflect GOV.UK docker as the preferred way to run GOV.UK apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
 # Publishing API
 
-The Publishing API aims to provide _workflow as a service_ so that common
-publishing features can be written once and used by all publishing applications
-across Government. Content can be stored and retrieved using the API and
-workflow actions can be performed, such as creating a new draft or publishing an
-existing piece of content.
+The Publishing API aims to provide _workflow as a service_ so that common publishing features can be written once and used by all publishing applications across Government. Content can be stored and retrieved using the API and workflow actions can be performed, such as creating a new draft or publishing an existing piece of content.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,10 +1,8 @@
 # Publishing API
 
-The Publishing API aims to provide _workflow as a service_ so that common
-publishing features can be written once and used by all publishing applications
-across Government. Content can be stored and retrieved using the API and
-workflow actions can be performed, such as creating a new draft or publishing an
-existing piece of content.
+The Publishing API aims to provide _workflow as a service_ so that common publishing features can be written once and used by all publishing applications across Government. Content can be stored and retrieved using the API and workflow actions can be performed, such as creating a new draft or publishing an existing piece of content.
```
benthorner, line 3: (We've deviated so much from this aim I don't think it's at all true, but definitely not something to fix in this PR!)

benthorner (approved):
Thanks for splitting into small commits, which made this much easier to review ⭐. Just a couple of more minor comments you might want to consider before merging:

- The "Further documentation" heading doesn't render properly for me (appears with `###`). I'm assuming it's just me, but worth checking for special chars or something like that.

- The end of the README links to CONTRIBUTING.md, which isn't really what it claims to be (and is also quite generic). How about moving this to `docs/data-migrations.md`?

thomasleese:
> * The end of the README links to CONTRIBUTING.md, which isn't really what it claims to be (and is also quite generic). How about moving this to `docs/data-migrations.md`?

The reason those docs were part of `CONTRIBUTING.md` was to make sure that people read them when opening PRs. A few years ago, we noticed a lot of people would raise PRs against this repo with database migrations which just perform data migrations (and often wouldn't end up working correctly) which is why we wrote those docs.

ChrisBAshton:
Thanks for the review, Ben.

Makes sense, Tom. I think I agree that the CONTRIBUTING.md is perhaps too specific to data migrations, and that a rethink of approach would be good. I'd suggest if it becomes an issue again, we can look at adding a warning to the PR template.

PR #1939, opened 2021-04-26
Remove legacy /healthcheck route

barrucadu:
govuk-puppet and govuk-aws have now been updated to only use the new
routes, so the old one is no longer needed.

See RFC 141 for more information.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)


erino (approved):

erino:
I think `app/models/healthcheck/queue_latency.rb` and `spec/models/healthcheck/queue_latency_spec.rb` can be deleted too?

PR #1984, opened 2021-10-12
Find final redirect path before presenting to Content Store

ChrisBAshton:
When we send the redirect payload to Content Store, we want to specify the final redirect destination of the content, rather than force multiple redirects on the user. This is to prevent `ERR_TOO_MANY_REDIRECTS` browser errors which prevent users from being able to follow some particularly long redirect chains.

Before this can be merged:

> I've put together a draft PR which does roughly what we're hoping to accomplish here, i.e. find the last redirect in the chain when presenting to Content Store.
>
> It will require manual testing with a real world example such as one of the long redirect chains on this card. I'm conscious that there may be some performance issues with the recursive lookup.
>
> It also only really works for Unpublishing which have one redirect, but `redirects` is an array, which suggests there are cases where there is more than one redirect to consider per Unpublishing.
>
> I also want to add a safety clause to prevent circular redirects, as otherwise we'd never find the 'latest' path to redirect to.
>
> Finally, we'll want to apply this code to existing redirect chains, probably by writing a custom rake task.

Trello: https://trello.com/c/vFuzQYFB/2628-investigate-a-fix-for-long-redirect-chains-in-the-publishing-api-5

### DIFF
--- a/app/errors/redirect_loop_error.rb
+++ b/app/errors/redirect_loop_error.rb
@@ -1,0 +1,2 @@
+class RedirectLoopError < RuntimeError
+end
--- a/app/presenters/redirect_presenter.rb
+++ b/app/presenters/redirect_presenter.rb
@@ -61,10 +61,13 @@ private
 
   def latest_redirect_paths
     present[:redirects].map do |redirect|
+      original_redirect = redirect[:destination]
       while (corresponding_edition = Edition.where(base_path: redirect[:destination]).order(:updated_at).last)
         break unless corresponding_edition && corresponding_edition.unpublishing.present?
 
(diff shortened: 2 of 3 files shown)


ChrisBAshton:
After a discussion with Kevin & Barbara, we've concluded that this behaviour would be better encapsulated in Router rather than here. That said, the underlying reason for why we're having to do this is due to a bug in Whitehall, so we'll attempt to fix the bug and then fix the existing occurrences of these long redirect chains. See https://trello.com/c/vFuzQYFB/2628-investigate-a-fix-for-long-redirect-chains-in-the-publishing-api-5 for details.

PR #1994, opened 2021-11-19
Remove Content API Prototype

kevindew:
This prototype is a relic of a project from 2017 and is not used by the
app or tested. It should have been deleted in 2017 really 🤷.

The idea behind this originally (for any archaeologists) was that GOV.UK
would have a V2 of the Content API which supported accessing historical
data [1]. The early journey to having this was the prototyping of that
Content API by iterating the old one in Publishing API. It was
undetermined whether a future iteration of the Content API would be a
separate application (a la Content Store) or part of Publishing API

[1]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-081-approach-for-next-iteration-of-content-api.md


spikeheap (approved):
👍 this matches what I found independently. It's nice to see the prototype was written in a way that made it easy to remove 🙇 😍

PR #1996, opened 2021-12-02
Remove misleading noise from RSpec test

ollietreend:
One of the RSpec tests printed an error message directly to the console. This was misleading behaviour because it made it look like there was an error occuring within the tests – despite RSpec reporting 0 failures at the end of the run.

It turns out the printed error message was actually expected as part of the test. The `print` statement was probably added to help with debugging during development, and then was never removed.


kevindew (approved):
Oh yikes, nice catch!

PR #2011, opened 2022-01-19
Redirect legacy Whitehall HTML attachment paths in Publishing API

baisa:
We need a one-off rake task in Publishing API, to update all legacy redirects for non-English Whitehall HTML attachments to redirect just to the content ID based URL.

[Trello card](https://trello.com/c/Qehuzldr/2780-redirect-legacy-whitehall-html-attachment-paths-in-publishing-api-5)


app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    editions.each do |edition|
+      update_redirects(edition)
+      Commands::V2::RepresentDownstream.new.call([edition.content_id], queue: DownstreamQueue::HIGH_QUEUE)
+    end
+  end
+
+  def self.update_redirects(edition)
+    return if edition.redirects.blank?
+
+    next_base_path = edition.redirects.first[:destination]
```
ChrisBAshton, line 13: We have to be careful assuming that the `redirects` array only has one element in it. I seem to remember that's almost always the case, but we should probably do a check here (and in the loop below) to exit if there is >1 redirect in the array, and log it in the job so that we can investigate manually later.

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -10,4 +10,11 @@
     described_class.update_redirects(edition_1)
     expect(edition_1.reload.redirects.first[:destination]).to eq edition_4.base_path
   end
+
+  it "raises an error if caught in a redirect loop" do
```
ChrisBAshton, line 14: Has a redirect loop happened in your tests?

I think we'll need to avoid raising an exception, as we wouldn't want the rest of the job to fail. Probably best just to skip over that edition and log the details.

ChrisBAshton (commented):
Great stuff pushing this forward, @baisa 👍

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.describe OneOffRedirectFixService do
+  it "should find the last edition and update redirects" do
+    edition_1 = create(:redirect_live_edition, destination: "/redirected-path-1", base_path: "/original-base-path")
+    create(:redirect_live_edition, destination: "/redirected-path-2", base_path: "/redirected-path-1")
+    create(:redirect_live_edition, destination: "/redirected-path-3", base_path: "/redirected-path-2")
+    edition_4 = create(:live_edition, base_path: "/redirected-path-3")
+
+    described_class.update_redirects(edition_1)
+    expect(edition_1.reload.redirects.first[:destination]).to eq edition_4.base_path
```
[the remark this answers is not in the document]
baisa: Good point!

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,27 @@
...
+  def self.fix_redirects!
+    editions = Edition.where(document_type: "redirect").joins(:document).where.not("documents.locale = 'en'")
+    editions.each do |edition|
+      update_redirects(edition)
+      Commands::V2::RepresentDownstream.new.call([edition.content_id], queue: DownstreamQueue::HIGH_QUEUE)
+    end
+  end
+
+  def self.update_redirects(edition)
+    return if edition.redirects.blank?
```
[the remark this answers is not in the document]
baisa: Good point @ChrisBAshton

app/services/one_off_redirect_fix_service.rb
baisa: done 👍

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      if next_base_path == edition.base_path
+        Rails.logger.info "[RedirectDepthFix] edition #{edition.id} and #{next_edition.id} has circular dependencies"
+        return false
+      end
+    end
+
+    redirect = edition.redirects.first
+    redirect[:destination] = next_base_path
+    edition.redirects = [redirect]
```
ChrisBAshton, line 38: Can this be shortened to this?

```suggestion
    edition.redirects.first[:destination] = next_base_path
```

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
+class OneOffRedirectFixService
+  def self.fix_redirects!
+    editions = Edition.where(document_type: "redirect").joins(:document).where.not("documents.locale = 'en'")
+    editions.each do |edition|
+      updated = update_redirects(edition)
+      Commands::V2::RepresentDownstream.new.call([edition.content_id], queue: DownstreamQueue::LOW_QUEUE) if updated
```
ChrisBAshton, line 6: Might be nice to keep track of a `editions_updated += 1` variable here, and output it at the end of the task, so we know how many have been updated 👀

ChrisBAshton (commented):
Thanks @baisa - this is looking close to completion! 👍 Just a couple of comments below.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
...
+        Rails.logger.info "[RedirectDepthFix] edition #{next_edition.id} has more then one redirect"
+        return false
+      end
+
+      next_base_path = next_edition.redirects.first[:destination]
+
+      if next_base_path == edition.base_path
+        Rails.logger.info "[RedirectDepthFix] edition #{edition.id} and #{next_edition.id} has circular dependencies"
+        return false
+      end
```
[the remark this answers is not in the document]
baisa: It did not happen so I have deleted this bit.

app/services/one_off_redirect_fix_service.rb
baisa: If we shorten this code, the edition will not get updated, we have to update destination value first.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
+class OneOffRedirectFixService
+  def self.fix_redirects!
+    editions = Edition.where(document_type: "redirect").joins(:document).where.not("documents.locale = 'en'")
```
ChrisBAshton, line 3: We should have a test that this doesn't affect English editions.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+    Rails.logger.info "#{editions_updated} editions have been updated."
+  end
+
+  def self.update_redirects(edition)
+    return false if edition.redirects.blank?
+
+    if edition.redirects.count > 1
+      Rails.logger.info "[RedirectDepthFix] edition #{edition.id} has more than one redirect."
```
ChrisBAshton, line 19: We should have a test that editions with >1 redirect get skipped over.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      return false
+    end
+
+    next_base_path = edition.redirects.first[:destination]
+
+    while (next_edition = Edition.where(base_path: next_base_path).order(:updated_at).last)
+      break unless next_edition && next_edition.redirects.present?
+
+      if next_edition.redirects.count > 1
+        Rails.logger.info "[RedirectDepthFix] edition #{next_edition.id} has more than one redirect. Unable to process earlier edition #{edition.id}"
```
ChrisBAshton, line 29: We should have a test that editions that have 1 redirect, with a more recent edition that have >1 redirects, get skipped over.

ChrisBAshton (approved):
Hi @baisa - the code looks good 👍 I've just realised we're missing test coverage for 3 bits of logic, which would be great to add before merging. If not, as this is a temporary task, we could get away with a manual test run on integration. I'll leave it up to you - feel free to Slack me.

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    described_class.update_redirects(edition_1)
+    described_class.update_redirects(edition_2)
+    described_class.update_redirects(edition_3)
+    expect(edition_1.reload.redirects.first[:destination]).to eq edition_4.base_path
+    expect(edition_2.reload.redirects.first[:destination]).to eq edition_4.base_path
+    expect(edition_3.reload.redirects.first[:destination]).to eq edition_4.base_path
+  end
+
+  it "should not update redirects for English edition" do
+    document = create(:document)
```
ChrisBAshton, line 19: Could we make this more explicit by providing `locale: "en"`? Rather than relying on that being set by default? Otherwise it's not really clear how this test is different to the one above it 🤔  Would be nice to have one that tests the opposite behaviour for non-English locales too. I notice in the first test we're just testing the `update_redirects` method directly, which is fine, but we're only really half-testing the `fix_redirects!` method here.

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    edition_1 = create(:redirect_live_edition, destination: "/redirected-path-1", base_path: "/original-base-path", document: document)
+    create(:live_edition, base_path: "/redirected-path-2")
+
+    described_class.fix_redirects!
+    expect(edition_1.reload.redirects.first[:destination]).to eq "/redirected-path-1"
+  end
+
+  it "should skip updating edition before the one and the one with more than 1 redirect" do
```
ChrisBAshton, line 27: This test looks like we're expecting editions 1-4 to be unchanged at the end of the `update_redirects` calls. (Which is also the behaviour I'd expect). So I'd rephrase the test name to:

```suggestion
  it "should avoid updating editions when any editions in the chain have more than 1 redirect" do
```

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    edition_2.save!(validate: false)
+    edition_3 = create(:redirect_live_edition, destination: "/redirected-path-3", base_path: "/redirected-path-2")
+    edition_4 = create(:live_edition, base_path: "/redirected-path-3")
+
+    described_class.update_redirects(edition_1)
+    described_class.update_redirects(edition_2)
+    described_class.update_redirects(edition_3)
+    expect(edition_1.reload.redirects.first[:destination]).to eq edition_2.base_path
+    expect(edition_2.reload.redirects.first[:destination]).not_to eq edition_4.base_path
+    expect(edition_3.reload.redirects.first[:destination]).to eq edition_4.base_path
```
ChrisBAshton, line 43: We could also make these assertions clearer by making a note of the original values, i.e.

```suggestion
    expect(edition_1.reload.redirects).to eq edition_1_redirects
    expect(edition_2.reload.redirects).to eq edition_2_redirects
    expect(edition_3.reload.redirects).to eq edition_3_redirects
```

Otherwise at first glance, it looks like we're expecting the values to have changed.

ChrisBAshton (commented):
Thanks for adding those tests. Some minor comments on them, then this is good to go!

ChrisBAshton:
Just tried running this on Integration: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/303763/console

It processed several successfully until here:

```
09:04:36 User excluded error: #<ActiveRecord::RecordInvalid: Validation failed: Redirects path cannot equal the destination>
09:04:36 rake aborted!
09:04:36 ActiveRecord::RecordInvalid: Validation failed: Redirects path cannot equal the destination
```

Looks like we will need to capture that circular redirect after all.

baisa:
> Just tried running this on Integration: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/303763/console
> 
> It processed several successfully until here:
> 
> ```
> 09:04:36 User excluded error: #<ActiveRecord::RecordInvalid: Validation failed: Redirects path cannot equal the destination>
> 09:04:36 rake aborted!
> 09:04:36 ActiveRecord::RecordInvalid: Validation failed: Redirects path cannot equal the destination
> ```
> 
> Looks like we will need to capture that circular redirect after all.

I have added it back

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    edition_2.save!(validate: false)
+    edition_3 = create(:redirect_live_edition, destination: "/redirected-path-3", base_path: "/redirected-path-2")
+    create(:live_edition, base_path: "/redirected-path-3")
+
+    described_class.update_redirects(edition_1)
+    described_class.update_redirects(edition_2)
+    described_class.update_redirects(edition_3)
+    expect(edition_1.reload.redirects).to eq edition_1.redirects
+    expect(edition_2.reload.redirects).to eq edition_2.redirects
+    expect(edition_3.reload.redirects).to eq edition_3.redirects
```
ChrisBAshton, line 63: I'm not 100% sure if this works - it feels risky. We're basically relying on `edition_n` to not fetch things fresh from the database, but to return what it has in memory. If the `redirects` method was ever changed, this test would give us false sense of security, because it would actually be saying "expect [specific thing from database] to eq [specific thing from database]", which would always be `true`. That way, if calling `update_redirects` _does_ actually change the `redirects` values, the test would continue to pass when we don't want it to.

I think it would be safer to cache the values in temporary variables (e.g. `edition_1_redirects = edition_1.redirects` and then do assertions against that (`expect(edition_1.reload.redirects).to eq(edition_1_redirects)`.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      return false
+    end
+
+    next_base_path = edition.redirects.first[:destination]
+
+    while (next_edition = Edition.where(base_path: next_base_path).order(:updated_at).last)
+      break unless next_edition && next_edition.redirects.present?
+
+      if next_base_path == edition.base_path
+        Rails.logger.info "[RedirectDepthFix] edition #{edition.id} and #{next_edition.id} has a circular redirect"
```
ChrisBAshton, line 29: Nitpick: in each `Rails.logger.info`, we have a `[RedirectDepthFix]` prefix. I think that prefix only really applies to this one (circular redirect == infinite depth), whereas the other two should probably be changed to something like `[RedirectsValueInvalid]`.

spec/services/one_off_redirect_fix_service_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    described_class.update_redirects(edition_1)
+    described_class.update_redirects(edition_2)
+    described_class.update_redirects(edition_3)
+    expect(edition_1.reload.redirects.first[:destination]).to eq edition_4.base_path
+    expect(edition_2.reload.redirects.first[:destination]).to eq edition_4.base_path
+    expect(edition_3.reload.redirects.first[:destination]).to eq edition_4.base_path
+  end
+
+  it "should not update edition if redirects path equals the destination" do
```
ChrisBAshton, line 18: Thanks for adding this test 👍

ChrisBAshton (commented):
Thanks for pushing up the fix, @baisa - I'll try running it on Integration now.

app/services/one_off_redirect_fix_service.rb
```diff
@@ -0,0 +1,48 @@
...
+        return false
+      end
+
+      next_base_path = next_edition.redirects.first[:destination]
+    end
+
+    redirect = edition.redirects.first
+    redirect[:destination] = next_base_path
+    edition.redirects = [redirect]
+    edition.save!
```
ChrisBAshton, line 44: Hmm it failed with the same [validation error](https://github.com/alphagov/publishing-api/blob/71a3407ad33a5379e16d8163c7f4850036936974/app/validators/routes_and_redirects_validator.rb#L138), unfortunately.

Might be worth capturing the exception to find out more:

```suggestion
    begin
      edition.save!
    rescue ActiveRecord::RecordInvalid => e
      Rails.logger.info "[RedirectDepthFix] #{edition.id} could not be updated: #{e.message}"
      return false

    end
```

baisa: I have added this - it run successfully on Integration https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/304128/console

ChrisBAshton (approved):
Nice one! I've checked the Jenkins output and ~15k editions were successfully updated. Just 25 manual ones to look at, which we can do later.

This looks good to merge and run on Production 👍

PR #2026, opened 2022-02-10
Prepare Publishing API for replatforming

bilbof:
- Uses multi-stage build for Dockerfile and using slim-buster cuts image size by 2/3rds
- Configure Puma using GovukAppConfig gem


sengi (approved):
🚀

PR #2048, opened 2022-03-15
Upgrade to Rails 7

brucebolt:
Upgrades to Rails 7.

Also makes the changes needed to move from the Classic autoloader to Zeitwerk, as [Classic is no longer supported by Rails 7](https://rubyonrails.org/2021/9/3/autoloading-in-rails-7-get-ready).

[Trello card](https://trello.com/c/IMxcB3Nm/272-upgrade-publishing-api-from-rails-61x-to-70x)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", "6.1.5"
+gem "rails", "7.0.2.3"
```
benjamineskola, line 3: Would it be worth making this a `~>` definition at this point?

```suggestion
gem "rails", "~> 7.0.2.3"
```

brucebolt: Unfortunately not, as bundler may downgrade Rails if a dependency requires an older version.  It seems fixing the release is the safest option.

benjamineskola:
Nitpick: in [`config/application.rb`, L41–44](https://github.com/alphagov/publishing-api/pull/2048/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9) are FIXME comments that have been addressed by this PR:

```ruby
    # FIXME: the 3 lines below will be uncommented as part of the restructuring work.
    # config.eager_load_paths += Dir["#{config.root}/app/queries"]
    # config.eager_load_paths += Dir["#{config.root}/app/commands"]
    # config.eager_load_paths += Dir["#{config.root}/app/presenters"]
```

benjamineskola (approved):

PR #2050, opened 2022-03-16
Add optimalworkshop.com to allowlist of external domains

hannako:
The explore team is adding a recruitment banner to subsets of pages
on govuk. The banner contains a link to the online survey at the
above domain. To save us having to redeploy the frontend applications
every time the survey URL is updated we would like to set up a redirect
in short-url manager. Once we have completed this phase of testing
optimalworkshop.com will be removed from the allowlist.

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -1,5 +1,5 @@
 class RoutesAndRedirectsValidator < ActiveModel::Validator
-  EXTERNAL_HOST_ALLOW_LIST = %w[.gov.uk .judiciary.uk .nhs.uk .ukri.org .nationalhighways.co.uk].freeze
+  EXTERNAL_HOST_ALLOW_LIST = %w[.gov.uk .judiciary.uk .nhs.uk .ukri.org .nationalhighways.co.uk optimalworkshop.com].freeze
 
   def validate(record, base_path: nil)
     base_path = record.base_path if base_path.nil?
--- a/spec/support/routes_and_redirects_validator.rb
+++ b/spec/support/routes_and_redirects_validator.rb
@@ -212,6 +212,7 @@ RSpec.shared_examples_for RoutesAndRedirectsValidator do
           https://www.nhs.uk/
           https://www.ukri.org/
           https://www.nationalhighways.co.uk/
+          https://GDSUserResearch.optimalworkshop.com/treejack/1234
         ].each do |destination|
           edition.redirects = [{ path: "#{subject.base_path}/new", type: "exact", destination: destination }]
 


chao-xian:
Do we know if we're allowed to add this domain to the list?

kevindew:
This seems a bit fishy as it will allow every redirect to the Publishing API to redirect to this location. I'm somewhat thinking if we can avoid using Short URL Manager for this it'd be preferable - but I don't know the details.

How often are we expecting this URL to change?

hannako:
@kevindew I planned to post in govuk developers for a sense check on if this is a bad idea. The URL will need to change "a few times over the next couple of weeks". I will get a bit more clarity on that. But it's certainly not a daily task - more like weekly.

We have the banner (which uses the [intervention component](https://github.com/alphagov/government-frontend/blob/main/app/views/layouts/application.html.erb#L34-L40)) on a subset of pages rendered by government-frontend and frontend. I wanted to avoid adding anything to the gem as it's painful to release a gem bump just to update a URL. Updating the hardcoded url in two apps is not a big deal or a lot of work, so if this is too risky I'll close the PR.

kevindew:
> @kevindew I planned to post in govuk developers for a sense check on if this is a bad idea. The URL will need to change "a few times over the next couple of weeks". I will get a bit more clarity on that. But it's certainly not a daily task - more like weekly.
> 
> UPDATE: 4 times in the next 2 weeks, and potentially another 3 times after that
> 
> We have the banner (which uses the [intervention component](https://github.com/alphagov/government-frontend/blob/main/app/views/layouts/application.html.erb#L34-L40)) on a subset of pages rendered by government-frontend and frontend. I wanted to avoid adding anything to the gem as it's painful to release a gem bump just to update a URL. Updating the hardcoded url in two apps is not a big deal or a lot of work, so if this is too risky I'll close the PR.

Sure, I can see how it's a tricky thing to work with.

I agree about not wanting to update and release the Gem I think you do right to avoid that.

I think you should take the approach of updating government_frontend and frontend each time and have them inject that information into the component. 

I guess at an ideal level this would be managed by a CMS - potentially the sort of global data that is used to render headers and footers (a death of static thing) and failing that it seems best placed in a major frontend app, which probably exposes that frontend/government-frontend have very similar responsibilities.

hannako:
Thanks @kevindew 

> I think you should take the approach of updating government_frontend and frontend each time and have them inject that information into the component.

This is how it's being done at the [moment](https://github.com/alphagov/government-frontend/blob/main/app/views/layouts/application.html.erb#L34-L40) . Although I should put those strings into locale files! I will close this and related https://github.com/alphagov/short-url-manager/pull/672

PR #2056, opened 2022-03-31
Docker/end-to-end test related cleanups

kevindew:
The primary motivation of this PR is to add curl as a dependency of the Publishing API Dockerfile so that curl can be used as a healthcheck for the container as per: https://github.com/alphagov/publishing-e2e-tests/commit/e5f1925c2f2b70723ba320e1df2978a2db997954

However I found a few yaks to shave along the way. I removed nodejs from the Dockerfile (not too sure why it was once needed), removed an old docker test file that hasn't been used in years, and added a user to the DB seeds which fixes a race condition that can break the publishing-e2e-tests


davidgisbey (approved):
lgtm!

PR #2070, opened 2022-04-22
Migrate HMRC manuals from 'live' to 'beta'

1pretz1:
HMRC manuals are rendered with a 'beta' banner due to custom logic in
Manuals Frontend, as we're migrating to Government Frontend we need to
set the phase value to persist the beta status.

There are questions around whether these pages should still be in 'beta'
but for now we need to keep the same phase.

Data migration template shamelessly copied from https://github.com/alphagov/publishing-api/pull/2047 (thanks @edwardkerry 😄)

Trello:
https://trello.com/c/sCpFwYz6/1213-migrate-hmrc-manuals-from-live-to-beta


edwardkerry (approved):
Looks like it'll do what we need!

PR #2079, opened 2022-05-30
Skip if document type is a redirect

1pretz1:
Currently the task updates the edition if the document type is
'redirect'. This doesn't cause any issues downstream but the
`rendering_app` is populated when it should be `nil`.


KludgeKML (approved):
LGTM

PR #2084, opened 2022-06-16
Combine rails_helper and spec_helper files into one 

kevindew:
It is likely that most, if not all, of the tests in this repo require
configuration from both of these of files and having two is a source of
confusion and decision points.

This combines them into a single file as per the GOV.UK Rails app
conventions: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#testing-utilities

I also reduced the amounts of RSpec comments that explain how you might
use the file.


BeckaL (approved):
All looks good to me, a nice simplification 😄

PR #2085, opened 2022-06-16
Remove corporate info fix

BeckaL:
This was likely a one-off rake task so tidying up.


kevindew (approved):

JonathanHallam (approved):
LGTM, looking at the [original commit message](https://github.com/alphagov/publishing-api/commit/7c4d61e238f9489d6a1e8517b576fc51b7d266bc), the bug this was negating has been fixed

PR #2086, opened 2022-06-16
Remove consistency checker

BeckaL:
This removes the task as well as the related checker and loader. These were added in 2017 but are now unused, so removing these to tidy up the project.


kevindew (approved):
This looks good to me. I remember the good intentions when this was added but in practice it was something that we didn't have the resources to monitor.

We can always revert it should we want something similar.

JonathanHallam (approved):
Nice one thank you :) 

I wonder if there's a way that we can track how often things like this are being used so that this kind of work is easier in the future

PR #2087, opened 2022-06-16
Removing link_setter.rb

BeckaL:
Tidying up as this is unused and was probably missed from an earlier commit.

This also has the benefit of boosting our test coverage ahead of CD.


kevindew (approved):

JonathanHallam (approved):
I've searched through our code and can't find any examples of this actually being used

PR #2088, opened 2022-06-16
Remove set last edited at task

BeckaL:
This was likely a one off task that can now be removed.


kevindew (approved):
Looks good to remove. Seems to reference a table that doesn't exist anymore - were it reference a real table it'd be extremely dangerous 😱

JonathanHallam (approved):
Nice one, thanks :)

PR #2089, opened 2022-06-16
Remove remove_invalid_worldorg_drafts task

BeckaL:
This was likely a one-off task so tidying up.


JonathanHallam (approved):
Yup happy with this one. The original commit message makes it look like this wasn't temporary so much as there to sort out problems that occur from a bug. I'm assuming that bug has been fixed but we can always revert this if we need the task back

PR #2090, opened 2022-06-16
Add pull request template

BeckaL:
Add template to warn about continuous deployments. 


JonathanHallam (approved):
Niiiice! Although you probably want to add a do not merge flag to this until we're ready to turn CD on for publishing api

PR #2094, opened 2022-06-17
Remove unused put links route

BeckaL:
This was introduced for backwards compatibility in 2016 but logs show this is now unused, and has been fully replaced by the patch endpoint. It is unused by GDS api adaptors, which instead [uses the patch route](https://github.com/alphagov/gds-api-adapters/blob/355ae4ca64bcc9491d4d30bb011efe62a80512af/lib/gds_api/publishing_api.rb#L322-L330). This route was not listed in the [api docs](https://github.com/alphagov/publishing-api/blob/main/docs/api.md) so there is no change to that as part of this PR.

The following screenshots show usage of PUT (legacy) vs PATCH (current) endpoints over the last 7 days, demonstrating that PUT is no longer used:

PUT request hits over the last 7 days showing 0 hits
PATCH request hits over the last 7 days showing 235,024 hits

Doing this as part of work to bring publishing api up to the standards required for CD. This was a public route so covered under the pact test requirements, but we should remove it rather than adding pact tests for it :-) 

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,8 +28,6 @@ Rails.application.routes.draw do
         get "/links/:content_id", to: "link_sets#get_links"
         get "/expanded-links/:content_id", to: "link_sets#expanded_links"
         patch "/links/:content_id", to: "link_sets#patch_links"
-        # put is provided for backwards compatibility.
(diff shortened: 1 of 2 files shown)


JonathanHallam (approved):
Perfect, thanks :)

PR #2096, opened 2022-06-17
Remove import route

BeckaL:
As noted in the comments for the import command, this was a temporary addition to support maslow migration, and since we are now well past March 2017 we can remove the endpoint and command.

Doing this as part of work to bring publishing api up to the standards required for CD. This was a public route so covered under the pact test requirements, but we should remove it rather than adding pact tests for it :-) 

Query used to check for usage on kibana: `application: publishing-api AND request: POST\ \/v2\/content\/*\/import*`

gds api adaptors PR: https://github.com/alphagov/gds-api-adapters/pull/1147


JonathanHallam (approved):
👍

PR #2100, opened 2022-06-21
Redirect corrupted routes

1pretz1:
These sections have been left in a strange state in Router after all
live HMRC manuals were republished [1]. These section routes have a
`redirect_to` field populated, but aren't handled as redirects, rather
they're processed like normal backend routes.

This appears to have happened as a result of Publishing API thinking
these sections are normal published editions and so the republish job,
republished them and ignored the set redirect for these sections.

Not sure how this has happened as the current tasks to redirect a
section wouldn't leave the sections in this state in Publishing API [2].

Maybe these were redirects were made before the tasks existed and not
persisted properly though the stack.

The below query in Router API returns the problem routes.

```
Route.where(backend_id: "government-frontend", handler: "backend").where(incoming_path: /^\/hmrc.*/).nin(redirect_to: nil).pluck(:incoming_path, redirect_to)
```

By checking if the sections are present on the owning manual, we can
work out whether each of these sections should be redirected, or not.

As HMRC Manuals API doesn't have a database, the data migration makes
sense to live in Publishing API. It also follows the same method to
implement a redirect (put then publish) as HMRC manuals API [2], rather
than unpublish with a redirect.

Trello:
https://trello.com/c/ZhJ2oj7N/1339-resolve-redirect-issues-for-manuals

[1]: https://github.com/alphagov/publishing-api/pull/2047
[2]: https://github.com/alphagov/hmrc-manuals-api/blob/main/lib/tasks/redirect_sections.rake#L2

### DIFF
--- a/lib/tasks/tmp_unpublish_and_redirect_hmrc_manuals.rake
+++ b/lib/tasks/tmp_unpublish_and_redirect_hmrc_manuals.rake
@@ -1,0 +1,137 @@
+# Generated from live Router API (routes which have a redirect, but handler is backend so redirect isn't actioned):
+# routes = Route.where(backend_id: "government-frontend", handler: "backend").where(incoming_path: /^\/hmrc.*/).nin(redirect_to: nil).pluck(:incoming_path, redirect_to)
+# routes.map { |x| { base_path: x[0], redirect_to: x[1] }}
+routes = [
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

PR #2116, opened 2022-07-20
Fix flaky test by resolving change note accepting empty timestamps

kevindew:
This provides a fix for [a test](https://github.com/alphagov/publishing-api/blob/c92cc3101346c2cb1075c359d32851b970874b3a/spec/integration/message_queue_publishing_spec.rb#L4-L17) that fails in about 10% of Publishing API builds. The test relies on random data and thus exposes an issue in the code:

```
  1) Message queue publishing puts the correct message on the queue
     Failure/Error: expect(errors).to eql([])

       expected: []
            got: [{:failed_attribute=>"TypeV4", :fragment=>"#/details/change_history/0/public_timestamp", :message=>"T...76f-aca8114b1247#", :schema=>#<Addressable::URI:0x16c60 URI:a4923dd9-f21e-5f89-b76f-aca8114b1247#>}]

       (compared using eql?)

       Diff:
       @@ -1,6 +1,11 @@
       -[]
       +[{:failed_attribute=>"TypeV4",
       +  :fragment=>"#/details/change_history/0/public_timestamp",
       +  :message=>
       +   "The property '#/details/change_history/0/public_timestamp' of type null did not match the following type: string in schema a4923dd9-f21e-5f89-b76f-aca8114b1247#",
       +  :schema=>
       +   #<Addressable::URI:0x16c60 URI:a4923dd9-f21e-5f89-b76f-aca8114b1247#>}]

     # ./spec/integration/message_queue_publishing_spec.rb:36:in `block in ensure_message_queue_payload_validates_against_notification_schema'
     # ./spec/integration/message_queue_publishing_spec.rb:30:in `ensure_message_queue_payload_validates_against_notification_schema'
     # ./spec/integration/message_queue_publishing_spec.rb:17:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:85:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:85:in `block (2 levels) in <top (required)>'
```

The cause of this issue is that Publishing API will present change notes to the frontend and message queue when a change note does not have a timestamp. This is invalid as the schema dictates that [a change history will have a timestamp](https://github.com/alphagov/govuk-content-schemas/blob/0e748b701f657ebb41f871a8d8515d2eb3872228/formats/shared/definitions/publishing_api_base.jsonnet#L96-L116).

The fix for this is two-fold: 

1) Accept that in the draft process we can have change notes without a timestamp and filter them out of data sent to frontend
2) At the point of publishing set any change notes that are missing a timestamp.

I also had to refactor some funky tests (that I wrote 🤦) to be able to implement it.

In practice this issue seems to have caused a very limited impact on the Publishing API with only about 40 documents existing with invalid change notes. This is likely because the change note feature of Publishing API is rather niche and not used by Whitehall and other big Publishing apps.

More info in the commits and definitely easiest reviewed commit by commit.

### DIFF
--- a/app/presenters/change_history_presenter.rb
+++ b/app/presenters/change_history_presenter.rb
@@ -30,6 +30,7 @@ module Presenters
         .joins(:edition)
         .where(editions: { document: document })
         .where("user_facing_version <= ?", version_number)
+        .where.not(public_timestamp: nil)
         .order(:public_timestamp)
     end
 
--- a/app/models/edition/timestamps.rb
+++ b/app/models/edition/timestamps.rb
@@ -58,6 +58,10 @@ class Edition::Timestamps
       end
     end
 
+    edition.change_note&.then do |change_note|
+      change_note.update!(public_timestamp: edition.public_updated_at) unless change_note.public_timestamp
+    end
+
     edition.save!
   end
 end
--- a/spec/factories/change_note.rb
+++ b/spec/factories/change_note.rb
@@ -1,3 +1,5 @@
 FactoryBot.define do
-  factory :change_note
+  factory :change_note do
+    public_timestamp { 1.day.ago }
+  end
 end
--- a/spec/models/edition/timestamps_spec.rb
+++ b/spec/models/edition/timestamps_spec.rb
@@ -173,5 +173,19 @@ RSpec.describe Edition::Timestamps do
         expect(edition.public_updated_at).to eq(current_time)
       end
     end
+
+    it "sets the public_timestamp of a change note without one" do
+      edition = create(:edition, public_updated_at: "2022-07-18")
+      change_note = create(:change_note, edition: edition, public_timestamp: nil)
+      expect { described_class.live_transition(edition, "minor") }
+        .to change { change_note.reload.public_timestamp }.to(edition.public_updated_at)
+    end
+
+    it "doesn't change the public_timestamp of a change note with one already set" do
+      edition = create(:edition)
+      change_note = create(:change_note, edition: edition, public_timestamp: "2022-07-18")
+      expect { described_class.live_transition(edition, "minor") }
+        .not_to(change { change_note.reload.public_timestamp })
+    end
   end
 end
--- a/spec/models/edition/timestamps_spec.rb
+++ b/spec/models/edition/timestamps_spec.rb
@@ -4,221 +4,173 @@ RSpec.describe Edition::Timestamps do
   let(:current_time) { Time.zone.now }
 
   describe "#edited" do
-    let(:edition) { build(:edition, update_type: update_type) }
-    let(:previous_live_version) do
-      build(
-        :edition,
-        publishing_api_first_published_at: "2017-01-01",
-        first_published_at: "2017-04-01",
-        major_published_at: "2017-11-11",
-      )
-    end
-
-    let(:update_type) { "major" }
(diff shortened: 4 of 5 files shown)


app/presenters/change_history_presenter.rb
```diff
@@ -30,6 +30,7 @@ def change_notes
         .joins(:edition)
         .where(editions: { document: document })
         .where("user_facing_version <= ?", version_number)
+        .where.not(public_timestamp: nil)
```
BeckaL, line 33: I've gone back and forth on my thinking about this - I keep having an inkling that there's a better way to deal with this, but not come to definite conclusions. 

I think basically my concern is whether this will cause any issues for draft content. I know there's currently an issue anyway since these draft documents with un-stamped change notes will fail to be sent in the same way that this test fails, but I'm also wondering whether filtering them out will also cause us issues at some point. 

I guess we couldn't give these timestamps in the same way as we do for published editions, since the edition doesn't have a public timestamp. But wondering whether we could just give them _some_ timestamp, or enforce timestamps on the frontend, or solve in some other way. Or whether the implications are fine?

So more of a discussion point given that you've thought about this a lot, than a definite suggestion.

BeckaL:
Thanks for looking at this! I actually [created a card](https://trello.com/c/2XPRkxU9/135-flakey-test-in-publishing-api) for this a few weeks ago if you want some trello satisfaction. 

Fix for second bug here looks good. There's something about the first fix that feels instinctively off to me - may well be the case that this is the best way, have tried to explain what doesn't quite sit right with me above.

app/presenters/change_history_presenter.rb
kevindew: Thanks for giving this some thought. I can understand your concern with this, there is definitely something "sweep this under the carpet" with a query condition like this.

I can explain some of my rationale though. I see the primary concern is having the Publishing API produce a change history that creates a valid schema. So I'm prioritising outputting only valid data over outputting complete data. This seems preferable to me as it holds up Publishing API's side of the contract. Changing the schema also seems a growing bit of work which creates confusing scenarios (change notes without timestamps).

The tricky thing, as you've addressed, is that the state of the item can be valid and not have a timestamp. So we can end up in the draft situation. I think that's ok to not have the change note show before the date is available.

You're probably right that one day an issue will occur. I'd hope it's only some confusion of a draft for someone.

Annoyingly though this change note stuff is niche, because apps like Whitehall send a whole change history in their payload (if anything I think this ChangeNote system is bad but that's a longer story) which I think somewhat explains why this hasn't caused user issues.

BeckaL: Sorry for the slow reply @kevindew, and thanks for the explanation.

That makes sense - I think in a perfect world it'd be nice for the schemas to represent reality, so having a draft schema vs published schema. But I can see how this can get confusing, especially since I've been looking at the current state of confusion in schemas! 

So I think this is a pragmatic approach, we can review in future, and is definitely an improvement on now when the draft content with changenotes would error.

BeckaL (approved):
Happy to approve, thanks for explanation about approach above. 

Do we need to republish the ~40 affected drafts?

kevindew:
Thanks @BeckaL I'm not too concerned as their drafts so will plan to go ahead without.

PR #2117, opened 2022-07-20
Update to Ruby 3.1.2

kevindew:
This updates this application to use the current version of Ruby. This work was put together as a proof of concept for confirming that our [install of Ruby 3.1.2](https://github.com/alphagov/packager/pull/200) was successful. It now makes sense to use this to start the adoption of Ruby 3.1.x.

Note, nearly all the changes in this PR are due to rubocop preferring the shorter Ruby syntax of `{ item: }` than `{ item: item }` which was introduced in Ruby 3.1


app/adapters/content_store.rb
```diff
@@ -5,17 +5,17 @@ class ContentStore
     def self.put_content_item(base_path, content_item)
       CommandError.with_error_handling do
         PublishingAPI.service(:live_content_store).put_content_item(
-          base_path: base_path,
-          content_item: content_item,
+          base_path:,
```
JonathanHallam, line 8: I don't especially like the new syntax and am pretty much guaranteed to forget every time 😆

JonathanHallam (approved):
Happy with this, makes in docker okay. That rubocop commit though!!!

app/adapters/content_store.rb
kevindew: Yes, it looks a bit confusing because of it's similarity with required keyword argument syntax.

The [JavaScript equivalent shorthand syntax](https://attacomsian.com/blog/javascript-object-property-shorthand#:~:text=The%20object%20literal%20property%20value,make%20the%20object%20initialization%20shorter.&text=The%20property%20value%20shorthand%20syntax,value%20as%20a%20property%20value.) has been all the rage for a while so I expect it'll quickly become the norm in Ruby and we'll become very used to it.

PR #2151, opened 2022-09-14
Add royal.uk to the external links allow list

AgaDufrat:
Related to: https://github.com/alphagov/whitehall/pull/6815


ollietreend (dismissed):

ollietreend:
@AgaDufrat it turns out we actually don't need this after all.

This would allow us to redirect to royal.uk URLs, but in our case we only need to publish links to it. So no need for this validation to change.

kevindew:
Sorry my bad! It was only news to me that we actually had two set of external host allow lists in Whitehall.

PR #2152, opened 2022-09-29
Make Dockerfile compatible with govuk-ruby-base.

sengi:
- Stop doing `mkdir /app`, since govuk-ruby-base now does this for us (and `WORKDIR` creates its directory anyway if it's doesn't exist).
- Avoid installing a couple of packages which are now in the base image.
- Bring `.dockerignore` into line with most of the other repos.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,11 @@
+.dockerignore
 .git
 .gitignore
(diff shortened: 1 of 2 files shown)


martynipratt (approved):
I approve this

PR #2158, opened 2022-10-05
Replace `Plek.current` with `Plek.new`

brucebolt:
Usage of `Plek.current` was marked as deprecated in https://github.com/alphagov/plek/pull/94, therefore we should remove uses of the old syntax now we've upgraded Plek, else we'll log lots of deprecation warnings.


BeckaL (approved):
🙌

PR #2173, opened 2022-10-26
Don't set storage backend addresses in the Dockerfile.

sengi:
Setting `RABBITMQ_URL` in the Dockerfile makes it very difficult to use `RABBITMQ_HOSTS` etc. in Kubernetes.

This change doesn't affect anything in EC2.

Requires alphagov/publishing-e2e-tests/pull/518.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,10 @@ COPY . /app
 
 FROM $base_image
 
-# TODO: don't set DATABASE_URL, PORT or RABBITMQ_URL here. Set them in publishing-e2e-tests.
-ENV DATABASE_URL=postgresql://postgres@postgres/publishing-api PORT=3093
-ENV RABBITMQ_URL=amqp://guest:guest@rabbitmq:5672 RABBITMQ_EXCHANGE=published_documents
-
 ENV GOVUK_CONTENT_SCHEMAS_PATH=/govuk-content-schemas
 ENV GOVUK_APP_NAME=publishing-api
+# TODO: move this default for RABBITMQ_EXCHANGE to config/initializers/services.rb.
+ENV RABBITMQ_EXCHANGE=published_documents
 
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 COPY --from=builder /app /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,6 @@ FROM $base_image
 
 ENV GOVUK_CONTENT_SCHEMAS_PATH=/govuk-content-schemas
 ENV GOVUK_APP_NAME=publishing-api
-# TODO: move this default for RABBITMQ_EXCHANGE to config/initializers/services.rb.
-ENV RABBITMQ_EXCHANGE=published_documents
 
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 COPY --from=builder /app /app/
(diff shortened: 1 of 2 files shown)


kevindew:
Change looks good, but think e2e will need the env var setting in [publishing_api](https://github.com/alphagov/publishing-e2e-tests/blob/main/docker-compose.yml#L337) and [publishing_api_worker](https://github.com/alphagov/publishing-e2e-tests/blob/main/docker-compose.yml#L337).

This did catch my eye about you having a problem with RABBITMQ_URL though, what problems did you face?

I ask as we're actually discussing a bit about the env var vars for RabbitMQ over in https://github.com/alphagov/govuk_message_queue_consumer/pull/78 - one of the things I've theorised is that all our individual variables RABBITMQ_HOSTS, RABBITMQ_USERNAME etc are a bit annoying and life would be simpler adopting RABBITMQ_URL since it's [a standard](https://www.rabbitmq.com/uri-spec.html) whereas the other env vars seem to be our own creation and probably best deprecated.

sengi:
> Change looks good, but think e2e will need the env var setting in [publishing_api](https://github.com/alphagov/publishing-e2e-tests/blob/main/docker-compose.yml#L337) and [publishing_api_worker](https://github.com/alphagov/publishing-e2e-tests/blob/main/docker-compose.yml#L337).

Oh whoops, I didn't spot that it was only set for search-api and not everything. Thanks!

> This did catch my eye about you having a problem with RABBITMQ_URL though, what problems did you face?

The problem I ran into was that it's hard to unset an environment variable in Kubernetes when it's been set in the container image. I suspected (though didn't verify) that RABBITMQ_URL was taking precedence as the Rabbit client still wasn't working after setting RABBITMQ_HOST etc. and checking that the pod was able to reach that address. I thought I'd attack the problem from both sides anyway in case I'm wrong about which one takes precedence :)

> I ask as we're actually discussing a bit about the env var vars for RabbitMQ over in [alphagov/govuk_message_queue_consumer#78](https://github.com/alphagov/govuk_message_queue_consumer/pull/78) - one of the things I've theorised is that all our individual variables RABBITMQ_HOSTS, RABBITMQ_USERNAME etc are a bit annoying and life would be simpler adopting RABBITMQ_URL since it's [a standard](https://www.rabbitmq.com/uri-spec.html) whereas the other env vars seem to be our own creation and probably best deprecated.

Aha, I actually already have [a PR to try using RABBITMQ_URL instead](https://github.com/alphagov/govuk-helm-charts/pull/659) so it's great to hear that that's the expected way! It'd be great to get rid of our somewhat convoluted config code for those RABBITMQ_HOST etc. vars.

kevindew:
Oh great good to hear it re RABBITMQ_URL env var. Do let me know if you have problems. I am thinking about deprecating the other ones so would be good to know if there's a need for them. It'd also make deprecation all the easier if you're already on RABBITMQ_URL

sengi:
RFAL - tests are passing now :)

kevindew (approved):
Change looks good.

I am wondering if we should just do the TODO though - I have a bit of an aversion to TODO comments as they rarely get done. 

I'm wondering if the change could be just:

```
                  elsif ENV["RABBITMQ_URL"] && ENV["RABBITMQ_EXCHANGE"]
                    { exchange: ENV["RABBITMQ_EXCHANGE"] }
```

to:

```
                  elsif ENV["RABBITMQ_URL"]
                    { exchange: ENV.fetch("RABBITMQ_EXCHANGE", "published_documents") }
```

Since we don't have RABBITMQ_URL set in puppet_aws world, that seems a pretty low risk change to make.

sengi:
Thanks! I'm all for doing TODOs when they doable :) Done.

PR #2181, opened 2022-11-02
Mark all routes as 'gone', not just the base path

brucebolt:
If an edition has more than one route (e.g. an atom feed, in addition to the base path), we currently only mark the base path as 'gone' when unpublishing the content item.  This is a bug, as we would expect all routes to be unpublished in the same way.

This fixes that bug by ensuring we unpublish all routes.

[Trello card](https://trello.com/c/ul5lL6y3)


kevindew (approved):
Looks good to me, thanks for this.

Are we worried this might produce any surprises for content without a base_path? I guess they just have an empty array of routes?

brucebolt:
> Are we worried this might produce any surprises for content without a base_path? I guess they just have an empty array of routes?

They do indeed have an empty array, which is the same behaviour as we were previously creating (and there's already a test to check for this).

PR #2190, opened 2022-11-11
Lint: use list syntax for Dockerfile CMD.

sengi:
Generated with `gsed -i 's/^CMD bundle exec puma$/CMD [bundle, exec, puma]/'`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,4 +18,4 @@ ENV GOVUK_APP_NAME=publishing-api
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 COPY --from=builder /app /app/
 WORKDIR /app
-CMD bundle exec puma
+CMD ["bundle", "exec", "puma"]


kevindew (approved):
This looks good, it'd be useful to explain why this is better than the previous option though.

I'm a bit uncomfortable with lint fixes that aren't enforced by a linter as a future person may have a different preference.

sengi:
Both good questions! This particular change is a widely-accepted best practice. I haven't found a good-enough linter yet though. I've been running https://github.com/hadolint/hadolint locally but it has too many false positives to be a net benefit as a pre-merge check.

kevindew:
> Both good questions! This particular change is a widely-accepted best practice. I haven't found a good-enough linter yet though. I've been running https://github.com/hadolint/hadolint locally but it has too many false positives to be a net benefit as a pre-merge check.

Sure thing, well it seems good to go in 👍 

As a nitpicker, it'd be nicer if the commit message said why it was the better practice or linked to something that explained it - I had to lookup https://docs.docker.com/engine/reference/builder/#cmd for context. But I appreciate that as this is multiple repos that might be too arduous a task so don't let that suggestion stop you merging unless you strongly agree it'd be valuable.

PR #2199, opened 2022-11-23
Update publishing api following changes to govuk_schemas and jenkinslib

BeckaL:
This tidies up publishing api for two recent changes to govuk_schemas and jenkinslib:

1. Update to govuk_schemas - The new version of the govuk_schemas gem allows for the path to content schemas to be manually set, rather than using an environment variable (which happens for the deployed version of the app) or using a fallback (which happens for local development). This change means that this path will be used for both the running app and local development.
2. Remove a parameter that is no longer required (the behaviour is now the default in jenkinslib)

[Trello](https://trello.com/c/ZUK2duYI/361-add-govuk-content-schemas-to-publishing-api-repo)


spec/spec_helper.rb
```diff
@@ -33,11 +33,10 @@
   raise <<-MESSAGE
     No schemas found
 
-    The Publishing API tests require that the govuk-content-schemas
+    The Publishing API tests require that content schemas
     are available. These are accessed through the GovukSchemas gem,
```
brucebolt, line 37: Whilst you're editing this file, could `GovukSchemas` be changed to `govuk_schemas` to actually match the name of the gem.

brucebolt (approved):

kevindew (commented):
Looks good, just a small suggestion.

spec/spec_helper.rb
BeckaL: Good point, have opted to remove this message as per Kevin's suggestion below

PR #2201, opened 2022-11-23
Update topic_groups definitions: add content_ids, remove description 

AgaDufrat:
## Context

When a content item is tagged to curated Mainstream browse or Specialist topic and this content item’s slug changes the link disappears from the topic page. This is because the curated lists/groups are based on base paths. 
The “contents” slugs don’t get updated as a part of dependency resolution in Publishing API.

We want to base it on content_ids which are less subjected to change (if at all).

## Changes proposed 

- Mark content_ids as not-deprecated.
The content_ids property was added to enable support of service_manual_groups as it was decided that the should use the existing topic groups definition according to https://github.com/alphagov/govuk-content-schemas/pull/203. The schema was then split into "topic_groups" and “service_manual_topic_groups”  in https://github.com/alphagov/govuk-content-schemas/commit/7bad6a241d3b3e178b66df9e81fd9cc8edc54e95 and the properties marked deprecated for later removal.

- Update the description in line with the recently unified topic terminology
- Remove deprecated `description` property from "topic_groups"

### Notes
- There will be a follow-up PR to add `content_ids` to the required parameters
- We will remove `contents` property after the migration

## Related PRs
- https://github.com/alphagov/collections-publisher/pull/1749
- https://github.com/alphagov/collections/pull/3056
---

### DIFF
--- a/content_schemas/dist/formats/mainstream_browse_page/frontend/schema.json
+++ b/content_schemas/dist/formats/mainstream_browse_page/frontend/schema.json
@@ -667,10 +667,6 @@
               "$ref": "#/definitions/absolute_path"
             }
           },
-          "description": {
-            "description": "DEPRECATED",
-            "type": "string"
-          },
           "name": {
             "type": "string"
           }
--- a/content_schemas/dist/formats/mainstream_browse_page/frontend/schema.json
+++ b/content_schemas/dist/formats/mainstream_browse_page/frontend/schema.json
@@ -657,10 +657,11 @@
         "additionalProperties": false,
         "properties": {
           "content_ids": {
-            "description": "DEPRECATED",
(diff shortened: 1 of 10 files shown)


kevindew:
This looks good, should we be marking the contents field as deprecated? As I guess we don't want both of contents and content_ids?

I feel like we could use a description field on content_ids that says what its purpose is too. Something like "the content_id of content that is in the topic group". A thing that is unclear to me too is whether we care about the ordering of content_id so it'd be good to specify that in the description.

kevindew:
> Will add the description! I thought we may need a short period of having both `contents` and `content_ids` before we change what's used for rendering in Collections.

Yup I expect we do too, be just good to mark that contents is the old way and to be removed.

kevindew (approved):

PR #2204, opened 2022-11-24
Only run dependent app builds when schemas change

BeckaL:
This PR updates publishing api's jenkinsfile to only run dependent apps' CI builds when there are changes to content schemas. This will speed up our builds, which have slowed down since merging in content schemas.

In this PR, which itself contains no changes to schemas, you can see that no dependent builds are being triggered, and the CI log for the build contains the message "no changes to schemas detected, skipping dependent apps stage".

The dependent build behaviour is demonstrated in [this PR](https://github.com/alphagov/publishing-api/pull/2206), where you can see the dependent builds triggered and reported on. Additionally, the CI log contains the message "Schemas change detected in PR: triggering dependent builds".

This can probably be further used to only trigger building the schemas in the PR when changes are detected. Just introducing this as a first step so we can observe its effects, and since it's the biggest speed up win.

[Trello](https://trello.com/c/iZup32bI/362-only-run-schemas-related-tests-and-build-steps-in-publishing-api-when-changes-are-made)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,6 +86,11 @@ def checkGeneratedSchemasAreUpToDate(govuk) {
     }
 }
 
+boolean schemasDirChangedInBranch(branchName) {
+  noChangesToSchemasOnBranch = sh(script: "git diff --exit-code origin/main -- 'content_schemas/dist/' ${branchName} --  'content_schemas/dist/' ", returnStatus: true) == 0
+  return !noChangesToSchemasOnBranch
+}
+
 def checkSchemaDependentProjects() {
 // Run schema tests outside of 'node' definition, so that they do not block the
 // original executor while the downstream tests are being run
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -86,6 +86,11 @@ def checkGeneratedSchemasAreUpToDate(govuk) {
     }
 }
 
+boolean schemasDirChangedInPr(branchName) {
```
brucebolt, line 89: I'd recommend naming this `schemasDirChangedInBranch` instead of `schemasDirChangedInPr` as we run a CI build each time a branch is pushed, even if there isn't a PR.
```suggestion
boolean schemasDirChangedInBranch(branchName) {
```

brucebolt (commented):
Looks good to me.  I've just made one suggestion about the naming of a method.

brucebolt (approved):

PR #2208, opened 2022-11-28
Add GitHub Action workflow for verifying pact tests

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This moves the pact verification to a GitHub Action and will no longer be performed by a Jenkins build. This workflow will also be called by GDS API Adapters as part of its CI build to assert that newly generated pacts are valid against this app. To avoid running this same task twice `pact:verify` is no longer part of the Jenkins test actions and instead the other rake default steps are included.

For more details see https://github.com/alphagov/gds-api-adapters/pull/1175 where the wider change is being co-ordinated.

### DIFF
--- a/.github/workflows/pact-verify.yml
+++ b/.github/workflows/pact-verify.yml
@@ -1,0 +1,56 @@
+# Pact verify workflow
+#
+# This workflow asserts that Pact contract tests are valid against this
+# codebase. It is trigged when changes are made to this project and it
+# is explicitly called by GDS API Adapters when changes are made there.
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM - but what's the plan for the Publishing API / Content Store pact tests?

kevindew:
> LGTM - but what's the plan for the Publishing API / Content Store pact tests?

Thanks, I'm considering that out of scope for now just because GDS API Adapters is so big. In theory could add GitHub Actions for it though.

PR #2211, opened 2022-11-30
Add licence_transaction schema

1pretz1:
Adds schema for the new licence_transaction document type. We'll need the following Specialist Publisher metadata fields:

* country
* sector
* activity 
* continuation_link
* will_continue_on
* licence_identifier

We only hardcode values for 'country' as the values aren't likely to change. `continuation_link` (URL), `will_continue_on` (URL hint) and `licence_identifier` will be user inputted.  We also allow any value for the sector and activity list as 1) it is not yet confirmed and 2) we expect the list to change in time and so only having to update it in Specialist Publisher skips a bit of admin. 

Trello:
https://trello.com/c/mGJ5DIsI/1659-create-licence-document-type-in-specialist-publisher

### DIFF
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -73,6 +73,7 @@
 - knowledge_alpha
 - licence
 - license_finder
+- licence_transaction
 - local_transaction
 - maib_report
 - mainstream_browse_page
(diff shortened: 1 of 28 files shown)


baisa (approved):
Looks good @1pretz1 !

PR #2223, opened 2022-12-12
Add vhost for rabbitmq

BeckaL:
As part of switching to amazonmq, we are now providing many of the env variables needed for bunny via a rabbitMQ URL, rather than individually. However, we still need to explicitly provide the vhost in this case, since this is not contained in the URL.


BeckaL:
Closing since we're now going to add a different vhost into the url (this can actually be specified via the url as it turns out, just '/' as vhost is a special case that you have to url encode to use

PR #2236, opened 2022-12-29
Add fields for Worldwide Organisations

brucebolt:
In https://github.com/alphagov/whitehall/pull/7191, we have added additional fields to be presented as part of Worldwide Organisation content items.  This adds those additional fields to the schema.

Ultimately, this work will enable us to get the content for Worldwide Organisations into the content item, meaning the rendering can be moved out of Whitehall.

[Trello card](https://trello.com/c/3r6TFSCF)

### DIFF
--- a/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
+++ b/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
@@ -8,10 +8,7 @@
       type: "string",
     },
     crest: {
-      type: [
-        "string",
-        "null",
-      ],
+      type: "string",
       enum: [
         "bis",
         "dit",
@@ -24,7 +21,6 @@
         "so",
         "ukaea",
         "wales",
-        null,
       ],
     },
     image: {
--- a/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
+++ b/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
@@ -1,0 +1,32 @@
+{
+  type: "object",
+  properties: {
+    formatted_title: {
+      type: "string",
+    },
+    crest: {
+      type: [
+        "string",
+        "null",
+      ],
+      enum: [
+        "bis",
+        "dit",
+        "eo",
+        "hmrc",
+        "ho",
+        "mod",
+        "portcullis",
+        "single-identity",
+        "so",
+        "ukaea",
+        "wales",
+        null,
+      ],
+    },
+    image: {
+      "$ref": "#/definitions/image",
+    },
+  },
+  description: "The organisation's logo, including the logo image and formatted name.",
+}
--- a/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
+++ b/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
@@ -3,6 +3,7 @@
   required: [
     "formatted_title",
   ],
+  additionalProperties: false,
   properties: {
     formatted_title: {
       type: "string",
--- a/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
+++ b/content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
@@ -1,5 +1,8 @@
 {
   type: "object",
+  required: [
+    "formatted_title",
+  ],
   properties: {
     formatted_title: {
       type: "string",
--- a/content_schemas/formats/worldwide_organisation.jsonnet
+++ b/content_schemas/formats/worldwide_organisation.jsonnet
@@ -4,6 +4,7 @@
       type: "object",
       additionalProperties: false,
       properties: {
+        logo: (import "shared/definitions/_organisation_logo.jsonnet"),
         ordered_corporate_information_pages: {
           type: "array",
           items: {
--- a/content_schemas/formats/worldwide_organisation.jsonnet
+++ b/content_schemas/formats/worldwide_organisation.jsonnet
@@ -30,6 +30,9 @@
     },
     links: (import "shared/base_links.jsonnet") + {
       corporate_information_pages: "Corporate information pages for this Worldwide Organisation"
+      ordered_contacts: "Contact details for this Worldwide Organisation",
+      sponsoring_organisations: "Sponsoring organisations for this Worldwide Organisation",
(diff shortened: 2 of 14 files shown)


content_schemas/formats/shared/definitions/_logo.jsonnet
```diff
@@ -0,0 +1,32 @@
+{
```
kevindew, line 1: Shall we call this file organisation logo? I'm wondering if logo is a bit too generic as it's definitely constrained to a few scenarios.

content_schemas/formats/shared/definitions/_logo.jsonnet
```diff
@@ -0,0 +1,32 @@
...
+  type: "object",
```
kevindew, line 2: Are any of the properties here required?

content_schemas/formats/shared/definitions/_logo.jsonnet
```diff
@@ -0,0 +1,32 @@
...
+  properties: {
+    formatted_title: {
+      type: "string",
+    },
+    crest: {
```
kevindew, line 7: This looks a bit complex with allowing string and null and then having null in the enum. Though this is just a move I wonder if we can simplify it.

I wonder if it'd be better to require a string for this and then to just not set the property all together if this isn't defined (i.e. capitalise on it not being a required property)

content_schemas/formats/shared/definitions/_logo.jsonnet
```diff
@@ -0,0 +1,32 @@
+{
+  type: "object",
+  properties: {
```
kevindew, line 3: Should this also have additional properties set to false?

content_schemas/formats/worldwide_organisation.jsonnet
```diff
@@ -4,6 +4,26 @@
...
+            additionalProperties: false,
+            required: [
+              "title",
+              "href",
+            ],
+            properties: {
+              title: {
+                type: "string",
+              },
+              href: {
```
kevindew, line 21: I'm probably drifting into the optimistic here, but are these corporate information pages available via the links hash? I'm a bit concerned that title and href are rather delicate attributes that have a tight coupling.

I'm wondering if this could be done entirely through expanded links, assuming title is not customised and if title is customised here whether the href could be a content_id instead so the href can be pulled out of expanded links

kevindew (commented):
Looks good 👍  just a few suggestions.

kevindew:
Oh and for future, when editing schemas it's a bit easier to review if you have a single commit at the end that regenerates the schemas, otherwise it's quite hard to determine what are your changes amongst the regeneration ones (aside, I dream of us not committing the generated schemas and them being a deploy time artefact)

content_schemas/formats/worldwide_organisation.jsonnet
brucebolt: Unfortunately the title comes from Whitehall arbitrarily defining what to include here (sometimes it doesn't match the name of the document, being pulled in from the locale files).  It's the same for organisation pages too.

content_schemas/formats/shared/definitions/_logo.jsonnet
brucebolt: Good idea, I'll rename this.

content_schemas/formats/worldwide_organisation.jsonnet
kevindew: Sure, that seems fine, would the href be able to be a content_id then so the document can be matched from the links rather than a hardcoded URL?

brucebolt:
Thanks @kevindew, looking at addressing your comments now.

> when editing schemas it's a bit easier to review if you have a single commit at the end that regenerates the schemas

Noted.  I'll update this so it's easier to re-review.

content_schemas/formats/shared/definitions/_logo.jsonnet
brucebolt: Looks like the title is required, but the crest and image are optional.  I'll add this as required.

content_schemas/formats/shared/definitions/_logo.jsonnet
brucebolt: I've removed both `null` from this property as it's optional anyway.

content_schemas/formats/shared/definitions/_logo.jsonnet
brucebolt: Yes, it should.  Have added that in.

content_schemas/formats/worldwide_organisation.jsonnet
brucebolt: That should be possible.  I'll have to add the pages in as links first.

brucebolt: I've added these in as links and replaced the `href` in the details hash with a content ID.

content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
```diff
@@ -10,7 +10,6 @@
     crest: {
       type: [
         "string",
-        "null",
       ],
```
kevindew, line 14:
```suggestion
      type: "string"
```

minor, you don't need to define this in an array.

kevindew (approved):
Looks great 👍 thanks for the changes. Just a minor syntactic issue.

Do we know if the existing organisation type is sending any data with a crest of null? I imagine we need to ensure Whitehall is omitting the fields when the data doesn't exist?

content_schemas/formats/shared/definitions/_organisation_logo.jsonnet
brucebolt: Thanks.  I'll fix this.

brucebolt:
> Do we know if the existing organisation type is sending any data with a crest of null? I imagine we need to ensure Whitehall is omitting the fields when the data doesn't exist?

That shouldn't be happening as [we are using `compact` on the logo hash](https://github.com/alphagov/whitehall/blob/4aec232784a537e581b4962ba5c49f9b0b5366ea/app/presenters/publishing_api/organisation_presenter.rb#L77-L81) so the nil values are all being removed.

kevindew:
Great stuff 👍 merge away

PR #2237, opened 2022-12-29
Add youtube video url and make images optional for promotional features

davidgisbey:
## Description

We're giving executive organisations the option to provide a Youtube video URL instead of an image so they can embed
it on the org page.

This PR does adds functionality to allow a promotional feature to either have an image or a youtube video id passed in the items hash, but not both.

## Related PRs

Collections - https://github.com/alphagov/collections/pull/3116
Whitehall - https://github.com/alphagov/whitehall/pull/7193

## Trello card 

https://trello.com/c/hEEcf7QA/1007-add-a-video-field-to-whitehall-promotional-features

### DIFF
--- a/content_schemas/examples/organisation/frontend/number_10.json
+++ b/content_schemas/examples/organisation/frontend/number_10.json
@@ -1574,10 +1574,7 @@
             "title": "",
             "href": "",
             "summary": "The Number 10 media blog provides an opportunity to share the government’s position on a wide range of issues directly to the public.",
-            "image": {
-              "url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/promotional_feature_item/image/67/No10-door-image.png",
-              "alt_text": ""
-            },
+            "youtube_video_id": "fFmDQn9Lbl4",
             "double_width": false,
             "links": [
               {
--- a/content_schemas/formats/organisation.jsonnet
+++ b/content_schemas/formats/organisation.jsonnet
@@ -86,54 +86,10 @@
               items: {
                 type: "array",
                 items: {
-                  type: "object",
-                  additionalProperties: false,
-                  required: [
-                    "summary",
-                    "double_width",
+                  anyOf: [
+                    { "$ref": "#/definitions/promotional_feature_item_image" },
+                    { "$ref": "#/definitions/promotional_feature_youtube" },
                   ],
-                  properties: {
-                    title: {
-                      type: [
-                        "string",
-                        "null",
-                      ],
-                    },
-                    href: {
-                      type: [
-                        "string",
-                        "null",
-                      ],
-                    },
-                    summary: {
-                      type: "string",
-                    },
-                    image: {
-                      "$ref": "#/definitions/image",
-                    },
-                    double_width: {
-                      type: "boolean",
-                    },
-                    links: {
-                      type: "array",
-                      items: {
-                        type: "object",
-                        additionalProperties: false,
-                        required: [
-                          "title",
-                          "href",
-                        ],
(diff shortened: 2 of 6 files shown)


content_schemas/formats/organisation.jsonnet
```diff
@@ -140,7 +140,20 @@
...
                     },
                     image: {
-                      "$ref": "#/definitions/image",
+                      "oneOf": [
+                        {
+                          "$ref": "#/definitions/image",
+                        },
+                        {
+                          "type": "null"
+                        }
```
kevindew, line 149: I wonder if we need this to be allowed to be null. As this is not a required property we just need to not set the attribute rather than sending a null value.

content_schemas/formats/organisation.jsonnet
```diff
@@ -140,7 +140,20 @@
...
+                      ]
+                    },
+                    youtube_video_id: {
+                      type: [
+                        "string",
+                        "null",
```
kevindew, line 155: Similarly here I imagine we can just set this to a string if it exists otherwise not set the attribute, as it's also not a required property.

kevindew (commented):
Looks good, just some questions on the null handling and the multiple ways things can be done.

davidgisbey:
@kevindew would you mind having a look at this if you get a change please 🙏

content_schemas/examples/organisation/frontend/number_10.json
```diff
@@ -1574,10 +1574,7 @@
             "title": "",
             "href": "",
             "summary": "The Number 10 media blog provides an opportunity to share the government’s position on a wide range of issues directly to the public.",
-            "image": {
-              "url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/promotional_feature_item/image/67/No10-door-image.png",
-              "alt_text": ""
-            },
+            "youtube_video_id": "fFmDQn9Lbl4",
```
kevindew, line 1577: Nice! I imagine F&V can use this example schema to develop against (well if such facilities still exist - they did once)

kevindew (commented):
I think I mislead you a bit on this in my advice, sorry about that. Aside from that it's looking good and could just be rejigged a bit to be a bit simpler.

content_schemas/formats/organisation.jsonnet
```diff
@@ -73,70 +73,10 @@
...
           type: "array",
           items: {
-            type: "object",
-            additionalProperties: false,
-            required: [
-              "title",
-              "items",
+            oneOf: [
+              { "$ref": "#/definitions/image_promotional_feature" },
+              { "$ref": "#/definitions/youtube_promotional_feature" },
```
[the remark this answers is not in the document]
davidgisbey: I've realised that oneOf doesn't work here as it expects all promotional feature items to be either `promotional_feature_item_image` or `promotional_feature_youtube`

Using anyOf works though, and unless i'm misunderstanding how it works would still blow up as both don't allow additional properties.

davidgisbey:
Thanks for the review. I've pushed up the changes.

content_schemas/examples/organisation/frontend/number_10.json
davidgisbey: Yup absolutely 👍

content_schemas/formats/organisation.jsonnet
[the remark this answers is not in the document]
kevindew: Oh yes, good catch. Yes anyOf sounds correct.

kevindew (approved):
Looks good to me 👍

PR #2246, opened 2023-01-09
Upgrade sidekiq

JonathanHallam:
Upgrades govuk sidekiq to the latest version


Gemfile
```diff
@@ -12,7 +12,7 @@ gem "govspeak"
 gem "govuk_app_config"
 gem "govuk_document_types"
 gem "govuk_schemas"
-gem "govuk_sidekiq"
+gem "govuk_sidekiq", "6.0.0"
```
brucebolt, line 15: Do we need to pin this to exactly version 6.0.0?

kevindew: Nope, it's against the rules: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html

config/initializers/sidekiq.rb
```diff
@@ -6,6 +6,7 @@
 end
 
 Sidekiq.configure_server do |config|
+  config.log_formatter = Sidekiq::Logger::Formatters::JSON.new
```
kevindew, line 9: I'd have thought you'd not need this: https://github.com/alphagov/govuk_sidekiq/blob/77298fcf54a66514a384e34a35b643be3c16d2d3/lib/govuk_sidekiq/sidekiq_initializer.rb#L15

config/initializers/sidekiq.rb
```diff
@@ -6,6 +6,7 @@
 end
```
kevindew, line 6: Shall we add: https://github.com/alphagov/link-checker-api/pull/616/files#diff-fe63ad82c410f651197078bd21118c1fa7a9dbb0a468e90940e76ba6c07e87eaR22-R24 to prevent config regressions?

kevindew (commented):
Nice work on this. I was just doing so consistency checks with link-checker-api.

I think we also need to remove: https://github.com/alphagov/publishing-api/blob/main/config/sidekiq.yml#L4-L6 as it'll be redundant.

brucebolt (commented):
As we're using this PR as an example for other teams, I think a little more explanation in the commit messages as to why we're making that change would be useful, as it'll help the other teams to understand why they need to make the same change in their apps.

app/commands/base_command.rb
```diff
@@ -17,7 +17,6 @@ def self.call(payload, downstream: true, nested: false, callbacks: [], **options
           ).call
         end
       end
-
```
BeckaL, line 20: This looks like a slip - can we remove?

BeckaL (commented):
Generally looks good to me, echoing Bruce's comments about putting some more detail into the commit messages, especially if we're going to use this as an example for other teams. For example, would be good to explain exactly why we need to stringify keys.

Gemfile
JonathanHallam: woops, sorry

config/initializers/sidekiq.rb
JonathanHallam: oops, good shout

app/commands/base_command.rb
JonathanHallam: 👍

brucebolt (approved):
Looks good to me 👍.  Thanks for updating the commit messages, they're much more informative now.

sihugh:
As this PR is an example, I wanted to reference this logging fix which has been necessary on some other apps: https://github.com/alphagov/account-api/pull/577

PR #2254, opened 2023-01-11
Add Content Schema tests

theseanything:
When content schemas change we need to run tests for all the dependant
apps. This uses two workflows as GitHub limits the number of workflow
calls to 20.

### DIFF
--- a/.github/workflows/test-content-schemas-1.yml
+++ b/.github/workflows/test-content-schemas-1.yml
@@ -1,0 +1,113 @@
+name: Test Content Schemas (1)
+
+on:
+  workflow_dispatch: {}
+  push:
+    branches:
+      - main
+    paths:
+      - "content_schemas/**"
+      - "!content_schemas/README.md"
+  pull_request:
+    paths:
+      - "content_schemas/**"
+      - "!content_schemas/README.md"
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-content-schemas-1.yml
```diff
@@ -0,0 +1,113 @@
...
+  pull_request:
+    paths:
+      - "content_schemas/**"
+      - "!content_schemas/README.md"
+
+jobs:
+  test-collections-publisher:
+    name: Test Collections Publisher
+    uses: alphagov/collections-publisher/.github/workflows/rspec.yml@main
+    with:
```
sengi, line 20: optional: worth using a YAML anchor to factor out the repeated block?

sengi (approved):

.github/workflows/test-content-schemas-1.yml
theseanything: Good shout! Unfortunately, at the moment, [workflows don't support YAML anchors](https://github.com/actions/runner/issues/1182) 😞

sengi: Ohh sorry hadn't come across that. Thanks!

PR #2257, opened 2023-01-12
More efficient approach to checking /v2/editions pagination

kevindew:
This replaces the pagination approach used in keyset pagination with one that:

a) doesn't look up pages if it can infer whether we're on first or last page (based on any params for first page, less than per_page results for last page)
b) does a basic unordered exists? query for checking for further pages rather than the full one previously used - this saves computation for determining ordering

This allows the performance of these queries to be improved. The motivation for adding this was adding a new field and discovering that the pagination checks performed worse than the initial query. I've then broken this off into a distinct commit and it's likely the full benefits of this will demonstrated in a later commit.

Benchmarks are shown below.

The motivation for this change was to improve the performance of the /v2/editions endpoint for the [Contentful Spike](https://trello.com/c/Yt4gZGRl/373-configure-publishing-api-for-contentful-listener-api). However this does not need to be part of that work and seems better merged in independently, rather than lingering away in a spike branch. 

---

Before:

```

publishing-api=# EXPLAIN ANALYZE SELECT "content_id", "editions"."updated_at", "editions"."id" FROM "editions" INNER JOIN "documents" ON "documents"."id" = "editions"."document_id" WHERE "editions"."state" IN ('draft', 'published', 'unpublished') AND "documents"."locale" = 'cy' AND ((editions.updated_at, editions.id) > ('2018-08-24T11:05:17.251903Z', '3308026')) ORDER BY "editions"."updated_at" ASC, "editions"."id" ASC LIMIT 1;
                                                                                QUERY PLAN

--------------------------------------------------------------------------------------------------------------------------
-------------------------------------------------
 Limit  (cost=0.86..819.81 rows=1 width=28) (actual time=851.106..851.108 rows=1 loops=1)
   ->  Nested Loop  (cost=0.86..11771551.45 rows=14374 width=28) (actual time=851.102..851.103 rows=1 loops=1)
         ->  Index Scan using index_editions_on_updated_at_and_id on editions  (cost=0.43..9484246.68 rows=634163 width=16
) (actual time=0.268..827.864 rows=2211 loops=1)
               Index Cond: (ROW(updated_at, id) > ROW('2018-08-24 11:05:17.251903'::timestamp without time zone, 3308026))
               Filter: ((state)::text = ANY ('{draft,published,unpublished}'::text[]))
               Rows Removed by Filter: 46764
         ->  Index Scan using documents_pkey on documents  (cost=0.43..3.61 rows=1 width=20) (actual time=0.010..0.010 row
s=0 loops=2211)
               Index Cond: (id = editions.document_id)
               Filter: ((locale)::text = 'cy'::text)
               Rows Removed by Filter: 1
 Planning Time: 1.019 ms
 Execution Time: 851.245 ms
(12 rows)
```

After:

```
publishing-api=# EXPLAIN ANALYZE SELECT 1 AS one FROM "editions" INNER JOIN "documents" ON "documents"."id" = "editions"."document_id" WHERE "editions"."state" IN ('draft', 'published', 'unpublished') AND "documents"."locale" = 'cy' AND ((editions.updated_at, editions.id) > ('2018-08-24T11:05:17.251903Z', '3308026')) LIMIT 1;
                                                                          QUERY PLAN

--------------------------------------------------------------------------------------------------------------------------
-------------------------------------
 Limit  (cost=0.43..44.65 rows=1 width=4) (actual time=0.380..0.387 rows=1 loops=1)
   ->  Nested Loop  (cost=0.43..635622.72 rows=14374 width=4) (actual time=0.379..0.379 rows=1 loops=1)
         ->  Seq Scan on documents  (cost=0.00..23408.65 rows=24057 width=4) (actual time=0.059..0.059 rows=1 loops=1)
               Filter: ((locale)::text = 'cy'::text)
               Rows Removed by Filter: 53
         ->  Index Scan using index_editions_on_document_id_and_state on editions  (cost=0.43..25.42 rows=3 width=4) (actu
al time=0.318..0.318 rows=1 loops=1)
               Index Cond: ((document_id = documents.id) AND ((state)::text = ANY ('{draft,published,unpublished}'::text[]
)))
               Filter: (ROW(updated_at, id) > ROW('2018-08-24 11:05:17.251903'::timestamp without time zone, 3308026))
 Planning Time: 2.861 ms
 Execution Time: 0.549 ms
(10 rows)
```

### DIFF
--- a/app/queries/keyset_pagination.rb
+++ b/app/queries/keyset_pagination.rb
@@ -32,7 +32,6 @@ module Queries
 
       if before
         @previous = before
-        @order = order == :asc ? :desc : :asc
         @direction = :backwards
       else
         @previous = after
@@ -57,17 +56,31 @@ module Queries
     end
 
     def is_first_page?
-      @is_first_page ||= ordered_initial_results.empty? ||
-        KeysetPagination.new(
-          client, per_page: 1, before: next_before_key
-        ).empty?
+      if @is_first_page.nil?
+        # We can assume that a lack of pagination data means we're on the first page
+        if previous.nil? || ordered_initial_results.empty?
+          @is_first_page = true
+        else
+          where = where_clause(greater_than: order != :asc)
+          @is_first_page = !client.initial_query.where(where, *next_before_key).exists?
+        end
+      end
+
+      @is_first_page
     end
 
     def is_last_page?
-      @is_last_page ||= ordered_initial_results.empty? ||
-        KeysetPagination.new(
-          client, per_page: 1, after: next_after_key
-        ).empty?
+      if @is_last_page.nil?
+        if ordered_initial_results.empty? || ordered_initial_results.count < per_page
+          # If we've got less results than a page we can assume this is the last page
+          @is_last_page = true
+        else
+          where = where_clause(greater_than: order == :asc)
+          @is_last_page = !client.initial_query.where(where, *next_after_key).exists?
+        end
+      end
+
+      @is_last_page
     end
 
     def key_fields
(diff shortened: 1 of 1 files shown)


app/queries/keyset_pagination.rb
```diff
@@ -57,17 +56,31 @@ def next_after_key
     end
 
     def is_first_page?
-      @is_first_page ||= ordered_initial_results.empty? ||
-        KeysetPagination.new(
-          client, per_page: 1, before: next_before_key
-        ).empty?
+      if @is_first_page.nil?
```
BeckaL, line 59: Could this be simplified to 
```    def is_first_page?
      @if_first_page ||=
        # We can assume that a lack of pagination data means we're on the first page
        if previous.nil? || ordered_initial_results.empty?
          true
        else
          where = where_clause(greater_than: order != :asc)
          !client.initial_query.where(where, *next_before_key).exists?
        end
    end```

and same with the is_last_page method?

BeckaL (approved):
Just a suggested refactor, but will approve and let you decide, logic looks good to me

app/queries/keyset_pagination.rb
kevindew: We can't unfortunately as, in the event of a false value, the code we be re-executed each time it's called. I.e if @if_first_page is set to false then `@if_first_page ||=` will be executed each time the method is called. 

Instead this code checks uses a check on whether the ivar is nil which will mean the execution is skipped when it is true or false.

kevindew: Thanks, life would certainly be easier if there was an operator that acted as a "set this if it's not already set" it's frustrating when you want to memoise false and nil values.

PR #2260, opened 2023-01-17
Add schemas for historic appointment pages

brucebolt:
This adds the `historic_appointments` and `historic_appointment` schemas, which will be used for the 'Past Prime Ministers' (i.e. https://www.gov.uk/government/history/past-prime-ministers) and 'Past Prime Minister' pages (e.g. https://www.gov.uk/government/history/past-prime-ministers/neville-chamberlain).

They are named as such so they can be re-used if the 'Past Chancellors' and 'Past Foreign Secretaries' pages are ever moved from being hardcoded.

The contents of the schema will be added in later work, when we begin to build the presenter in Whitehall.

[Trello card](https://trello.com/c/xUout0dR)

### DIFF
--- a/content_schemas/dist/formats/historic_appointment/frontend/schema.json
+++ b/content_schemas/dist/formats/historic_appointment/frontend/schema.json
@@ -1,0 +1,797 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "required": [
+    "base_path",
+    "content_id",
+    "description",
+    "details",
+    "document_type",
+    "links",
+    "locale",
+    "public_updated_at",
+    "schema_name",
+    "title",
+    "updated_at"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "analytics_identifier": {
+      "$ref": "#/definitions/analytics_identifier"
+    },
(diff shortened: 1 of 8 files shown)


content_schemas/formats/historic_appointments.jsonnet
```diff
@@ -0,0 +1,9 @@
+(import "shared/default_format.jsonnet") + {
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {},
```
BeckaL, line 6: I'm a bit confused here - is the point of this PR just to add a basic schema, and then iterate it afterwards? Why no properties defined here?
Was imagining the shape of the schema would roughly be a list of time periods, which have titles, and then contain list of appointment details.
Same question with the other format you added

brucebolt: > I'm a bit confused here - is the point of this PR just to add a basic schema, and then iterate it afterwards?

Yes, my thought was to create an empty schema then we can add to it when updating the Whitehall presenter.  I could add the fields now if that's the preferred outcome for this card?

BeckaL (approved):
LGTM as basis of these two schemas

PR #2301, opened 2023-03-03
Fix AWS auth methods other than secret key.

sengi:
The way we were initialising the AWS SDK was preventing any authentication methods other than secret key from working.

Only pass an access key / secret key if it's actually specified in the environment, otherwise [let the SDK work as documented](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/#Configuration).

This allows, for example, authentication via EC2 instance profile creds so that we can stop using long-lived secret keys.


dj-maisy (approved):
Bit easier to read. Nice. 👍

PR #2302, opened 2023-03-03
Fix magic class names for DB migrations.

sengi:
7e66d46 / #2048 broke database migrations whose class names contained the string `Json`, because it changed the logic for magically generating the camel-cased class name from the source filename.

Update the class names to match what Rails now expects them to be.

Perfect illustration of why such magic is a terrible idea, but Rails makes it hard to avoid this sort of thing.

Tested: `rails db:migrate` succeeds on integration with this patch.

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -10,7 +10,6 @@
 #   inflect.uncountable %w( fish sheep )
 # end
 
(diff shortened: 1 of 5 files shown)


JonathanHallam (approved):
nice one, thanks Chris

PR #2303, opened 2023-03-06
Add schema for 'How Government Works' page

brucebolt:
This is currently hardcoded in Whitehall but will need a content item to enable rendering in another application.

Note that all the details fields are optional as they will not be present when "reshuffle mode" is switched on.

[Trello card](https://trello.com/c/ckwRKhIR)

### DIFF
--- a/content_schemas/formats/how_government_works.jsonnet
+++ b/content_schemas/formats/how_government_works.jsonnet
@@ -1,0 +1,76 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: [
+    "how_government_works"
+  ],
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {
+        department_counts: {
+          description: "Number of departments by type",
+          type: "object",
+          properties: {
+            ministerial_departments: {
+              description: "Number of ministerial departments",
+              type: "integer",
+            },
+            non_ministerial_departments: {
+              description: "Number of non-ministerial departments",
+              type: "integer",
+            },
+            agencies_and_public_bodies: {
+              description: "Number of agencies and public bodies",
+              type: "integer",
+            }
+          },
+          required: [
+            "ministerial_departments",
+            "non_ministerial_departments",
+            "agencies_and_public_bodies",
+          ],
+        },
+        ministerial_role_counts: {
+          description: "Number of ministerial role appointments by type",
+          type: "object",
+          properties: {
+            prime_minister: {
+              description: "Number of current people in the prime minister role",
+              type: "integer",
+            },
+            cabinet_ministers: {
+              description: "Number of current people in cabinet roles",
+              type: "integer",
+            },
+            other_ministers: {
+              description: "Number of current people in non-cabinet ministerial roles",
+              type: "integer",
+            },
+            total_ministers: {
+              description: "Number of current ministers",
+              type: "integer",
+            }
+          },
+          required: [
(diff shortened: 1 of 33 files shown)


content_schemas/formats/how_government_works.jsonnet
```diff
@@ -0,0 +1,73 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: [
+    "how_government_works"
+  ],
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
```
BeckaL, line 8: Do we want to add the reshuffle mode as a required field to this details hash?

BeckaL (commented):
In general this looks good to me, a suggestion to add something into the required properties for details.

The only other things that appear on the page that I could see being added to this schema (famous last words TM) is all the currently hard-coded organisations that appear on the page as examples (e.g. "Some departments, like the [Ministry of Defence](https://www.gov.uk/government/organisations/ministry-of-defence), cover the whole UK"). If they get reslugged / name changed, they're not going to get updated on this page. We could add them in as links e.g. `"uk_wide_org_example": MOD_CONTENT_ID` etc, but this still requires an element of hard-coding this end (similar to [this](https://github.com/alphagov/whitehall/blob/15ed64ee5616e097eba47b325c12333ec4f826db/app/presenters/publishing_api/operational_field_presenter.rb#L5)). So yeah, we're not going to get the benefits of dependency resolution here, but maybe we tolerate that.

brucebolt:
> The only other things that appear on the page that I could see being added to this schema (famous last words TM) is all the currently hard-coded organisations that appear on the page as examples (e.g. "Some departments, like the [Ministry of Defence](https://www.gov.uk/government/organisations/ministry-of-defence), cover the whole UK"). If they get reslugged / name changed, they're not going to get updated on this page.
> 
> We could add them in as links e.g. `"uk_wide_org_example": MOD_CONTENT_ID` etc, but this still requires an element of hard-coding this end (similar to [this](https://github.com/alphagov/whitehall/blob/15ed64ee5616e097eba47b325c12333ec4f826db/app/presenters/publishing_api/operational_field_presenter.rb#L5)). So yeah, we're not going to get the benefits of dependency resolution here, but maybe we tolerate that.

I did think about adding these, but then realised there are 10 organisations named in this page, so we'd need to include them all.  In many changes (e.g. DIT being replaced by DBT), the existing organisation isn't renamed but replaced with a new one, so this would need manual changes anyway.  Additionally, at the moment, the names and slugs are hardcoded, and we update them manually when changes are made (e.g. [changing Her Majesty's Inspectorate of Prisons to His Majesty's Inspectorate of Prisons](https://github.com/alphagov/whitehall/commit/8597bb1120a6232a94fed03e92d86e4bfcdc9c4c#diff-f4c7ac711dede2bb393689aff5b8be18efa5fa73a5361b34d71f14f93c2191d8)).  So I don't think these are worth adding.  Maybe Find & View might want to review the content on this page once we've migrated it, to see what is actually the user need of these examples?

content_schemas/formats/how_government_works.jsonnet
brucebolt: Good idea 👍

BeckaL:
> I did think about adding these...

Yep, agree with all your points here. Happy to leave as is.

BeckaL (approved):
🥳

PR #2314, opened 2023-03-14
Add temporary rake task for FOI request

brucebolt:
We have received a FOI request for a report on updates to ministers on GOV.UK since 2022-01-01.

This rake task will allow us to produce that report.

The task will be deleted after it has been run once.

[Trello card](https://trello.com/c/pTGig4hE)

### DIFF
--- a/lib/tasks/temp_foi.rake
+++ b/lib/tasks/temp_foi.rake
@@ -1,0 +1,30 @@
+desc "Temporary rake task to extract data for a FOI request"
+task temp_foi: :environment do
+  role_appointments = Edition
+    .where("published_at >= ?", "2022-01-01")
+    .where(schema_name: "role_appointment", content_store: "live")
+    .order(:published_at)
+
+  ministerial_role_appointments = role_appointments.select do |role_appointment|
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
Looks good to me, nice one! I wonder if we'll eventually want to present the timestamps in a more human readable format, but I guess we can do that with the data we get.

lib/tasks/temp_foi.rake
```diff
@@ -0,0 +1,30 @@
...
+    role_edition.document_type == "ministerial_role"
+  end
+
+  final = ministerial_role_appointments.map do |role_appointment|
+    {
+      name: role_appointment.title,
+      started_on: role_appointment.details[:started_on],
+      ended_on: role_appointment.details[:ended_on],
+      current: role_appointment.details[:current],
+      published_at: role_appointment.public_updated_at,
```
BeckaL, line 21: Sorry - a thought here post approval, should we we using the published at here instead?

brucebolt: Yes, you're right.  I don't think there's much difference, but I'll use `published_at` anyway.

PR #2316, opened 2023-03-15
Create BulkOrganisationUpdater

baisa:
[Trello card](https://trello.com/c/d8AMqhuq/1192-machinery-of-government-retagging-mainstream-documents)


lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def find_organisations(row)
+      all_organisations = Queries::GetLinkables.new(document_type: "organisation").call
+      organisations = row.fetch("All organisations")
+      return [] unless organisations
+
+      new_organisations = []
+      organisations.split(",").map do |slug|
+        organisation = all_organisations.detect { |org| org.base_path == "/government/organisations/#{slug}" }
```
ollietreend, line 50: I think this code might struggle with rogue whitespace.

Given an example row:

| Path | All organisations |
| --- | --- |
| the-warm-home-discount-scheme | department-for-energy-security-and-net-zero, department-for-business-energy-and-industrial-strategy |

There are spaces in between each of the slugs.

When we split them by `","`:

```ruby
"dept-one, dept-two, dept-three".split(",") # => ["dept-one", " dept-two", " dept-three"]
```

Note the `space` character at the beginning of the second and third slugs.

I think this will result in problems when trying to match the organisation by `base_path`:

```ruby
"/government/organisations/#{slug}" # => "/government/organisations/ dept-two"
```

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -0,0 +1,67 @@
...
+        if organisation.nil?
+          puts "error: #{slug}: could not find edition"
```
ollietreend, line 52: This error could be misleading.

```suggestion
          puts "error: #{slug}: could not find organisation"
```

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+
+    def find_organisations(row)
+      organisations = row.fetch("All organisations")&.strip
+      return [] unless organisations
+
+      new_organisations = []
+      organisations.split(",").map do |slug|
+        organisation = @all_organisations.detect { |org| org.base_path.strip == "/government/organisations/#{slug}".strip }
```
ollietreend, line 51: I think we're using `strip` in the wrong places right now. We need to strip the `slug` variable, and nothing else.

I've also switched the use of `map` for `each` since we're not concerned with the output value. Instead, we're appending to the `new_organisations` array, so `each` is fine here.

Sorry to be pernickety 🙈 but how about this?

```suggestion
      organisations.split(",").map(&:strip).each do |slug|
        organisation = @all_organisations.detect { |org| org.base_path == "/government/organisations/#{slug}" }
```

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -0,0 +1,68 @@
...
+    attr_reader :filename
+
+    def process_row(row)
+      document = find_document(row)
+      return unless document
+
+      organisations = find_organisations(row)
+      update_document(document, organisations)
+
+      puts "Sucessfully updated"
```
ollietreend, line 31: How about including something identifiable, so we know which row has successfully updated?

```suggestion
      path = row.fetch("Path")&.strip
      puts "Sucessfully updated: #{path}"
```

ollietreend (approved):
This looks good to me @baisa 👍🏻 

As you've said in the description: if we decide to keep this Rake task, then we should add test coverage.

PR #2328, opened 2023-03-23
Add schema for world index page

brucebolt:
This schema will be used for the content item hosted at `/world`.

[Trello card](https://trello.com/c/7kBEEWUL)

### DIFF
--- a/content_schemas/formats/world_index.jsonnet
+++ b/content_schemas/formats/world_index.jsonnet
@@ -1,0 +1,73 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: [
+    "world_index"
+  ],
+
+  definitions: (import "shared/definitions/_whitehall.jsonnet") + {
+    details: {
+      type: "object",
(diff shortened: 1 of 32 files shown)


content_schemas/formats/world_index.jsonnet
```diff
@@ -0,0 +1,73 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: [
+    "world_index"
```
BeckaL, line 3: I like this, I guess if we used this consistently in other schemas we wouldn't have any new document type modifying almost every schema. Maybe a future improvement for the backburner somewhere.

BeckaL (approved):
All looks good to me, I like that we're being on it specifying things as required fields, makes the schemas (theoretically) a lot more useful

brucebolt:
The only failing test is against the frontend app, and the same test is failing on the main branch of frontend, so this should be safe to merge.

PR #2330, opened 2023-03-28
Add migration to remove path reservation for `/world`

brucebolt:
There is a path reservation for `/world` reserved by Content Tagger. No content is currently published to this path and there is no intention for Content Tagger to use this path for anything.

Therefore deleting it so Whitehall is able to publish the world index page at the same path.

Adding this as a migration so there is a record of when this was removed (and why), and also to make it safer to run than executing commands on a console.

[Trello card](https://trello.com/c/7kBEEWUL)


BeckaL (approved):
Looks good to me. 

I'm sure I read somewhere in the replatforming stuff that the new platform didn't have the same guarantees about when a migration is run so it was more important to add the `down` migration, but a) I can't find reference to where that was said and b) it won't matter for this one since it's not adding any application code. So writing this down to get this into other people's heads / so I have a higher chance of remembering when it matters.

PR #2344, opened 2023-04-18
Update link expansion rules for ministers index page

brucebolt:
This updates the expansion rules to include the `role_appointment` and `role` for each person who needs to be included in the ministers index page.

It also gives us their status as a privy counsellor and their whip payment status.


chrislo (approved):
Looks good to me. A minor stylistic point about the commit message, here we say, for example: 

> the role and role appointment for each person listed on the ministers index page.

and I wonder if we could add something for our future selves that explains why? Maybe it's a user need we're addressing, or someone/department has asked us? It might help our future selves to understand the changes if we have a bit of that context.

PR #2349, opened 2023-04-25
Add srcset support to the logo image for topical events

KludgeKML:
Draft for this https://trello.com/c/uW5kvqga/1961-using-srcset-functionality-to-make-images-less-blurry-on-govuk to demonstrate the concept

- [ ] This application is owned by the publishing platform team. Please let us know in #govuk-publishing-platform when you raise any PRs.

### DIFF
--- a/content_schemas/dist/formats/topical_event/frontend/schema.json
+++ b/content_schemas/dist/formats/topical_event/frontend/schema.json
@@ -490,15 +490,7 @@
           "format": "date-time"
         },
         "image": {
-          "properties": {
-            "alt_text": {
-              "type": "string"
-            },
-            "url": {
-              "type": "string",
-              "format": "uri"
-            }
-          }
+          "$ref": "#/definitions/image_with_srcset"
         },
         "ordered_featured_documents": {
           "description": "A set of featured documents to display for the Topical Event.",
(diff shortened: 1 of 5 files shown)


kevindew:
Hey interesting stuff here @KludgeKML, I've been talking about this with Kuba about this in the lead-up to this work.

An early bit of feedback is that I think it's likely overly prescriptive that the content item is tailored towards the srcset attribute. I imagine it'll be more friendly to use if we have an array of images each with URL and dimensions. Then we could potentially replace concepts like "high_resolution_url" with just plucking the item with greatest dimensions from the list.

The other thing, which harks back to my chats with Kuba, is that to experiment with fixing this for the coronation page we don't need to have this data flow through the full stack. All the filenames are predictable so we can prove the srcset is a good approach without needing to change the rest of the stack. As we already have a custom hack for the coronation (bunting) it didn't seem a net increase in tech debt and would be faster to get a public version (and would likely inform which of the file sizes we'd want to use in srcset)

KludgeKML:
Yeah, it should probably be providing the image sizes/urls in a more generic array that doesn't directly reference srcset. 

Given that this is on publishing's radar to do properly and we already have hacks in place for the coronation logo, I'll use the predictable image sizes in the frontend to simulate the data coming through: https://github.com/alphagov/collections/pull/3237

Closing this PR.

kevindew:
Thanks @KludgeKML 

> Given that this is on publishing's radar to do properly and we already have hacks in place for the coronation logo, I'll use the predictable image sizes in the frontend to simulate the data coming through: [alphagov/collections#3237](https://github.com/alphagov/collections/pull/3237)

Just to add some clarity here, this isn't currently on publishing's radar to do properly. I think once we've got a good proof of concept for Topical Events we'll want to get it through the stack properly and I'm expecting this to be initiated by your team (as it's not on a backlog anywhere). Happy to chat over Slack if this isn't clear.

PR #2354, opened 2023-05-02
Update embassies_index schema

chrisroos:
Trello link: https://trello.com/c/MMUcRbIT

We are moving the rendering of the [embassies index page](https://www.gov.uk/world/embassies) from Whitehall to a rendering app (probably Collections).

The embassies index page displays information about the embassy or embassies available in each World Location. Each World Location can have either:

- a list of one or more local offices that British nationals can contact for assistance (e.g. Australia has "Canberra", "Sydney" etc.)
- a single remote office that British nationals can contact for assistance (e.g. Afghanisation currently has a remote office in Qatar)
- no offices that a British national can contact (in which case we currently display "contact local authorities").

We've modelled these three cases in this schema.

We've tested that the schema is fit for purpose by generating a JSON document conforming to the schema in Whitehall (see WIP PR https://github.com/alphagov/whitehall/pull/7596) and then rendering that JSON document in the Collections app (see WIP PR https://github.com/alphagov/collections/pull/3227).

### DIFF
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -37,6 +37,7 @@
 - drcf_digital_markets_research
 - drug_safety_update
 - email_alert_signup
+- embassies_index
 - employment_appeal_tribunal_decision
 - employment_tribunal_decision
 - equality_and_diversity
--- a/content_schemas/dist/formats/embassies_index/frontend/schema.json
+++ b/content_schemas/dist/formats/embassies_index/frontend/schema.json
@@ -73,6 +73,7 @@
         "drcf_digital_markets_research",
         "drug_safety_update",
         "email_alert_signup",
+        "embassies_index",
         "employment_appeal_tribunal_decision",
         "employment_tribunal_decision",
         "equality_and_diversity",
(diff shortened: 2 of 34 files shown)


BeckaL (commented):
This all looks really good to me, thanks for excellent commit messages and a well thought through schema. 

Given this is a change to an existing schema that's in use, I'm wondering if it'd be nicer to make a backwards compatible change here i.e. first make the missing field in the whitehall test optional, merge in the whitehall change, and then switch it to mandatory. 

These dependent changes used to be a bit easier in our old CI system where you could re-run CI specifying a different branch of e.g. whitehall to run tests against, so you could get your passing test suite before merging. As it is at the moment I think (?) we'd have to remove branch protection rules to merge this in. Maybe we could have a chat through?

chrisroos:
Thanks, @BeckaL!

I've modified the commit that updates the embassies_index schema and have removed the requirement for the `world_locations` to be present. I'll add this back once we've made the corresponding change to Whitehall.

BeckaL (approved):
Lovely, looks good to me!

chrisroos:
Awesome. Thanks, @BeckaL :+1:

PR #2361, opened 2023-05-10
Add schema for Worldwide Offices

jkempster34:
https://trello.com/c/CUp57yUx

### Add schema format for Worldwide Offices
We are moving the rendering of Worldwide Organisations out of Whitehall. As
part of this, we also need to move the related Worldwide Office pages.

### Generate schemas for Worldwide Offices
Generate schemas from the format added in the previous commit.

---

### DIFF
--- a/content_schemas/examples/worldwide_office/frontend/worldwide_office.json
+++ b/content_schemas/examples/worldwide_office/frontend/worldwide_office.json
@@ -1,0 +1,17 @@
+{
+  "base_path": "/world/organisations/british-consulate-general-atlanta/office/british-consulate-general-atlanta",
+  "content_id": "e9789264-2d9c-4776-8ecc-2f7aa681d456",
+  "description": "",
+  "locale": "en",
+  "publishing_app": "whitehall",
+  "rendering_app": "collections",
+  "schema_name": "worldwide_office",
+  "document_type": "worldwide_office",
+  "title": "British Consulate General Atlanta",
+  "public_updated_at": "2016-09-14T18:19:27+01:00",
+  "updated_at": "2016-09-14T10:37:00Z",
+  "details": {
+    "access_and_opening_times": "<div class=\"govspeak\"><p>Telephone: 9.00am to 4.00pm Monday to Friday (except Public Holidays), Emergency Consular Service for British nationals available out of hours.</p></div>"
+  },
+  "links": {}
+}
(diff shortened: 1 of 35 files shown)


content_schemas/examples/worldwide_office/frontend/worldwide_office.json
```diff
@@ -0,0 +1,17 @@
+{
```
jkempster34, line 1: This example is a bit basic for now, but I figure I'll revisit it with some real data once I've published some offices.

content_schemas/formats/worldwide_office.jsonnet
```diff
@@ -0,0 +1,20 @@
+(import "shared/default_format.jsonnet") + {
+  definitions: (import "shared/definitions/_whitehall.jsonnet") + {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {
+        access_and_opening_times: {
+           type: [
+             "string",
+             "null",
```
brucebolt, line 10: I _think_ we might not need to specify `null` here, since the field isn't being marked as optional.  Maybe that could be tested by having two examples: one with `access_and_opening_times` and one without?

jkempster34: We have examples of both:
- [Required properties with null specified](https://github.com/alphagov/publishing-api/blob/067cd760d76e91d97b47d5e8edecff2ac0fe5ea7/content_schemas/dist/formats/answer/notification/schema.json#L562)
- [Optional properties with null specified](https://github.com/alphagov/publishing-api/blob/067cd760d76e91d97b47d5e8edecff2ac0fe5ea7/content_schemas/formats/shared/default_properties/publisher.jsonnet#L12)

I think this is right, `access_and_opening_times` is currently optional (because it's not in the `required` array), and we _do_ want to be able to provide a `null` value here (which means it needs to be [optional](https://json-schema.org/understanding-json-schema/reference/object.html#required:~:text=In%20JSON%20a%20property%20with%20value%20null%20is%20not%20equivalent%20to%20the%20property%20not%20being%20present.%20This%20fails%20because%20null%20is%20not%20of%20type%20%E2%80%9Cstring%E2%80%9D%2C%20it%E2%80%99s%20of%20type%20%E2%80%9Cnull%E2%80%9D) with a `null` type)

brucebolt: Perfect, thanks for confirming!

PR #2401, opened 2023-06-08
Present all redirects consistently

BeckaL:
This PR ensures that all redirects in publishing api are treat consistently in terms of how they are presented downstream.

### Context

Redirects can be represented in publishing api in one of two ways:

1. As an unpublished edition, with an unpublishing record with a type of 'redirect'. These result from an 'unpublish' request to publishing api.
2. As a published edition, with a schema name and document type of 'redirect'. These result from a 'put content' request to publishing api.

We think that there is a use case for both, with 1) primarily being for pieces of content that were previously live on gov.uk and now redirected, and 2) being for urls that were never pieces of content on gov.uk, but where we want to ensure a redirect occurs. In practice, we think we have a lot of cases of 2) which should in fact be 1), because publishing apps call publishing api inconsistently in the case of redirects.

While both result in the same behaviour on gov.uk i.e. a redirected page, they are treated inconsistently in publishing api, which determines the presenter to use to represent the edition downstream (to content store and to the rabbit mq queues) based on whether it is unpublished or not. This means that for redirects that fall into the second category, we treat them as normal bits of live content. This ends up writing content which fails schema validation.

### What's the impact of the current situation / what are we fixing

This schema validation does not stop anything from being published, but it does result in us logging errors (this is how this was originally noticed) - see [this example](https://kibana.logit.io/s/13d1a0b1-f54f-407b-a4e5-f53ba653fac3/app/discover#/doc/8ac115c0-aac1-11e8-88ea-0383c11b333a/filebeat-2023.06.12?id=yyi_rogB4AU214aYzl9d). Additionally, having content in a state that doesn't match what we say about it in schemas is confusing.

### Considerations with this PR for the reviewer

I think this change will not cause issues. One thing I did think about was whether we'd lose any important information from switching some content items from the edition presenter to the more sparse redirect presenter. 

Here's an example of a published redirect content item: https://www.gov.uk/api/content/topic/immigration-operational-guidance
Here's an example of an unpublished redirect content item:
https://www.gov.uk/api/content/guidance/apply-to-join-the-physics-itt-for-engineers-pilot

Looking at these two specific examples, it looks like we're not going to be losing any useful information by switching the presenter - we lose links, but this is good since they don't exist in the schema and don't make sense for a redirect. I've added in the first published at date in this PR in case useful for metadata on the redirect, but it's not strictly necessary.

So I think this is all OK, but would welcome any thoughts about unintended consequences.

[Trello](https://trello.com/c/RbFmTlla/1834-investigate-why-redirects-from-collections-publisher-results-in-notification-schema-errors)

### DIFF
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -45,7 +45,11 @@ private
   end
 
   def redirect_presenter
-    Presenters::RedirectPresenter.from_unpublished_edition(edition)
+    if unpublishing
+      Presenters::RedirectPresenter.from_unpublished_edition(edition)
+    else
+      Presenters::RedirectPresenter.from_published_edition(edition)
+    end
   end
 
   def gone_presenter
@@ -67,6 +71,7 @@ private
   end
 
   def message_queue_presenter
+    return redirect_presenter if edition.document_type == "redirect"
     return content_presenter unless unpublished?
 
     case unpublishing.type
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -61,6 +61,7 @@ private
   end
 
   def content_store_presenter
+    return redirect_presenter if edition.document_type == "redirect"
     return content_presenter unless unpublished?
 
     case unpublishing.type
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -45,7 +45,7 @@ private
   end
 
   def redirect_presenter
-    Presenters::RedirectPresenter.from_edition(edition)
+    Presenters::RedirectPresenter.from_unpublished_edition(edition)
   end
 
   def gone_presenter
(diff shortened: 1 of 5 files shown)


brucebolt:
> I think this change will not cause issues. One thing I did think about was whether we'd lose any important information from switching some content items from the edition presenter to the more sparse redirect presenter.
> 
> Here's an example of a published redirect content item: https://www.gov.uk/api/content/topic/immigration-operational-guidance Here's an example of an unpublished redirect content item: https://www.gov.uk/api/content/guidance/apply-to-join-the-physics-itt-for-engineers-pilot
> 
> Looking at these two specific examples, it looks like we're not going to be losing any useful information by switching the presenter - we lose links, but this is good since they don't exist in the schema and don't make sense for a redirect. I've added in the first published at date in this PR in case useful for metadata on the redirect, but it's not strictly necessary.

I agree.  I can't see anything that would break as I can't see the links actually being used anywhere.

brucebolt (approved):
These changes all seem to make sense to me.  Thanks for raising this @BeckaL.

PR #2413, opened 2023-06-15
Fix bug when filtering by org and all locales

1pretz1:
Previously, if the query included `locales: all` and contained link_filters then we'd receive an SQL error:

```
PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "documents"
LINE 1: ...s" INNER JOIN link_sets ON link_sets.content_id = documents....
```

This only occurred when using the `all` value for a locale due to the `documents` join being skipped [1] and so then the link_sets join failed [2] due to missing the `documents` table declaration.

[1]: https://github.com/alphagov/publishing-api/blob/ff491b745bf7cfdd3ff8fba6dabbdabbe40c9150/app/queries/get_content_collection.rb#L52
[2]: https://github.com/alphagov/publishing-api/blob/ff491b745bf7cfdd3ff8fba6dabbdabbe40c9150/app/models/link.rb#L13

### DIFF
--- a/app/queries/get_content_collection.rb
+++ b/app/queries/get_content_collection.rb
@@ -46,10 +46,10 @@ module Queries
     )
 
     def editions
-      scope = Edition.all
+      scope = Edition.with_document
       scope = scope.where(document_type: document_types) if document_types.any?
       scope = scope.where(publishing_app:) if publishing_app
-      scope = scope.with_document.where("documents.locale": locale) unless locale == "all"
+      scope = scope.where("documents.locale": locale) unless locale == "all"
       scope = Link.filter_editions(scope, link_filters) if link_filters.present?
       scope
     end
--- a/spec/queries/get_content_collection_spec.rb
+++ b/spec/queries/get_content_collection_spec.rb
@@ -19,11 +19,16 @@ RSpec.describe Queries::GetContentCollection do
         document_type: "mainstream_browse_page",
         schema_name: "mainstream_browse_page",
       )
-      create(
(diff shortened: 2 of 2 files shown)


app/queries/get_content_collection.rb
```diff
@@ -50,7 +50,7 @@ def editions
       scope = scope.where(document_type: document_types) if document_types.any?
       scope = scope.where(publishing_app:) if publishing_app
       scope = scope.with_document.where("documents.locale": locale) unless locale == "all"
-      scope = Link.filter_editions(scope, link_filters) if link_filters.present?
+      scope = Link.filter_editions(scope.with_document, link_filters) if link_filters.present?
```
kevindew, line 53: I think rather than having `scope.with_document` in a couple of places we should just update the initial scope to be `Edition.with_document` instead of `Edition.all` - as I can't imagine a scenario where we'd not query with a document anyway.

kevindew (commented):
Nice job fixing the bug, just making an implementation suggestion

app/queries/get_content_collection.rb
1pretz1: Sure, I initially added another `join` as we don't technically need the `documents` join for certain queries but if you think it is unlikely then adding to the default scope makes the code more readable and less error prone 👍

1pretz1:
Thanks for the review @kevindew! I've fixed up based on your suggestion.

kevindew (approved):

PR #2422, opened 2023-06-23
Correct document type on call for evidence examples

beccapearce:
We have two fields on call for evidence one is called schema and one is called document_type. We need document_type to reflect the state of the call for evidence so it can be correctly rendered.

This corrects where the document_type has been put as the schema so we could work on the code in another app.

Trello card: https://trello.com/c/hxD7EECb/1467-update-document-types-for-call-for-evidence-in-publishing-api


ryanb-gds (approved):
Looks good to me

PR #2423, opened 2023-06-23
Add the field held on another site to example call for evidence

beccapearce:
We have added the field `held_on_another_website` to the example open_call_for_evidence so that it can be used to test in the frontend rendering.

https://trello.com/c/ur96Ex4v/1466-add-the-field-held-on-another-website-to-open-call-for-evidence-example-in-publishing-api


ryanb-gds (approved):
Looks good. As discussed, might be worth considering whether the schema definition for this field should be just "string" or "string or null", but that's a broader question that would affect consultations as well

PR #2424, opened 2023-06-23
Remove public feedback from call for evidence

beccapearce:
When a closed call for evidence has results published to it, we only upload an outcome, therefore the public feedback options can be removed from the content_schemas.

trello card: https://trello.com/c/LVOk25hB/1476-remove-public-feedback-from-call-for-evidence-in-publishing-api


ryanb-gds (approved):
Looks like it should do the trick

PR #2431, opened 2023-06-30
Replaced consultation with call for evidence in call for evidence examples

ryanb-gds:
Without this, the goverment-frontend tests that check the call for evidence rendering were failing


minhngocd (approved):
LGTM 🎉

PR #2453, opened 2023-07-20
Add case type options to cma case schema

ryanb-gds:
I have recently added two new case type options to the Competition and Markets Authority case specialist content type. I was unaware at the time that the case types options were enumerated in the Publishing API content schema, resulting in the CMA raising a ticket when they encountered a rather unpleasant looking JSON Schema validation error message. I'm hopeful this will resolve the issue.

Trello ticket from Zendesk report: https://trello.com/c/UD22Wsh0
Original CMA request Trello ticket: https://trello.com/c/r7sZCKI4


JonathanHallam (approved):
Nice one, thank you :)

PR #2458, opened 2023-07-27
Add task to create code-snippet for email-alert-api queries

KludgeKML:
It's currently awkward to find out how many people would be notified via email-alert-api if a particular content item is updated. The information flows one way (down the message queue) to email alert service and is stored in email-alert-api as a ContentChange. A [recent PR](https://github.com/alphagov/email-alert-api/pull/1976) added a rake task to easily query Email Alert API for the numbers of people in subscriber lists for changes that had already gone out, but we have no way of answering the number of people affected by a hypothetical change (a question asked occasionally on second line).

Since Publishing API and email-alert-api can't talk directly, I've added a rake task which will generate a code snippet to paste into the email-alert-api console, which then gives you a list of affected subscriber lists.

Example usage: You want to know how many people would be notified about a change to "/government/news/ifrs-17-gad-helps-develop-new-guidance"

On **publishing-api**, run the rake task:
```
rails email_reports:get_subscriber_list_code[/government/news/ifrs-17-gad-helps-develop-new-guidance]
```

this will output code similar to this:

```
To use: Copy everything between the start and end block markers, and
paste into an email-alert-api console

=== START BLOCK ===
lists = SubscriberListQuery.new(
  content_id: "70379c1a-4ed3-4296-b0ad-3afcf310d52e",
  tags: {:user_journey_document_supertype=>"thing", :email_document_supertype=>"announcements", :government_document_supertype=>"news-stories", :content_purpose_subgroup=>"news", :content_purpose_supergroup=>"news_and_communications", :content_store_document_type=>"news_story"},
  links: {:government=>["d4fbc1b9-d47d-4386-af04-ac909f868f92"], :organisations=>["02e01dc9-a4c4-46d9-b496-a618c89afe62"], :original_primary_publishing_organisation=>["02e01dc9-a4c4-46d9-b496-a618c89afe62"], :primary_publishing_organisation=>["02e01dc9-a4c4-46d9-b496-a618c89afe62"], :taxons=>["f3caf326-fe33-410f-b7f4-553f4011c81e"], :user_journey_document_supertype=>"thing", :email_document_supertype=>"announcements", :government_document_supertype=>"news-stories", :content_purpose_subgroup=>"news", :content_purpose_supergroup=>"news_and_communications", :content_store_document_type=>"news_story", :taxon_tree=>["f3caf326-fe33-410f-b7f4-553f4011c81e", "f3f4b5d3-49c4-487b-bd5b-be75f11ec8c5", "e48ab80a-de80-4e83-bf59-26316856a5f9"]},
  document_type: "news_story",
  email_document_supertype:  "announcements",
  government_document_supertype:  "news-stories",
).lists
=== END BLOCK ===
```

...copy and paste this into the application console of **email-alert-api**. lists will now contain a list of subscriber lists that would be notified, which can be manipulated to (eg) give you the breakdown of number of emails that would be sent:

```
        total_subs = lists.sum { |l| l.subscriptions.active.count }
        immediately_subs = lists.sum { |l| l.subscriptions.active.immediately.count }
        daily_subs = lists.sum { |l| l.subscriptions.active.daily.count }
        weekly_subs = lists.sum { |l| l.subscriptions.active.weekly.count }
```

### DIFF
--- a/lib/tasks/email_reports.rake
+++ b/lib/tasks/email_reports.rake
@@ -1,0 +1,54 @@
+namespace :email_reports do
+  desc "Given a path, creates a chunk of code that can be pasted into email-alert-api to get subscriber lists that would be alerted by a change to that path"
+  task :get_subscriber_list_code, %i[govuk_path] => :environment do |_, _args|
+    puts_subscriber_list_query(govuk_path)
+  end
+end
+
+def puts_subscriber_list_query(govuk_path)
+  edition = Edition.live.where(base_path: govuk_path).last
+  ds_payload = DownstreamPayload.new(edition, 1, draft: false)
+  payload = ds_payload.message_queue_payload
+
+  puts("To use: Copy everything between the start and end block markers, and")
+  puts("paste into an email-alert-api console")
+  puts("")
+  puts("=== START BLOCK ===")
+  puts("lists = SubscriberListQuery.new(")
+  puts("  content_id: \"#{payload[:content_id]}\",")
+  puts("  tags: #{(payload[:tags] || {}).merge(additional_items(payload))},")
+  puts("  links: #{(payload[:links] || {}).merge(additional_items(payload).merge(taxon_tree: taxon_tree(payload)))},")
+  puts("  document_type: \"#{payload[:document_type]}\",")
+  puts("  email_document_supertype:  \"#{payload['email_document_supertype']}\",")
+  puts("  government_document_supertype:  \"#{payload['government_document_supertype']}\",")
+  puts(").lists")
+  puts("=== END BLOCK ===")
+end
+
+def additional_items(payload)
+  {
+    user_journey_document_supertype: payload["user_journey_document_supertype"],
+    email_document_supertype: payload["email_document_supertype"],
+    government_document_supertype: payload["government_document_supertype"],
+    content_purpose_subgroup: payload["content_purpose_subgroup"],
+    content_purpose_supergroup: payload["content_purpose_supergroup"],
+    content_store_document_type: payload[:document_type],
+  }
+end
+
+def taxon_tree(payload)
+  [payload[:expanded_links][:taxons].first[:content_id]] + get_parent_links(payload[:expanded_links][:taxons].first)
+end
+
+def get_parent_links(taxon_struct)
+  return [] unless taxon_struct[:links].key?(:parent_taxons)
+  return [] unless taxon_struct[:links][:parent_taxons].any?
+
+  tree = []
+  taxon_struct[:links][:parent_taxons].each do |parent_taxon|
+    tree += [parent_taxon[:content_id]]
+    tree += get_parent_links(parent_taxon)
+  end
+
+  tree
+end


ChrisBAshton (commented):
Thanks for raising this, @KludgeKML! But I'm a bit uncomfortable with the idea of using one app to generate a code snippet designed to be run against another app. It's brittle (if an implementation in Email Alert API changes, we'd have to change this code too), and it's also quite a clunky workflow.

It looks as though you can grab everything you need from the content API, e.g. https://www.gov.uk/api/content/government/news/ifrs-17-gad-helps-develop-new-guidance (or https://draft-origin.publishing.service.gov.uk/api/content/government/news/ifrs-17-gad-helps-develop-new-guidance, if you're wanting a more accurate figure, e.g. they're about publish a change with a new tag).

From a quick look, the only data you can't get explicitly from content API is this bit:

```
  email_document_supertype:  "announcements",
  government_document_supertype:  "news-stories",
```

...but I'm hoping they can be derived from the `document_type` or similar.

TLDR I think this could be a rake task in email-alert-api, taking the same `govuk_path` parameter but querying the content API...?

KludgeKML:
Yeah, I'm also not exactly thrilled by this code generation system, but perhaps we can do it properly if I can get some 

I did initially think I could get everything from Content API, but:

1)  when I tried I was missing a bunch of those supertypes, and they did seem to make a big difference (I think using content API I managed to match about 7 of the 22 lists that page would trigger)
2) Content Api is technically still in beta IIRC, and I know that there's a bit of discussion about what happens to it. 

Not saying that's not a better idea, just there are some questions.

Do we know how many supertypes, etc, there are? Is it possible we can easily derive them from the document type?

ChrisBAshton:
Looks like there's a [complete list of supertypes](https://github.com/alphagov/govuk_document_types/blob/0b6b2280a94c6090339afc89b5f1db93ed664a86/data/supertypes.yml#L70-L161) available in the [govuk_document_types](https://rubygems.org/gems/govuk_document_types) gem. Would that work?🤞 

I wouldn't worry too much about Content API going away. It's a consideration, but not one I'd worry about for the sake of a non-critical rake task.

kevindew:
Ooh this is a clever idea, but I do +1 Chris' suggestion. Architecturally it is definitely advantageous that Publishing API doesn't have a coupling to Email Alert API (we've already intentionally avoided that with the message queue abstract interface)

> 1. when I tried I was missing a bunch of those supertypes, and they did seem to make a big difference (I think using content API I managed to match about 7 of the 22 lists that page would trigger)

The data that is in the Content Store is the same data structure as the DownstreamPayload class you're using. It includes the supertypes code from the gem: https://github.com/alphagov/publishing-api/blob/bb07dee07ab4196338d189534f16e8c6bcddcf03/app/presenters/edition_presenter.rb#L52

I'm not sure the cause of the issues you experienced but, unless I'm missing something, in theory the data you get from this task with regards to supertypes should match the Content Store.

> 2. Content Api is technically still in beta IIRC, and I know that there's a bit of discussion about what happens to it.

The beta aspect of that is for citizens to use it, and reflects our position of not committing to backwards compatibility (and likely a few other things we don't really commit to). But the use of it in our systems is vital - the whole of www.gov.uk is dependent on it - so it's a very safe thing to use.

KludgeKML:
Replaced by: https://github.com/alphagov/email-alert-api/pull/1979

PR #2466, opened 2023-08-04
Specify document types for schemas that lack them 

kevindew:
These schemas previously allowed any document type, this was incorrect
as it would make no sense for say a topical_event to have a
document_type of news_article.

The reason they were configured with such a range of document types was
because the default behaviour was for any document_type to be allowed
and this is what occurs when a document_type is omitted. This default is
consistent with past behaviour where any combination of schema_name and
document_type was allowed.

I verified the appropriate document types with Publishing API:

```
irb(main):006:0> Edition.where(schema_name: "embassies_index", state: %w[draft published]).group(:document_type).count
=> {"embassies_index"=>1}
irb(main):007:0> Edition.where(schema_name: "field_of_operation", state: %w[draft published]).group(:document_type).count
=> {"field_of_operation"=>5}
irb(main):008:0> Edition.where(schema_name: "historic_appointments", state: %w[draft published]).group(:document_type).count
=> {"historic_appointments"=>1}
irb(main):009:0> Edition.where(schema_name: "special_route", state: %w[draft published]).group(:document_type).count
=> {"search"=>1, "special_route"=>76}
irb(main):010:0> Edition.where(schema_name: "topical_event", state: %w[draft published]).group(:document_type).count
=> {"topical_event"=>126}
irb(main):011:0> Edition.where(schema_name: "world_location_news", state: %w[draft published]).group(:document_type).count
=> {"world_location_news"=>320}
irb(main):012:0> Edition.where(schema_name: "worldwide_office", state: %w[draft published]).group(:document_type).count
=> {"worldwide_office"=>609}
irb(main):013:0> Edition.where(schema_name: "worldwide_organisation", state: %w[draft published]).group(:document_type).count
=> {"worldwide_organisation"=>547}
```

(It's a bit annoying about that search one)

Ideally we'd remove this default behaviour and instead default an
omitted schema_name to be the document_type. However this is complicated
by the generic and generic_with_external_links document types which
don't have corresponding document types.

These are annoying thing to hold us back as they seem to be mostly dead
document types that should be deleted but I can still see evidence of
them in alphagov code so I'm not sure it's 100% safe to remove them.

```
irb(main):001:0> Edition.where(document_type: "generic").count
=> 0
irb(main):002:0> Edition.where(document_type: "generic_with_external_related_links").count
=> 39
irb(main):003:0> Edition.where(document_type: "generic_with_external_related_links").group(:state).count
=> {"superseded"=>34, "unpublished"=>5}
```


ChrisBAshton (approved):
Thanks Kevin - finally got around to looking at this today and it looks eminently sensible to me.

(Would be great if we could track down that one `special_route` edition that has a `document_type` of `search`, and kill it! But that's for another day)

I'm minded to let @JonathanHallam review and do the merging, given he's closer to the inner workings of Publishing API 🙏

PR #2471, opened 2023-08-11
Add 'open_filter_on_load' property to finder

jon-kirwan:
## What

https://trello.com/c/p6FCqlHs/2042-expand-filter-and-location-facets-by-default-on-mobile

Add `open_filter_on_load` property to open filters, on page load, on mobile. Configured for **Find a licence** (specialist finder) only.

## Why

During user research, a common observation has emerged that mobile users do not interact with the 'Filter' button.

## Anything else

https://github.com/alphagov/finder-frontend/pull/3131
https://github.com/alphagov/specialist-publisher/pull/2369

### DIFF
--- a/content_schemas/dist/formats/finder/frontend/schema.json
+++ b/content_schemas/dist/formats/finder/frontend/schema.json
@@ -347,6 +347,9 @@
         "no_index": {
           "type": "boolean"
         },
+        "open_filter_on_load": {
+          "type": "boolean"
+        },
         "reject": {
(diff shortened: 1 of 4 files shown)


MartinJJones (commented):
It appears that `open_on_load` is only being used to open the mobile filters modal on page load, is it worth updating this to something like `open_mobile_filter_on_load`?

MartinJJones (approved):
Looks good to me 👍

PR #2492, opened 2023-09-06
Revert "Add new field to edition links"

hannako:
Reverts alphagov/publishing-api#2491


aldavidson:
We no longer need this, a subsequent PR fixed the build on main

hannako:
This revert is no longer required - we were reverting because of a failing test on main branch of publishing api which we suspected had been introduced by #2491. But this was not the case - it was a change made to the github actions config in smart answers.

PR #2515, opened 2023-10-04
Add `lock_ttl` to downstream workers

brucebolt:
This adds an expiry time to the worker locks. This means that if a lock fails to be unset, we will automatically expire it after one day.

We know that locks sometimes remain on jobs after they have executed but have yet to identify the cause.

This should alleviate the issue by expiring these locks after one day, meaning the document preview can be generated again after that day.

We will separately identify and fix the issue causing the locks to not be removed.

[Trello card](https://trello.com/c/IUj0VRw8)


richardTowers (approved):
Yeah, this seems sensible to me.

As I understand it, the locking is there to prevent relatively short term race conditions (order of seconds / minutes), so having them expire after a whole day should be no problem.

PR #2517, opened 2023-10-05
Reduce `lock_ttl` and make the value global

brucebolt:
In https://github.com/alphagov/publishing-api/pull/2515, we added an expiry time to the locking of downstream workers, as some locks are not being removed when the job executes. We are still unsure of the reason for this.

Having examined the logs further, this affects all types of worker, not just the downstream workers.

Therefore making the `lock_ttl` value global for all Sidekiq workers.

Also reducing the expiry time to one hour, as all the workers we have should complete within that timescale.

[Trello card](https://trello.com/c/IUj0VRw8)


jkempster34 (approved):
Makes sense to me

PR #2533, opened 2023-10-30
Disallow a conflicting draft editions

brucebolt:
Currently, a publishing application can send a draft edition to Publishing API, which has the same base path as a live edition but with a different content ID.

This results in a valid draft edition being created in Publishing API.

However after this occurs, neither the original live edition with the original content ID or the draft edition can have a new edition created, as there will be a conflict in content IDs.

This updates the validation to disallow a draft edition with a different content ID if there is already a published edition at that base path.

[Trello card](https://trello.com/c/QRFqcMKd)

### DIFF
--- a/app/queries/base_path_for_state.rb
+++ b/app/queries/base_path_for_state.rb
@@ -36,7 +36,7 @@ module Queries
     end
 
     def allowed_states(state)
-      state == "draft" ? %w[draft] : %w[published unpublished]
+      state == "draft" ? %w[draft published] : %w[published unpublished]
     end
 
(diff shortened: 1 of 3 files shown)


brucebolt:
Closing as this causes issues with the pact tests.  Maybe there is a reason we permit draft editions that "take over" a base path, even if they have a different content ID?

brucebolt:
Closing as this causes issues with the pact tests. Maybe there is a reason we permit draft editions that "take over" a base path, even if they have a different content ID?

PR #2541, opened 2023-11-08
Add ability to requeue only published content

csutter:
For our initial bulk import of data into `search-api-v2`, we are only
interested in published content that is visible to users, and having to
consume all content is slowing down the initial import significantly and
putting unnecessary load on Publishing API.

Also cleans up and refactors the Rake tasks, and adds tests for this
functionality.

- Add new `queue:requeue_all_the_published_things` task to specifically
  requeue only published editions
- Refactor as much logic as possible from `queue:*` Rake tasks into a
  library class so that shared stuff doesn't have to be repeated and its
  effects can actually be tested
- Use `Edition.live` scope instead of `.where(content_store: :live)`
- Stop joining on Document with `.with_document` scope as no attributes
  from that model are used
- Resolve some inaccurate comments and descriptions for Rake tasks
  (published vs live)

### DIFF
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -1,5 +1,3 @@
-MAX_IMPORT_QUEUE_SIZE = 1000
-
 namespace :queue do
   desc "Watch the queue, and print messages on the console"
   task watcher: :environment do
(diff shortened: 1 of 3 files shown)


lib/tasks/queue.rake
```diff
@@ -30,27 +28,17 @@ namespace :queue do
...
-    scope.find_each do |edition|
-      RequeueContent.perform_async(edition.id, version)
-    end
+    RequeueContentByScope.new(
+      Edition.live.where(document_type: document_types),
+    ).call
   end
 
-  desc "Add all published editions to the message queue with a specified routing key"
+  desc "Add all live editions to the message queue with a specified routing key"
```
mike29736, line 41: :cool:

mike29736 (approved):
Apologies that this review took so long. When I did get to it on my to-do list, I basically had to learn what 100% of this meant :mortar_board: 

I agree with you on the removal of the `with_document` scope here. I can only imagine it was either copy-paste or that we were once plagued by Editions with no Documents..?

PR #2566, opened 2023-12-06
Optionally send PublishIntent additions/deletions to content store asynchronously

aldavidson:
Our [plan for safely swapping content-store over to a PostgreSQL database](https://docs.google.com/document/d/14B3YE978TDbujYnGHL0gm_gnf_EwD5MSHx7iRF24-90/edit#heading=h.wctys5xv7gdn) relies on us being able to suspend all writes/deletions for the length of time needed to run a full Mongo-to-Postgres export. This is currently around 90 minutes, but may be longer for the live content-store as it gets heavy sustained traffic.

When running this on the draft content-store, we found that some publish-intent writes were still going through even when the publishing-api workers were disabled. The numbers were low (less than 10 in 2hrs) but they still need to be handled.

We decided that the best option was to temporarily change publishing-api to send these updates to content-store asynchronously, for the duration of the switchover. (see the [Trello card](https://trello.com/c/a7ngC11v/952-figure-out-how-to-handle-observed-puts-of-publish-intents-while-publishing-api-workers-were-disabled) for full details). When the workers get switched back on at the end of the switchover, these updates will be pulled off the queue and sent to the content-store automatically.

We think this is safe, as publishing-api [already checks for uniqueness](https://github.com/alphagov/publishing-api/blob/d8ded888f56a2ae7909a8baf45968ed15e90423a/app/models/path_reservation.rb#L25-L30) of a path reservation before propagating to the content-store, so the risk of an out-of-order execution resulting in a problem should be very low.

Update 8th Dec 16:39:
I've successfully tested this with a scheduled publish on integration - 
image
...which showed up correctly in content-store:
```
> PublishIntent.order('created_at desc').first(1).pluck('base_path', 'created_at')
=> [["/government/news/test-new-news-story-5", Fri, 08 Dec 2023 16:32:50.131144000 UTC +00:00]]
```

So we'll merge this as close as possible to the start of the maintenance period on Monday 

### DIFF
--- a/app/commands/delete_publish_intent.rb
+++ b/app/commands/delete_publish_intent.rb
@@ -2,7 +2,14 @@ module Commands
   class DeletePublishIntent < BaseCommand
     def call
       if downstream
-        PublishingAPI.service(:live_content_store).delete_publish_intent(base_path)
+        enqueue = ENV.fetch("ENQUEUE_PUBLISH_INTENTS", false)
+        if enqueue == "true"
+          DeletePublishIntentWorker.perform_async(
+            "base_path" => base_path,
+          )
+        else
+          PublishingAPI.service(:live_content_store).delete_publish_intent(base_path)
+        end
       end
 
       Success.new({})
--- a/app/commands/put_publish_intent.rb
+++ b/app/commands/put_publish_intent.rb
@@ -5,7 +5,15 @@ module Commands
 
       if downstream
         payload = publish_intent
-        Adapters::ContentStore.put_publish_intent(base_path, payload)
+        enqueue = ENV.fetch("ENQUEUE_PUBLISH_INTENTS", false)
+        if enqueue == "true"
+          PutPublishIntentWorker.perform_async(
+            "base_path" => base_path,
+            "payload" => payload.to_json,
+          )
+        else
+          Adapters::ContentStore.put_publish_intent(base_path, payload)
+        end
       end
 
       Success.new(payload)
--- a/app/workers/put_publish_intent_worker.rb
+++ b/app/workers/put_publish_intent_worker.rb
@@ -1,0 +1,25 @@
+require "sidekiq-unique-jobs"
+
+class PutPublishIntentWorker
+  include DownstreamQueue
+  include Sidekiq::Worker
+  include PerformAsyncInQueue
+
+  sidekiq_options queue: HIGH_QUEUE,
+                  lock: :until_executing,
+                  lock_args_method: :uniq_args,
+                  on_conflict: :log
+
+  def self.uniq_args(args)
+    [
+      args.first["base_path"],
+      name,
+    ]
+  end
+
+  def perform(args = {})
+    Adapters::ContentStore.put_publish_intent(args["base_path"], JSON.parse(args["payload"]))
+  rescue AbortWorkerError => e
+    notify_airbrake(e, args)
+  end
+end
--- a/spec/commands/delete_publish_intent_spec.rb
+++ b/spec/commands/delete_publish_intent_spec.rb
@@ -9,6 +9,8 @@ RSpec.describe Commands::DeletePublishIntent do
     }
   end
 
+  let(:content_store) { PublishingAPI.service(:live_content_store) }
+
   it "responds successfully" do
     result = described_class.call(payload)
     expect(result).to be_a(Commands::Success)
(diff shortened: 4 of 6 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@
 /coverage
 /spec/reports
 spec/examples.txt
+.byebug_history
```
JonathanHallam, line 17: Should probably just be it's own commit

app/commands/put_publish_intent.rb
```diff
@@ -5,7 +5,15 @@ def call
 
       if downstream
         payload = publish_intent
-        Adapters::ContentStore.put_publish_intent(base_path, payload)
+        enqueue = ENV.fetch("ENQUEUE_PUBLISH_INTENTS", false)
+        if enqueue == "true"
```
JonathanHallam, line 9: is there a reason this is stored as a string rather than a boolean?

app/commands/delete_publish_intent.rb
```diff
@@ -2,7 +2,14 @@ module Commands
   class DeletePublishIntent < BaseCommand
```
JonathanHallam, line 2: Can you add something in the commit message about why we're doing this please

spec/commands/put_publish_intent_spec.rb
```diff
@@ -18,7 +18,7 @@
   context "when the downstream flag is set to false" do
     it "does not send any downstream requests" do
       expect(Adapters::ContentStore).not_to receive(:put_publish_intent)
-      
+
```
JonathanHallam, line 21: same as comment above about linting

JonathanHallam (changes requested):
Few small questions/ tidy ups to be done but otherwise, good, thanks Al :)

.gitignore
```diff
@@ -14,6 +14,7 @@
 /coverage
```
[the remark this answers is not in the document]
aldavidson: I've squashed and rebased, with a much more detailed multi-line-aware message

.gitignore
aldavidson: yeah, I've removed this change & rebased.

app/commands/put_publish_intent.rb
aldavidson: Yes - it's passed through as an environment variable from the Kubernetes Helm chart, which means it's ultimately going to passed through to the app as a string value:

```
$ RANDOM_ENV_VAR=true irb
irb(main):001> ENV['RANDOM_ENV_VAR']
=> "true"
```

spec/commands/put_publish_intent_spec.rb
```diff
@@ -17,10 +17,35 @@
...
-
+      expect(Adapters::ContentStore).not_to receive(:put_publish_intent)
+      
       described_class.call(payload, downstream: false)
     end
   end
+
+  context "when the downstream flag is set to true" do
+    context "when the ENQUEUE_PUBLISH_INTENTS flag is present" do
+      context "and the ENQUEUE_PUBLISH_INTENTS flag is true" do
```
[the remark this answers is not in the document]
aldavidson: Fair point - I've removed the redundant line 27 (and its equivalent in `delete_publish_intent_spec.rb`) and added examples for when the flag is not set

app/commands/delete_publish_intent.rb
aldavidson: I've added a much more detailed explanation in the commit message.

app/commands/put_publish_intent.rb
```diff
@@ -9,7 +9,7 @@ def call
         if enqueue == "true"
           PutPublishIntentWorker.perform_async(
             "base_path" => base_path,
-            "payload" => payload.to_json
+            "payload" => payload.to_json,
```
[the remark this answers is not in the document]
aldavidson: I rebased the two commits into one

app/workers/put_publish_intent_worker.rb
```diff
@@ -5,12 +5,11 @@ class PutPublishIntentWorker
   include Sidekiq::Worker
   include PerformAsyncInQueue
 
-
   sidekiq_options queue: HIGH_QUEUE,
                   lock: :until_executing,
                   lock_args_method: :uniq_args,
                   on_conflict: :log
-  
+
```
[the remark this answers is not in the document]
aldavidson: I rebased the two commits into one

spec/commands/delete_publish_intent_spec.rb
```diff
@@ -9,17 +9,52 @@
...
+  context "when the downstream flag is set to false" do
+    it "does not send any downstream requests" do
+      expect(content_store).not_to receive(:delete_publish_intent)
 
       described_class.call(payload, downstream: false)
     end
   end
+
+  context "when the downstream flag is set to true" do
+    context "when the ENQUEUE_PUBLISH_INTENTS flag is present" do
```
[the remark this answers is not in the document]
aldavidson: done - see my reply to your previous comment above on `put_publish_intent_spec.rb`

spec/commands/put_publish_intent_spec.rb
aldavidson: Should be clearer now after rebasing

app/commands/put_publish_intent.rb
JonathanHallam: totally agree

spec/commands/put_publish_intent_spec.rb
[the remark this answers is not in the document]
JonathanHallam: I think something slightly funny has happened with the commits here because these changes are in the first one and then removed in the 2nd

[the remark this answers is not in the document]
aldavidson: I'll squash them both down to one commit before merging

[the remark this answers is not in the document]
aldavidson: done

JonathanHallam (approved):
thanks!

PR #2603, opened 2024-01-15
Provide token to `delete-artifact` workflow item

brucebolt:
In https://github.com/alphagov/publishing-api/pull/2595, we updated `delete-artifact` from version 3 to 4.

However this version [now uses the GitHub REST API](https://github.com/GeekyEggo/delete-artifact/pull/17) and therefore requires a token. Therefore updating the workflow to provide this token.


aldavidson (approved):
LGTM :+1:

PR #2608, opened 2024-01-25
Add destination to redirect validation errors

richardTowers:
It's usually not obvious when you're looking at one of these 422 unacceptable entity validation messages in Sentry or Kibana what the payload was, so an error message that just says "internal redirects cannot end with /" can be hard to diagnose.

Adding the destination in these cases should make it much more obvious what the validator is upset about.

We already include the destination in error messages in the reject_query_parameters_and_fragment method, so I don't see any reason why it would be a problem to include them here too.

### DIFF
--- a/app/validators/routes_and_redirects_validator.rb
+++ b/app/validators/routes_and_redirects_validator.rb
@@ -147,7 +147,7 @@ private
       errors.add(:redirects, "path cannot equal the destination") if path == destination
       return unless errors.empty?
 
-      errors.add(:redirects, "destination invalid") if invalid_destination?(destination)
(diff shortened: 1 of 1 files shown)


jkempster34 (approved):
Surprised that no change in specs was needed, maybe we have some missing coverage here

Otherwise, this makes sense to me!

richardTowers:
Yeah, some of the tests don't check the error messages by the looks of it - https://github.com/alphagov/publishing-api/blob/main/spec/support/routes_and_redirects_validator.rb#L161

I'm okay with that, as having to update the tests whenever you update the messages is a bit annoying.

PR #2638, opened 2024-02-14
Add office fields to worldwide organisation details

brucebolt:
This will allow us to store the Worldwide Office details in the Worldwide Organisation, meaning we can represent this content as a single multi-part page.

The worldwide office detail is moved into a shared definition, so it can shared between the two types of office attached to a Worldwide Organisation and the existing Worldwide Office format.

[Trello card](https://trello.com/c/9ivR4TGQ)

### DIFF
--- a/content_schemas/formats/shared/definitions/_worldwide_office.jsonnet
+++ b/content_schemas/formats/shared/definitions/_worldwide_office.jsonnet
@@ -1,0 +1,54 @@
+{
+  type: "object",
+  additionalProperties: false,
+  properties: {
+    access_and_opening_times: {
+      type: [
+        "string",
+        "null",
+      ],
+      description: "The access and opening times for this Worldwide Office.",
+    },
+    contact_content_id: {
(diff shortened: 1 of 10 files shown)


content_schemas/formats/shared/definitions/_worldwide_office.jsonnet
```diff
@@ -0,0 +1,54 @@
+{
+  type: "object",
+  additionalProperties: false,
+  properties: {
+    access_and_opening_times: {
+      type: [
+        "string",
```
jkempster34, line 7: I'm not sure I understand the benefit, but can these use the `body_html_and_govspeak` definition? That seems the most appropriate

brucebolt: I'm not sure we could do that in this shared definition, as I think it'll affect the way we need to present the data from Whitehall, which will include modifying the existing worldwide offices.

jkempster34 (approved):

PR #2643, opened 2024-02-20
Allow translations to be nested under their parent document

brucebolt:
We have a validation that currently prevents translated documents being nested within their parent document.
    
E.g. the following was previously invalid:
- Base path: /world/organisation/my-org.fr
- Route: /world/organisation/my-org/offices/local-office.fr

We must allow these to enable us to represent multi-part content with translations in a single content item.

[Trello card](https://trello.com/c/9ivR4TGQ)


JonathanHallam (approved):
Thanks!

PR #2664, opened 2024-03-05
Add substitute event for queue consumers

callumknights:
This PR is largely a copy of #2652 that I made to get my head around Richards thinking. 

Trello: https://trello.com/c/xaN2Dhwk/1005-emit-an-event-when-documents-are-unpublished-through-subsitution

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -177,6 +177,7 @@ module Commands
         # This enables changing the locale of some content where a
         # published edition for the previous locale still exists.
         published_edition_for_different_locale.substitute
+        SubstitutionHelper.substitute_message(published_edition_for_different_locale)
       end
 
       def set_timestamps
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -60,6 +60,10 @@ private
     Presenters::VanishPresenter.from_edition(edition)
   end
 
+  def substitute_presenter
+    Presenters::SubstitutePresenter.from_edition(edition)
+  end
+
   def content_store_presenter
     if unpublishing
       return redirect_presenter if unpublishing.type == "redirect"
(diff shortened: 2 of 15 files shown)


app/downstream_payload.rb
```diff
@@ -79,7 +79,10 @@ def message_queue_presenter
     when "redirect" then redirect_presenter
     when "gone" then gone_presenter
     when "vanish" then vanish_presenter
-    else content_presenter
+    when "substitute" then vanish_presenter
```
brucebolt: Should this not be `substitute_presenter`?

app/commands/v2/publish.rb
```diff
@@ -177,6 +177,7 @@ def clear_published_item_of_different_locale_but_matching_base_path
         # This enables changing the locale of some content where a
         # published edition for the previous locale still exists.
         published_edition_for_different_locale.substitute
+        SubstitutionHelper.substitute_message(published_edition_for_different_locale)
```
brucebolt, line 180: Could the first line of the commit message be shortened here, so it fits onto one line?

content_schemas/dist/formats/substitute/notification/schema.json
```diff
@@ -2,29 +2,13 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "required": [
-    "analytics_identifier",
```
brucebolt: Maybe this commit could be merged into [`Add schema for new substitute type`](https://github.com/alphagov/publishing-api/pull/2664/commits/0659f302cc25e604d66f7ca292c2cbc958fc59a8), so these lines are never added in the first place?

app/downstream_payload.rb
brucebolt: Oh, I see the `substitute_presenter` method was added in a later commit.  Maybe this first commit should just add the logging, then the `when "substitute"` case could be added in the same commit as the `substitute_presenter` method?

brucebolt (commented):
Looks good 👍.  I've just added a few really small comments about the layout of the commits, but otherwise the code looks fine.

I think you'll also need to auto-regenerate the schemas again, because it looks like `topic_taxonomy_taxons` have been removed since they were last generated and there's now a failing test because of that.

app/commands/v2/publish.rb
callumknights: Done

content_schemas/dist/formats/substitute/notification/schema.json
callumknights: Done as above 👍

brucebolt (approved):
Thanks, the two commits are much easier to follow.

richardTowers:
I think there's also some refactoring that could be done to move some of the duplicated code into SubstitutionHelper, which I've sketched out on [this branch](https://github.com/alphagov/publishing-api/compare/add-substitute-event..towers/add-substitute-event-2?expand=1) - I think that could be taken out of scope an addressed in a future PR though.

richardTowers (approved):
Looks good to me - thanks for the perseverence on this 🙇🏻 

It's probably worth getting back to @csutter as this gets deployed, as he might have an idea of a good way to test it / let us know whether the new behaviour works well enough for the search use case.

PR #2666, opened 2024-03-06
Allow external redirects without a subdomain

brucebolt:
The external redirects list currently requires the domain to have a subdomain for the redirect to be valid, e.g. https://www.nationalhighways.co.uk is valid but https://nationalhighways.co.uk is not.

Removing the leading dot so we can permit redirects that don't contain a subdomain.

[Trello card](https://trello.com/c/LaeOzwYT)


jkempster34 (commented):
Will this validation now allow `evilgov.uk` where it wouldn't before?

brucebolt:
> Will this validation now allow `evilgov.uk` where it wouldn't before?

You're right, it will.  I've added this to the tests and updated the `government_domain?` method to handle this better.  We'll now expect the dot unless the entire host matches something on the allow list.

PR #2669, opened 2024-03-12
Remove specialist topic code from Publishing API 

unoduetre:

## What

When all specialist topic pages have been archived and redirected, we need to remove the specialist topic related code from Publishing API. This application is owned by Publishing Platform.

## Why

To remove technical debt

[Trello ticket](https://trello.com/c/2LcbNRSQ/2417-remove-specialist-topic-code-from-publishing-api-l)


content_schemas/dist/formats/email_alert_signup/frontend/schema.json
```diff
@@ -293,7 +293,6 @@
         "email_alert_type": {
           "type": "string",
           "enum": [
-            "topics",
```
hannako, line 304: I strongly suspect this has nothing to do with specialist topics. Although I've not dug into it enough to find out where this schema is being used (will do so and let you know) ....if for example on [this](https://www.gov.uk/email-signup?link=/world/afghanistan) page you click **confirm** you can see that the params contain a topic attribute. Ie in the domain of email alerts, topic != specialist_topic.

content_schemas/formats/shared/base_links.jsonnet
```diff
@@ -1,22 +1,21 @@
 {
```
hannako: https://github.com/alphagov/publishing-api/pull/2669/commits/d85bfbfb4212073e09106c4eaf318ba4ba28ae13 is a big commit and hard to review. I've always been slightly confused by the difference between base_links and whitehall_edition_links. What do you think about splitting this commit:
- Remove topic from `base_links`, and run `rake build_schemas` as per the [docs](https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#general-workflow). What I don't think that will do is update the examples so you'd have to add those to the commit manually. Perhaps by running `rake` and seeing which fail?
- Remove topic from `whitehall_edition_links` and same as above.

content_schemas/dist/formats/call_for_evidence/frontend/schema.json
```diff
@@ -1020,12 +1020,6 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "additional_topics": {
```
hannako, line 1030: I'll come back to this, or have you already checked? Ie do you have anything that confirms these fields (additional_topics and primary_topics) are specialist topics?

hannako (commented):
Appreciate this is only draft, but wanted to point out a few things before you got too far!

content_schemas/dist/formats/answer/frontend/schema.json
```diff
@@ -149,10 +149,6 @@
           "description": "Prototype-stage taxonomy label for this content item",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
-        "topic_taxonomy_taxons": {
```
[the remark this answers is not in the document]
unoduetre: Reverted. I've created a normal commit as a revert commit would generate too many conflicts.

content_schemas/dist/formats/call_for_evidence/frontend/schema.json
unoduetre: I've grepped through email apps and found nothing. If you know that it is used somewhere else, please, let me know.

content_schemas/formats/shared/base_links.jsonnet
unoduetre: I will split it later before putting this PR as ready for review.

content_schemas/dist/formats/email_alert_signup/frontend/schema.json
unoduetre: Please, let me know ther result of your investigation. This was done by analogy to the other email stuff as suggested by the following comment (which is not directly related though): https://github.com/alphagov/publishing-api/blob/9760686d780a9eec2539ffa3912fbb5bce9fb830/content_schemas/formats/shared/definitions/email_alert.jsonnet#L29. If you think it's something different from the other one, I will revert this change.

content_schemas/dist/formats/call_for_evidence/frontend/schema.json
unoduetre: Also, we have contract tests, but if they cannot be relied on, please, let me know.

content_schemas/dist/formats/email_alert_signup/frontend/schema.json
unoduetre: Also, we have contract tests, but if they cannot be relied on, please, let me know.

hannako: Ah yes I think you're right 👍

hannako:
@unoduetre I know this is still a draft, and that you've previously said you want the schemas to reflect the state of the database and therefore not split the PR. But at 378 files and 11 commits-ish I really think this PR is too big to review or safely merge:

- if something breaks as a result of shipping this PR it will be very hard to figure out which change was responsible. 
- I think it's better practice to have a single pr for the database migration alone so that if we need to roll anything back we don't add complexity. 

My suggestion: We could remove topics from whitehall_edition links in one PR, remove it from base_links in a second PR, remove additional_topics etc in a third. Is there something I'm missing here that makes that impossible?

content_schemas/formats/shared/base_links.jsonnet
unoduetre: I've separated dist files (they are autogenerated) from others, which should make it easier to review.

unoduetre:
As discussed with @hannako, this PR need to be split into a number of separate PRs.

unoduetre:
I've added additional commits as I've found during review one more thing I missed earlier.

content_schemas/formats/email_alert_signup.jsonnet
```diff
@@ -37,7 +37,6 @@
         email_alert_type: {
           type: "string",
           enum: [
-            "topics",
```
hannako, line 40: Let's cherry pick https://github.com/alphagov/publishing-api/pull/2669/commits/ad572f9b2142c2d0377ec4026a366c4c9b9b8545 and  https://github.com/alphagov/publishing-api/pull/2669/commits/2a8893b1b39a60c845542f76ed94b4fc9b3d5cfd into its own work: "Remove `topics` from email_alert_signup schema and associated examples". I'm not sure we've removed this field from the publishing applications as per the documentation for [removing-a-field-from-a-content-schema ](https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#removing-a-field-from-a-content-schema) so it would be nice to have this as a standalone PR, so we can link to relevant changes in other apps.

content_schemas/dist/formats/call_for_evidence/frontend/schema.json
hannako: I'm a bit confused by https://github.com/alphagov/publishing-api/pull/2669/commits/ddaf127ef7da368238d376a16312dd0701c2b41f and https://github.com/alphagov/publishing-api/pull/2669/commits/513f78f0285b85b093b1747286b22b8bfe8eb64b

It looks like we need to remove topics, additional_topics and primary topic from the tags attribute of the email alert schema. And the relevant examples need to be updated. But I'm struggling to figure out if the examples you've added are including this schema. If they are, then it might be that there's some additional code changes needed which would form part of this change. 

So let's make this it's own PR. As yet I've not removed any code from email alert service or email alert api that might be using these fields. So hopefully the contract tests will fail in a helpful and informative way so we can link all the relevant work.

db/migrate/20240314140958_delete_specialist_topics_from_database.rb
```diff
@@ -0,0 +1,43 @@
+class DeleteSpecialistTopicsFromDatabase < ActiveRecord::Migration[7.1]
```
hannako, line 1: As mentioned previously - I think it's safer to have a standalone PR for the database migration. And I'd definitely want to chat this PR through with someone from Publishing Platform first.

hannako (commented):
Thanks for such a thorough piece of investigation @unoduetre really amazing attention to detail!

I have obviously suggested a rework, but hopefully it will make things easier rather than more complicated. Am very happy to pair on any aspect.

unoduetre:
This PR has been split into the following PRs:

* https://github.com/alphagov/publishing-api/pull/2677
* https://github.com/alphagov/publishing-api/pull/2678
* https://github.com/alphagov/publishing-api/pull/2679
* https://github.com/alphagov/publishing-api/pull/2680
* https://github.com/alphagov/publishing-api/pull/2682
* https://github.com/alphagov/publishing-api/pull/2683
* https://github.com/alphagov/publishing-api/pull/2684

PR #2680, opened 2024-03-18
2417 remove specialist topic code from publishing api l 4

unoduetre:

Part of

[🏔Epic | Remove specialist topic code from Publishing API [L]EPICS 🌀](https://trello.com/c/2LcbNRSQ/2417-%F0%9F%8F%94epic-remove-specialist-topic-code-from-publishing-api-l "‌")

## What

Remove topics from email\_alert\_signup schema

## Why

Retiring specialist topics
[Trello card](https://trello.com/c/lJFANf8k/2488-remove-topics-from-emailalertsignup-schema-m-l)

## How

[https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#removing-a-field-from-a-content-schema](https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#removing-a-field-from-a-content-schema "‌")

### DIFF
--- a/content_schemas/dist/formats/email_alert_signup/frontend/schema.json
+++ b/content_schemas/dist/formats/email_alert_signup/frontend/schema.json
@@ -301,7 +301,6 @@
         "email_alert_type": {
           "type": "string",
           "enum": [
-            "topics",
             "policies",
             "countries"
           ]
--- a/content_schemas/dist/formats/email_alert_signup/notification/schema.json
+++ b/content_schemas/dist/formats/email_alert_signup/notification/schema.json
@@ -392,7 +392,6 @@
         "email_alert_type": {
           "type": "string",
           "enum": [
-            "topics",
             "policies",
             "countries"
           ]
--- a/content_schemas/dist/formats/email_alert_signup/publisher_v2/schema.json
+++ b/content_schemas/dist/formats/email_alert_signup/publisher_v2/schema.json
@@ -180,7 +180,6 @@
         "email_alert_type": {
           "type": "string",
           "enum": [
-            "topics",
             "policies",
             "countries"
           ]
--- a/content_schemas/formats/email_alert_signup.jsonnet
+++ b/content_schemas/formats/email_alert_signup.jsonnet
@@ -37,7 +37,6 @@
         email_alert_type: {
           type: "string",
           enum: [
-            "topics",
             "policies",
             "countries",
           ],


hannako:
I did some digging into where this schema is being used:
- It is used by one of the three email signup routes supported by [email alert frontend](https://github.com/alphagov/email-alert-frontend/blob/main/README.md?plain=1#L9), ie signups from `<base_path>/email-signup`, [see controller](https://github.com/alphagov/email-alert-frontend/blob/bea75b7d43519bc97c108fb2a2d36dab1c28be24/config/routes.rb#L6)
- this route is currently only used by travel advice pages eg https://www.gov.uk/api/content/foreign-travel-advice/france/email-signup
- the concept of the `email-alert-type` is I think legacy code from when signups were possible from policy pages (clues [here](https://github.com/alphagov/email-alert-frontend/commit/e89bb9a0efbacfc3582bda27f10facc0d2f425f1?diff=split&w=0), [here](https://github.com/alphagov/email-alert-frontend/pull/17) and [here](https://github.com/alphagov/email-alert-api/pull/93) ) so _probably_ the whole key could be removed. As it seems like on travel advice pages, the links contain the `countries` key.
- Removing the whole key is out of scope of this work, but just noting here to make the review easier:

**TLDR**: I don't think the `topic` field was being used anywhere anyway so there are no consequences of removing it.

hannako (approved):
I did some digging into where this schema is currently being used.
- It's used on travel advice sign up pages only. eg https://www.gov.uk/api/content/foreign-travel-advice/afghanistan/email-signup
- It doesn't look like the email_alert_type field is even used tho, as the countries key is in top level links
- I think the `policies` key (and potentially the whole `email_alert_type`) key could be removed too (clues [here](https://github.com/alphagov/email-alert-frontend/commit/e89bb9a0efbacfc3582bda27f10facc0d2f425f1?diff=split&w=0) [here](https://github.com/alphagov/email-alert-frontend/pull/17) and [here](https://github.com/alphagov/email-alert-api/pull/93)) but that's out of scope of this work. https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/116490261/Policies

TLDR: I don't think the `topic` key was being used anyhow, so I can't foresee any unexpected consequences of removing it.

jkempster34 (approved):

PR #2684, opened 2024-03-18
[TODO] Remove specialist topics from database

unoduetre:


hannako:
Sorry @unoduetre am only getting to this now...

I don't think we need/should do this:

- When a document is unpublished/redirected it remains in the database. We should follow that pattern. 
- Publishing API is the source of truth, it stores all of the documents, along with their current state (via the Edition) so deleting these documents means we are losing access that that history.

@brucebolt do you agree?

JonathanHallam:
@hannako I agree, not sure we need to do this. I assume all the specialist topics are already marked as withdrawn or unpublished so I don't think there's a need to do this

unoduetre:
Closing the PR.

PR #2686, opened 2024-03-19
Use edition links for Worldwide Offices

brucebolt:
We are working to make Worldwide Organisations and their offices editionable.

Therefore we need to be able to link the links to editions, rather than documents, to prevent draft updates becoming live before publication.

[Trello card](https://trello.com/c/5Nr1NieG)


JonathanHallam (approved):
thanks!

PR #2687, opened 2024-03-19
Remove non-edition links for worldwide offices

brucebolt:
In https://github.com/alphagov/whitehall/pull/8910, we removed the presentation of non-edition links for worldwide offices.

Therefore we can remove them from the schema.

Depends on https://github.com/alphagov/whitehall/pull/8910.

[Trello card](https://trello.com/c/5Nr1NieG)


JonathanHallam (approved):
👍

PR #2725, opened 2024-05-03
Add Rake task to requeue all ever published editions

csutter:
We currently have two Rake tasks that requeue messages for content, one for all live editions, and another for all live and published editions. This adds a third one to requeue messages for all editions that are either published or unpublished, or in other words, all editions that are or have been published at one point or another.

For downstream applications that rely on incoming messages from Publishing API to keep their internal state consistent with upstream, it can be useful to also requeue messages for unpublished content - as they can be used to delete content where they may have missed the unpublishing message for whatever reason.

### DIFF
--- a/lib/tasks/queue.rake
+++ b/lib/tasks/queue.rake
@@ -51,6 +51,20 @@ namespace :queue do
     RequeueContentByScope.new(Edition.live, action: args.action).call
   end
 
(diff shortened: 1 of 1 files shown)


JonathanHallam (approved):
The idea of running this scares me a bit but I don't see the harm in having it given we already have the other tasks to republish subsets!

PR #2747, opened 2024-05-29
Associate Document, LinkSet and Link

richardTowers:
Add [rails associations](https://guides.rubyonrails.org/association_basics.html) to the Document, LinkSet and Link models to make interacting with them a little simpler.

I'm only making use of these in one place right now, because it was an obvious improvement. I imagine there are quite a few places in the Queries module where these associations would help us avoid bits of inline SQL in `joins()` statements, but a lot of that code is quite... scary.

My actual motivation for this is that we're going to want to write a complicated query for the GraphQL work which does something like "given this content id, get all the linked documents, and then find the published edition for each of them". This query will be a lot easier to write if we can use associations instead of inline SQL.

While I was thinking about 857b5c1 I think I spotted a problem with that method, so I'm going to raise another PR to add a warning to that one to see if it's used.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -4,6 +4,7 @@ class Document < ApplicationRecord
   belongs_to :owning_document, class_name: "Document", optional: true
   has_many :editions
   has_one :link_set, primary_key: "content_id", foreign_key: "content_id", inverse_of: :document
+  has_many :link_set_links, class_name: "Link", through: :link_set, source: :links
   has_one :draft, -> { where(content_store: "draft") }, class_name: "Edition"
   has_one :live, -> { where(content_store: "live") }, class_name: "Edition"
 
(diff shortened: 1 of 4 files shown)


app/models/document.rb
```diff
@@ -3,6 +3,7 @@ class Document < ApplicationRecord
 
   belongs_to :owning_document, class_name: "Document", optional: true
   has_many :editions
+  has_one :link_set, primary_key: "content_id", foreign_key: "content_id", inverse_of: :document
```
brucebolt, line 6: I've often wondered what the point of `LinkSet` actually is, since it appears to hold no more data than the `Document` itself.  But I guess that's something for another time.

brucebolt (approved):
Looks good to me.  Really like the removal of that raw SQL!

You've got a `WIP` on the final commit message.

PR #2750, opened 2024-06-03
Schema for popular links

anatron:


content_schemas/formats/popular_links.jsonnet
```diff
@@ -0,0 +1,40 @@
+{
+  document_type: "popular_links",
```
richardTowers, line 2: I think there's a case to make that this should be named and implemented a bit more generically.

If we called it (say) `"link_collection"` then it could potentially be used by other content items that need to include a collection of links.

By calling it "popular links" we kind of restrict it to just this one use case.

In terms of implementation, I think we'd need to drop the `minItems: 6` / `maxItems: 6` constraint if we wanted it to be generic. I think that's fine - if we want exactly 6 items in the popular links document for the homepage, we can enforce that in Publisher. Just because we can validate it here, doesn't mean we should.

content_schemas/formats/popular_links.jsonnet
```diff
@@ -0,0 +1,40 @@
...
+  routes: "optional",
+  rendering_app: "optional",
+  title: "optional",
+  redirects: "optional",
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {
+        links_details: {
```
syed-ali-tw, line 13: We think there could be a better name for this field and we welcome any feedback on improving this name, the use of word link is very common inside publishing api and hence the hesitance.

richardTowers, line 13: This name feels a little repetitive - `details: { links_details: ...`. I appreciate you don't really want just `links`, as that's going to get confused with the whole concept of `links`.

Maybe we could use `link_collection` for this key, whether or not we call the document type itself `link_collection`?

content_schemas/formats/popular_links.jsonnet
```diff
@@ -0,0 +1,40 @@
...
+              title: {
+                type: "string"
+              },
+              url: {
+                type: "string"
+              }
+            }
+          }
+        },
+        language: {
```
richardTowers, line 34: Is language the same thing as `locale`?

If so, we shouldn't need it inside details - the usual way would be to have different `documents` for the different `locale`s.

content_schemas/formats/homepage.jsonnet
```diff
@@ -6,5 +6,7 @@
       properties: {},
     },
   },
-  links: {},
+  links: {
+    popular_links: "Home page popular links"
```
richardTowers, line 10: I mentioned below that I think `link_collection` might be a better name for the document type here, _but_ that doesn't mean you can't call the link from the homepage to the link_collection `popular_links`.

Consider [ordered_cabinet_ministers](https://github.com/alphagov/publishing-api/blob/a74f0a509020a20d460f45c07d3227d71d919605/content_schemas/formats/ministers_index.jsonnet#L20) on the ministers index page - the document type it's linking to is going to be Person, but the name of the link is `ordered_cabinet_ministers`. I think that's the right way to represent the semantics of the thing you're linking to.

So I would call the document type `link_collection`, but I would call this link type `popular_links`.

content_schemas/formats/popular_links.jsonnet
```diff
@@ -0,0 +1,40 @@
+{
```
richardTowers, line 1: All the other jsonnet files start with `(import "shared/default_format.jsonnet") + {`. I'm not sure whether this is strictly necessary, but I think unless you've got a good reason to diverge it's best to be consistent.

richardTowers (changes requested):

content_schemas/formats/popular_links.jsonnet
anatron: Accepted

content_schemas/formats/homepage.jsonnet
anatron: Accepted

content_schemas/formats/homepage.jsonnet
```diff
@@ -6,5 +6,7 @@
       properties: {},
     },
   },
-  links: {},
+  links: {
+    link_collection: "Collection of links to be used to display popular links on homepage"
```
richardTowers, line 10: I would prefer this link to still be called `popular_links` I think - it doesn't have to match the document type.

The difference is between:

```
{
  "base_path": "/",
  "document_type": "homepage",
  "links": {
    "link_collection": [
      {
        "document_type": "link_collection",
        ...
```

and:


```
{
  "base_path": "/",
  "document_type": "homepage",
  "links": {
    "popular_links": [
      {
        "document_type": "link_collection",
        ...
```

syed-ali-tw: Ah ok, thats interesting, make sense.

content_schemas/formats/link_collection.jsonnet
```diff
@@ -0,0 +1,35 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: "link_collection",
+  base_path: "optional",
+  routes: "optional",
+  rendering_app: "optional",
+  title: "optional",
```
richardTowers, line 6: I wonder if we should make this mandatory...

Just thinking if we _do_ ever end up using `link_collection` for more than one content item, it might get hard to tell them apart if we don't strongly encourage the publishing apps to identify them somehow...

I guess also we could use the title for the "Popular on GOV.UK" heading on the homepage... If we ended up using the popular links in other places, would we want to use the same title everywhere, and be able to update it once? Maybe?

syed-ali-tw: Yeah, that makes sense again, title kind of becomes the identifier. thank you.

richardTowers (approved):
Looks great - thanks for taking the time to think this through!

PR #2763, opened 2024-06-07
Overhaul of link expansion docs

ryanb-gds:
This commit makes the differences between link set links and edition links more explicit. It also explains why publishing applications might use one or the other of those categories of link and why each of them exists.

It also attempts to explain dependency resolution in a bit more detail.

This commit also includes a large number of edits for clarity.


docs/link-expansion.md
```diff
@@ -25,22 +25,13 @@
...
-used to determine and expand links at the point of request.
+Link expansion is the process of converting the stored links of an edition into a JSON representation containing more information about the linked content items. The expansion process occurs when sending an edition downstream to the
+[Content Store][content-store].
 
 A closely related process to this is
-[dependency-resolution](dependency-resolution.md). This is something of a
-reversal in link expansion and has the role of determining which editions
-will need re-presenting to the content store as a result of an update to a
-document.
+[dependency resolution](dependency-resolution.md). This is the opposite of link expansion. When a document is updated, Publishing API works out which documents link to the updated document, and presents the linked documents to Content Store.
```
ryanb-gds, line 34: I wanted to add something here about the limitations of dependency resolution in cases where the cardinality of the links has changed or the ordering of the links, but I couldn't actually rationalise what the limitation is.

Theoretically when an edition has a new link added to it, dependency resolution should run to ensure that any content items which link to the edition are presented to content store with their latest expanded links. Is that not how things work in reality?

Obviously when it's a new document we can't have linked to it yet so dependency resolution can't work, but that can be managed through Publishing apps (though at this point if Publishing apps are doing callbacks to update related items on create, they might as well do them on update and delete as well, so dependency resolution isn't abstracting away a lot of complexity for them).

docs/link-expansion.md
```diff
@@ -107,41 +98,31 @@ Within a `links` JSON object there are keys which indicate the type of link
...
-These links are added via the
-[`patch-link-set`](api.md#patch-v2linkscontent_id) endpoint. They are
-associated with a `content_id`, which therefore associates them with all
-the locales of a document.
+In cases when there are edition links and link set links which have the same
+`link_type`, the edition links will take precedence during link expansion.
 
-They are typically used for taxonomy based links and aren't related to the
-specific content of an edition.
+Edition links should be the default approach to link creation, because link set links have the disadvantage of applying to all editions and locales of a document, which is generally not what is required in most use cases. However, edition links have a substantial limitation, which is that recursive link expansion is not applied to them. If a document needs to access data from documents more than one link away, then link set links must be used.
```
brucebolt, line 113: Could you link to the section on recursive link expansion here, just in case the reader isn't sure what that is.

docs/link-expansion.md
```diff
@@ -107,41 +98,31 @@ Within a `links` JSON object there are keys which indicate the type of link
...
 
-They do not follow a draft/published workflow, once added they will apply to
-editions of a document on both draft and live content stores.
+There was an attempt to allow recursive expansion of edition links, but we weren't able to complete it. See <https://github.com/alphagov/publishing-api/pull/2605>.
```
brucebolt, line 115: It's worth noting that we found another (albeit convoluted) way of solving the problem we faced, without needing recursive link expansion.  We probably could've got the recursive expansion working if we'd dedicated a lot more time to it.

brucebolt (approved):
Thanks for this 🎖️ .  These are much clearer and easier to understand now.

Gweaton (approved):
I've found this a fair bit clearer than the previous iteration, and it's nicely structured. Thanks for making these changes 👍

PR #2767, opened 2024-06-11
Add API methods to get schema information

pezholio:
This adds some API methods to return information about the schemas that Publishing API knows about, as well as fetching a schema by name. This allows us to get up to date schema information without having to have a local copy of the schemas stored somewhere and using the [GOV.UK Schemas](https://github.com/alphagov/govuk_schemas/) gem.

This will allow the content modelling team to have a way to create many different types of reusable content using the Publishing API as a single source of truth for the schema, rather than having to keep fields, validation rules etc in two places.

Happy to chat through in more detail if anyone needs any more context / has any questions!

### DIFF
--- a/app/services/schema_service.rb
+++ b/app/services/schema_service.rb
@@ -5,4 +5,11 @@ class SchemaService
                         .select { |k| k.ends_with?("schema.json") }
                         .transform_keys { |k| format_regex.match(k)[1] }
   end
+
+  def self.find_schema_by_name(name)
+    GovukSchemas::Schema.find(publisher_schema: name)
+  rescue Errno::ENOENT
+    message = "Could not find publisher schema with name: #{name}"
+    raise CommandError.new(code: 404, message:)
+  end
 end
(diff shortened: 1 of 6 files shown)


spec/controllers/v2/schemas_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.describe V2::SchemasController do
```
Harriethw, line 1: does there also need to be a request spec?

app/services/schema_service.rb
```diff
@@ -0,0 +1,8 @@
+class SchemaService
+  def self.all_schemas
+    format_regex = Regexp.new(".+\/formats\/([a-z_]+)\/.+")
```
Harriethw, line 3: so here we are using the file path to get the type of object the schema is for? Is there any other way to get the type, e.g. in the schema itself with a `title` key for instance? Just wondering if there's a risk depending on a file path set in another gem which could easily change...

spec/controllers/v2/schemas_controller_spec.rb
pezholio: I guess I'm kinda doing a request spec over a controller spec here, as I guess controller specs are more designed to be "unit-type" tests, so I think I'll switch over 👍

app/services/schema_service.rb
Harriethw: After chat with @pezholio There is no codified place for the name/type except in folder name

PR #2769, opened 2024-06-12
Republish drafts of other translations on PUT of new translation

brucebolt:
When a draft translation is added and another translation of the same content ID exists, we need to send the existing translation(s) downstream so their `available_translations` links can be updated to include the new translation.

We're not currently doing this, which leads to a situation where the link to the new translation never appears on the draft of the existing translations.

[Trello card](https://trello.com/c/8St6Zw4w)

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -4,6 +4,8 @@ module Commands
       def call
         PutContentValidator.new(payload, self).validate
 
+        update_draft_items_of_different_locale_but_matching_content_id
+
         remove_previous_path_reservations
         reserve_current_path
         clear_draft_items_of_same_locale_and_base_path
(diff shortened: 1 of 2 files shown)


app/commands/v2/put_content.rb
```diff
@@ -200,6 +202,29 @@ def clear_draft_item_of_different_locale_but_matching_base_path
         )
       end
 
+      def update_draft_items_of_different_locale_but_matching_content_id
+        return unless payload[:base_path]
+        return if Document.where(content_id: payload[:content_id], locale: payload[:locale]).any?
```
Gweaton, line 207: Is it worth adding a couple of tests for these early returns? Might also help serve as documentation.

Gweaton (approved):
I think this makes sense, just have a couple of qs/comments

app/commands/v2/put_content.rb
```diff
@@ -200,6 +202,29 @@ def clear_draft_item_of_different_locale_but_matching_base_path
...
+
+        draft_editions_for_different_locale = Document
+          .where(content_id: payload[:content_id])
+          .where
+          .not(locale: payload[:locale])
+          .map(&:draft)
+          .compact
+
+        after_transaction_commit do
```
Gweaton, line 216: What's the transaction here? Is it that the rest of the `call` method executes successfully and we only want to run this if it does? Or is it scoped to this method?

app/commands/v2/put_content.rb
brucebolt: Yes, that's a good idea, I'll add them.

app/commands/v2/put_content.rb
brucebolt: Yes, it's after the `call` (I think it's [this line](https://github.com/alphagov/publishing-api/blob/5ad088fe246c3dcaa4f73efdb9e637df6b35eaed/app/commands/base_command.rb#L21) that executes the block), so we only send the payload to content store once the data has actually been committed to the database.

PR #2772, opened 2024-06-17
Add pact tests for new Schema endpoints

tahb:
## Changes in this PR

* [We've recently added new Schema endpoints](https://github.com/alphagov/publishing-api/pull/2767) for listing and getting individual schemas
* We now want to add [helpers into the GDS API Adapters](https://github.com/alphagov/gds-api-adapters/pull/1275) to support this and so have to come back to add Pact tests here

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -56,6 +56,32 @@ Pact.provider_states_for "GDS API Adapters" do
     end
   end
 
+  provider_state "there are publisher schemas" do
+    set_up do
+      schemas = {
+        "/govuk/publishing-api/content_schemas/dist/formats/email_address/publisher_v2/schema.json": {
+          type: "object",
+          required: %w[a],
+          properties: {
+            email_address: { "some" => "schema" },
+          },
+        },
+        "/govuk/publishing-api/content_schemas/dist/formats/tax_license/publisher_v2/schema.json": {
+          type: "object",
+          required: %w[a],
+          properties: {
+            tax_license: { "another" => "schema" },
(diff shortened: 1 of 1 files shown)


app/services/schema_service.rb
```diff
@@ -9,7 +9,7 @@ def self.all_schemas
   def self.find_schema_by_name(name)
     GovukSchemas::Schema.find(publisher_schema: name)
   rescue Errno::ENOENT
-    message = "Could not find publisher schema with name: #{name}"
-    raise CommandError.new(code: 404, message:)
+    msg = "Could not find publisher schema with name: #{name}"
+    raise CommandError.new(code: 404, message: msg)
```
tahb, line 13: @pezholio curious about this one! Is this a thing?!

pezholio: It's a pattern I've seen elsewhere in the app and seems to work well

pezholio: Sorry, missed the context from the commit. This is indeed a thing, introduced in Ruby 3.1 - https://dev.to/baweaver/ruby-3-1-shorthand-hash-syntax-first-impressions-19op

tahb: Oh nice, I somehow couldn't find that in my search. 

> Posted on 12 Sept 2021

Slightly late to this party! I suppose it's okay once you know but otherwise not in love with it. As it's used elsewhere I'll drop this commit.

pezholio (approved):
LGTM 👍

Gweaton (approved):
Looks good 👍 just one q but nothing blocking

PR #2776, opened 2024-06-20
Add Content Block Email Address content type

pezholio:
The Content Modelling Team are working on a spike to add the ability to publishing reusable pieces of content (“Content Blocks”) via Whitehall. The first content type we’re looking at is Email addresses, so we’re adding a simple content schema to cover this functionality.

### DIFF
--- a/content_schemas/examples/content_block_email_address/publisher_v2/example.json
+++ b/content_schemas/examples/content_block_email_address/publisher_v2/example.json
@@ -1,0 +1,11 @@
+{
+  "locale": "en",
+  "schema_name": "content_block_email_address",
+  "document_type": "content_block_email_address",
+  "title": "Government Digital Service - General contact",
+  "description": "General contact email address for Government Digital Service",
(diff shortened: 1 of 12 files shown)


content_schemas/examples/content_block_email_address/publisher_v2/example.json
```diff
@@ -0,0 +1,11 @@
+{
+  "locale": "en",
+  "schema_name": "content_block_email_address",
+  "document_type": "content_block_email_address",
+  "title": "Government Digital Service",
+  "description": "Government Digital Service",
```
Gweaton, line 6: What's description being used for here? And `title` I suppose, unless that's supposed to represent the name of the thing that has an email address?

pezholio: I've updated this to be clearer. It's still very much in flux though

Gweaton (approved):

PR #2787, opened 2024-06-26
Temporarily pin ruby version to 3.3.1

Gweaton:
Deployments to integration are currently broken due to the `jsonnet` gem
failing to install.

We believe this is due to an issue with recent govuk-ruby-builder
images, and using 3.3.3 (the latest patch at time of writing) causes the
`jsonnet` gem to fail to install. This is currently being investigated
but I'm pinning Ruby back to 3.3.1 to unblock deployments.

Issue open here: https://github.com/alphagov/govuk-ruby-images/issues/96

[Relevant slack conversation](https://gds.slack.com/archives/C013F737737/p1719399161016749)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-ARG ruby_version=3.3
+# TODO: unpin obsolete base image once https://github.com/alphagov/govuk-ruby-images/issues/96 is resolved
+ARG ruby_version=3.3.1
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
 ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:$ruby_version
 


brucebolt (approved):

yndajas (approved):
Nice investigation! A few minor things on the commit message:
- Is `govuk-ruby builder images` meant to be `govuk-ruby-images builds`?
- I think it might be easier to scan for the motivation for this change if you mention at the start that the Publishing API integration deployments are broken. Currently "to unblock deployments" is just a short few words right at the end
- Might also be useful to link to [the commit you referenced on Slack](https://github.com/alphagov/publishing-api/pull/2773/commits/4c0afcbaaf61f6bc92dbdf443db6dab663450fba) as the change that broke things? The issue doesn't mention it

sengi (approved):

PR #2788, opened 2024-06-26
Use libjsonnet from the base image.

sengi:
https://github.com/yugui/ruby-jsonnet/ appears to be unmaintained and pins an obsolete version of libjsonnet++ which doesn't compile with g++13 (which we recently switched to in govuk-ruby-images because it's the default C++ compiler in Ubuntu 24.04).

Thankfully the ruby-jsonnet build has an option to tell it not to try to build its own obsolete version of libjsonnet and use a preinstalled one.

As a bonus side-effect, this speeds up the build because we no longer have to build libjsonnet++.

Tested: `bundle install` succeeds with https://github.com/alphagov/govuk-ruby-images/pull/97

Rollout: needs alphagov/govuk-ruby-images#97 to have [landed](https://github.com/alphagov/govuk-ruby-images/actions/runs/9682506418) on ghcr.io.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-# TODO: unpin obsolete base image once https://github.com/alphagov/govuk-ruby-images/issues/96 is resolved
-ARG ruby_version=3.3.1
+ARG ruby_version=3.3
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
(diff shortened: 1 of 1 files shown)


sengi:
Whoops forgot the headers.

theseanything (approved):

Gweaton (approved):
Nice work, thanks for jumping on this.

PR #2792, opened 2024-06-27
Don’t require a rendering app for content blocks

pezholio:
Content blocks won’t be rendered by a specific app, so we don’t need a rendering app to be given.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -37,6 +37,7 @@ class Edition < ApplicationRecord
 
   NON_RENDERABLE_FORMATS = %w[redirect gone].freeze
   NO_RENDERING_APP_FORMATS = %w[contact external_content role_appointment].freeze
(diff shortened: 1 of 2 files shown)


app/models/edition.rb
```diff
@@ -246,6 +247,10 @@ def renderable_content?
   end
 
   def requires_rendering_app?
-    renderable_content? && NO_RENDERING_APP_FORMATS.exclude?(document_type)
+    !is_contact_block? && renderable_content? && NO_RENDERING_APP_FORMATS.exclude?(document_type)
+  end
+
+  def is_contact_block?
```
brucebolt, line 253:
```suggestion
  def is_content_block?
```

pezholio: Good catch!

brucebolt (approved):

PR #2794, opened 2024-07-01
Spike/look into embedding content

pezholio:
This is a (possibly quite naive) expression of how I think content embedding might work. The flow works by looking for content in the format `{{embed:CONTENT_TYPE:UUID}}` within govspeak / HTML and follows the following flow:

```mermaid
graph TD;
id1[Content sent to API]-->id2[Links extracted from content]-->id3[Govspeak transformed]-->id4[Placeholders replaced with embedded content]-->id5[Transformed content sent to Content API];
```

It currently only supports embeds in a `body` attribute, but I think we can easily make this work with any field within a content type's payload. 

Current assumption is that once the links are extracted, this makes the content a dependency on the embedded content, so when changes are made, the content is re presented for publishing. Could do with testing this assumption though.

Thoughts welcome! 👍

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -87,15 +87,22 @@ module Commands
       end
 
       def create_links(edition)
-        return if payload[:links].nil?
+        links = (payload[:links] || {}).merge(fetch_embedded_content(edition))
+        return if links.empty?
 
-        payload[:links].each do |link_type, target_link_ids|
+        links.each do |link_type, target_link_ids|
           edition.links.create!(
             target_link_ids.map.with_index do |target_link_id, i|
               { link_type:, target_content_id: target_link_id, position: i }
             end,
           )
         end
+      end
+
+      def fetch_embedded_content(edition)
+        return {} if edition[:details]["body"].nil?
+
+        ContentEmbedService.new(edition[:details]["body"]).fetch_links
       end
 
       def create_redirect
--- a/app/services/content_embed_service.rb
+++ b/app/services/content_embed_service.rb
@@ -1,0 +1,56 @@
+class ContentEmbedService
+  include ApplicationHelper
+
+  SUPPORTED_DOCUMENT_TYPES = %w[contact].freeze
+  UUID_REGEX = /([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/
+  EMBED_REGEX = /({{embed:(#{SUPPORTED_DOCUMENT_TYPES.join('|')}):#{UUID_REGEX}}})/
+
+  def initialize(body)
+    @body = body
+  end
+
+  def render
+    match_data = @body.scan(EMBED_REGEX)
+
+    match_data.each do |match|
+      embed_code = match[0]
+      document_type = match[1]
+      content_id = match[2]
+
+      edition = find_edition(document_type, content_id)
+      # TODO: Link Edition to document
+      replace_embed_code_with_content(embed_code, edition)
+    end
+
+    @body
+  end
+
+private
+
+  def find_edition(document_type, content_id)
+    Edition.with_document.find_by(
+      state: "published",
+      content_store: "live",
+      document_type:,
+      documents: { content_id: },
+    )
+    ## TODO: What do we do if the edition can't be found?
+  end
+
+  def replace_embed_code_with_content(embed_code, edition)
+    @body.gsub!(
+      embed_code,
+      renderer.render(partial: view_for_document_type(edition.document_type), locals: { details: edition.details }, formats: [:html]),
+    )
+  end
+
+  def renderer
+    @renderer ||= ActionController::Base
+  end
+
+  def view_for_document_type(document_type)
+    {
+      "contact" => "contacts/contact",
+    }[document_type]
+  end
+end
--- a/app/services/content_embed_service.rb
+++ b/app/services/content_embed_service.rb
@@ -5,21 +5,42 @@ class ContentEmbedService
   UUID_REGEX = /([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/
   EMBED_REGEX = /({{embed:(#{SUPPORTED_DOCUMENT_TYPES.join('|')}):#{UUID_REGEX}}})/
 
+  class EmbeddedEdition
+    attr_reader :embed_code
+
+    def initialize(embed_code:, document_type:, content_id:)
+      @embed_code = embed_code
+      @document_type = document_type
+      @content_id = content_id
+    end
+
+    def edition
+      @edition ||= Edition.with_document.find_by(
+        state: "published",
+        content_store: "live",
+        document_type: @document_type,
+        documents: { content_id: @content_id },
+      )
+    end
+  end
+
   def initialize(body)
     @body = body
   end
 
+  def embedded_editions
+    @embedded_editions ||= @body.scan(EMBED_REGEX).map do |match|
+      EmbeddedEdition.new(
+        embed_code: match[0],
+        document_type: match[1],
+        content_id: match[2],
+      )
+    end
+  end
+
   def render
-    match_data = @body.scan(EMBED_REGEX)
-
-    match_data.each do |match|
-      embed_code = match[0]
-      document_type = match[1]
-      content_id = match[2]
-
-      edition = find_edition(document_type, content_id)
-      # TODO: Link Edition to document
-      replace_embed_code_with_content(embed_code, edition)
+    embedded_editions.each do |embedded_edition|
+      replace_embed_code_with_content(embedded_edition.embed_code, embedded_edition.edition)
     end
 
     @body
--- a/app/services/content_embed_service.rb
+++ b/app/services/content_embed_service.rb
@@ -6,7 +6,7 @@ class ContentEmbedService
   EMBED_REGEX = /({{embed:(#{SUPPORTED_DOCUMENT_TYPES.join('|')}):#{UUID_REGEX}}})/
 
   class EmbeddedEdition
-    attr_reader :embed_code
+    attr_reader :embed_code, :document_type, :content_id
 
     def initialize(embed_code:, document_type:, content_id:)
       @embed_code = embed_code
@@ -35,6 +35,12 @@ class ContentEmbedService
         document_type: match[1],
         content_id: match[2],
       )
+    end
+  end
+
+  def fetch_links
+    embedded_editions.group_by(&:document_type).transform_values do |items|
+      items.map(&:content_id)
     end
   end
 
--- a/spec/integration/put_content/content_with_embedded_content_spec.rb
+++ b/spec/integration/put_content/content_with_embedded_content_spec.rb
@@ -1,0 +1,21 @@
+RSpec.describe "PUT /v2/content when embedded content is provided" do
+  include_context "PutContent call"
+
+  context "with embedded content" do
+    let(:contact) { create(:edition, state: "published", content_store: "live", document_type: "contact") }
+    let(:document) { create(:document) }
+    let!(:link) { document.content_id }
+
+    before do
+      payload.merge!(document_type: "press_release", schema_name: "news_article", details: { body: "{{embed:contact:#{contact.document.content_id}}}" })
+    end
+
+    it "should create links" do
+      expect {
+        put "/v2/content/#{content_id}", params: payload.to_json
+      }.to change(Link, :count).by(1)
+
+      expect(Link.find_by(target_content_id: contact.content_id)).to be
+    end
+  end
+end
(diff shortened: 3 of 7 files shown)


richardTowers:
Before looking at the substance of this, I'll just note that I did a quick check to see if the `{{` syntax would conflict with anything:

```
select * from editions
where editions.state = 'published'
and editions.details ->> 'body' LIKE '%{{%'
```

Two pages - https://www.gov.uk/government/publications/open-contracting and https://www.gov.uk/hmrc-internal-manuals/international-manual/intm820513

... so `{{embed:CONTENT_TYPE:UUID}}` would definitely be unique / wouldn't conflict with existing content, but `{{just_some_string}}` would have a couple of collisions. We can bikeshed the syntax later though.

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,83 @@
+class ContentEmbedService
+  include ApplicationHelper
+
+  SUPPORTED_DOCUMENT_TYPES = %w[contact].freeze
+  UUID_REGEX = /([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/
+  EMBED_REGEX = /({{embed:(#{SUPPORTED_DOCUMENT_TYPES.join('|')}):#{UUID_REGEX}}})/
```
richardTowers, line 6: A few thoughts about the format here (some of this is bike shedding)

- `{{` vs `[` or `[[` - we'd probably want to do some user research on these formats - content designers are not used to entering curly braces currently, as markdown / govspeak doesn't use them. Although ideally people won't be typing them out by hand (visual editor / insert object menu item), if it looks like the `{{` syntax is scary / confusing we should consider alternatives. We'll have to make sure whatever we choose doesn't collide with existing syntax.
- The `embed:type:guid` syntax comes from ADR004, but it doesn't really justify why that syntax is needed / preferable. Particularly if we're using curly braces, I don't think `embed:` is needed to disambiguate from other things. I think the `type` bit is probably helpful namespacing so the users can see what type of embed it is, and I guess it makes the edition search a little faster too.
- We might want to be able to specify which field we're referring to (here we're using `title`, but you could imagine wanting to do `{{contact:government_digital_service:office_address}}` or `{{contact:government_digital_service:foi_email_address}}` (i.e. referring to specific fields on complex objects). I think it would be overcomplicating things to allow for JMESPath expressions or anything like that.
- I think the use of UUIDs will be problematic for a couple of UX reasons - firstly they're intimidating, secondly they're very hard to review in 2i. I think we should aim to use friendly IDs instead, although that will open several new cans of worms (e.g. should friendlyID be part of a document, or an edition? If an edition, can it change from one edition to the next? do friendlyIDs have to be globally unique? etc.)

All of that said, I think this pattern is fine for the purposes of this spike. Just wanted to get those thoughts down / start a discussion if people disagree on any bits.

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,83 @@
...
+    def initialize(embed_code:, document_type:, content_id:)
+      @embed_code = embed_code
+      @document_type = document_type
+      @content_id = content_id
+    end
+
+    def edition
+      @edition ||= Edition.with_document.find_by(
+        state: "published",
+        content_store: "live",
```
richardTowers, line 20: Eventually we'll want this to work for draft content too - i.e. if we're processing a draft edition, we should find the latest draft or published Editions for each embed. Looks like it will be a little bit annoying to have to pass that context through the DetailsPresenter unfortunately.

pezholio:
I'm not sure why the [pacts are failing](https://github.com/alphagov/publishing-api/actions/runs/9741662807/job/26881326658?pr=2794) - it seems like we're passing an empty array in the `body` part of details, which is already covered by the presenter specs 🤔

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,83 @@
...
+private
+
+  def find_edition(document_type, content_id)
+    Edition.with_document.find_by(
+      state: "published",
+      content_store: "live",
+      document_type:,
+      documents: { content_id: },
+    )
+    ## TODO: What do we do if the edition can't be found?
```
richardTowers, line 64: Some kind of error I guess...

We might have to think about the UX here too, as I suspect this processing will be happening inside a sidekiq job, so it might not be possible to communicate back to the publishing app that the user has referred to a UUID that doesn't match up with an edition in the right state for the right document type.

We'll also have to think about edge cases like "if someone unpublishes or withdraws an object that's embedded in other editions, what should happen when those editions are republished?". I feel like we'll want some kind of equivalent to `ON DELETE RESTRICT` in SQL, where it's not possible to unpublish / otherwise remove an object if its embedded in any published / draft editions.

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def replace_embed_code_with_content(embed_code, edition)
+    @body.gsub!(
+      embed_code,
+      renderer.render(partial: view_for_document_type(edition.document_type), locals: { details: edition.details }, formats: [:html]),
```
richardTowers, line 70: Nit, but I think the partial is probably overkill here - could just do:

```suggestion
      edition.details[:title]
```

... and worry about what to do for document types where we want a different field than title later.

app/commands/v2/put_content.rb
```diff
@@ -98,6 +99,12 @@ def create_links(edition)
         end
       end
 
+      def fetch_embedded_content(edition)
+        return {} if edition[:details]["body"].nil?
+
+        ContentEmbedService.new(edition[:details]["body"]).fetch_links
```
richardTowers, line 105: I think you might need the same `body.is_a?(Array)` check as you have in DetailsPresenter here (or maybe pull it down into ContentEmbedService)

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      EmbeddedEdition.new(
+        embed_code: match[0],
+        document_type: match[1],
+        content_id: match[2],
+      )
+    end
+  end
+
+  def fetch_links
+    embedded_editions.group_by(&:document_type).transform_values do |items|
```
richardTowers, line 42: This assumes a 1-1 match of document_type -> link_type, but I wonder whether we should have a special link_type for embeds instead.

i.e. this

```mermaid
erDiagram
    Guide ||--o{ Contact : embed
```

Instead of this:

```mermaid
erDiagram
    Guide ||--o{ Contact : contact
```

pezholio: Oh good point!

richardTowers: We'd need to add the new link type to all the schemas for it to be permitted. I think here: https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/shared/base_edition_links.jsonnet

mike21573: I am curious about implementation options that avoid using links. I'm very new to them in both theory and practice, so I don't have a strong sense of how good a fit they are yet. I'm digging through the docs now and trying to answer the question: are we in fact talking about introducing a new concept here? Link expansion and dependency resolution were what made me think of links for embedded content, but on reflection I'm not sure how central those mechanisms are to what a link actually is.

(For me this doesn't impact the value of this spike at all, by the way)

richardTowers: For me, it's really appealing to use the existing concept of links. Partly because we'd dependency resolution "for free", and we'll want to be able to do that (i.e. when an embedded object changes, update all the pages which embed it). Partly because piggy backing on the documents / editions concept (which we have to do if we're using links) means we can benefit from solved problems like "how do I schedule updates of this thing?". But also generally, embedded content just feels like a specialisation of linked content to me.

@ryanb-gds recently updated [the link expansion docs](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html), so those are 100% worth a read if you haven't already. There are certainly some problems with links as they currently work (link set links belonging to a content item, rather than an edition, edition links not being expanded recursively), but at least for the simple case of a direct embed, those won't be a problem.

All that said, it's quite possible I've overlooked some issue with links for this use case, and I'd be very interested in hearing about edge cases or things which we might not be able to achieve with this model.

app/services/content_embed_service.rb
yndajas: I wondered about this, but would we want draft editions to be able to embed unpublished content? What if the edition with embedded content gets published without a published edition of the embedded content existing? Presumably this problem is already solved for existing needs

pezholio: I think that's a bit of a noodle scratcher as far as UX goes - if users embed content, what would they expect to appear? My instinct would be the latest live version, but I think we'd have to test that assumption. 

If we're looking at replicating the behaviour that Contact embedding gives us in Whitehall, then there's currently _no_ concept of "draft" editions, as everything is sent directly to live, so, for now, this replicates what we already have

app/services/content_embed_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      replace_embed_code_with_content(embed_code, edition)
+    end
+
+    @body
+  end
+
+private
+
+  def find_edition(document_type, content_id)
+    Edition.with_document.find_by(
```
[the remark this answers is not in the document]
pezholio: I don't _think_ so, I don't think we'd really want to encourage any non-editionable content in Publishing API at all ideally. Certainly, the content block work we're looking at will all be editionable.

app/services/content_embed_service.rb
```diff
@@ -5,21 +5,42 @@ class ContentEmbedService
...
+      )
+    end
+  end
 
-      edition = find_edition(document_type, content_id)
-      # TODO: Link Edition to document
-      replace_embed_code_with_content(embed_code, edition)
+  def render
+    embedded_editions.each do |embedded_edition|
+      replace_embed_code_with_content(embedded_edition.embed_code, embedded_edition.edition)
```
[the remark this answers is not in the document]
pezholio: That's a good shout, this really is an initial spike, so it's all up for grabs really! 👍

app/services/content_embed_service.rb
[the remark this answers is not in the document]
yndajas: I think I'm probably conflating editionable-ness in Whitehall with editionable-ness in Publishing API here in any case

app/services/content_embed_service.rb
richardTowers: Yeah, there are some complexities here for sure. I think something like:

- If a document is in draft, it should be able to refer to other documents still in draft, and it should appear together with the latest draft editions when viewed through the draft frontend / content store. I believe this currently works for edition links, but I haven't got an example to hand.
- There should be some way for the publishing UI to hint that embedded things which are in draft either don't have published versions yet, or are newer than the live versions (so the value once published may be different, unless the draft embed is published)
- If you try to publish an edition that embeds documents which are still in draft, you should get an error, telling you that you need to publish the embedded documents before you can publish the embedding document.
- It should not be possible to unpublish / withdraw documents which are embedded in other documents (so if you really wanted to do that, you'd have to find all the usages and inline the values first)

app/commands/v2/put_content.rb
```diff
@@ -98,6 +99,12 @@ def create_links(edition)
         end
       end
 
+      def fetch_embedded_content(edition)
+        return {} if edition[:details]["body"].nil?
```
richardTowers, line 103: We should think about whether we want to support things other than `body` - in particular `parts` in multi part content will probably need support for embeds.

spec/integration/put_content/content_with_embedded_content_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe "PUT /v2/content when embedded content is provided" do
+  include_context "PutContent call"
+
+  context "with embedded content" do
+    let(:contact) { create(:edition, state: "published", content_store: "live", document_type: "contact") }
+    let(:document) { create(:document) }
+    let!(:link) { document.content_id }
+
+    before do
+      payload.merge!(document_type: "press_release", schema_name: "news_article", details: { body: "{{embed:contact:#{contact.document.content_id}}}" })
```
richardTowers, line 10: We should probably have another test for when body is an array like `[{content_type: "text/html"}]`

spec/integration/put_content/content_with_embedded_content_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe "PUT /v2/content when embedded content is provided" do
+  include_context "PutContent call"
+
+  context "with embedded content" do
+    let(:contact) { create(:edition, state: "published", content_store: "live", document_type: "contact") }
+    let(:document) { create(:document) }
+    let!(:link) { document.content_id }
```
richardTowers, line 7: We should have a test where we already have some existing links in the payload to check they're being merged correctly.

app/commands/v2/put_content.rb
```diff
@@ -87,9 +87,10 @@ def check_update_type
       end
 
       def create_links(edition)
-        return if payload[:links].nil?
+        links = (payload[:links] || {}).merge(fetch_embedded_content(edition))
```
richardTowers, line 90: We need to think about when embedded editions would be removed - Bruce says it's a bit weird in that you need to provide a key with an empty value to clear out an existing link (rather than just not providing it).

PR #2806, opened 2024-07-18
Add Content Block Tax Rate content type

Harriethw:
The Content Modelling Team are working on a spike to add the ability to publishing reusable pieces of content (“Content Blocks”) via Whitehall. This adds a simple initial schema to help us work with a type of tax rate.

Added to Whitehall here: https://github.com/alphagov/whitehall/pull/9278 

Some things I found out:
* I wanted to make the `rate` field an integer and a percentage - so max 100 value. However, our current form submites everything as strings, e.g. 100 becomes "100" so we would need to do more work to make this possible.
* I was also not able to give fields a min/max length because the tests generated didn't seem to take notice of these paramteres and were generating data above the min length - would need to look into this.

### DIFF
--- a/content_schemas/formats/content_block_tax_rate.jsonnet
+++ b/content_schemas/formats/content_block_tax_rate.jsonnet
@@ -1,0 +1,24 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: "content_block_tax_rate",
+  base_path: "forbidden",
+  routes: "forbidden",
+  rendering_app: "forbidden",
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      required: ["band", "income", "rate"],
+      properties: {
+        band: {
+          type: "string",
+        },
+        income: {
+          type: "string",
+        },
+        rate: {
+          type: "string",
+        }
+      },
+    },
+  },
+}
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -23,6 +23,7 @@
 - closed_consultation
 - cma_case
 - content_block_email_address
+- content_block_tax_rate
 - complaints_procedure
 - completed_transaction
 - consultation
--- a/content_schemas/dist/formats/content_block_tax_rate/notification/schema.json
+++ b/content_schemas/dist/formats/content_block_tax_rate/notification/schema.json
@@ -32,7 +32,7 @@
       "$ref": "#/definitions/analytics_identifier"
     },
     "base_path": {
-      "$ref": "#/definitions/absolute_path"
+      "type": "null"
     },
     "content_id": {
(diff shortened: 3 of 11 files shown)


content_schemas/formats/content_block_tax_rate.jsonnet
```diff
@@ -0,0 +1,24 @@
...
+      properties: {
+        band: {
+          type: "string",
+        },
+        income: {
+          type: "string",
+        },
+        rate: {
+          type: "string",
+        }
```
pezholio, line 20: This is an interesting one, because I think we're mixing up metadata and actual data here. To me, the `band` and the `income` would be metadata here, as they wouldn't actually be returned to the user. My mental model was more like we'd have two content types, `tax_income` and `tax_rate`, with individual body fields.

To use the example of a table of tax rates, I'd expect it to be used like this:

Band 	| Taxable income 	| Tax rate |
---------|----|----|
Personal Allowance | {{embed:tax_income:personal-allowance}} | {{embed:tax_rate:personal-allowance}}
Basic rate | {{embed:tax_income:basic-rate}} | {{embed:tax_rate:basic-rate}}
Higher rate | {{embed:tax_income:higher-rate}} | {{embed:tax_rate:higher-rate}}
Additional rate | {{embed:tax_income:additional-rate}} | {{embed:tax_rate:additional-rate}}

That said (and this is a thing that's just occurred to me), if we went with this approach we _could_ potentially have a more three-dimensional view and do something like this:


Band 	| Taxable income 	| Tax rate |
---------|----|----|
Personal Allowance | {{embed:tax_rate:personal-allowance:income}} | {{embed:tax_rate:personal-allowance:rate}}
Basic rate | {{embed:tax_rate:basic-rate:income}} | {{{embed:tax_rate:basic-rate:rate}}
Higher rate | {{embed:tax_rate:higher-rate:income}} | {{embed:tax_rate:higher-rate:rate}}
Additional rate | {{embed:tax_rate:additional-rate:income}} | {{embed:tax_rate:additional-rate:rate}}

Which could be quite cool. I think we'd need to run this via Rik first, as well as think how this would work within the rest of the user journey. Potentially super powerful though, and would also scale to things like contacts etc too

Harriethw: hmm, I just thought that that entire row on this page for example could be a bit of re-usable content in other Editions https://www.gov.uk/income-tax-rates 

Is there another block type that's less controversial we could use that would give us a schema with multiple fields? I think it would be really useful to have a 'straw man' to help us see how the different types work.

pezholio: How about telephone number?

Harriethw:
Closing for https://github.com/alphagov/publishing-api/pull/2807

PR #2807, opened 2024-07-18
Add Content Block Postal Address type

Harriethw:
The Content Modelling Team are working on a spike to add the ability to publishing reusable pieces of content (“Content Blocks”) via Whitehall. This adds a simple initial schema to help us work with a type of postal address.

There is an existing `post_addresses` object in the `contact` schema which is shaped slightly differently - I thought it better to go with the shape that GDS currently suggestions in their docs https://design-system.service.gov.uk/patterns/addresses/ to make it easier for our creation forms to comply with GDS standards.

We could add/change these fields in the future.

Things to note:
* The schema generation seems to order fields alphabetically, so the actual schema served to our frontend puts `County` as the first field 🤔

### DIFF
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -23,6 +23,7 @@
 - closed_consultation
 - cma_case
 - content_block_email_address
+- content_block_postal_address
 - complaints_procedure
 - completed_transaction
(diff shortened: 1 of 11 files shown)


pezholio (approved):
LGTM 👍

PR #2813, opened 2024-07-26
Allow embedding of content within document body

callumknights:
This allows content items to be embedded within other content, using the syntax `{{embed:CONTENT_TYPE:UUID}}` in the `body` field.

There are a number of assumptions here:

- The content being embedded must exist in a live state in the same language as the content it is being embedded within.
- Only the title will currently be embedded.  Future work will extend to include other fields.
- Only the `body` fields can contain embedded content.  Future work will extend to include other fields.
- Only documents of type `contact` can be embedded.  This can be extended by updating the `SUPPORTED_DOCUMENT_TYPES` constant.

This is based on the spike by @pezholio in https://github.com/alphagov/publishing-api/pull/2794.

### DIFF
--- a/app/presenters/content_embed_presenter.rb
+++ b/app/presenters/content_embed_presenter.rb
@@ -1,0 +1,54 @@
+module Presenters
+  class ContentEmbedPresenter
+    def initialize(edition)
+      @edition = edition
+    end
+
+    def render_embedded_content(details)
+      return details if details[:body].nil?
+
+      details[:body] = if details[:body].is_a?(Array)
+                         details[:body].map do |content|
+                           {
+                             content_type: content[:content_type],
+                             content: render_embedded_editions(content[:content]),
+                           }
+                         end
+                       else
+                         render_embedded_editions(details[:body])
+                       end
+
+      details
+    end
+
+  private
+
+    def render_embedded_editions(content)
+      embedded_content_references = EmbeddedContentFinderService.new.find_content_references(content)
+      embedded_content_references_by_content_id = embedded_content_references.index_by(&:content_id)
+
+      target_content_ids = @edition
+        .links
+        .where(link_type: "embed")
+        .pluck(:target_content_id)
+
+      embedded_edition_ids = ::Queries::GetEditionIdsWithFallbacks.call(
+        target_content_ids,
+        locale_fallback_order: [@edition.locale, Edition::DEFAULT_LOCALE].uniq,
+        state_fallback_order: %w[published],
+      )
+
+      embedded_editions = Edition.where(id: embedded_edition_ids)
+
+      embedded_editions.each do |embedded_edition|
+        embed_code = embedded_content_references_by_content_id[embedded_edition.content_id].embed_code
+        content = content.gsub(
+          embed_code,
+          embedded_edition.title,
+        )
+      end
+
+      content
+    end
+  end
+end
--- a/app/presenters/details_presenter.rb
+++ b/app/presenters/details_presenter.rb
@@ -2,17 +2,19 @@ require "govspeak"
 
 module Presenters
   class DetailsPresenter
-    attr_reader :content_item_details, :change_history_presenter
+    attr_reader :content_item_details, :change_history_presenter, :content_embed_presenter
 
-    def initialize(content_item_details, change_history_presenter)
+    def initialize(content_item_details, change_history_presenter, content_embed_presenter)
       @content_item_details = SymbolizeJSON.symbolize(content_item_details)
       @change_history_presenter = change_history_presenter
+      @content_embed_presenter = content_embed_presenter
     end
 
     def details
       @details ||=
         begin
-          updated = recursively_transform_govspeak(content_item_details)
+          updated = content_embed(content_item_details).presence || content_item_details
+          updated = recursively_transform_govspeak(updated)
           updated[:change_history] = change_history if change_history.present?
           updated
         end
@@ -44,6 +46,10 @@ module Presenters
       else
         obj.map { |o| recursively_transform_govspeak(o) }
       end
+    end
+
+    def content_embed(content_item_details)
+      @content_embed ||= content_embed_presenter&.render_embedded_content(content_item_details)
     end
 
     def change_history
--- a/app/presenters/queries/expanded_link_set.rb
+++ b/app/presenters/queries/expanded_link_set.rb
@@ -61,7 +61,7 @@ module Presenters
           end
 
           if hash[:details]
-            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil, nil).details
+            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil, content_embed_presenter(hash[:content_id], hash[:locale])).details
           end
         end
       end
@@ -76,6 +76,11 @@ module Presenters
 
       def translations
         available_translations.translations
+      end
+
+      def content_embed_presenter(content_id, locale)
+        edition = Document.find_by(content_id:, locale:).live
+        Presenters::ContentEmbedPresenter.new(edition) if edition
       end
     end
   end
--- a/app/presenters/queries/expanded_link_set.rb
+++ b/app/presenters/queries/expanded_link_set.rb
@@ -61,7 +61,7 @@ module Presenters
           end
 
           if hash[:details]
-            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil).details
+            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil, nil).details
           end
         end
       end
--- a/app/services/embedded_content_finder_service.rb
+++ b/app/services/embedded_content_finder_service.rb
@@ -1,0 +1,45 @@
+class EmbeddedContentFinderService
+  ContentReference = Data.define(:document_type, :content_id, :embed_code)
+
+  SUPPORTED_DOCUMENT_TYPES = %w[contact].freeze
+  UUID_REGEX = /([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/
+  EMBED_REGEX = /({{embed:(#{SUPPORTED_DOCUMENT_TYPES.join('|')}):#{UUID_REGEX}}})/
+
+  def fetch_linked_content_ids(body, locale)
+    content_references = if body.is_a?(Array)
+                           body.map { |hash| find_content_references(hash[:content]) }.flatten
+                         else
+                           find_content_references(body)
+                         end
+    return [] if content_references.empty?
+
+    check_all_references_exist(content_references, locale)
+    content_references.map(&:content_id)
+  end
+
+  def find_content_references(body)
+    body.scan(EMBED_REGEX).map { |match| ContentReference.new(document_type: match[1], content_id: match[2], embed_code: match[0]) }.uniq
+  end
+
+private
+
+  def check_all_references_exist(content_references, locale)
+    found_editions = live_editions(content_references, locale)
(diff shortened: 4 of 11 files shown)


app/presenters/content_embed_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def render_embedded_editions(content)
+      embedded_content_references = EmbeddedContentFinderService.new.find_content_references(content)
+      embedded_content_references_by_content_id = embedded_content_references.index_by(&:content_id)
+      embedded_editions = @edition
+        .links
+        .where(link_type: "embed")
+        .map(&:target_documents)
+        .map { |target_documents| target_documents.find_by(locale: @edition.locale) || target_documents.find_by(locale: Edition::DEFAULT_LOCALE) }
+        .map(&:live)
```
richardTowers, line 34: There's a bit of an `n+1` here I think. I'd expect this to make:

- 1 query to get `links.where(link_type: "embded")`
- n queries to get target documents for each link
- n queries to get the live edition for each target document

This is probably fine for now, as n probably won't be large (as I don't expect documents to have that more than a few embeds).

I think you could do it in one query that gets all the possible locales, and then find the correct locales in ruby.

Something like:

```ruby
edition_ids = @edition
  .links
  .where(link_type: "embed")
  .joins(target_documents: :live)
  .where(target_documents: { locale: locales })
  .select("editions.id")
# Note ^ the use of select rather than pluck here means this is still
# an `ActiveRecord_AssociationRelation`, rather than an array of ids,
# so it won't make a separate SQL query
  
embedded_editions = Edition.where(id: edition_ids)
# Generates:
# SELECT "editions".* FROM "editions"
# WHERE "editions"."id" IN (
#   SELECT "editions"."id" FROM "links"
#   INNER JOIN "documents" "target_documents"
#     ON "target_documents"."content_id" = "links"."target_content_id"
#   INNER JOIN "editions"
#     ON "editions"."content_store" = 'live'
#     AND "editions"."document_id" = "target_documents"."id"
#   WHERE "links"."edition_id" = 1
#   AND "links"."link_type" = 'embed'
#   AND "target_documents"."locale" IN ('cy', 'en')
# )

# Exercise for the reader: we'll have multiple editions where there are multiple locales. Get the best edition for the locales we want.
```

richardTowers: Alternatively, you could use the same trick as we do here: https://github.com/alphagov/publishing-api/blob/c3bfff2652016ec4acbd2883dd22786ab7b8ea7d/app/queries/get_edition_ids_with_fallbacks.rb#L14 which is complicated but presumably works...

Could even use `Queries::GetEditionIdsWithFallbacks` directly I guess, but you might end up making one or two unnecessary queries there.

app/presenters/queries/expanded_link_set.rb
```diff
@@ -61,7 +61,7 @@ def present_expanded_link(link_hash)
           end
 
           if hash[:details]
-            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil).details
+            hash[:details] = Presenters::DetailsPresenter.new(hash[:details], nil, nil).details
```
richardTowers, line 64: I guess this means that we won't expand embed references inside linked `body` elements, so we might end up with things like:

```ruby
{
  content_type: "person",
  # ...
  links: {
    role: [{
      content_type: "role",
      # ...
      details: {
        body: "Some text where we haven't expanded an embed: {{embed:contact:00000000-0000-0000-0000-000000000000}}"
        # ...
      }
    }]
  }
}
```

I think that's fine for now, maybe we'll revisit it later.

app/presenters/details_presenter.rb
```diff
@@ -2,17 +2,19 @@
...
       @content_item_details = SymbolizeJSON.symbolize(content_item_details)
       @change_history_presenter = change_history_presenter
+      @content_embed_presenter = content_embed_presenter
     end
 
     def details
       @details ||=
         begin
-          updated = recursively_transform_govspeak(content_item_details)
+          updated = content_embed(content_item_details).presence || content_item_details
```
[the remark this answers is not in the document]
brucebolt: I think we chose to keep the name `updated` as it reflects the the details have been updated with the embedded content (as they were previously being updated with the transformed govspeak).

app/services/embedded_content_finder_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    if found_editions.count != content_references.count
+      not_found_content_ids = content_references.map(&:content_id) - found_editions.map(&:content_id)
+      raise CommandError.new(
+        code: 422,
+        message: "Could not find any live editions in locale #{locale} for: #{not_found_content_ids.join(', ')}, ",
+      )
+    end
+  end
+
+  def editions(content_references, locale)
```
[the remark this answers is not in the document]
brucebolt: I've renamed this method to `live_editions`.

spec/integration/put_content/content_with_embedded_content_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    before do
+      payload.merge!(document_type: "press_release", schema_name: "news_article", details: { body: "{{embed:contact:#{first_contact.document.content_id}}} {{embed:contact:#{second_contact.document.content_id}}}" })
+    end
+
+    it "should create links" do
+      expect {
+        put "/v2/content/#{content_id}", params: payload.to_json
+      }.to change(Link, :count).by(2)
+
+      expect(Link.find_by(target_content_id: first_contact.content_id)).to be
```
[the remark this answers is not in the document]
brucebolt: I think the intention here (and elsewhere in this file) is for `to be` to mean `not_to be_nil`.  I've updated this so it's clearer.

spec/presenters/content_embed_presenter_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      details: details.deep_stringify_keys,
+      links_hash: {
+        embed: [embedded_content_id],
+      },
+    )
+  end
+  let(:details) { {} }
+
+  before do
+    document = create(:document, content_id: embedded_content_id)
```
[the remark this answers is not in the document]
brucebolt: Oops, look like the name `document` was used twice to mean different things.  I've updated this to give the name `embedded_document` for the thing that is being embedded into the other document.

spec/presenters/details_presenter_spec.rb
```diff
@@ -179,5 +183,62 @@
...
+        end
+      end
+
+      context "when we're passed an empty array" do
+        let(:edition) do
+          create(:edition, details: { body: [] })
+        end
+        let(:edition_details) { edition.details }
+
+        it "embeds the contact details" do
```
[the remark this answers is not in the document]
brucebolt: I've updated the test name, as I think this was just incorrectly copied from a previous example.

app/presenters/content_embed_presenter.rb
brucebolt: I've gone for the approach of using `Queries::GetEditionIdsWithFallbacks` directly, as that seems to be the least complicated and DRY-est way of getting the embedded editions.

I think that approach should also make it easier for us to deal with embedding draft content (should we wish to) in the future.

app/presenters/content_embed_presenter.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+    def render_embedded_editions(content)
+      embedded_content_references = EmbeddedContentFinderService.new.find_content_references(content)
+      embedded_content_references_by_content_id = embedded_content_references.index_by(&:content_id)
+
+      target_content_ids = @edition
+        .links
+        .where(link_type: "embed")
+        .map(&:target_content_id)
```
richardTowers, line 33: nit: I think you could do 
```suggestion
        .pluck(:target_content_id)
```

which would avoid allocating Link objects for each link and then calling `#target_content_id` on each one (instead just returning the content ids directly from the SQL query.

app/presenters/content_embed_presenter.rb
```diff
@@ -0,0 +1,54 @@
...
+        state_fallback_order: %w[published],
+      )
+
+      embedded_editions = Edition.where(id: embedded_edition_ids)
+
+      embedded_editions.each do |embedded_edition|
+        embed_code = embedded_content_references_by_content_id[embedded_edition.content_id].embed_code
+        content = content.gsub(
+          embed_code,
+          embedded_edition.title,
```
richardTowers, line 47: Ah, I think this is actually not the right field on contact to be replacing things with. `title` will be something like "Government Digital Service", rather than "10 Whitechapel High Street" like we probably want.

The way Whitehall does it is:

- [Render the contacts/contact partial](https://github.com/alphagov/whitehall/blob/81beec81ef99e1c8d230f028a8b725e406770371/app/helpers/govspeak_helper.rb#L160-L161)
- Which [is actually proper complicated](https://github.com/alphagov/whitehall/blob/main/app/views/contacts/_contact.html.erb)
- Complicated further by the fact that [the code to render addresses](https://github.com/alphagov/whitehall/blob/main/app/helpers/admin/contacts_helper.rb#L2) [is itself very complicated](https://github.com/alphagov/whitehall/blob/main/lib/address_formatter/formatter.rb#L3)

🤔  we could merge this with it just using title, but it's probably not actually what we want. On the other hand, I don't think publishing-api having a bunch of knowledge of how to render addresses is desirable - I'd be somewhat resistant to copying the templating code over from Whitehall.

I think the simplest option I can think of would be to add an "embeddable_html" field to Contact when Whitehall presents Contacts to publishing-api, and then use that. That way Whitehall can continue doing the rendering of Contact HTML (even though that's not very nice either), and publishing-api doesn't need to get rendering app capabilities.

Longer term with the design of this feature, I wonder if this is a case for passing the `{{embed:...}}` syntax all the way through to the frontend, and having the rendering app do the rendering of contact data into HTML (having followed the expanded link).

I'm sort of leaning towards getting this PR merged and then fixing this in a subsequent PR though.

richardTowers (approved):
A couple of nitpicks, and I'm afraid one actual issue - `edition.title` is not the value we want for the replacements.

I'm inclined to merge this even with this issue, as it won't cause any harm since the feature won't be used. We'll need to have a bit of a discussion about the best way forward on how to render the HTML for contact embeds I think.

app/presenters/content_embed_presenter.rb
```diff
@@ -0,0 +1,54 @@
...
+      embedded_content_references_by_content_id = embedded_content_references.index_by(&:content_id)
+
+      target_content_ids = @edition
+        .links
+        .where(link_type: "embed")
+        .map(&:target_content_id)
+
+      embedded_edition_ids = ::Queries::GetEditionIdsWithFallbacks.call(
+        target_content_ids,
+        locale_fallback_order: [@edition.locale, Edition::DEFAULT_LOCALE],
```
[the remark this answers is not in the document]
brucebolt: Good spot, I'll use `uniq` here.

app/presenters/content_embed_presenter.rb
brucebolt: Nice, will update this.

app/presenters/content_embed_presenter.rb
brucebolt: We'll leave this as `title` for now (on the basis that no-one will actually be embedding anything into live content) but will write up a card to implement `embedded_html` in the schemas and include that in the document instead.

>Longer term with the design of this feature, I wonder if this is a case for passing the {{embed:...}} syntax all the way through to the frontend, and having the rendering app do the rendering of contact data into HTML (having followed the expanded link).

This makes a lot of sense to me.  Particularly as the frontend app is better placed to do any conversion to HTML (rather than the current situation of the publishing app or Publishing API doing this, then any frontend changes needing all content to be republished).

PR #2818, opened 2024-07-30
Sort shared schema definitions into alphabetical order

brucebolt:
There is no reason for these to be in any specific order, so sorting to be alphabetical. This makes it easier to work with.


yndajas (approved):
Love a good alphabetical sort!

PR #2819, opened 2024-07-30
Add `embed` to all content schemas

brucebolt:
This will allow us to store embedded content as a link on documents.


richardTowers (approved):
Looks good to me.

I guess we should probably add some examples which include embed links / embed references at some point, but that should probably be a separate PR.

PR #2823, opened 2024-08-01
Improve link handling in content endpoint

pezholio:
The Content Modelling team have a need to list what documents have a dependency on a given content block, including the Title, Content Type and associated Organisation (this could be via `primary_publishing_organisation` or another link, such as `sponsoring_organisation` for Worldwide Organisations for example.

We initially tried creating a new endpoint (see #2822), but we've since discovered the [`GET /v2/content`](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#get-v2content) endpoint does 90% of what we need to do. However, we need to make some changes to:

- Allow the `link_*` query option to support Edition links, as well as Linkset links, this is because embedded content will be stored as an Edition link
- Return Linkset links as well as Edition links in the responses - currently only Edition links are returned, which means we can't always identify the associated Organisation for a document

I did consider expanding the links at this point, but this could be a bit over the top given that Organisations are already handled in Whitehall, so as long as we have a Content ID, we should be able to fetch the associated Organisation easily.

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -21,10 +21,13 @@ class Link < ApplicationRecord
       logger.warn("filter_editions called with multiple filters. These will be ANDed together in a way that probably isn't what we want. Filters were: #{filters.inspect}")
     end
 
-    scope = scope.joins(document: :link_set_links)
+    scope = scope.joins("LEFT JOIN links edition_links ON edition_links.edition_id = editions.id")
+    scope = scope.left_joins(document: :link_set_links)
 
     filters.each do |link_type, target_content_id|
-      scope = scope.where(links: { link_type:, target_content_id: })
+      scope = scope
+                .where(edition_links: { link_type:, target_content_id: })
+                .or(scope.where(links: { link_type:, target_content_id: }))
     end
 
     scope
--- a/app/presenters/queries/content_item_presenter.rb
+++ b/app/presenters/queries/content_item_presenter.rb
@@ -148,6 +148,8 @@ module Presenters
           nil
         when :links
           nil
+        when :link_set_links
+          nil
         when :total
           nil
         else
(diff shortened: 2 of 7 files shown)


Harriethw:
It doesn't look like there is documentation in [the Pact docs](https://govuk-pact-broker-6991351eca05.herokuapp.com/pacts/provider/Publishing%20API/consumer/GDS%20API%20Adapters/latest#a_get_entries_request_given_a_content_item_exists_in_multiple_locales_with_content_id:_bed722e6-db68-43e5-9079-063f623335a7) for querying `v2/content` with the `link_*` param - would it be worth adding to them? I think that would help me at least understand what change is being made.

Harriethw:
Looks good from my perspective :) But would want some eyes from elsewhere in Publishing who will know the service better than me

app/models/link.rb
```diff
@@ -21,10 +21,13 @@ def self.filter_editions(scope, filters)
       logger.warn("filter_editions called with multiple filters. These will be ANDed together in a way that probably isn't what we want. Filters were: #{filters.inspect}")
     end
 
-    scope = scope.joins(document: :link_set_links)
+    scope = scope.joins("LEFT JOIN links edition_links ON edition_links.edition_id = editions.id")
```
Gweaton, line 24: Might be worth adding the context to the commit msg as to why this change is happening, i.e. because embedded content will be stored as an Edition link

Gweaton (commented):
The approach here makes sense, but I'm still not 100% sure what the impact of these changes might be on API consumers e.g. if it could slow things down due to the larger volume of data coming back from these JOINs. Will see if there's anyone in my team who might be able to help.

brucebolt (commented):
The code changes all look fine to me, but I'd echo @Gweaton's comment about any impact this will have on API consumers.

Particularly I'm wondering if you confident this won't start returning additional links (i.e. the edition links) that may or may not be needed in other consumers of this endpoint?

pezholio:
@brucebolt Cool - thanks for looking, and good call. Do you think it's sufficient to maybe add `primary_publishing_organisation` (which is the link we're interested in) as a supported argument for `fields`, so we don't have to unnecessarily load all the links?

brucebolt:
> @brucebolt Cool - thanks for looking, and good call. Do you think it's sufficient to maybe add `primary_publishing_organisation` (which is the link we're interested in) as a supported argument for `fields`, so we don't have to unnecessarily load all the links?

That's a possibility.  Or you could return the edition links as a separate value (to maintain backward compatibility for any other API consumers) and then merge them together in your consuming app.

pezholio:
@brucebolt @Gweaton This is ready for another look now

brucebolt (approved):

PR #2829, opened 2024-08-02
add `primary_publishing_organisation` to Content Block Email

Harriethw:
We would like to associate a primary organisation with a Content Block Email Address, on the Edition level.
This adds the field to the schema to make this possible.

Once this is added we can send the `primary_publishing_organisation` as a `link` when we publish a Content Block edition - the WIP PR is here https://github.com/alphagov/whitehall/pull/9338 

Questions:

* Because this has been added under `edition_links`, when this is sent to the Publishing API a `Link` between the `Edition`'s content_id and the `Organisation`'s content_id. This seems to follow other models like the [News Article.](https://github.com/alphagov/publishing-api/blob/9c2e19da5a091a6416aba2e3b8f8f39c1dfa03c5/content_schemas/formats/news_article.jsonnet) But there is also an `organisation` type of `Links` that are connected with `Documents` - Should we stick with Edition-level relationship at the moment, assuming that for Content Blocks that makes sense for now? E.g. one version of an Edition could have a different Organisation to another version of the Edition

### DIFF
--- a/content_schemas/dist/formats/content_block_email_address/notification/schema.json
+++ b/content_schemas/dist/formats/content_block_email_address/notification/schema.json
@@ -85,10 +85,6 @@
           "description": "Link type automatically added by Publishing API",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
-        "embed": {
-          "description": "Content that will be embedded within the document, using embed tags.",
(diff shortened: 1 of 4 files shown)


pezholio (approved):
LGTM - I agree with having Edition links for Content Blocks, as it can change over time 👍

PR #2833, opened 2024-08-07
Fix issue with embedded content on body being an array

brucebolt:
We used a symbol as the hash key for the body, when it is actually a string.

No test cases covered this scenario, so adding a relevant test too.

### DIFF
--- a/app/services/embedded_content_finder_service.rb
+++ b/app/services/embedded_content_finder_service.rb
@@ -7,7 +7,7 @@ class EmbeddedContentFinderService
 
   def fetch_linked_content_ids(body, locale)
     content_references = if body.is_a?(Array)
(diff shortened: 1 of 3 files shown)


callumknights (approved):

richardTowers:
Are we going to have the same issue [here](https://github.com/alphagov/publishing-api/blob/main/app/presenters/content_embed_presenter.rb#L8-L19), except it will just do nothing instead of blowing up?

brucebolt:
The values here are using symbol keys (because of [this line](https://github.com/alphagov/publishing-api/blob/9bb0a4cdeaa93c4892b161e104497e9abdcfac0a/app/presenters/details_presenter.rb#L8)) so this is fine to remain a symbol.

PR #2838, opened 2024-08-09
Add a shared `content_block` format

pezholio:
This allows us to share definitions across all content blocks, and define new ones more easily. There are minimal changes to the generated schemas, apart from adding some links to the Postal Address Schema, and adding the default links back to the Email Address schema (which we deleted accidentally).


Harriethw (approved):
another good idea 🫡

PR #2846, opened 2024-08-19
Content modelling/417 Get embedded content by target_content_id

tahb:
# What does this do?

Given a `target_content_id` which is the edition of the content block/embeddable content, it will find and return any other live editions [where it has been embedded and a link set up](https://github.com/alphagov/publishing-api/pull/2813).

For The content Object store within Whitehall we only need a few fields compared to the existing `content/:id` endpoint: `title`, `base_path`, `document_type`, `target_content_id` and a couple of fields about the associated organisation.

With this data we’ll present a table that allows reusable content publishers to see where their content is currently being included.

# Why a new endpoint?

We have previously tried to use [the `/v2/content/:content_id` endpoint](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#put-v2contentcontent_id). Though it includes a lot more `Edition` data than we need it also returned the publishing organisation link and gave us pagination and sorting for free. **This worked locally but on[ Integration this timed out](https://gds.slack.com/archives/C0776B04EJU/p1723559847781489) and our implementation (which is flagged off in production) is currently broken.**

We tracked down the performance problem to [a subquery used](https://github.com/alphagov/publishing-api/blob/3f4cc8523b1713aefc1e8f70950e1a81cfa56f22/app/presenters/queries/content_item_presenter.rb#L190) when the `link_set_links` `field` is passed through. [It appears we added this a few weeks ago in an effort to extend the endpoint for our purposes](https://github.com/alphagov/publishing-api/pull/2823). As a result of the data that follows I think this change should be reverted and [I'll propose it as a separate change](https://github.com/alphagov/publishing-api/pull/2847).

We did some local [benchmarking](https://ruby-doc.org/stdlib-2.7.1/libdoc/benchmark/rdoc/Benchmark.html) and found the query was slowing down notably after a few hundred thousand records:

```
2024-08-15 17:24:20 Document count: 178924
2024-08-15 17:24:20 Edition count: 178925
2024-08-15 17:24:20 Link count: 178855
                                      user            system      total          real
2024-08-15 17:24:20   0.003363   0.012408   0.015771 (  0.334085)
1.92s in the browser

2024-08-15 17:29:01 Document count: 228173
2024-08-15 17:29:01 Edition count: 228174
2024-08-15 17:29:01 Link count: 228106
                                      user            system      total          real
2024-08-15 17:29:01   0.019349   0.013283   0.032632 (  0.499379)
3.8s in the browser
```

We stopped seeding at this point as we could see that seeding more would eventually result in a timeout.

On integration the comparable data counts at the time were:

```
39_953_449 links
14_170_691 editions
1_323_401 documents 
931_902 link sets
```

It seems reasonable that there might be more than hundreds of thousands of
documents and millions of links in production. We can't find an obvious
fault in the existing subquery. Instead of risking a change to it we
create a new endpoint which does just the thing we need.

On the same data set we see a marked improvement:
```
2024-08-16 11:19:27 Document: 230366
2024-08-16 11:19:27 Edition: 230365
2024-08-16 11:19:27 Link: 230295
                                      user            system      total          real
2024-08-16 11:19:27   0.000593   0.000170   0.000763 (  0.001478)
```

A negative side effect of this is that we won't get pagination and
sorting for free.

## Naming

We continue to use the relatively new term of `embedded`. [It was the language used on the original linking work](https://github.com/alphagov/publishing-api/pull/2813) and also is what we use as our embed references: `{{embed…}}`.

We sit alongside many other endpoints and services which are named "link", "linkables", "link sets" and "link set links". Though we of course use `links` under the hood to find embedded content we try to abstract that in order to more clearly distinguish this.

### DIFF
--- a/app/presenters/queries/embedded_content_presenter.rb
+++ b/app/presenters/queries/embedded_content_presenter.rb
@@ -1,0 +1,43 @@
+module Presenters
+  module Queries
+    class EmbeddedContentPresenter
+      def self.present(target_edition_id, host_editions)
+        new(target_edition_id, host_editions).present
+      end
+
+      def initialize(target_edition_id, host_editions)
+        self.target_edition_id = target_edition_id
+        self.host_editions = host_editions.to_a
+      end
+
+      def present
+        {
+          content_id: target_edition_id,
+          total: host_editions.count,
+          results:,
+        }
+      end
+
+      def results
+        return [] unless host_editions.any?
+
+        host_editions.map do |edition|
+          {
+            title: edition.title,
+            base_path: edition.base_path,
+            document_type: edition.document_type,
+            primary_publishing_organisation: {
+              content_id: edition.primary_publishing_organisation_content_id,
+              title: edition.primary_publishing_organisation_title,
+              base_path: edition.primary_publishing_organisation_base_path,
+            },
+          }
+        end
+      end
+
+    private
+
+      attr_accessor :target_edition_id, :host_editions
+    end
+  end
+end
--- a/app/queries/get_embedded_content.rb
+++ b/app/queries/get_embedded_content.rb
@@ -1,0 +1,39 @@
+module Queries
+  class GetEmbeddedContent
+    attr_reader :target_content_id
+
+    def initialize(target_content_id)
+      self.target_content_id = target_content_id
+    end
+
+    def call
+      if Document.find_by(content_id: target_content_id).nil?
+        message = "Could not find an edition to get embedded content for"
+        raise CommandError.new(code: 404, message:)
+      end
+
+      Presenters::Queries::EmbeddedContentPresenter.present(
+        target_content_id,
+        host_editions,
+      )
+    end
+
+  private
+
+    def host_editions
+      @host_editions ||= Edition.live
+        .joins(:links)
+        .joins("LEFT JOIN links AS primary_links ON primary_links.edition_id = editions.id AND primary_links.link_type = 'primary_publishing_organisation'")
+        .joins("LEFT JOIN documents ON documents.content_id = primary_links.target_content_id")
+        .joins("LEFT JOIN editions AS org_editions ON org_editions.document_id = documents.id")
+        .where(links: { link_type: embedded_link_type, target_content_id: })
+        .select(
+          "editions.id, editions.title, editions.base_path, editions.document_type, primary_links.target_content_id AS primary_publishing_organisation_content_id",
+          "org_editions.title AS primary_publishing_organisation_title",
+          "org_editions.base_path AS primary_publishing_organisation_base_path",
+        )
+    end
+
+    attr_writer :target_content_id
+  end
+end
--- a/app/queries/get_embedded_content.rb
+++ b/app/queries/get_embedded_content.rb
@@ -1,9 +1,10 @@
 module Queries
   class GetEmbeddedContent
-    attr_reader :target_content_id
+    attr_reader :target_content_id, :states
 
     def initialize(target_content_id)
       self.target_content_id = target_content_id
+      self.states = %i[draft published]
     end
 
     def call
@@ -21,7 +22,7 @@ module Queries
   private
 
     def host_editions
-      @host_editions ||= Edition.live
+      @host_editions ||= Edition.where(state: states)
         .joins(:links)
         .joins("LEFT JOIN links AS primary_links ON primary_links.edition_id = editions.id AND primary_links.link_type = 'primary_publishing_organisation'")
         .joins("LEFT JOIN documents ON documents.content_id = primary_links.target_content_id")
@@ -34,6 +35,10 @@ module Queries
         )
     end
 
-    attr_writer :target_content_id
+    def embedded_link_type
+      "embed"
+    end
+
+    attr_writer :target_content_id, :states
   end
 end
--- a/spec/queries/get_embedded_content_spec.rb
+++ b/spec/queries/get_embedded_content_spec.rb
@@ -1,5 +1,22 @@
 RSpec.describe Queries::GetEmbeddedContent do
   describe "#call" do
+    let(:organisation) do
+      create(:live_edition,
+             title: "bar",
+             document_type: "organisation",
+             schema_name: "organisation",
+             base_path: "/government/organisations/bar")
+    end
+
+    let(:content_block) do
+      create(:live_edition,
+             document_type: "content_block_email_address",
+             schema_name: "content_block_email_address",
+             details: {
+               "email_address" => "foo@example.com",
+             })
+    end
+
     it "returns data prepared by the presenter" do
       target_content_id = SecureRandom.uuid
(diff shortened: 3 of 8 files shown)


app/presenters/queries/embedded_content_presenter.rb
```diff
@@ -0,0 +1,43 @@
+module Presenters
+  module Queries
+    class EmbeddedContentPresenter
+      def self.present(target_edition_id, host_edition)
```
Gweaton, line 4: Should the second parameter be `host_editions` (plural) or are we only passing one edition here? In the test we're passing multiple - should this be renamed, or am I misunderstanding here?

app/queries/get_embedded_content.rb
```diff
@@ -0,0 +1,39 @@
...
+    def call
+      if Document.find_by(content_id: target_content_id).nil?
+        message = "Could not find an edition to get embedded content for"
+        raise CommandError.new(code: 404, message:)
+      end
+
+      Presenters::Queries::EmbeddedContentPresenter.new(
+        target_content_id,
+        host_editions,
+      ).present
```
Gweaton, line 18: Can/should we use the `.present` class method here, or is there a reason we need to use the instance method?

app/queries/get_embedded_content.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+  private
+
+    def host_editions
+      @host_editions ||= Edition.live
```
Gweaton, line 24: Would we ever want people to be able to embed content in draft docs e.g. for previewing what's being published? Or are we starting with `live` for now?

spec/queries/get_embedded_content_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    context "when the target_content_id doesn't match a Document" do
+      it "returns 404" do
+        expect { described_class.new(SecureRandom.uuid).call }.to raise_error(CommandError) do |error|
+          expect(error.code).to eq(404)
+          expect(error.message).to eq("Could not find an edition to get embedded content for")
+        end
+      end
+    end
+
+    context "when the target_content_id does not match a live edition" do
```
Gweaton, line 29: Is it more accurate to say something like `when the target_content is embedded in non-live editions`...`it does not include the non-live editions in the results`?

I was initially reading this as us searching for the `content_block` via the `target_content_id`, so this might make that a bit clearer.

Gweaton (commented):
This looks good to me - just a couple of small questions/comments. Well done for wrangling that complex query!

app/queries/get_embedded_content.rb
```diff
@@ -0,0 +1,39 @@
...
+        .joins(:links)
+        .joins("LEFT JOIN links AS primary_links ON primary_links.edition_id = editions.id AND primary_links.link_type = 'primary_publishing_organisation'")
+        .joins("LEFT JOIN documents ON documents.content_id = primary_links.target_content_id")
+        .joins("LEFT JOIN editions AS org_editions ON org_editions.document_id = documents.id")
+        .where(links: { target_content_id: }) # TODO: Inlcude `link_type: "embedded"`` when it is supported.
```
richardTowers, line 29: I believe it is supported, only it's called "embed", rather than "embedded" (added in 372952b1efb68b3d8f78b855d9aa3202a46fd488). Unless I misunderstand?

richardTowers (commented):
Just one drive by comment - I'll leave it to George to do the proper review. Generally looks good to me. Agree with the logic behind making this a separate endpoint.

richardTowers:
Ah, I think I know why [the subquery](https://github.com/alphagov/publishing-api/blob/3f4cc8523b1713aefc1e8f70950e1a81cfa56f22/app/presenters/queries/content_item_presenter.rb#L190) performed so terribly.

[This line](https://github.com/alphagov/publishing-api/blob/3f4cc8523b1713aefc1e8f70950e1a81cfa56f22/app/presenters/queries/content_item_presenter.rb#L195) is ambiguous about what it's referring to as `content_id` - should it be `link_sets.content_id = link_sets.content_id` or `link_sets.content_id = subquery.content_id`. To us it's obvious it should be the latter, but I suspect Postgres is doing the former.

You might find you can get good performance if you update it to `link_sets.content_id = subquery.content_id` to be explicit.

app/queries/get_embedded_content.rb
tahb: Embedding should work for draft documents so there is an argument that content block publishers might want to see ALL documents regardless of state.

[For our use case I've gone back to clarify what we need as our default case](https://gds.slack.com/archives/C062Z1FA2ES/p1724239210827159). 

In any case I can imagine that consumers might want embedded content in either state. I'll open this up to everything and then we can filter it down later 👍

app/presenters/queries/embedded_content_presenter.rb
tahb: Yes good spot! Thank you.

spec/queries/get_embedded_content_spec.rb
tahb: Yes this wasn't worded well was it! As i'm changing the behaviour to include draft and published I'll revamp the tests in a new commit which hopefully makes it clearer 👍

app/queries/get_embedded_content.rb
tahb: I think I followed [this Presenter as a template](https://github.com/alphagov/publishing-api/blob/6cee925ae331e7cc059b1432033df8d1d343a103/app/presenters/queries/link_set_presenter.rb#L12) which had the instance method and then I didn't have a test to drive me to add the `self.present(…)` that calls the initialiser internally. I see now that other presenters are doing that so I'll add that too to keep it consistent 👍

app/queries/get_embedded_content.rb
tahb: I've changed it to do draft and live, but not superseded or unpublished (though we don't have a factory for the latter).

app/queries/get_embedded_content.rb
tahb: Oh super, I didn't realise this had happened already. Thank you 🙌

tahb:
@Gweaton Thank you for the review! This is definitely in a better place.

I've addressed everything and notably we now:

- return draft _and_ published host content now
- only check for links with the `embed` type and updated/corrected tests

Ready for another review please! 🙏

app/queries/get_embedded_content.rb
```diff
@@ -0,0 +1,44 @@
+module Queries
+  class GetEmbeddedContent
+    attr_reader :target_content_id, :states
+
+    def initialize(target_content_id)
+      self.target_content_id = target_content_id
+      self.states = %i[draft published]
```
tahb, line 7: I haven't gone as far as making these fully configurable as I can't verify whether we actually need it! That said this is hopefully where it can be injected in from a query param in future 🤞

Gweaton (approved):
Thanks for making those changes - looks good to me!

PR #2847, opened 2024-08-20
Revert "Improve link handling in content endpoint"

tahb:
Reverts alphagov/publishing-api#2823

During the testing of this change on integration we found we got timeouts when trying to send the `field`: `link_set_links`.

[We have proposed an alternative way for us to find embedded content via a new endpoint altogether](https://github.com/alphagov/publishing-api/pull/2846). More information such as benchmarking can be found there.

As this doesn't work on integration it is unlikely that any API consumers other than us have started using this.

I asked in [govuk-publishing-platform](https://gds.slack.com/archives/C03D792LYJG/p1724151195088429) if there was a formal process for deprecating this and @callumknights suggested a straight revert should be fine.

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -21,13 +21,10 @@ class Link < ApplicationRecord
       logger.warn("filter_editions called with multiple filters. These will be ANDed together in a way that probably isn't what we want. Filters were: #{filters.inspect}")
     end
 
-    scope = scope.joins("LEFT JOIN links edition_links ON edition_links.edition_id = editions.id")
-    scope = scope.left_joins(document: :link_set_links)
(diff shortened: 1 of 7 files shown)


JonathanHallam (approved):
have a green tick :)

tahb:
Since being reverted @richardTowers [may have found the flaw that might have improved the performance](https://github.com/alphagov/publishing-api/pull/2846#issuecomment-2301815794). Should this work need to go back in for any reason (perhaps someone else wants to query by the `link_set_links` field?) then the suggestion is worth investigating further!

That said I think the new approach we're taking of a dedicated endpoint for embedded content is a good idea so we'll leave this reverted.

PR #2855, opened 2024-08-28
Content modelling/450 pact helpers for getting content with embeds

tahb:
Add Pact helpers for getting content which embeds other content

This should support the GDS API Adapter tests being written here[1].

[1] https://github.com/alphagov/gds-api-adapters/pull/1283


spec/service_consumers/pact_helper.rb
```diff
@@ -650,4 +650,35 @@ def url_encode(str)
...
+      primary_publishing_organisation = create(:live_edition,
+                                               document: organisation_document,
+                                               title: "bar",
+                                               document_type: "organisation",
+                                               schema_name: "organisation",
+                                               base_path: "/bar")
+      create(
+        :link,
+        edition: edition_with_embedded_edition,
+        link_type: :primary_publishing_organisation,
```
richardTowers, line 677: Do we need a link with `link_type: :embed` somewhere?

tahb: Yes we do 👍 I added this [to the new factory](https://github.com/alphagov/publishing-api/pull/2855/files#diff-c4ac97010572ffaccb617167de9370025111bac1634d097ae013d54f41852b05R105) which is perhaps too mysterious and could be made explicit?

callumknights (approved):

spec/service_consumers/pact_helper.rb
richardTowers: Ohh, I see - that's fine, I just didn't follow it all the way through. No need to be any more explicit I don't think.

PR #2875, opened 2024-09-20
Add licensify as a rendering app

theseanything:
This will allow us to publish a special route the ensures that we have routing information for licensify in the Content Store. This supports work to allow router to use the content store as source of truth for routing information. Also in ensures that in case of disaster recovery, we can easily republish routes to licensify.

https://github.com/alphagov/special-route-publisher/pull/352


yndajas:
@theseanything would you mind if we mark this as draft (looks like there's one failing check that needs addressing), so that Seal stops giving us daily alerts about it for now?

PR #2882, opened 2024-09-24
(562) Ensure only live organisations appear in join

pezholio:
When embedded content has a primary publication organisation, the JOIN was pulling in all draft and superseded editions in, resulting in duplicate rows (so if an organisation has 400 superseded editions and 1 live edition we get 401 rows!). This alters the JOIN to only pull in the live edition of a primary publishing organisation, meaning we only get one row per embedded content.

I’ve also updated the tests to create draft and superseded editions for the organisation, so we can ensure that we have a test suite that better represents reality.


Harriethw (approved):
🐛 🔨

PR #2887, opened 2024-09-25
Tidy special route publisher and related files

KludgeKML:
Tidy up some things missed in the move of special route publisher

- move non-task files from /lib/tasks to /lib (and their respective spec files)
- consolidate the two special route yaml files into one (and their respective fixtures) and filter in/out homepage based on document type == homepage
- fix order of values in publishing_app.jsonnet
- namespace special route tasks


lib/data_sanitizer.rb
```diff
@@ -0,0 +1,10 @@
+class DataSanitizer
```
brucebolt, line 1: I think all of these files would be better namespaced under something like `SpecialRoutes`, then in `lib/special_routes` to keep them separate from the other Publishing API core behaviour.

KludgeKML: Well these two (DataSanitizer and DatabaseRecordValidator) aren't special routes related - they just suffered from the same problem (being non-rake files in the /tasks directory).

I could leave them where they are (don't tinker with them in this PR)? then just touch the SpecialRoutesPublisher library file and move it to lib/special routes (rename it SpecialRoutes:Publisher?)

brucebolt: Ah, sorry, didn't quite notice that.  I think they should be ok with the base directory then, since they're probably part of the core Publishing API rake tasks.

brucebolt (approved):

PR #2889, opened 2024-09-25
Add favicon special route for frontend

KludgeKML:
Frontend now handles favicons, (https://github.com/alphagov/frontend/pull/4241) so add a special route to point /favicon.ico there

### DIFF
--- a/lib/data/special_routes.yaml
+++ b/lib/data/special_routes.yaml
@@ -245,3 +245,10 @@
   :additional_routes:
     - :base_path: "/when-do-the-clocks-change.json"
       :type: "exact"
+
+- :content_id: "5fc8b4bc-f899-4d4e-ae10-f9beeb172f50"
+  :base_path: "/favicon.ico"
+  :title: "Favicon"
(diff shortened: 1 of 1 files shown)


lib/data/special_routes.yaml
```diff
@@ -244,4 +244,10 @@
...
     - :base_path: "/when-do-the-clocks-change.json"
-      :type: "exact"
\ No newline at end of file
+      :type: "exact"
+
+- :content_id: "5fc8b4bc-f899-4d4e-ae10-f9beeb172f50"
+  :base_path: "/favicon.ico"
+  :title: "Favicon"
+  :rendering_app: "frontend"
+  :description: "The favicon is the image displayed in locations such as the browser tabs."
```
hannako, line 253: This route is already in the db though right? So do we need to add `:override_existing: true` ?

KludgeKML: Oh, good spot. Thanks! Fixed.

hannako (approved):
LGTM

PR #2895, opened 2024-09-30
Add landing page schema / document type

richardTowers:
Like the coronavirus landing page schema, this will permit details objects of any structure. This is to give us an initial blank canvas while we establish the correct content model, and may be restricted further down the line.

### DIFF
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -78,6 +78,7 @@
 - independent_report
 - international_development_fund
 - international_treaty
+- landing_page
 - licence
 - license_finder
 - licence_transaction
--- a/content_schemas/dist/formats/generic/frontend/schema.json
+++ b/content_schemas/dist/formats/generic/frontend/schema.json
@@ -114,6 +114,7 @@
         "independent_report",
         "international_development_fund",
         "international_treaty",
+        "landing_page",
         "licence",
         "license_finder",
         "licence_transaction",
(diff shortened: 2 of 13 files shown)


leenagupte (commented):
I have a question about the description of the change which I've tagged you in elsewhere.

Also, could an example content item be added?.

richardTowers:
Description edited.

I don't think it's worth adding an example content item until we know what the structure is going to be, no? Otherwise it'll just be a confusing / imperfect mirror of the structure we've hardcoded in frontend.

leenagupte:
> Description edited.
> 
> I don't think it's worth adding an example content item until we know what the structure is going to be, no? Otherwise it'll just be a confusing / imperfect mirror of the structure we've hardcoded in frontend.

I meant an example for the basic content item, with an empty details hash. 

A fake content item is being duplicated in a few places in the frontend tests at the moment:

- [model spec](https://github.com/alphagov/frontend/blob/1d8104f7ebbd31c251dd1787172653d2b566738c/spec/models/landing_page_spec.rb)
- [system spec](https://github.com/alphagov/frontend/blob/1d8104f7ebbd31c251dd1787172653d2b566738c/spec/system/landing_page_spec.rb)
- [request spec](https://github.com/alphagov/frontend/blob/1d8104f7ebbd31c251dd1787172653d2b566738c/spec/requests/landing_page_spec.rb)

It'd be nicer if one example could be pulled in rather.

richardTowers:
Ah, fair enough. Yeah, I can add that.

richardTowers:
Moving back to draft while I investigate the failing test

richardTowers:
Frontend tests fixed in https://github.com/alphagov/frontend/pull/4267
Unrelated whitehall tests fixed in https://github.com/alphagov/whitehall/pull/9506

Ready for review again!

leenagupte (approved):
Thanks for adding an example ⭐

PR #2896, opened 2024-09-30
Account api special routes

KludgeKML:
Almost all account-api routes go to frontend, and a bunch of them are already in the file, because they've been duplicated in the past between account-api and frontend. So add the missing ones: /account, /api/personalisation/check-email-subscription - which does go to account-api, the only route that does, /sign-in and /sign-in/redirect)

(Also: remove two routes that no longer exist in frontend, and tidy up the file to make things a bit easier to find).

### DIFF
--- a/lib/data/special_routes.yaml
+++ b/lib/data/special_routes.yaml
@@ -14,16 +14,6 @@
   :title: "Account home page"
   :rendering_app: "frontend"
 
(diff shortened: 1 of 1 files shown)


leenagupte (approved):
Thanks for alphabetising the routes ⭐

PR #2903, opened 2024-10-02
Relax Publishing API schema for specialist documents

lauraghiorghisor-tw:
# Context
When creating or editing Specialist Documents, we need to make some changes in Publishing API and regenerate the schemas. The schemas, and the underlying files that help generate the schemas, where the manual changes are made, are quite verbose when it comes to specifying facets and document types.

To reduce effort in creating and maintaining specialist document configuration, we can reduce the specificity of the schemas. We believe validating that facets and document type are simple strings, rather than maintaining a full list of options to validate against, is valuable enough. This is what these commits achieve.

# Changes
- remove enums from `_specialist_document `shared definitions, so that facets are only validated as strings
- remove `document_type` enums from `specialist_document` schemas, so that it's now only validated as a string
- remove specialist document types from `allowed_document_types` list
- add custom logic for specialist documents, so that the schema generators behave differently when dealing with them
- add test to cover new exception logic

# Note - alternative path we've looked into
We also considered maintaining the `document_type` enum and validation, instead reducing maintenance cost by removing the `allowed_document_types` list. The list is used as an additional "whitelisting" check, and as a fallback when a document type is not provided. There is no actual schema in `publishing_api` that does not have a `document_type` specified, meaning we could raise an error if the document type is not provided, instead of defaulting to allow list logic [here](https://github.com/alphagov/publishing-api/blob/e54eb4d21cabd2b5172393b79082bff44b92cae7/lib/schema_generator/format.rb#L174). 

Nonetheless, there are some `generic` and `generic_with_external_related_links` schemas that do make use of the allowed types as they do not provide a document type. We could provide any random string as a type, but it seems other publishing apps use these generic schemas to run tests, making this change rather costly as it involves fixing many failures across several repos. 

# Links
[Trello](https://trello.com/c/sFqXJwDv/2815-relax-publishing-api-schema-for-specialist-documents)


minhngocd (approved):
Looks good to me! would be nice to also get an eye from Publishing Platform as an FYI and see if anyone wants to raise any concerns. 

At the rate of change of specialist document schema types, relaxing the schemas will give us better value in reduced cost of change - with relatively low risk as communications are between internal systems, and any mistakes would happen during generation of a new specialist document type, which is currently a highly manual process with lots of manual validation.

brucebolt (approved):
Looks good to me 👍.

I've never seen the benefit of hardcoding these lists in the schema anyway, especially since the requests come from a known source (one of our own apps that we control).

kevindew:
Argh - this has broken our build on GOV.UK Chat. As we validate that the document types we care about are actually valid document types

It's a bit weird that we now have a method `GovukSchemas::DocumentTypes.valid_document_types` that now returns an incomplete list - could we consider reverting this until we can either deprecate or have alternative means to resolve the API issues this causes on GovukSchemas?

kevindew:
Also expect this will break the dev docs: https://github.com/search?q=repo%3Aalphagov%2Fgovuk-developer-docs+allowed_document_types&type=code

kevindew:
Understanding this a bit more it looks like this would remove what is our only list definitive list of document types for GOV.UK so would strongly suggest rethinking this as we'll lose that resource. 

It feels like you could save a maintenance overhead by only updating one list.

Also seems an unlikely situation to have to add a new specialist document type and not make any schema changes - given it's a new type of content, something in the schemas should know about that.

lauraghiorghisor-tw:
Linking #2917 for reference on what we reverted.

lauraghiorghisor-tw:
> Understanding this a bit more it looks like this would remove what is our only list definitive list of document types for GOV.UK so would strongly suggest rethinking this as we'll lose that resource.
> 
> It feels like you could save a maintenance overhead by only updating one list.
> 
> Also seems an unlikely situation to have to add a new specialist document type and not make any schema changes - given it's a new type of content, something in the schemas should know about that.


We were not aware that the allowed list is used as a reference definitive list. We even looked into removing the allowed list completely (broke many tests that use the generic schemas). It's these generic (and generic with links) schemas that use the entire list of allowed documents. Otherwise all other ("real") schemas get their own `document_type` field which is used by the schema generators. You can see in #2917 that the generic schemas were the only ones to change upon reintroducing the specialist document types in the allowed list. 

This unlikely situation is the "relaxation" we've introduced. Upon adding a new specialist document type, none of the specialist document schemas will change.

PR #2904, opened 2024-10-03
Content Modelling/593 New update type for updating Content Blocks and their Host Documents

Harriethw:
For more context of the Content Block work, please see the `content_object_store` engine in Whitehall https://github.com/alphagov/whitehall/blob/3375d41d987f9790e410eeb8bc6bc79aebfc1cab/docs/adr/0004-content-object-store-added-with-a-rails-engine.md

For the long version of why we are making this change see the ADR https://github.com/alphagov/publishing-api/blob/content-modelling/593-content-block-update-type/docs/arch/adr-006-new-event-type-for-content-blocks.md 

when a Content Block is updated, we want to support features that relate to the updating of the Block and its Host Documents (Documents that have the Block in their content, currently identified as Links).

As Content Blocks are a new type of content, we think it makes sense that a new update type is issued, because we want to diverge in the near future from the existing update flow for Editions, and it semantically is a different type of update - `major`/`minor` updates do not mean anything (currently) in Content Block land.

This PR introduces two new update types:
1. `content_block` to track when a content block is being updated
2. `host_content` to track when a host document is being updated

without (yet) changing any of the functionality related to `major` updates. 

This will enable us to make changes in the near future such as adding Change Notes and listening to the update type to send emails. (as in ADR).

The work to use this type is in Whitehall PR here https://github.com/alphagov/whitehall/pull/9497 

Spike branch to show how we might create ChangeNotes in Publishing API https://github.com/alphagov/publishing-api/compare/content-modelling/spike-change-notes?expand=1

## questions

* This would mean that the `major_published_at` and `public_updated_at` fields on an Edition might not be updated correctly https://github.com/alphagov/publishing-api/blob/df7eae7b8cd1160aeda2f54c963bf08162d84c7a/app/models/edition/timestamps.rb#L43 (do we care about that?)
* Are there any other knock on effects of adding new update types?

----

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -11,8 +11,10 @@ module Commands
 
     private
 
+      UPDATE_TYPES_WITH_CHANGE_NOTES = %w[major content_block].freeze
+
       def publish_edition
-        delete_change_notes unless update_type == "major"
+        delete_change_notes unless UPDATE_TYPES_WITH_CHANGE_NOTES.include?(update_type)
         previous_edition.supersede if previous_edition
 
         unless edition.pathless?
@@ -137,7 +139,7 @@ module Commands
       end
 
       def valid_update_types
-        %w[major minor republish links]
+        %w[major minor republish links content_block]
       end
 
       def already_published?
--- a/app/sidekiq/downstream_live_job.rb
+++ b/app/sidekiq/downstream_live_job.rb
@@ -83,7 +83,8 @@ private
   end
 
   def enqueue_dependencies
-    DependencyResolutionJob.perform_async(
+    dependency_job_klass = @message_queue_event_type == "content_block" ? HostContentUpdateJob : DependencyResolutionJob
+    dependency_job_klass.perform_async(
       "content_store" => "Adapters::ContentStore",
       "content_id" => content_id,
       "locale" => locale,
--- a/app/sidekiq/host_content_update_job.rb
+++ b/app/sidekiq/host_content_update_job.rb
@@ -1,0 +1,18 @@
+class HostContentUpdateJob < DependencyResolutionJob
+private
+
+  def downstream_live(dependent_content_id, locale)
+    return if draft?
+
+    DownstreamLiveJob.perform_async_in_queue(
+      DownstreamLiveJob::LOW_QUEUE,
+      "content_id" => dependent_content_id,
+      "locale" => locale,
+      "message_queue_event_type" => "host_content",
+      "update_dependencies" => false,
+      "dependency_resolution_source_content_id" => content_id,
+      "source_command" => source_command,
+      "source_fields" => source_fields,
+    )
+  end
+end
--- a/docs/arch/adr-006-new-event-type-for-content-blocks.md
+++ b/docs/arch/adr-006-new-event-type-for-content-blocks.md
@@ -1,0 +1,70 @@
+# Decision Record: Add new event type to add custom dependency resolution behaviour when a Content Block is updated
+
+## Context
+
+### Existing behaviour
+
+The way content blocks are handled leans on a lot of existing functionality within the publishing platform. 
+Previously to the introduction of content blocks, documents had (and still have!) the concept of “links”, so when a 
+document is changed, this triggers the republication of all the documents that are associated with that page.
+
+A good example of this is an “Organisation” page (for example, the Cabinet Office). The Cabinet Office organisation 
+page, as well as listing some information about the organisation, also lists the ministers for that organisation.
+
+When the frontend fetches the organisation from the content store, it also gets the ministers (stored as links). 
+This information is static within the content store, so if information about those ministers changes, we need to 
+send the organisation page back to the Content Store along with the new minister information.
+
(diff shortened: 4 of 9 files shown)


app/sidekiq/host_content_update_job.rb
```diff
@@ -0,0 +1,33 @@
+class HostContentUpdateJob < DependencyResolutionJob
```
pezholio, line 1: I know this isn't doing a ton at the mo, but it'd be good to get some tests for this class, just so we're documenting behaviour

Harriethw: ahh yeah good point I forgot to come back to this!

docs/arch/adr-006-new-event-type-for-content-blocks.md
```diff
@@ -0,0 +1,67 @@
...
+a dependent document is updated. This has a couple of issues:
+
+1. The code is already quite verbose, and this adds a bunch of complexities
+2. We need to reach into the payload and generate a new changenote for each dependent bit of content, which causes 
+   misdirection and is again, quite messy
+
+We therefore suggest updating the Publishing API code to send a new `host_content_update` event type when a piece of 
+dependent content is updated.
+
+We will then be able to update the Email Alerts Service to subscribe to this event type, and also add subscription
```
pezholio, line 49: I think we're going to skip the subscription functionality within the publishing apps, so let's leave this bit out and only cover emails

app/sidekiq/downstream_live_job.rb
```diff
@@ -83,15 +83,27 @@ def assign_attributes(attributes)
...
-      "content_store" => "Adapters::ContentStore",
-      "content_id" => content_id,
-      "locale" => locale,
-      "orphaned_content_ids" => orphaned_content_ids,
-      "source_command" => source_command,
-      "source_document_type" => edition.document_type,
-      "source_fields" => source_fields,
-    )
+    if @message_queue_event_type == "content_block"
+      HostContentUpdateJob.perform_async(
```
pezholio, line 87: Tiny thing here, but I'm wondering if we can set the job's class based on the event type, something like:

```ruby
dependency_job_klass = @message_queue_event_type == "content_block" ? HostContentUpdateJob : DependencyResolutionJob
dependency_job_klass.perform_async(
  "content_store" => "Adapters::ContentStore",
  "content_id" => content_id,
  "locale" => locale,
  "orphaned_content_ids" => orphaned_content_ids,
  "source_command" => source_command,
  "source_document_type" => edition.document_type,
  "source_fields" => source_fields,
)
```

pezholio (approved):
LGTM - just a tiny suggestion code-wise

app/commands/v2/publish.rb
```diff
@@ -12,7 +12,7 @@ def call
     private
 
       def publish_edition
-        delete_change_notes unless update_type == "major"
+        delete_change_notes unless update_type == "major" || update_type == "content_block"
```
JonathanHallam, line 15: I think this is fine for now but if this sets a precedent we might want something like
```
UNDELETABLE_CHANGE_NOTE_UPDATE_TYPES = %w(major content_block).freeze
...
delete_change_notes unless UNDELETABLE_CHANGE_NOTE_UPDATE_TYPES.include(update_type)
```

naming things is hard🙈

JonathanHallam (commented):
I've made one really tiny non blocking comment. @Gweaton could you have a look over this as well just to see what the crossover is (if any) with the friendly ID work

app/sidekiq/host_content_update_job.rb
```diff
@@ -0,0 +1,18 @@
...
+private
+
+  def downstream_live(dependent_content_id, locale)
+    return if draft?
+
+    DownstreamLiveJob.perform_async_in_queue(
+      DownstreamLiveJob::LOW_QUEUE,
+      "content_id" => dependent_content_id,
+      "locale" => locale,
+      "message_queue_event_type" => "host_content",
```
Gweaton, line 11: **question/suggestion**: given this is doing the same as `DependencyResolutionJob`'s `downstream_live` method but passing in a different `message_queue_event_type`, could it be an idea to extract that to a shared method with an argument for `message_queue_event_type`? Or can you see this behaviour diverging in future updates?

This would prevent changes to `DependencyResolutionJob`'s behaviour being missed when updating host content, but this may be by design!

Gweaton (approved):
This looks good - I can't see any conflicts with the work we've been doing to support friendly content IDs (aka `contentIdAliases`) at the moment, due to the way publishing uses `content_id`s under the hood anyway, and I think it still should.

Just one non-blocking suggestion if it's possible/nice.

app/commands/v2/publish.rb
Harriethw: this is a good point, I meant to come back to this line as I didn't like magic strings floating around. I've called the list `UPDATE_TYPES_WITH_CHANGE_NOTES` but not sure if that's a better name :)

Harriethw: done! have put at top of private block to keep the scope limited

PR #2911, opened 2024-10-09
Create human readable content aliases on `PUT Content`

Gweaton:
[Trello card](https://trello.com/c/lgu8vhzK/1402-finish-up-friendly-id-work)

## Changes in this PR

At the moment, Publishing API uses "content ids" (UUIDs) to identify bits of content. When embedding content into the body or title of an Edition, users currently enter this content id as part of the embed code (e.g `{{embed:email:f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a}}`, which is then expanded under the `embed` key of the `Edition`'s `links` attribute.

Here we introduce a human-readable alias for this content id to make it easier for content editors to embed content and make sense of this embedded string.

We're not expanding this anywhere yet, and it's an optional parameter, so this shouldn't have any impact on existing API calls or embedding of content.

### DIFF
--- a/app/commands/v2/put_content.rb
+++ b/app/commands/v2/put_content.rb
@@ -16,6 +16,7 @@ module Commands
         edition = create_or_update_edition
         set_timestamps(edition)
 
+        create_content_id_alias
         update_content_dependencies(edition)
 
         orphaned_links = link_diff_between(
@@ -96,6 +97,19 @@ module Commands
             end,
           )
         end
+      end
+
+      def create_content_id_alias
+        content_id_alias_name = payload[:content_id_alias]
+        return unless content_id_alias_name
+
+        if (content_id_alias = ContentIdAlias.find_by(name: content_id_alias_name))
+          return if content_id_alias.content_id == payload[:content_id]
+
+          raise CommandError.new(code: 422, message: "ContentIdAlias with name \"#{content_id_alias_name}\" exists for a different content ID.")
+        end
+
+        ContentIdAlias.create!(name: content_id_alias_name, content_id: payload[:content_id])
       end
 
       def fetch_embedded_content(edition)
--- a/app/models/content_id_alias.rb
+++ b/app/models/content_id_alias.rb
@@ -1,0 +1,3 @@
+class ContentIdAlias < ApplicationRecord
+  validates :name, uniqueness: true, presence: true
+end
--- a/content_schemas/formats/content_block_postal_address.jsonnet
+++ b/content_schemas/formats/content_block_postal_address.jsonnet
@@ -20,7 +20,7 @@
         },
         postcode: {
           type: "string"
-        }
+        },
       },
     },
   },
--- a/content_schemas/formats/shared/definitions/content_id_alias.jsonnet
+++ b/content_schemas/formats/shared/definitions/content_id_alias.jsonnet
@@ -1,0 +1,16 @@
+{
+  content_id_alias: {
+    description: "Human-readable alias for a Content ID, used when embedding content. Should only be supplied when updating Content Blocks.",
+    type: "string",
+  },
+  content_id_alias_optional: {
+    anyOf: [
+      {
+        "$ref": "#/definitions/content_id_alias",
+      },
+      {
+        type: "null",
+      },
+    ],
+  },
+}
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_02_20_153403) do
+ActiveRecord::Schema[7.2].define(version: 2024_10_09_141509) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -42,6 +42,15 @@ ActiveRecord::Schema[7.1].define(version: 2024_02_20_153403) do
     t.datetime "created_at", precision: nil
     t.datetime "updated_at", precision: nil
     t.index ["edition_id"], name: "index_change_notes_on_edition_id"
+  end
+
+  create_table "content_id_aliases", force: :cascade do |t|
+    t.string "name", null: false
+    t.uuid "content_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["content_id"], name: "index_content_id_aliases_on_content_id"
+    t.index ["name"], name: "index_content_id_aliases_on_name", unique: true
   end
 
   create_table "documents", id: :serial, force: :cascade do |t|
--- a/content_schemas/dist/formats/answer/publisher_v2/schema.json
+++ b/content_schemas/dist/formats/answer/publisher_v2/schema.json
@@ -35,6 +35,9 @@
         "string"
       ]
     },
+    "content_id_alias": {
+      "type": "null"
+    },
     "description": {
       "$ref": "#/definitions/description_optional"
     },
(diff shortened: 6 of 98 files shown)


content_schemas/formats/shared/definitions/content_id_alias.jsonnet
```diff
@@ -0,0 +1,5 @@
+{
```
Gweaton, line 1: Do I need to move the addition of this new field into its own PR as per [the documentation here](https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#general-workflow) or is it ok to sit within this PR?

brucebolt: I'd leave it within this PR, as the schema change is related to the other commits.

app/commands/v2/put_content.rb
```diff
@@ -98,6 +99,17 @@ def create_links(edition)
         end
       end
 
+      def create_content_id_alias
+        content_id_alias = payload.dig(:details, :content_id_alias)
+        return unless content_id_alias && EmbeddedContentFinderService::SUPPORTED_DOCUMENT_TYPES.include?(payload[:document_type])
+
+        if ContentIdAlias.find_by(name: content_id_alias)
```
pezholio, line 106: Will this fail if I publish more than one edition for the same document? This reads like we're expecting editions to have unique aliases, which I don't think is right.

brucebolt (commented):
Breaking this down into smaller PRs was definitely a good idea, since it's made this much easier to review 🎖️

pezholio (commented):
This looks good on the whole. I am worried about that check, but I could be barking up the wrong tree!

I guess if we're enforcing uniqueness on the API side, we should probably ensure the publishing app creates unique aliases, much in the same way as slugs

app/models/content_id_alias.rb
```diff
@@ -0,0 +1,8 @@
+class ContentIdAlias < ApplicationRecord
+  validates :name, uniqueness: true
+  validate :name_cannot_be_blank
+
+  def name_cannot_be_blank
+    errors.add(:name, "Name cannot be blank") if name.strip.empty?
```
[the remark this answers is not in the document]
Gweaton: Ah yeah, that's exactly what we need. I was scouring the docs for something like that but missed that note about `presence` 🤦

db/schema.rb
```diff
@@ -98,6 +107,15 @@
...
   end
 
+  create_table "embedded_content_references", force: :cascade do |t|
+    t.string "friendly_id", null: false
+    t.uuid "content_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["content_id"], name: "index_embedded_content_references_on_content_id"
+    t.index ["friendly_id"], name: "index_embedded_content_references_on_friendly_id"
+  end
```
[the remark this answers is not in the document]
Gweaton: oops, this must've slipped in from before the great renaming.

app/models/content_id_alias.rb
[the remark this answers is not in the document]
yndajas: Ooh, didn't realise that was whitespace-proof!

app/commands/v2/put_content.rb
yndajas: Yeah I think we should only error if the content ID differs:
- if the alias and content ID match an existing `ContentIdAlias`, do nothing
- if the alias matches an existing `ContentIdAlias`, error
- if the alias doesn't exist, create one for the given content ID

app/commands/v2/put_content.rb
```diff
@@ -98,6 +99,19 @@ def create_links(edition)
...
 
+      def create_content_id_alias
+        content_id_alias = payload.dig(:details, :content_id_alias)
+        return unless content_id_alias
+
+        return if ContentIdAlias.find_by(name: content_id_alias, content_id: payload[:content_id])
+
+        if ContentIdAlias.find_by(name: content_id_alias)
+          raise CommandError.new(code: 422, message: "ContentIdAlias with name \"#{content_id_alias}\" exists for a different content ID.")
+        end
```
brucebolt, line 110: Minor non-blocking suggestion: this will make two database queries that are almost identical. Could it be condensed into one?
```suggestion
        if content_id_alias = ContentIdAlias.find_by(name: content_id_alias)
          return if content_id_alias.content_id == payload[:content_id]
          
          raise CommandError.new(code: 422, message: "ContentIdAlias with name \"#{content_id_alias}\" exists for a different content ID.")
        end
```

brucebolt (approved):

app/commands/v2/put_content.rb
yndajas: Yeah I did think about that - your suggestion looks good!

content_schemas/formats/content_block_postal_address.jsonnet
```diff
@@ -20,6 +20,9 @@
         },
         postcode: {
           type: "string"
+        },
+        content_id_alias: {
+          "$ref": "#/definitions/content_id_alias",
```
pezholio, line 25: This feels to be like a metadata element, so should ideally sit outside the `details` object. I _think_ it should be just a case of adding it to the top level, so it looks like this:

```
(import "shared/content_block.jsonnet") + {
  document_type: "content_block_postal_address",
  content_id_alias: {
    "$ref": "#/definitions/content_id_alias",
  },
  definitions: {
    details: {
      type: "object",
      additionalProperties: false,
      required: ["line_1", "town_or_city", "postcode"],
      properties: {
        line_1: {
          type: "string",
        },
        line_2: {
          type: "string",
        },
        town_or_city: {
          type: "string",
        },
        county: {
          type: "string",
        },
        postcode: {
          type: "string"
        },
      },
    },
  },
}
```

pezholio: Thinking about it, that should probably sit in the `shared/content_block.jsonnet` file too

pezholio: I've added a couple of temp commits (for ease of rebase) to demonstrate what I'm thinking.

Basically, I've added a `content_id_alias` field to the [publisher default properties](https://github.com/alphagov/publishing-api/blob/2481672d732cfd01db9af0097308e5275818b275/content_schemas/formats/shared/default_properties/publisher.jsonnet#L35), and [made them forbidden by default](https://github.com/alphagov/publishing-api/blob/2481672d732cfd01db9af0097308e5275818b275/content_schemas/formats/shared/default_format.jsonnet#L14). The Content Block format (which inherits from the default) then [overrides this to make it optional (we might want to make it mandatory later)](https://github.com/alphagov/publishing-api/blob/2481672d732cfd01db9af0097308e5275818b275/content_schemas/formats/shared/content_block.jsonnet#L5)

Hope this makes sense! Any questions, let me know! 👍

content_schemas/formats/shared/definitions/content_id_alias.jsonnet
```diff
@@ -1,5 +1,16 @@
 {
   content_id_alias: {
+    description: "Human-readable alias for a Content ID, used when embedding content. Should only be supplied when updating Content Blocks.",
     type: "string",
   },
+  content_id_alias_optional: {
```
pezholio, line 6: I think I follow this now, so if we were going to make it required, we'd have something like this?

```
content_id_alias_required: {
  "$ref": "#/definitions/content_id_alias",
}
```

pezholio: Oh wait, I see, all the magic is in the `Format` class - that makes sense 👍

callumknights (approved):

PR #2915, opened 2024-10-14
Update docs re: updating content schemas

Gweaton:
These docs didn't explain the steps involved in adding new fields to the `default_format.jsonnet` schema, which we had to do when adding `content_id_alias` as a new field outside of those already declared there.

This hopefully clarifies things, in particular the need to add new methods to the `SchemaGenerator::Format` class which was hidden away.

### DIFF
--- a/docs/content_schemas/changing-an-existing-content-schema.md
+++ b/docs/content_schemas/changing-an-existing-content-schema.md
@@ -1,50 +1,75 @@
 # Changing an existing content schema
 
-## General workflow
+## When not adding or removing top-level fields
 
-* create a branch
-* make changes to the relevant schema fragment under `formats/<format name>.jsonnet`
-* change the relevant frontend examples (or add a new example)
-* run `rake build_schemas` to compile a new schema.json
-* run `rake` to test the example against the schema
-* commit and push
-* open a PR
-* watch the multi-build status to see if all apps are compatible with the change
-* if the builds are successful, it can be merged
+This describes how to add or remove fields that extend the top-level fields
+defined in `default_format.jsonnet`. For example, if you wanted to submit an
+additional field via a form in a publishing app, you might add a field nested
+within `details` (a field already defined in `default_format.jsonnet`).
 
-## Adding a new field to a content schema
+1. If you're adding or removing a required field:
+   - Update publishing app(s) to provide/stop providing the field
+   - If removing a field, update frontend app(s) to no longer access the
+     soon-to-be-removed field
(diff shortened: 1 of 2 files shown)


docs/content_schemas/changing-an-existing-content-schema.md
```diff
@@ -1,50 +1,64 @@
 # Changing an existing content schema
 
-## General workflow
+## When `default_format.jsonnet` does not need updating
```
pezholio, line 3: I wonder whether "When not adding/removing top-level fields" would be a better description?

docs/content_schemas/changing-an-existing-content-schema.md
```diff
@@ -1,50 +1,64 @@
...
+   for the field
+1. Add, remove, or update the relevant example schemas in
+   `content_schemas/examples/**/<example_name>.jsonnet`.
+1. Run `rake build_schemas` to compile the updated schemas in
+   `content_schemas/dist/` and validate the example schemas.
+1. Check that all the CI workflows pass. These will verify that the content
+   schemas are compatible with the suite of publishing apps
 
-Follow the General workflow described above. You are now free to add support for the field to apps. If you are adding a mandatory field you will need to additionally:
+## When `default_format.jsonnet` needs updating
```
pezholio, line 30: Similarly, something like "When adding or removing top-level fields" might read better here

docs/content_schemas/changing-an-existing-content-schema.md
```diff
@@ -1,50 +1,64 @@
...
-If the field was mandatory:
-
-* Remove the field from the `required` attribute by following the General workflow
-  * Remove the field from examples to ensure that the frontend can handle the field being optional
-* Deploy any frontend changes
-* Change the publisher to stop sending the field
-* Deploy the publisher
-* Optional: change the records in content-store
-* Remove the field by following the General workflow
+1. If you're adding or removing a required field:
```
pezholio, line 39: It might be good to give an example in each of these steps, just to make it a bit clearer how to do this.

pezholio (commented):
A couple of comments here - would be good to get some examples in I think

docs/content_schemas/changing-an-existing-content-schema.md
Gweaton: We wrote this and then deleted it a few times, useful to have another opinion! I think we can put it back.

pezholio (approved):
Nice work!

PR #2917, opened 2024-10-16
Restore specialist document types in allowed list

lauraghiorghisor-tw:
We've recently removed the specialist document types from the allowed list, in an effort to relax the schema, so that it only checks that the document type is a string.

There are places that use this list, and it should include all the allowed types. We can keep the rest of our changes towards relaxing the schema.


ChrisBAshton (approved):

kevindew (approved):
Thanks this will fix our issue on GOV.UK Chat and I understand it'll stop the dev docs from breaking. Though expect you should have a Publishing Platform review too.

Would suggest you don't need the custom overrides like: https://github.com/alphagov/publishing-api/blob/4c12369a277df3833ed1c9e4f6657e43ddd2b532/lib/schema_generator/format.rb#L175-L177 anymore as the schema would be auto populated with the full list of document types so can be a bit stricter than fully permissive for less code.

lauraghiorghisor-tw:
> Original PR

Fair shout. Changes in #2918.

PR #2919, opened 2024-10-16
Optionally set an Author on an Edition

pezholio:
In Content Modelling, when viewing content that is dependent of a content block, we'd like to be able to see who last edited that content, so publishers can liaise with the "owner" of that content if, for example, a change to a content block will alter the meaning of the dependent content. 

As the Content Block Manager is publishing app agnostic, we currently have no way of knowing who last touched a piece of content, as this information is only stored within the publishing app itself.

This PR adds a new `last_edited_by` field which can optionally be sent to the Publishing API. In future this will allow us to use the [Embedded Content Endpoint](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#get-v2contentcontent_idembedded) to return information about authors (if it exists).

The next step will be to update the presenters in the various publishing apps - trialing with Whitehall first.

### DIFF
--- a/content_schemas/formats/shared/default_properties/publisher.jsonnet
+++ b/content_schemas/formats/shared/default_properties/publisher.jsonnet
@@ -23,6 +23,10 @@
     format: "date-time",
     description: "Last time when the content received a major or minor update.",
   },
+  last_edited_by_editor_id: {
+    type: "string",
+    description: "The UUID of the editor who edited the content.",
+  },
   previous_version: {
     type: "string",
   },
--- a/db/migrate/20241017080126_add_last_edited_by_editor_id_to_editions.rb
+++ b/db/migrate/20241017080126_add_last_edited_by_editor_id_to_editions.rb
@@ -1,0 +1,5 @@
+class AddLastEditedByEditorIdToEditions < ActiveRecord::Migration[7.2]
+  def change
+    add_column :editions, :last_edited_by_editor_id, :uuid
+  end
+end
--- a/spec/integration/put_content/new_edition_spec.rb
+++ b/spec/integration/put_content/new_edition_spec.rb
@@ -83,6 +83,8 @@ RSpec.describe "PUT /v2/content when the payload is for a brand new edition" do
     include_examples "creates a change note"
   end
 
+  include_examples "setting last_edited_by_editor_id"
+
   context "and the change history is in the details hash" do
     before do
       payload.delete(:change_note)
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -18,6 +18,7 @@ class Edition < ApplicationRecord
     document_type
     first_published_at
     last_edited_at
+    last_edited_by_editor_id
     major_published_at
     phase
     public_updated_at
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -9,6 +9,7 @@ module Presenters
       document_id
       id
       last_edited_at
+      last_edited_by_editor_id
       publishing_api_first_published_at
       major_published_at
       published_at
--- a/content_schemas/dist/formats/answer/publisher_v2/schema.json
+++ b/content_schemas/dist/formats/answer/publisher_v2/schema.json
@@ -58,6 +58,10 @@
       "type": "string",
       "format": "date-time"
     },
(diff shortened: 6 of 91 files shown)


app/commands/v2/put_content.rb
```diff
@@ -77,6 +78,16 @@ def set_timestamps(edition)
         Edition::Timestamps.edited(edition, payload, previously_published_edition)
       end
 
+      def set_publishing_user(edition)
+        return unless payload[:last_edited_by]
+
+        edition.author = Author.find_or_create_by(
+          name: payload[:last_edited_by][:name],
+          email: payload[:last_edited_by][:email],
```
yndajas, line 86: **suggestion:** multiple access

Is it worth assigning `payload[:last_edited_by]` to a variable before line 82, since it's accessed three times if present? We did similar in `create_content_id_alias`

spec/integration/put_content/new_edition_spec.rb
```diff
@@ -104,4 +110,42 @@
 
     include_examples "creates a change note"
   end
+
+  context "and the payload includes last_edited_by" do
```
yndajas, line 114: **issue:** test location

The tests look good in this third commit, but is this behaviour specific to new editions? I wonder if it belongs in the main `put_content_spec`, optionally in a `describe` block like we did for [optionally setting a content ID alias](https://github.com/alphagov/publishing-api/blob/dcdc40cdcb898b7816147d8c44225507788fb62a/spec/commands/v2/put_content_spec.rb#L107)?

yndajas (commented):
Generally looks good. A couple of things that might be good to address, and a couple of other more minor suggestions. I might also quickly run the broader idea past the team in our dev chat this afternoon

---

The labels/first lines of each of the comments here are following the [Conventional Comments](https://conventionalcomments.org/) format, which I'm trying out after watching [a useful DevOpsDays London talk](https://www.youtube.com/watch?v=5TnNlfPU7kY)

spec/integration/put_content/new_edition_spec.rb
pezholio: I've left the tests in the integration tests, but as a shared example for both new editions and editions with existing drafts.

content_schemas/formats/shared/default_properties/publisher.jsonnet
```diff
@@ -23,6 +23,10 @@
     format: "date-time",
     description: "Last time when the content received a major or minor update.",
   },
+  editor_id: {
+    type: "string",
+    description: "The UUID of the editor who updated the content.",
```
yndajas, line 28: **question:** intended use

The commit message now says "This takes the UUID of the user who _created_ the edition." while the description says "The UUID of the editor who _updated_ the content." Is it meant to be the creator or the last editor? I wonder if the field name could be more explicit, like before, e.g. `last_edited_by_id` or `last_edited_by_uuid`?

yndajas (commented):
Looks good, just a question about the field name/intended use, and a bit of leftover code that can be removed

content_schemas/formats/shared/definitions/publishing_api_base.jsonnet
```diff
@@ -164,6 +164,22 @@
       { type:"null" },
     ],
   },
+  author: {
```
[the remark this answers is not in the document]
pezholio: Good call!

content_schemas/formats/shared/default_properties/publisher.jsonnet
pezholio: I've gone with `last_edited_by_editor_id`, which also works well with the `last_edited_at` attribute

db/migrate/20241017080126_add_last_edited_by_editor_id_to_editions.rb
```diff
@@ -0,0 +1,7 @@
+class AddLastEditedByEditorIdToEditions < ActiveRecord::Migration[7.2]
+  def change
+    change_table :editions, bulk: true do |t|
+      t.uuid :last_edited_by_editor_id, null: true
+    end
```
yndajas, line 5: **suggestion (non-blocking):** simplification

This could probably just be an `add_column` given it's a single field

yndajas (approved):

PR #2925, opened 2024-10-18
Return `last_edited_at` in embedded content response

pezholio:
This builds on #2923 to also return the last time an edition was edited. I’ve had to change the integration tests to be a bit more verbose because of time precision shenanigans

### DIFF
--- a/spec/integration/embedded_content_spec.rb
+++ b/spec/integration/embedded_content_spec.rb
@@ -42,8 +42,10 @@ RSpec.describe "Embedded documents" do
 
   context "when an edition embeds a reference to the content block" do
     it "returns details of the edition and its publishing organisation in the results" do
+      last_edited_at = "2023-01-01T08:00:00.000Z"
       host_edition = create(:live_edition,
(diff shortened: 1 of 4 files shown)


spec/integration/embedded_content_spec.rb
```diff
@@ -44,6 +44,7 @@
     it "returns details of the edition and its publishing organisation in the results" do
       host_edition = create(:live_edition,
                             publishing_app: "whitehall",
+                            last_edited_at: 2.days.ago,
```
Harriethw, line 47: just curious, is there a reason to do 'days.ago'? if you gave it a fixed date/time would the date shenanigans be needed below, as you could just compare the exact value?

pezholio: Oooh, possibly. I'll give that a go. It's early 😅

pezholio: Done! I'm now setting the string and then parsing it, which saves a bit of messing around.

Harriethw (approved):
🕐

PR #2926, opened 2024-10-18
Allow documents links on landing pages

richardTowers:
... and also the standard set of Whitehall links.

We want landing pages to behave pretty much the same as document collections in terms of links, so matching most of those fields.

The exception to that is taxonomy_topic_email_override, which is really specific to the retirement of specialist topics, so not relevant to landing pages. We don't need that.


KludgeKML (approved):
LGTM!

PR #2928, opened 2024-10-18
Set up GraphQL

yndajas:
[Trello](https://trello.com/c/M5joCuKs/1431-set-up-graphql)

This sets up GraphQL in Publishing API, adding the dependency, running the install task, and then doing some minimal cleanup

---

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem "govuk_app_config"
 gem "govuk_document_types"
 gem "govuk_schemas"
 gem "govuk_sidekiq"
+gem "graphql"
 gem "jsonnet"
 gem "json-schema", require: false
 gem "oj"
--- a/app/graphql/publishing_api_schema.rb
+++ b/app/graphql/publishing_api_schema.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class PublishingApiSchema < GraphQL::Schema
+  query(Types::QueryType)
+
+  # For batch-loading (see https://graphql-ruby.org/dataloader/overview.html)
+  use GraphQL::Dataloader
+
+  # GraphQL-Ruby calls this when something goes wrong while running a query:
+  def self.type_error(err, context) # rubocop:disable Lint/UselessMethodDefinition
+    # if err.is_a?(GraphQL::InvalidNullError)
+    #   # report to your bug tracker here
+    #   return nil
+    # end
+    super
+  end
+
+  # Union and Interface Resolution
+  def self.resolve_type(_abstract_type, _obj, _ctx)
+    # TODO: Implement this method
+    # to return the correct GraphQL object type for `obj`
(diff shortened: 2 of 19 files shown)


app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,24 @@
...
+    field :nodes, [Types::NodeType, { null: true }], null: true, description: "Fetches a list of objects given a list of IDs." do
+      argument :ids, [ID], required: true, description: "IDs of the objects."
+    end
+
+    def nodes(ids:)
+      ids.map { |id| context.schema.object_from_id(id, context) }
+    end
+
+    # Add root-level fields here.
+    # They will be entry points for queries on your schema.
```
yndajas, line 22: This is deleted in the next PR but we could get rid of it here instead

config/routes.rb
```diff
@@ -15,6 +15,8 @@ def content_id_constraint(request)
 
     post "/lookup-by-base-path", to: "lookups#by_base_path"
 
+    post "/graphql", to: "graphql#execute"
```
yndajas, line 18: Should this be in the `v2` namespace?

app/graphql/publishing_api_schema.rb
```diff
@@ -0,0 +1,44 @@
...
+    #   # report to your bug tracker here
+    #   return nil
+    # end
+    super
+  end
+
+  # Union and Interface Resolution
+  def self.resolve_type(_abstract_type, _obj, _ctx)
+    # TODO: Implement this method
+    # to return the correct GraphQL object type for `obj`
```
yndajas, line 21: Note this outstanding TODO

Gemfile
```diff
@@ -45,3 +45,5 @@ group :development, :test do
   gem "timecop"
   gem "webmock", require: false
 end
+
+gem "graphql", "~> 2.3"
```
brucebolt, line 49: Could this be moved up into the main list further up the Gemfile.

app/graphql/publishing_api_schema.rb
brucebolt: I'm assuming this will get removed in a later PR when we put some things in this method?

app/graphql/types/query_type.rb
brucebolt: I agree that removing it here seems cleaner and will make the git history easier to follow.

config/routes.rb
brucebolt: I think it's fine to leave it outside that namesapce. `v2` was to indicate the second version of Publishing API's API, but this is the first iteration of the GraphQL API, which will remain separate from the other API.

brucebolt (approved):
All looks good to me.  Just added a few non-blocking comments.

app/graphql/publishing_api_schema.rb
yndajas: Yeah I think so!

PR #2929, opened 2024-10-18
Add GraphQL world index type and query

yndajas:
[Trello](https://trello.com/c/XRH0sRDE/1430-write-graphql-types-and-query-required-for-world-page)

This will allow us to get the data the frontend Collections app needs to render the gov.uk/world page using the GraphQL endpoint

---


app/graphql/types/query_type.rb
```diff
@@ -2,23 +2,16 @@
...
-    field :node, Types::NodeType, null: true, description: "Fetches an object given its ID." do
-      argument :id, ID, required: true, description: "ID of the object."
-    end
+    field :world_index, Types::WorldIndexType, null: false, description: "World index"
 
-    def node(id:)
-      context.schema.object_from_id(id, context)
+    def world_index
+      live_edition(content_id: "369729ba-7776-4123-96be-2e3e98e153e1")
     end
```
brucebolt, line 9: Is there a way this could be made more generic, so we don't need to hardcode content IDs for each page GraphQL will serve?

I'm thinking maybe we could have the base path as a query field, then we can grab the live edition based on the base path instead?

brucebolt: >Or just a generic edition query taking the base path?

That is more what I was thinking.  Maybe we could determine which type to use based on something we already know about that base path (e.g. the document type or schema name)?

yndajas: I've drafted a generic version here: https://github.com/alphagov/publishing-api/pull/2930

It's a little hacky under the hood, but the generic interface is probably nicer. Curious to hear your thoughts

I'll probably write some more unit tests if we do go down this avenue

brucebolt: I much prefer the approach in #2930, as it'll allow us to roll out more content types without needing to define additional query parameters (unless actually needed for filtering).

PR #2930, opened 2024-10-21
Add GraphQL world index type and query (alternative take)

yndajas:
[Trello](https://trello.com/c/XRH0sRDE/1430-write-graphql-types-and-query-required-for-world-page)

This will allow us to get the data the frontend Collections app needs to render the gov.uk/world page using the GraphQL endpoint

This is an alternative, more generic approach to #2929 (closes #2929). The more generic approach will reduce the overhead when adding further types: the main work will be defining the fields and how to retrieve them; we won't need to define an extra top-level API query interface per type

---

### DIFF
--- a/app/graphql/types/world_index_type.rb
+++ b/app/graphql/types/world_index_type.rb
@@ -2,23 +2,29 @@
 
 module Types
   class WorldIndexType < Types::EditionType
-    field :international_delegations, [WorldLocationType]
-    field :world_locations, [WorldLocationType]
+    class WorldLocation < Types::BaseObject
+      # moved this here in case we end up creating a WorldLocation < Types::EditionType type
+      field :active, Boolean, null: false
+      field :analytics_identifier, String
+      field :content_id, ID, null: false
+      field :iso2, String
+      field :name, String, null: false
+      field :slug, String, null: false
+      field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
+    end
+
+    field :body, String
+    field :international_delegations, [WorldLocation], null: false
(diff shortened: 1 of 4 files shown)


app/graphql/types/world_index_type.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def world_locations
+      simplify_world_locations(object.details[:world_locations])
+    end
+
+  private
+
+    def simplify_world_locations(world_locations)
+      world_locations.map { |world_location| world_location.slice(:active, :name, :slug) }
+    end
```
brucebolt, line 22: Do we need to do this step?  The current content item includes all the fields stored in the `details` for both world locations and international delegations, not just these three.

app/graphql/types/world_location_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Types
+  class WorldLocationType < GraphQL::Schema::Object
+    field :active, Boolean
+    field :name, String
+    field :slug, String
```
brucebolt, line 7: I think there's some fields missing here (`iso2`, `content_id`, `updated_at`, `analytics_id`). I'm not sure if they're used by the frontend apps at the moment, but they're definitely returned by the Content Store API, so we should probably still include them.

It does bring me back to the bigger question of whether we should be dynamically generating these from the schemas (otherwise the two things could end up out of sync), but that's definitely outside the scope of these proof of concept tickets!

brucebolt (commented):
This looks great!

I've just left a couple of questions about the fields we're returning, but otherwise should be good to go.

PR #2942, opened 2024-10-25
Add GraphQL world index types and query

Gweaton:
[Trello card](https://trello.com/c/XRH0sRDE/1430-write-graphql-types-and-query-required-for-world-page)

Closes #2929, closes #2930 

This is a more stripped-down-version of #2930, which added a few extra fields to the EditionType. As there's a fair bit more thought required for adding those fields, I felt it would help us get things unblocked slightly if we leave those extra fields for now and add them in future branches.

These new types will allow us to get the data the frontend Collections app needs to render the gov.uk/world page using the GraphQL endpoint.

As mentioned, the `EditionType` here is very minimal, to make this change as small as possible. We'll be adding additional fields to this in upcoming branches, but these extra fields aren't really necessary for the majority of the `/world` page in of itself, and there's a fair bit of complexity involved in some of the Edition fields e.g. Links that needs further thought.

### DIFF
--- a/app/graphql/types/world_index_type.rb
+++ b/app/graphql/types/world_index_type.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Types
+  class WorldIndexType < Types::EditionType
+    class WorldLocation < Types::BaseObject
+      field :active, Boolean, null: false
+      field :analytics_identifier, String
+      field :content_id, ID, null: false
+      field :iso2, String
+      field :name, String, null: false
+      field :slug, String, null: false
+      field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
+    end
+
+    field :body, String
+    field :international_delegations, [WorldLocation], null: false
+    field :world_locations, [WorldLocation], null: false
+
+    def self.base_path = "/world"
+
+    def international_delegations
+      object.details[:international_delegations]
+    end
+
+    def world_locations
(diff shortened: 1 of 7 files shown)


app/graphql/types/world_index_type.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Types
+  class WorldIndexType < Types::EditionType
+    class WorldLocation < Types::BaseObject
```
brucebolt, line 5: Will it be easy to move this type out of here and into it's own type file, if we decide to migrate world location pages to use GraphQL?

brucebolt (approved):
Looks good to me.  This seems to include (as far as I can see) all the same information as the existing content item, so should be fine for rendering this page.

brucebolt:
When you do merge this, could you remove the "(alternative, other take)" from the PR title first, just so it doesn't end up in the commit history and cause confusion in the future!

app/graphql/types/world_index_type.rb
Gweaton: This is actually a different kind of world location, just used for the world index page, so we're putting it in here to avoid a naming clash if we do create WorldLocation edition types.

brucebolt: Cool, that makes sense.  Thanks for explaining.

PR #2949, opened 2024-10-29
Add additional GraphQL Edition Type fields

Gweaton:
[Trello card](https://trello.com/c/e1Upvu9L/1441-add-remaining-graphql-edition-fields)

## Changes in this PR

Expands the edition type to include all the top-level properties that a content item contains, and most of the nested data. The queries sent by frontend apps will be able to pare down the data that gets sent to them from the Publishing API, but we are able to get almost all the same data as presented by the current content item if required.

The `withdrawn_notice` field has been implemented slightly differently here to how it is usually stored on a content item, in that it's now either a full object or `null` instead of the previous implementation where it was a full object or an empty object (`{}`) if one wasn't present. We've written in this way to comply with the recommended way of writing GraphQL schemas (see commit message for more details), but it will mean consumers may have to update the way they access this data.

#### Details

Currently a generic JSON type as it's an inherently flexible object that could have a wide range of fields.

It may be worth defining more specific types and moving the `details` keys up to the top level on each type for each page being built in future. This would allow for reduced complexity in querying the required fields.

We've used JSON to keep it flexible and to allow consumers to fall back on these fields if they need to.

#### Links

Needs further work to query and present the data that is currently presented in the content item so we've left the field off for now, rather than adding a non-so-useful placeholder. We'll need to add links when we add pages that use links like the ministers index page. It would be good to figure out how we can reduce the need for full link expansion and move to a more minimal on-demand version when we tackle this piece of work.

#### `publishing_scheduled_at` and `scheduled_publishing_delay_seconds`

These are not currently available in Publishing API. They are generated in Content Store, so we just return `null` here. If we do need to serve these in future, we'd need to find a way of storing these in Publishing API, but that's definitely out of scope for now.

### DIFF
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -2,6 +2,33 @@
 
 module Types
   class EditionType < Types::BaseObject
-    field :title, String
+    field :analytics_identifier, String
+    field :base_path, String
+    field :content_id, ID
+    field :description, String
+    field :details, GraphQL::Types::JSON, null: false
+    field :document_type, String
+    field :first_published_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :locale, String, null: false
+    field :phase, String, null: false
+    field :public_updated_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :publishing_app, String
+    field :publishing_request_id, String
+    field :publishing_scheduled_at, GraphQL::Types::ISO8601DateTime
+    field :rendering_app, String
+    field :scheduled_publishing_delay_seconds, Int
+    field :schema_name, String
+    field :title, String, null: false
+    field :updated_at, GraphQL::Types::ISO8601DateTime
+    field :withdrawn_notice, String
+
+    # Aliased by field methods for fields that are currently presented in the
+    # content item, but come from Content Store, so we can't provide them here
+    def not_stored_in_publishing_api
+      nil
+    end
+
+    alias_method :publishing_scheduled_at, :not_stored_in_publishing_api
+    alias_method :scheduled_publishing_delay_seconds, :not_stored_in_publishing_api
   end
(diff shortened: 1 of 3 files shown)


app/graphql/types/edition_type.rb
```diff
@@ -2,6 +2,35 @@
...
-    field :title, String
+    field :analytics_identifier, String
+    field :base_path, String
+    field :content_id, ID
+    field :description, String
+    field :details, GraphQL::Types::JSON, null: false
+    field :document_type, String
+    field :first_published_at, GraphQL::Types::ISO8601DateTime, null: false
+    # TODO: work out how to retrieve the links returned in the content item
+    field :links, [GraphQL::Types::JSON], null: false
```
brucebolt, line 18: Perhaps we should just remove links for now, as they aren't yet properly supported by this endpoint?

We may choose to present links in a completely different way once we get to other pages (e.g. the ministers index page) so a placeholder may not immediately be useful.

brucebolt (approved):
> It may be worth defining more specific types and moving the details keys up to the top level on each type for each page being built in future. This would allow for reduced complexity in querying the required fields.
> 
> We've used JSON to keep it flexible and to allow consumers to fall back on these fields if they need to.

Agreed.  We'll probably stumble upon the need for details to be moved to the top-level as we work on other pages.

>Needs further work to query and present the data that is currently presented in the content item. We've defaulted to the GraphQL JSON type again here, though it will always return an empty array until we query the relevant tables and join the data. We can expand on this in future work when we do need to render linked content. We might need to figure out if/how we can do this without relying on Link Expansion.

I'm in agreement that it makes sense to leave the links until later.  One of the aims of this work was the remove the need for rather blunt link expansion in it's current form and allow the requesting app to specify which fields it'd link expanded in the query.

PR #2950, opened 2024-10-30
Add pact test example with title

brucebolt:
This adds an example document for pact tests that has a title. This will be used in the GraphQL pact tests so we have some content to return in the response.

[Trello card](https://trello.com/c/jDF1nSQ0)

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -681,5 +681,16 @@ Pact.provider_states_for "GDS API Adapters" do
         target_content_id: primary_publishing_organisation.content_id,
       )
     end
+
+    provider_state "a published content item exists with content_id: 19ad249e-7ac4-4aa4-8ab4-b6c5f381c043" do
+      set_up do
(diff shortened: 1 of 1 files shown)


yndajas (approved):

spec/service_consumers/pact_helper.rb
```diff
@@ -681,5 +681,16 @@ def url_encode(str)
         target_content_id: primary_publishing_organisation.content_id,
       )
     end
+
+    provider_state "a published content item exists with content_id: 19ad249e-7ac4-4aa4-8ab4-b6c5f381c043" do
```
yndajas, line 685: **question:** pertinent field

Is the content ID what we'll care about wherever we need to use this, or will it be the base path?

yndajas: We do have [other examples that use base paths in their name](https://github.com/alphagov/publishing-api/pull/2950/files#diff-7d17ffbce13a8dc8820ff122a86f9a63284e76991d64a9981a7eb7c5697f6bfcR620-R638). I guess I was thinking this could be "a published content item exists with base_path /my-document"

PR #2957, opened 2024-10-31
(656) Get and cache pageviews for embedded content

pezholio:
~~***Note This PR is not ready for merge until the relevant environment vars have been added***~~

In content modelling, we have the requirement to return the number of pageviews that embedded content has had in the last 30 days. This allows editors to see at a glance how well used and how popular content that uses their block is, which will allow them to make a judgment call as to how impactful changes they make to a block will be.

The information we need is contained in a Google BigQuery database managed by the data team. I've added a service that makes the relevant query via the BigQuery API and returns the data. 

Rather than having to make these queries each time a request is made to the dependent content endpoint (and to allow sorting etc at the database level) I've set up a job that runs on a nightly basis to get all content that has an embed dependency and store the results in a `StatisticsCache` database table.

We can then add the relevant joins to the database query and return it in the body of the API response.

Going forward, we'd also like to call a background job to query the pageviews of a page when an embed is added too (rather than waiting until the nightly job runs), but I wanted to check the approach was sound before overcomplicating things even more!

Additionally, I've made some tweaks to the way the queries work to make the tests a little less verbose and clean up some of the logic.


app/controllers/v2/content_items_controller.rb
```diff
@@ -22,7 +22,7 @@ def linkables
     end
 
     def embedded
-      results = Queries::GetEmbeddedContent.new(
+      results = GetEmbeddedContentService.new(
```
Harriethw, line 25: definitely a good improvement to have a service etc. but the name still confuses me... would `GetHostContentService` be more accurate now?

app/services/get_embedded_content_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def call
+    if Document.find_by(content_id: target_content_id).nil?
+      message = "Could not find an edition to get embedded content for"
+      raise CommandError.new(code: 404, message:)
+    end
+
+    Presenters::EmbeddedContentPresenter.present(
+      target_content_id,
+      host_editions,
```
Harriethw, line 16: just a minor suggestion but I guess this isn't technically `editions` , but a specific type of Data object that has both Document and Edition fields, maybe should say `host_content`?

app/sidekiq/update_statistics_caches_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      editions = slice.map(&:edition)
+      page_views_for_slice = page_views(editions.map(&:base_path))
+      statistics_caches = page_views_for_slice.map do |result|
+        edition = editions.find { |e| e.base_path == result.path }
+        {
+          document_id: edition.document.id,
+          unique_pageviews: result.page_views,
+        }
+      end
+      StatisticsCache.upsert_all(statistics_caches, unique_by: [:document_id])
```
Harriethw, line 17: ooh cool I didn't know about `upsert_all`

spec/queries/get_embedded_content_spec.rb
```diff
@@ -96,51 +70,26 @@
...
-
-        allow(presenter_double).to receive(:present).and_return({}.to_json)
-
-        described_class.new(target_content_id).call
       end
     end
 
     context "when there are superceded editions that embed the target content" do
-      it "does not pass them to the presenter" do
+      it "does not return the superceded editions" do
```
Gweaton, line 92: **nit**: super[s]eded (sorry...)

Gweaton (approved):
I think this makes sense to me, nice work! Just have a couple of questions

app/sidekiq/update_statistics_caches_job.rb
```diff
@@ -0,0 +1,30 @@
+class UpdateStatisticsCachesJob
+  include Sidekiq::Job
+
+  BATCH_SIZE = 50
+
+  def perform
+    links.each_slice(BATCH_SIZE) do |slice|
```
Harriethw, line 7: would it be worth adding some logs here? I think otherwise jobs happen really silently and it's hard to follow them

pezholio: Good plan!

app/queries/get_embedded_content.rb
```diff
@@ -1,47 +1,82 @@
 module Queries
   class GetEmbeddedContent
-    attr_reader :target_content_id, :states
+    Result = Data.define(
```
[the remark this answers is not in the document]
pezholio: Yeah, I'm a big fan of Data classes! Coming back to Ruby after a couple of years on strongly-typed languages, passing round hashes kinda gave me the ick

app/controllers/v2/content_items_controller.rb
pezholio: Good plan 👍

app/presenters/embedded_content_presenter.rb
```diff
@@ -0,0 +1,45 @@
+module Presenters
+  class EmbeddedContentPresenter
+    def self.present(target_edition_id, host_editions)
+      new(target_edition_id, host_editions).present
+    end
+
+    def initialize(target_edition_id, host_editions)
+      self.target_edition_id = target_edition_id
```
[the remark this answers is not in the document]
pezholio: Nah, you're right, I'm just following existing patterns. I'd prefer an instance variable myself, so will fix

app/queries/get_embedded_content.rb
```diff
@@ -1,47 +1,82 @@
...
-          "statistics_caches.unique_pageviews AS unique_pageviews",
-        )
-    end
-
     def embedded_link_type
       "embed"
     end
 
-    attr_writer :target_content_id, :states
+    def binds
```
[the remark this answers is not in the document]
pezholio: I've refactored it a bit to be a bit clearer about what's going on. Documentation on this is sparse though, and I've only really seen it referenced in StackOverflow posts

pezholio:
I've added some safeguards to check the env vars are present before calling BigQuery (We'll need this in dev too), so I'm going to merge this.

PR #2960, opened 2024-11-01
(656) Update pageviews when publishing content with embeds

pezholio:
This follows on from #2957 to ensure the statistics for a content item that uses embeds get updated on publish. This is particularly useful when a page uses embeds for the first time, so there's minimal delay between the block being used and the page views showing in the Content Block Manager.


Harriethw (approved):
LGTM!

PR #2971, opened 2024-11-08
Add GraphQL roles prime minister page

mike3985:
https://trello.com/c/8e6tlsKw/1445-prime-minister-page-define-revise-graphql-types-add-type-to-editiontypeorsubtype

Most of the action is in the "Introduce RoleType for GraphQL prime-minister page" commit. If it's proving difficult to review, we can always add the link fields (and their associated type definitions) incrementally over a few commits. Let me know!

### DIFF
--- a/spec/graphql/types/query_type/edition_type_or_subtype_spec.rb
+++ b/spec/graphql/types/query_type/edition_type_or_subtype_spec.rb
@@ -3,9 +3,11 @@ RSpec.describe "Types::QueryType::EditionTypeOrSubtype" do
     it "includes all EditionType descendants" do
       Rails.application.eager_load!
 
-      expected = [Types::EditionType, *Types::EditionType.descendants]
+      expected = [Types::EditionType, *Types::EditionType.descendants].map(&:name)
+      actual = Types::QueryType::EditionTypeOrSubtype::EDITION_TYPES.map(&:name)
+      diff = expected - actual
 
-      expect(Types::QueryType::EditionTypeOrSubtype::EDITION_TYPES.sort).to eq(expected.sort)
+      expect(diff).to be_empty
     end
   end
 
--- a/spec/integration/graphql_spec.rb
+++ b/spec/integration/graphql_spec.rb
@@ -236,4 +236,216 @@ RSpec.describe "GraphQL" do
       expect(response.body).to eq(expected)
     end
   end
+
+  describe "roles" do
+    before do
+      role = create(
+        :live_edition,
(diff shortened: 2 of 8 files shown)


spec/integration/graphql_spec.rb
```diff
@@ -236,4 +236,217 @@
...
+      )
+    end
+
+    it "exposes Role-specific fields, generic Edition fields and Role's links" do
+      post "/graphql", params: {
+        query:
+          "{
+            edition(basePath: \"/government/ministers/prime-minister\") {
+              ... on Role {
+                basePath
```
mike3985, line 356: It turns out that we can easily make our fields snake-case... I think this would be nicer for our Rails-based consumers, but I don't know what the implications are for mobile apps.

Does anybody have an opinion about this?

spec/graphql/types/query_type/edition_type_or_subtype_spec.rb
```diff
@@ -3,9 +3,10 @@
     it "includes all EditionType descendants" do
       Rails.application.eager_load!
 
-      expected = [Types::EditionType, *Types::EditionType.descendants]
+      expected = [Types::EditionType, *Types::EditionType.descendants].map(&:name)
+      actual = Types::QueryType::EditionTypeOrSubtype::EDITION_TYPES.map(&:name)
 
-      expect(Types::QueryType::EditionTypeOrSubtype::EDITION_TYPES.sort).to eq(expected.sort)
+      expect(actual.sort).to eq(expected.sort)
```
JonathanHallam, line 9: I guess we could get rid of the sorting by doing something like 
```
expect(expected-actual).to be_empty
```

spec/integration/graphql_spec.rb
JonathanHallam: I mean i guess we're largely building this for ourselves, I'd say if it's easy enough we do it

JonathanHallam (approved):
I'm happy with this. I think you're definitely right in that the queries will need to be optimised but we can leave that for a later PR, thanks :)

spec/integration/graphql_spec.rb
```diff
@@ -236,4 +236,217 @@
...
+                  basePath
+                  title
+                }
+              }
+            }
+          }
+
+          fragment baseRoleAppointmentInfo on RoleAppointment {
+            endedOn
+            startedOn
```
yndajas, line 388: **question:** fragment

Do we need these for the `currentRoleAppointment`?

PR #2972, opened 2024-11-08
add `instructions_to_publishers` to schemas

Harriethw:
We want to add a field to all Content Blocks that
will be a short string containing an internal-only message.

This adds the field `instructions_for_publishers`
to a shared default schema file, and then sets it
as optional on content blocks.

Although this is a Content Block specific field,
it did not make sense to include it in the
`details` object because we are using that schema
for all custom public-facing fields on a block. We have already set precedent for adding Content
Block specific top level fields with
`content_id_alias`.

---


pezholio (approved):
Yep, I think this is the way to go - LGTM 👍

PR #2974, opened 2024-11-11
(657) Move rendering logic to external gem

pezholio:


Harriethw (approved):
Looks like you've captured everything!

callumknights (approved):
I like this idea, and I think it will make content-blocks a lot easier to understand in the long run. Nice work!

PR #2975, opened 2024-11-11
(678) Allow sorting and pagination on dependent content

pezholio:
This adds an `order` argument to sort dependent content by various fields, as well as pagination. 

We use the `order` query string param, prefixed with `-` for descending order queries, as this matches how other endpoints handle sorting

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -25,7 +25,7 @@ module V2
       results = GetHostContentService.new(
         path_params[:content_id],
         query_params[:order],
-        query_params[:page]
+        query_params[:page],
       ).call
 
       render json: results
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -25,6 +25,7 @@ module V2
       results = GetHostContentService.new(
         path_params[:content_id],
         query_params[:order],
+        query_params[:page]
       ).call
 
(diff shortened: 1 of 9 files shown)


app/services/get_host_content_service.rb
```diff
@@ -19,9 +18,33 @@ def call
 
 private
 
-  attr_writer :target_content_id
+  attr_accessor :target_content_id, :order
 
   def host_content
-    @host_content ||= Queries::GetEmbeddedContent.new(target_content_id).call
+    @host_content ||= Queries::GetEmbeddedContent.new(target_content_id, order_field:, order_direction:).call
```
Harriethw: maybe this is for another PR, but should we get around to renaming `GetEmbeddedContent` to `GetHostContent` as I think that's what it's doing?

app/controllers/v2/content_items_controller.rb
```diff
@@ -24,6 +24,7 @@ def linkables
     def embedded
```
Harriethw, line 24: another rename here, maybe it would be bigger change than I thought if we'd have to rename the endpoint 😬

Harriethw (approved):
Makes sense to me! nice tidying up of the SQL too.

app/services/get_host_content_service.rb
pezholio: Yeah, good call - I think I'll leave it for now though

PR #2989, opened 2024-11-19
(679) Get the number of instances on a page

pezholio:
This allows us to identify the amount of times an instance is used within a document. We've had to make some tweaks to ensure any duplicate links are not removed, so we are able to count the amount of times the link is used, but this seems to work without any issue.

### DIFF
--- a/app/controllers/v2/content_items_controller.rb
+++ b/app/controllers/v2/content_items_controller.rb
@@ -22,7 +22,7 @@ module V2
     end
 
     def embedded
-      results = GetHostContentService.new(
+      results = GetEmbeddedContentService.new(
         path_params[:content_id],
         query_params[:order],
         query_params[:page],
--- a/app/services/embedded_content_finder_service.rb
+++ b/app/services/embedded_content_finder_service.rb
@@ -8,20 +8,20 @@ class EmbeddedContentFinderService
   def fetch_linked_content_ids(details, locale)
     content_references = details.values.map { |value|
       find_content_references(value)
-    }.flatten.compact.uniq
+    }.flatten.compact
 
-    check_all_references_exist(content_references, locale)
+    check_all_references_exist(content_references.uniq, locale)
     content_references.map(&:content_id)
   end
 
   def find_content_references(value)
     case value
     when Array
-      value.map { |item| find_content_references(item) }.flatten
+      value.map { |item| find_content_references(item) }.flatten.uniq
     when Hash
-      value.map { |_, v| find_content_references(v) }.flatten
+      value.map { |_, v| find_content_references(v) }.flatten.uniq
     when String
-      value.scan(EMBED_REGEX).map { |match| ContentReference.new(document_type: match[1], content_id: match[2], embed_code: match[0]) }.uniq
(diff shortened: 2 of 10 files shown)


app/services/embedded_content_finder_service.rb
```diff
@@ -8,20 +8,20 @@ class EmbeddedContentFinderService
...
-    check_all_references_exist(content_references, locale)
+    check_all_references_exist(content_references.uniq, locale)
     content_references.map(&:content_id)
   end
 
   def find_content_references(value)
     case value
     when Array
-      value.map { |item| find_content_references(item) }.flatten
+      value.map { |item| find_content_references(item) }.flatten.uniq
```
Harriethw, line 20: is `find_content_references` basically recursively called until the `value` becomes a string? what's the impact of adding `uniq` on the end for arrays and hashes?

app/services/embedded_content_finder_service.rb
```diff
@@ -8,20 +8,20 @@ class EmbeddedContentFinderService
...
     when Hash
-      value.map { |_, v| find_content_references(v) }.flatten
+      value.map { |_, v| find_content_references(v) }.flatten.uniq
     when String
-      value.scan(EMBED_REGEX).map { |match| ContentReference.new(document_type: match[1], content_id: match[2], embed_code: match[0]) }.uniq
+      value.scan(EMBED_REGEX).map { |match| ContentReference.new(document_type: match[1], content_id: match[2], embed_code: match[0]) }
```
Harriethw, line 24: I guess there might not be an easier way of getting the count, and maybe not an issue now, but will having duplicate Links mean adding more links to iterate through when dependencies get updated? so if there are 10s of blocks in a host document, that could trigger 100s more calls and jobs throughout the host documents lifetime?

app/services/embedded_content_finder_service.rb
pezholio: So, this is mainly for multipart content, when there's a html and govspeak part. We don't want to class these as two instances

app/services/embedded_content_finder_service.rb
pezholio: Good call, but it looks like DependencyResolution will return an array of `content_id`s, which will then get uniqued out when we query (see https://github.com/alphagov/publishing-api/blob/4a679f153c286c248aa3f4fbf81b0a67278cede7/app/sidekiq/dependency_resolution_job.rb#L4) so I _think_ we're fine

Harriethw (approved):
Looks great - just a personal quibble with the renaming but I don't feel (too) strongly about it :)

app/controllers/v2/content_items_controller.rb
```diff
@@ -22,7 +22,7 @@ def linkables
     end
 
     def embedded
-      results = GetHostContentService.new(
```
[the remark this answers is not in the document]
Harriethw: oh I see the presenter and query use `GetEmbedded`, I thought we'd changed them 😅  - I think it would be worth doing a rename because we do need code to actually get embedded content versus getting host documents  (e.g. `EmbeddedContentFinderService`) and it could get confusing quickly..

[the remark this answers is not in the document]
pezholio: PR to do the rename is here https://github.com/alphagov/publishing-api/pull/2991

PR #2991, opened 2024-11-20
Rename `embedded` endpoint to `host_content`

pezholio:
This has always been a bit misleaseading/confusing, so let’s rename the endpoint and all associated classes / methods.

I’ve also added a route to point the old endpoint to the renamed controller method, which can be removed once we  update `gds-api-adapters`


Harriethw (approved):
nice, this is good to see 🧹

PR #2994, opened 2024-11-21
Add asset_manager_id to file attachments 

richardTowers:
Currently, the properties in the file attachment properties are insufficient to make a request to asset manager from the server to fetch the asset.

The API client wants to be called like this:

    GdsApi.asset_manager.media(asset_maanger_id, filename)

... but we don't have an asset_manager_id, we only have url (which include the id we want), and a bunch of other fields which don't. We shouldn't have to parse URLs to find IDs.

We get away with this because in most cases where we're using file assets, we redirect the user to the full asset URL (so we don't need to fetch it on the server side).

The one example I'm aware of where we do need to request assets on the server side is CSV previews, and we've ended up with a pretty wild workaround in that case. Specifically, we route some requests to the assets domains (e.g. assets.publishing.service.gov.uk) to frontend.  This way, the preview_url ends up being served by frontend itself. This means the URL gets parsed by Rails' routing logic, and we end up with the ID in a param which we can use to call asset manager: see https://github.com/alphagov/frontend/blob/8b075a3e26d775ad97b6b473ddaa22386e07cad2/app/controllers/csv_preview_controller.rb#L27

We'll need a corresponding commit in Whitehall (and maybe other publishing apps, if they have attachments?) to start sending this new field through to the API. It's optional, so this shouldn't break anything in the meantime.

EDIT: This description has been updated by @unoduetre to not include details which are no longer relevant.

### DIFF
--- a/content_schemas/dist/formats/call_for_evidence/frontend/schema.json
+++ b/content_schemas/dist/formats/call_for_evidence/frontend/schema.json
@@ -773,6 +773,9 @@
             "alternative_format_contact_email": {
               "type": "string"
             },
+            "asset_manager_id": {
+              "type": "string"
+            },
             "attachment_type": {
               "type": "string",
               "enum": [
--- a/content_schemas/dist/formats/call_for_evidence/notification/schema.json
+++ b/content_schemas/dist/formats/call_for_evidence/notification/schema.json
@@ -897,6 +897,9 @@
             "alternative_format_contact_email": {
               "type": "string"
             },
+            "asset_manager_id": {
+              "type": "string"
+            },
             "attachment_type": {
               "type": "string",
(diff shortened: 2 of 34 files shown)


catalinailie (approved):

brucebolt (commented):
Does this need to be in 3 commits, since the changes in the first two commits are overwritten in the third?

My recommendation would be to make the full change to `content_schemas/formats/shared/definitions/asset_links.jsonnet` in one commit, then regenerate the schemas using the rake task and commit this separately.

catalinailie:
Good point @brucebolt! I updated the commits to be tidier.

brucebolt (approved):

PR #3000, opened 2024-11-26
Rename date facet

GDSNewt:
Renames a facet in the new SFO Finder specialist finder.

[Previous PR](https://github.com/alphagov/publishing-api/pull/2968)
[Trello](https://trello.com/c/E7KMeGZo/3034-specialist-finder-sfo-cases)
[Zendesk](https://govuk.zendesk.com/agent/tickets/5958029)

PR #3001, opened 2024-11-26
Revert changes to test content schemas

GDSNewt:
We changed these in order to allow merging some changes in publishing api and specialist publisher. We are now reverting to main.

[PR that introduced the change](https://github.com/alphagov/publishing-api/pull/3000)


lauraghiorghisor-tw (approved):
LGTM 🤝

PR #3004, opened 2024-11-27
(731) Ensure only one Organisation is returned when fetching host content.

pezholio:
When fetching host content for an organisation which has a translation, multiple rows are returned, one per translation. The updates the join statement, so we only fetch organisation documents in the `en` locale.

Trello: https://trello.com/c/QgA7LneE/731-investigate-html-documents-coming-up-twice-for-location


Harriethw (approved):
nice!

PR #3005, opened 2024-11-27
(700) Return rollup data in `host-content` endpoint

pezholio:
Trello card: https://trello.com/c/eJTfFnFN/700-create-rollup-for-view-page

This updates the `host-content` endpoint to return a "rollup" of summed / counted data. The fields are:

* Views - the sum total of all pageviews for host documents
* Locations - the total count of all host documents
* Instances - the sum total of all the times the block has been used across all host documents
* Organisations - the number of unique organisations using the content block

### DIFF
--- a/app/queries/get_host_content.rb
+++ b/app/queries/get_host_content.rb
@@ -29,17 +29,17 @@ module Queries
     }.freeze
 
     FIELDS = [
-      { field: TABLES[:editions][:id], included_in_group?: true },
-      { field: TABLES[:editions][:title], included_in_group?: true },
-      { field: TABLES[:editions][:base_path], included_in_group?: true },
-      { field: TABLES[:editions][:document_type], included_in_group?: true },
-      { field: TABLES[:editions][:publishing_app], included_in_group?: true },
-      { field: TABLES[:editions][:last_edited_by_editor_id], included_in_group?: true },
-      { field: TABLES[:editions][:last_edited_at], included_in_group?: true },
+      { field: TABLES[:editions][:id], alias: "id", included_in_group?: true },
+      { field: TABLES[:editions][:title], alias: "title", included_in_group?: true },
+      { field: TABLES[:editions][:base_path], alias: "base_path", included_in_group?: true },
+      { field: TABLES[:editions][:document_type], alias: "document_type", included_in_group?: true },
+      { field: TABLES[:editions][:publishing_app], alias: "publishing_app", included_in_group?: true },
+      { field: TABLES[:editions][:last_edited_by_editor_id], alias: "last_edited_by_editor_id", included_in_group?: true },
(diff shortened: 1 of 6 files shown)


app/queries/get_host_content.rb
```diff
@@ -16,6 +16,13 @@ class GetHostContent
       :instances,
     )
 
+    Rollup = Data.define(
+      :views,
+      :locations,
+      :content_blocks,
```
Harriethw, line 22: should we keep it as `instances` if that's the term on the frontend as well?

app/queries/get_host_content.rb
```diff
@@ -29,17 +29,17 @@ class GetHostContent
...
+      { field: TABLES[:editions][:publishing_app], alias: "publishing_app", included_in_group?: true },
+      { field: TABLES[:editions][:last_edited_by_editor_id], alias: "last_edited_by_editor_id", included_in_group?: true },
+      { field: TABLES[:editions][:last_edited_at], alias: "last_edited_at", included_in_group?: true },
       { field: TABLES[:primary_links][:target_content_id], alias: "primary_publishing_organisation_content_id", included_in_group?: true },
       { field: TABLES[:org_editions][:title], alias: "primary_publishing_organisation_title", included_in_group?: true },
       { field: TABLES[:org_editions][:base_path], alias: "primary_publishing_organisation_base_path", included_in_group?: true },
-      { field: TABLES[:statistics_caches][:unique_pageviews], included_in_group?: true },
+      { field: TABLES[:statistics_caches][:unique_pageviews], alias: "unique_pageviews", included_in_group?: true },
       { field: TABLES[:documents][:content_id], alias: "host_content_id", included_in_group?: true },
       { field: TABLES[:editions][:id].count, alias: "instances", included_in_group?: false },
```
Harriethw, line 51: struggling to follow the whole query now, but is this counting the number of content blocks instead of the `instances` (if that means the amount of appearances in a host)? Shouldn't it be counting the links of type embed on a host (I think we create a new link every time a block is embedded even if it already exists)?

app/queries/get_host_content.rb
pezholio: Done! :

app/queries/get_host_content.rb
Harriethw: ah I see it's covered by a test so must be working :)

pezholio: Yeah, this query returns a row for each link initially, but we get rid of any duplicates with the call to `.group` here https://github.com/alphagov/publishing-api/blob/cd9ae9da86946a5a4933bb27563b0cfbe0eb8077/app/queries/get_host_content.rb#L155

Harriethw (approved):

PR #3008, opened 2024-11-28
Add support for news articles to GraphQL endpoint

brucebolt:
This allows us to retrieve news articles using a GraphQL query.

[Trello card](https://trello.com/c/RMkMvY8b)

### DIFF
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -43,6 +43,22 @@ module Types
     alias_method :publishing_scheduled_at, :not_stored_in_publishing_api
     alias_method :scheduled_publishing_delay_seconds, :not_stored_in_publishing_api
 
+    class Translation < Types::BaseObject
+      field :locale, String
+      field :base_path, String
+    end
+
+    class EditionLinks < Types::BaseObject
+      field :available_translations, [Translation]
+
+      def available_translations
+        Presenters::Queries::AvailableTranslations.by_edition(object)
+          .translations.fetch(:available_translations, [])
+      end
+    end
+
+    field :links, EditionLinks, method: :itself
+
   private
 
     def presented_edition
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -48,8 +48,65 @@ module Types
       field :base_path, String
     end
 
+    class GovernmentDetails < Types::BaseObject
+      field :current, Boolean
+    end
+
+    class GovernmentLink < Types::BaseObject
+      field :base_path, String
+      field :content_id, String
+      field :details, GovernmentDetails
+      field :title, String
+    end
+
+    class OrganisationLink < Types::BaseObject
+      field :base_path, String
+      field :content_id, String
+      field :title, String
+    end
+
+    class PersonLink < Types::BaseObject
+      field :base_path, String
+      field :content_id, String
+      field :title, String
+    end
+
+    class Taxon < Types::BaseObject
+      field :base_path, String
+      field :content_id, String
+      field :document_type, String
+      field :phase, String
+      field :title, String
+    end
+
+    class TaxonLink < Taxon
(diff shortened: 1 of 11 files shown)


app/graphql/types/edition_type.rb
```diff
@@ -43,6 +43,22 @@ def not_stored_in_publishing_api
...
+      field :base_path, String
+    end
+
+    class EditionLinks < Types::BaseObject
+      field :available_translations, [Translation]
+
+      def available_translations
+        Presenters::Queries::AvailableTranslations.by_edition(object)
+          .translations.fetch(:available_translations, [])
+      end
```
yndajas: **thought:** genericness

Is this link type available/present on all editions/types? If not, I've wondered before about a mixin approach for shared links that aren't quite generic enough to appear everywhere (e.g. `include AvailableTranslationsLinks` in the relevant type classes). This feels related to [the conversation over on the other PR](https://github.com/alphagov/publishing-api/pull/3024#discussion_r1871426511)!

Similar thoughts on the next commit - where/if to draw the line between generic and specific/how precise to be with our typing

spec/integration/graphql/news_article_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+    it "exposes news article specific fields in addition to generic edition fields" do
+      post "/graphql", params: {
+        query: <<~QUERY,
+          {
+              edition(
+                basePath: "/government/news/announcement",
+                contentStore: "live",
+              ) {
+                ... on Edition {
+                }
```
yndajas, line 90: **nitpick:** redundant code

I don't think this is necessary?

spec/integration/graphql/news_article_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+                      title
+                    }
+                    worldLocations {
+                      basePath
+                      title
+                    }
+                  }
+                  basePath
+                  title
+                  details
```
yndajas, line 134: **nitpick:** line order

When reviewing/outside of a code editor, I've lost a bit of context on the nesting here. Might be useful to move these simple fields to the top, before the complicated links field block

app/graphql/types/query_type.rb
```diff
@@ -4,10 +4,11 @@ module Types
   class QueryType < Types::BaseObject
     field :edition, EditionTypeOrSubtype, description: "An edition or one of its subtypes" do
       argument :base_path, String
+      argument :content_store, String, required: false, default_value: "live"
```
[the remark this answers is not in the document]
mike3985: Well, we're gonna have to face this decision at some point! I wonder if there's already a Content Store-free name for this concept in use?

spec/graphql/types/query_type_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    context "when there is a published edition and a newer draft edition" do
+      let(:document) { create(:document) }
+      let(:base_path) { "/foo" }
+      let(:live_edition) { create(:live_edition, document:, base_path:, user_facing_version: 1) }
+      let(:draft_edition) { create(:draft_edition, document:, base_path:, user_facing_version: 2) }
+
+      before do
+        live_edition.touch
+        draft_edition.touch
+      end
```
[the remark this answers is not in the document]
mike3985: If I've read this correctly, the `#touch` updates the record's `updated_at` in the database. Which, no, doesn't appear to be relevant to what's being tested.

This
```ruby
before do
  live_edition
  draft_edition
end
```
would be equivalent to using `let!`, right?

I'll replace it with a couple of `let!`s. I'm not a big `let!` fan personally, but it is already widely used in this test suite ✅

mike3985 (approved):
@yndajas, I've made the changes that you suggested and left your other comments unresolved. (I've linked to this pull request in the notes for our mobbing session, mentioning the relevance of the discussion here.)

PR #3009, opened 2024-11-28
Graphqulify ministers index

mike3985:
https://trello.com/c/PlR5Ken4/1452-querywinkle-the-ministers-index-page


app/graphql/types/ministers_index_type.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+
+module Types
+  class MinistersIndexType < Types::EditionType
```
brucebolt, line 4: There's a lot of stuff in this one class.  I think that's fine for now, but we'll want to remember to DRY-up some of this code as we introduce further document types that may need the same types.

brucebolt (approved):

PR #3049, opened 2024-12-17
Switch off auto-camelization for GraphQL fields

brucebolt:
GraphQL Ruby automatically [camelises field names](https://graphql-ruby.org/fields/introduction.html#field-names). We would like to switch this off and always use snake casing to match the field names currently present in content items.

Therefore applying this as a default to all our fields.

[Trello card](https://trello.com/c/RMkMvY8b)

### DIFF
--- a/app/graphql/types/base_argument.rb
+++ b/app/graphql/types/base_argument.rb
@@ -2,5 +2,8 @@
 
 module Types
   class BaseArgument < GraphQL::Schema::Argument
+    def initialize(*args, camelize: false, **kwargs, &block)
+      super
+    end
   end
 end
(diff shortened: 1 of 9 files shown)


app/graphql/types/base_argument.rb
```diff
@@ -2,5 +2,8 @@
 
 module Types
   class BaseArgument < GraphQL::Schema::Argument
+    def initialize(*args, camelize: false, **kwargs, &block)
+      super
+    end
```
yndajas, line 7: **question:** argument support

Does this definitely work for arguments? I can see that this is the suggested code for fields, but for arguments I can only see the suggestion for how to do it on a per-argument basis

yndajas (approved):
looks_good_to_me

app/graphql/types/base_argument.rb
brucebolt: I couldn't find any documentation for arguments, however since we're defining `Types::BaseArgument` as a argument class in our base field, I'm confident we're overriding the argument definition in the same way as we're doing for fields.

PR #3054, opened 2024-12-19
IPO trademark decisions finder

kashifatcha:
[Trello nested facets](https://trello.com/c/4RJusHld/3374-nested-facets-add-new-document)
[Trello IPO finder](https://trello.com/c/OA4zE8m0/3122-ipo-finder-request)


content_schemas/formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1169,6 +1172,49 @@
...
+        type: "string",
+      },
+      trademark_decision_mark: {
+        type: "string",
+      },
+      trademark_decision_class: {
+        type: "string",
+      },
+      trademark_decision_date: {
+        type: "string",
```
ryanb-gds, line 1195: We do have a date type I think - just checking we've made an intentional decision not to use it here?

content_schemas/formats/shared/definitions/nested_label_value_pair.jsonnet
```diff
@@ -0,0 +1,32 @@
+{
+  nested_label_value_pair: {
```
ryanb-gds, line 2: Would `label_value_pair_with_subfacets` be a better name for this at all?

ryanb-gds (approved):
Looks good, just want to check a couple of things but suspect no changes needed

content_schemas/formats/shared/definitions/nested_label_value_pair.jsonnet
lauraghiorghisor-tw: True, the "nested" is optional so that's misleading. Changed ✅

content_schemas/formats/shared/definitions/_specialist_document.jsonnet
minhngocd: "type: date" is on the finder schema as a separate property on the facet, i think the specialist document schema only cares that it is sending a date string of the correct format.

PR #3056, opened 2024-12-19
Use GraphQL-Ruby Dataloaders to improve Ministers Index performance

mike3985:
This reduces the number of database queries that are made in serving the
Ministers Index page from our /graphql endpoint. With my local data the
number when from 280 down to 28.

https://trello.com/c/dhngeJLd/1493-measure-and-improve-performance-of-ministers-index-graphql

### DIFF
--- a/app/graphql/sources/linked_to_editions_source.rb
+++ b/app/graphql/sources/linked_to_editions_source.rb
@@ -1,0 +1,29 @@
+module Sources
+  class LinkedToEditionsSource < GraphQL::Dataloader::Source
+    # rubocop:disable Lint/MissingSuper
+    def initialize(parent_object:)
+      @object = parent_object
+      @content_store = parent_object.content_store.to_sym
+    end
+    # rubocop:enable Lint/MissingSuper
+
+    def fetch(link_types)
+      all_links = @object.document.link_set_links
+        .includes(target_documents: @content_store) # content_store is :live or :draft (a Document has_one Edition by that name)
+        .where(link_type: link_types)
+        .where(target_documents: { locale: "en" })
+
+      link_types_map = link_types.index_with { [] }
+
+      all_links.each_with_object(link_types_map) { |link, hash|
+        hash[link.link_type].concat(editions_for_link(link))
+      }.values
+    end
+
+  private
+
+    def editions_for_link(link)
+      link.target_documents.map { |document| document.send(@content_store) }
+    end
+  end
+end
--- a/app/graphql/sources/person_current_roles_source.rb
+++ b/app/graphql/sources/person_current_roles_source.rb
@@ -1,0 +1,51 @@
+module Sources
+  class PersonCurrentRolesSource < GraphQL::Dataloader::Source
+    def fetch(person_content_ids)
+      all_roles = Edition
+        .live
+        .includes(
+          document: {
+            reverse_links: { # role -> role_appointment
+              link_set: {
+                documents: [
+                  :editions, # role_appointment
+                  :link_set_links, # role_appointment -> person
+                ],
+              },
+            },
+          },
+        )
+        .where(
+          document_type: "ministerial_role",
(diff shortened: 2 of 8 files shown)


.rubocop.yml
```diff
@@ -16,3 +16,7 @@ inherit_mode:
 #
 # See https://github.com/alphagov/rubocop-govuk/blob/main/CONTRIBUTING.md
 # **************************************************************
+
+Lint/MissingSuper:
```
brucebolt, line 20: Would it be possible to update the commit message headline for this commit, just to make it clearer what is being changed.

brucebolt: Or I wonder if the override can be specified in the class itself, to avoid adding an override here?

app/graphql/sources/linked_to_editions_source.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      all_links.each_with_object(link_types_map) { |link, hash|
+        hash[link.link_type].concat(editions_for_link(link))
+      }.values
+    end
+
+  private
+
+    def editions_for_link(link)
+      link.target_documents.map { |document| document.send(@content_store) }
```
brucebolt, line 24: Is `send` a safe thing to do based on a consumer's input?  I'm thinking a query could contain any value for `content_store` which we'll use to execute a method of the same name on the document.  Would it be better to explicitly write out `where(content_store: @content_store` here?

mike3985: That's smart thinking

.rubocop.yml
mike3985: I've dropped the commit and moved the override inline. I have a feeling that there's no way to win with this.

app/graphql/sources/linked_to_editions_source.rb
mike3985: We could check that the provided value is a member of `Edition.content_stores.keys`, and specify a default one to fallback to.

But taking a closer look, it's only the database query for the root Edition that uses the value from the GraphQL query. The `content_store` of the links fields is taken from the Edition's database record. So I'm going with this being fine for now, could be made more robust in the longer term, in case of changes to how its used.

app/graphql/sources/person_current_roles_source.rb
```diff
@@ -0,0 +1,51 @@
...
+          documents: { locale: "en" }, # role_appointment Documents
+          editions_documents: { document_type: "role_appointment" }, # role_appointment Editions
+          link_set_links: { target_content_id: person_content_ids, link_type: "person" },
+        )
+        .where("editions_documents.details ->> 'current' = 'true'") # editions_documents is the alias that Active Record gives to the role_appointment Editions in the SQL query
+        .order(reverse_links: { position: :asc })
+
+      ids_map = person_content_ids.index_with { [] }
+
+      all_roles.each_with_object(ids_map) { |role, hash|
```
JonathanHallam, line 31: I still don't have the best mental model for how this stuff is working and it's hurting my brain a little bit. I'm reasonably confident in it but i don't have time left this afternoon to get into my head well enough to approve. @brucebolt if you're more sure are you okay to give the 👍 ?

Commenting here because I'm reasonably confident this section is key as it's the bit that's flattening the layers out and doing the actual heavy lifting?

brucebolt: I'm not quite 100% confident in my understanding of how this all fits together, but it seems to make sense from my perspective.

It's also covered by the ministers index integration spec.

brucebolt (approved):

callumknights (approved):
I have only the loosest understanding of how the graphql dataloading works. But I can at least makes sense of the testing you have added @brucebolt 

Given what we discussed this morning I'm happy to approve based on that, but I'll definitely need to revisit this at some point.

PR #3063, opened 2024-12-30
Implement an edition type for all GraphQL edition-like documents

brucebolt:
This moves all edition types into a single type, to reduce duplication and minimise the amount of code required when adding support for content types to GraphQL.

The bespoke code for getting `role_appointments` on the ministers index page has been retained, as this was added to improve performance for that page due to the number of links and their complexity.

When we want to add a new content type that doesn't need any bespoke code, we just need to ensure all the relevant `details` fields are present in the Edition type.  All the link types have already been added and any links can be requested for any document type.  Without making the changes in this PR, we would need to create a new type for each document type, which would include a lot of code duplication where similar things exist across the types.

Also adds support for edition links and multi-level edition links.

[Trello card](https://trello.com/c/MSjuzc9M)


app/graphql/sources/reverse_linked_to_editions_source.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(parent_object:)
+      @object = parent_object
+      @content_store = parent_object.content_store.to_sym
+    end
+    # rubocop:enable Lint/MissingSuper
+
+    def fetch(link_types)
+      all_links = @object
+        .document
+        .reverse_links
```
mike3985: Are we missing some eager loading here?

```ruby
  .reverse_links
  .includes(source_documents: @content_store)
```

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
```diff
@@ -0,0 +1,20 @@
+RSpec.describe Sources::ReverseLinkedToEditionsSource do
+  it "returns the specified reverse link set links" do
+    target_document = create(:edition)
+    source_document_1 = create(:edition)
+    source_document_2 = create(:edition)
+    source_document_3 = create(:edition)
```
mike3985, line 6: It might be easier on my brain if these were called "source_edition"s

app/graphql/sources/reverse_linked_to_editions_source.rb
```diff
@@ -8,21 +8,31 @@ def initialize(parent_object:)
...
         .where(link_type: link_types)
 
-      link_types_map = link_types.index_with { [] }
+      link_set_links.each do |link|
+        link_types_map[link.link_type].concat(editions_for_link_set_link(link))
+      end
+
+      edition_links = Link
+        .where(target_content_id: @object.content_id, link_type: link_types)
+        .where.not(edition: nil)
```
mike3985, line 24: I've never worked with edition links. To me it looks like the first Active Record query ("link_set_links = ...") would return both link set links and edition links. Am I missing something?

mike3985 (commented):
This does look pretty good overall 👍

app/graphql/sources/reverse_linked_to_editions_source.rb
brucebolt: You're absolutely right.  I haven't noticed that the `reverse_links` relationship didn't involve `LinkSet` in any way.  I'll update this to just do the one query then grab the edition differently depending on the link type.

app/graphql/sources/reverse_linked_to_editions_source.rb
brucebolt: Yeah, I've added that in and it's substantially reduced the number of queries being made.

app/graphql/types/edition_type.rb
```diff
@@ -7,14 +7,167 @@ class WithdrawnNotice < Types::BaseObject
...
+
+      class Logo < Types::BaseObject
+        field :crest, String
+        field :formatted_title, String
+      end
+
+      class WhipOrganisation < Types::BaseObject
+        field :label, String
+        field :sort_order, Integer
+      end
```
yndajas, line 134: **nitpick:** line order

Should these go below the `field`s for consistency?

app/graphql/types/query_type/edition_type_or_subtype.rb
```diff
@@ -18,6 +15,8 @@ def resolve_type(object, _context)
           document_type = object.document_type
 
           matching_edition_subtype = Types::EditionType.descendants.find do |edition_subtype|
+            next unless edition_subtype.respond_to?(:document_types)
```
yndajas, line 18: **question:** use

What's this filter needed for?

spec/integration/graphql/news_article_spec.rb
```diff
@@ -86,10 +89,12 @@
                 base_path: "/government/news/announcement",
                 content_store: "live",
               ) {
-                ... on NewsArticle {
+                ... on Edition {
```
yndajas, line 92: **thought:** test suite structure

**Update:** I like your rewording in the world index spec: "has world index specific fields in the generic edition type". Perhaps that simple change in the rest of the type-specific specs would be sufficient, and it would allow us to keep some type-specific documentation even if there's no distinction in the implementation

---

I'd say that these tests and the separation of specs by type was motivated by the intention of mapping document types to GraphQL types. I'm not sure how much these specs and particularly tests like "exposes news article specific fields in addition to generic edition fields" make sense when everything is generic, since this behaviour is no longer specific to new article editions. Might be worth having a rethink about the edition/type-related tests

yndajas, line 92: **note:** syntax

If we're expecting a specific type and not accounting for any others, I don't think we really need the `...on` syntax

app/graphql/types/edition_type.rb
```diff
@@ -7,14 +7,167 @@ class WithdrawnNotice < Types::BaseObject
       field :withdrawn_at, GraphQL::Types::ISO8601DateTime
```
yndajas, line 7: **suggestion:** documenting our decisions

I'd be quite keen for this PR to include an ADR covering some of what you wrote in that Trello card about migrating to a god edition type, and our discussion of the pros and cons of the two different approaches

I think I missed the point at which we firmly decided to proceed with this approach, and I suspect future readers might want to know what motivated the shift from (possibly more Graphql-conventional?) stronger typing. The commit message here ("Move all edition fields onto a single edition type") does cover some of it, but it would be nice to capture the richness of your notes in Trello and our prior discussion

yndajas (commented):
This is a lot of work! A few thoughts and suggestions, but generally no major issues and I'm on board to give this approach a go

app/graphql/types/edition_type.rb
brucebolt: The classes used in the fields need to be defined before the fields themselves, otherwise we get an 'uninitialized constant' error.

spec/integration/graphql/news_article_spec.rb
brucebolt: I think this is needed as it remains possible to return a MinistersIndex type.

app/graphql/types/edition_type.rb
brucebolt: An ADR is a good idea.  I've added ADR-008 to explain the reasons behind the decision.

app/graphql/types/edition_type.rb
yndajas: Oh yeah, I misread these as methods!

spec/integration/graphql/news_article_spec.rb
yndajas: I think the `...on` syntax is more for when you're building a query that might be used for multiple different types, whereas this query is for a known (generic, not ministers index) type?

brucebolt: `EditionTypeOrSubtype` returns a union, which [the documentation](https://graphql.org/learn/queries/#inline-fragments) suggests will always need an inline fragment.

It's not ideal, but I guess this is our way of explicitly saying this is a request for the fields on an Edition.

docs/arch/adr-008-typing-for-graphql-editions.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Consequences
+
+We have removed the hardcoding and duplication amongst the varieties of edition we have currently implemented.
+
+These changes will make it easier to implement new document types, as the fields and links have already been defined.
+
+Since all links are now returned as an edition, any field can be selected from them, not just those that have been defined in the link type. This makes queries much more flexible, as the frontend application can request any fields it likes from the links, without any changes being made in Publishing API to support that field.
+
+If a frontend application queries for a field or link that does not exist on a given edition, it will receive an empty value for that field.
```
yndajas, line 29: **suggestion:** detail

I'd be keen for this to be a little more explicit on the pros and cons that we touched on in earlier discussions. I've written a summary of what I can recall below

---

We gain flexibility at the expense of feedback on which fields could be present for an edition of a given document/content type. With stronger typing, we receive errors guiding us on which fields are accepted for specific document/content types, but it involves writing a lot of bespoke types on the Publishing API side. Using a single edition type barring document/content types that need optimisation requires less work up front and technically provides more flexibility in queries, but that flexibility allows frontend apps to request fields that will always be null for a given document/content type (is there a benefit to the flexibility from the frontend perspective?). We get less feedback about which fields to expect: in place of errors when requesting fields that would never exist on a given document/content type, we get null data, and would need to inspect the content schemas to know if the reason for this is because a specific edition doesn't have data for a given field, or the given field is not accepted for any editions of the given document/content type

The single edition approach provides a similar level of feedback as retrieving data from the Content Store. Content Store data would never contain fields that would always be null for a given type, but it might also exclude optional/nullable fields. Using bracket notation on a content item from Content Store or the GraphQL endpoint would result in `nil` both when the field is absent and when it's null

We also floated the idea of a more compositional approach to types, whereby we define shared link types and field definitions once and then include them in specific types, but this would still have required more upfront work and might have resulted in some complex architecture

Ultimately we opted for the approach that would allow us to test GraphQL across all document/content types quickly

I wonder if some of this can be distilled into a concise table 🤔

JonathanHallam (approved):
Really nice work. Thanks Bruce. I'm approving this assuming that Ynda's comments will be looked at before merging

spec/integration/graphql/news_article_spec.rb
yndajas: Looks like the errors can be fixed by removing the duplicate `Image` and `logo` class definitions from the ministers index type, then using a namespaced classes for the fields: `EditionType::Details::Image` and `EditionType::Details::Logo`

image

brucebolt: I've updated the ministers index type to use the namespaced image and logo from the edition type, but I'm still getting the error `Selections can't be made directly on unions (see selections on EditionTypeOrSubtype)` when querying without an inline fragment.

PR #3067, opened 2025-01-02
Support multiple content types in details fields

brucebolt:
According to [ADR-004](https://github.com/alphagov/publishing-api/blob/main/docs/arch/adr-004-govspeak-and-embedded-content.md), it is possible to represent content in an array format containing multiple types. This can include already parsed govspeak in HTML format or only govspeak. In the latter case, Publishing API parses this and converts to HTML before sending the content to Content Store.

This replicates that behaviour for GraphQL responses, expanding this to all fields within `details`, not just the body.

As we are now using the `DetailsPresenter`, it also has the side effect of supporting embedded content and full change history in GraphQL responses.

[Trello card](https://trello.com/c/Rp5p0NTL)

### DIFF
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -153,17 +153,6 @@ module Types
       field :supports_historical_accounts, Boolean
       field :whip_organisation, WhipOrganisation
       field :world_locations, [EditionType], null: false
-
-      def body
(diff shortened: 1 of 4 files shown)


mike3985 (approved):
This looks much better than how I would've solved it 🙇

PR #3068, opened 2025-01-02
Return success from publish command when document already has a published edition and no draft exists

ryanb-gds:
Previously the publish command raised a conflict error in this case.

However, this error was rarely useful to the Publishing apps, as in almost all cases it is caused by publishing apps having to do their own dependency resolution due to shortcomings in Publishing API's dependency resolution, or by Sidekiq's "at least once" job running guarantees.

We therefore return a success response from the publish command and perform a no-op in this situation, rather than return an HTTP 409 error response for the publishing app to handle.

## Why

This causes occasional errors in Whitehall which trigger Sentry alerts. Whilst we could ignore 409 responses in Whitehall, it seems sensible to make a change on the Publishing API side. From a RESTful API perspective the request has been successful in that the document is in the published state.

For further discussion see this Trello card: https://trello.com/c/h7xkpcUj and [Slack thread](https://gds.slack.com/archives/C03D792LYJG/p1734448614763359) 

This will require an accompanying PR to update the Pact tests for Publishing API: https://github.com/alphagov/gds-api-adapters/pull/1319

### DIFF
--- a/app/commands/v2/publish.rb
+++ b/app/commands/v2/publish.rb
@@ -2,9 +2,11 @@ module Commands
   module V2
     class Publish < BaseCommand
       def call
-        validate
-        publish_edition
-        after_transaction_commit { send_downstream }
+        unless already_published? && edition.nil?
+          validate
+          publish_edition
+          after_transaction_commit { send_downstream }
+        end
 
         Success.new({ content_id: })
       end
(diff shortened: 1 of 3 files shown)


ryanb-gds:
Closing this PR because, as @mike3985 pointed out, this is in fact a breaking change to Publishing API (hence the failing Pact test). I therefore need to either declare a new version of the publish endpoint, or make the change non-breaking somehow.

PR #3077, opened 2025-01-10
(801) Add document type to HostContentUpdateJob events

pezholio:
Trello card: https://trello.com/c/iYVwlbzT/801-link-to-content-block-from-whitehall-history

This adds an additional `document_type` to the `HostContentUpdateJob` events introduced in https://github.com/alphagov/publishing-api/pull/2993. This will allow us to identify the type of block that has been updated when viewing a dependent document's history.


JonathanHallam (approved):
👍

PR #3081, opened 2025-01-14
Unpublish robots.txt

KludgeKML:
- robots.txt is now handled entirely within nginx, so the special route is unnecessary.
- a separate PR will be raised to remove this task once it's been run in production.


lib/tasks/data_hygiene.rake
```diff
@@ -40,4 +40,21 @@ namespace :data_hygiene do
   task :bulk_update_organisation, %i[csv_filename] => :environment do |_, args|
     DataHygiene::BulkOrganisationUpdater.call(args[:csv_filename])
   end
+
+  desc "Discard drafts (if present) and unpublish"
+  task :discard_drafts_and_unpublish, %i[content_id] => :environment do |_, args|
+    document = Document.where(content_id: args[:content_id]).first
```
brucebolt, line 46: You could get multiple documents returned here, since the same content ID is used for multiple locales.

I'd recommend returning all of them, then looping through to discard the drafts and unpublish each one.

KludgeKML: How does the unpublish work with locales? It's not clear to me how to pass a document object to Commands::V2::Unpublish (the examples I can find either pass in a content id, which obviously doesn't help, or an edition). Do I discard all the drafts and then pass in the Document's latest edition?

brucebolt: The discard and unpublish commands can both take a locale as an argument ([example](https://github.com/alphagov/publishing-api/pull/3047/files)).

KludgeKML:
@brucebolt After a long hiatus on this ticket, have finally updated with your suggestions and a decent test suite. Could you take another look please?

lib/tasks/data_hygiene.rake
```diff
@@ -40,4 +40,31 @@ namespace :data_hygiene do
   task :bulk_update_organisation, %i[csv_filename] => :environment do |_, args|
     DataHygiene::BulkOrganisationUpdater.call(args[:csv_filename])
   end
+
+  desc "Discard drafts (if present) and unpublish"
+  task :discard_drafts_and_unpublish, %i[content_id] => :environment do |_, args|
+    abort("Missing parameter: content_id") if args[:content_id].blank?
+
+    documents = Document.where(content_id: args[:content_id])
+    abort("Content ID #{args[:content_id]} not found") unless documents.count.positive?
```
brucebolt, line 49: Would `any?` here be a shorter way of achieving the same?

```suggestion
    abort("Content ID #{args[:content_id]} not found") unless documents.any?
```

brucebolt (commented):
This looks good to me.  Could you merge the commits, then I'll be happy to approve.

lib/tasks/data_hygiene.rake
KludgeKML: Ah yes, good point. I'll accept and then squish the commits.

brucebolt (approved):

PR #3094, opened 2025-01-21
Add rake task to get weekly edition counts

richardTowers:
A similar approach was used previously in analysis done in:

https://docs.google.com/spreadsheets/d/1ApBuQwsRLre4Pxau8Dw-XZbXr5eRg4MdS3-OA2pulh4/edit?gid=2078691047#gid=2078691047

... it's far from perfect, but counting the number of editions by their created_at date gives a reasonable proxy estimate of how many updates have been made in each publishing app each week.

I've been asked for this data again, so turning it from a one off script into a rake task.

Notes on the approach:

counting editions by created_at is more like a measure of how many drafts we started, rather than updates to live content we made. So this count will include drafts which were never published (but not drafts which were discarded, as those are deleted).

Various kinds of automation (rake tasks, database migrations) might have batch created huge numbers of editions, so there are a fair few outliers in this data. These can be ignored by a filter on count in post processing.


richardTowers:
I worked out how to do this using the BigQuery data extract, so this is no longer required:

https://docs.google.com/spreadsheets/d/1bQU20WZSvvFDx8fAYqms83tuHgBeYSKHl72BCbA8dc0/edit?gid=1784065235#gid=1784065235

It's possible we could do quite a few of these data extraction rake tasks using BigQuery TBH... 🤔

PR #3106, opened 2025-01-29
Batch link expansion queries

richardTowers:
(Note base branch is `add-graphql-optimisation-index`)

Previously LinkedToEditionsSource would make one query per edition (getting linked editions, given a list of link types).

We can do better than this by making it take a pair of edition and link type, and then execute a query like:

    SELECT *
    INNER JOIN "link_sets" ON "link_sets"."id" = "links"."link_set_id"
    INNER JOIN "documents" "target_documents" ON "target_documents"."content_id" = "links"."target_content_id"
    INNER JOIN "editions" ON "editions"."content_store" = 'live' AND "editions"."document_id" = "target_documents"."id"
    WHERE (
      ("link_sets"."content_id", "links"."link_type")
      IN
      (
        ('614abfd2-a26b-4df9-af6c-179e0fe1d99d','person'),
        ('2314c82e-2860-4cbe-9e5c-5320f64d0ac4','person'),
        ('a1052635-d603-40a1-b3f2-fb751b8ed9f9','person'),
        ...
      )
    )

This allows the dataloader to batch up all the editions and requested link types at each level of depth before sending one big query, which dramatically reduces the number of SQL queries needed.

For the /government/prime-minister page on my machine (with one additional tweak to avoid unnecessarily loading change notes), this reduces ActiveRecord time from ~500ms to ~50ms - around a 10x improvement.

Before:

    Completed 200 OK in 2694ms (Views: 1.5ms | ActiveRecord: 555.9ms (168 queries, 0 cached) | GC: 247.0ms)
    Completed 200 OK in 2787ms (Views: 2.1ms | ActiveRecord: 478.6ms (168 queries, 0 cached) | GC: 161.9ms)
    Completed 200 OK in 3145ms (Views: 2.3ms | ActiveRecord: 489.8ms (168 queries, 0 cached) | GC: 304.2ms)

After:

    Completed 200 OK in 1473ms (Views: 4.9ms | ActiveRecord: 147.0ms (12 queries, 0 cached) | GC: 61.4ms)
    Completed 200 OK in 1036ms (Views: 3.8ms | ActiveRecord: 63.3ms (12 queries, 0 cached) | GC: 70.3ms)
    Completed 200 OK in 1028ms (Views: 2.1ms | ActiveRecord: 46.1ms (12 queries, 0 cached) | GC: 41.9ms)

ActiveRecord doesn't seem to support the kind of tuple-inclusion query I'm making here. It does if all the columns are on the same table, but we need columns in different tables, so I've had to drop down to using Arel.

There might be a nicer way to do this, but I figured I'd get a draft up for some discussion first.


brucebolt:
I tried combining the link set and edition links into a single query .... but ended up making 19 SQL queries instead of 15.  I don't think the performance improvement will be significant by combining the two, so have left them as separate queries.

app/graphql/sources/linked_to_editions_source.rb
```diff
@@ -1,30 +1,35 @@
...
 
-      all_links = if @object.document.link_set
-                    edition_links.or(@object.document.link_set.links)
-                  else
-                    edition_links
-                  end
+      edition_links = Link
+        .joins(:edition, { target_documents: @content_store })
+        .includes(:edition, { target_documents: @content_store })
+        .where('("editions"."id", "links"."link_type") IN (?)', Arel.sql(edition_id_tuples.join(",")))
```
yndajas, line 16: **nitpick:** since the tuples are already strings arrays, you could maybe inline the `.join(",")` calls where you create the tuple variables?

yndajas (commented):
Haven't had time to finish looking at this, but no issues so far!

spec/graphql/sources/linked_to_editions_source_spec.rb
```diff
@@ -54,4 +54,27 @@
...
+
+    source_edition = create(:edition,
+                            content_store: "draft",
+                            links_hash: {
+                              "test_link" => [target_edition_1.content_id, target_edition_3.content_id],
+                            })
+
+    link_set = create(:link_set, content_id: source_edition.content_id)
+    create(:link, link_set: link_set, target_content_id: target_edition_2.content_id, link_type: "test_link")
+    create(:link, link_set: link_set, target_content_id: target_edition_4.content_id, link_type: "test_link")
```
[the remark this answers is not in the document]
brucebolt: Have updated that.  We used to have a rubocop rule to enforce this, but it seems to have been switched off recently.

app/graphql/sources/linked_to_editions_source.rb
brucebolt: Good idea.  I've updated this.

spec/graphql/sources/linked_to_editions_source_spec.rb
```diff
@@ -10,7 +10,7 @@
     create(:link, link_set: link_set, target_content_id: target_edition_3.content_id, link_type: "test_link")
 
     GraphQL::Dataloader.with_dataloading do |dataloader|
-      request = dataloader.with(described_class, parent_object: source_edition).request("test_link")
+      request = dataloader.with(described_class, content_store: source_edition.content_store).request([source_edition, "test_link"])
```
yndajas, line 13: **question:** are `.request` in this spec and `.load` in `BaseObject` both ending up at the `fetch` method in `LinkedToEditionsSource`? 🤯

yndajas (approved):

spec/graphql/sources/linked_to_editions_source_spec.rb
brucebolt: It appears to be, but I'm not sure why.  The same is done is the other dataloader tests too.

PR #3113, opened 2025-01-30
Make migrations concurrent

brucebolt:
The migrations to add new indexes were previously locking all writes to the editions table at the time of running.

As one of the migrations is slow (several minutes), this caused degraded performance of the application.

Making them concurrent to remove the need for the tables to be locked.

[Trello card](https://trello.com/c/74p1aw8n)

### DIFF
--- a/db/migrate/20250127105807_add_graphql_optimisation_index_to_editions.rb
+++ b/db/migrate/20250127105807_add_graphql_optimisation_index_to_editions.rb
@@ -1,7 +1,9 @@
 class AddGraphqlOptimisationIndexToEditions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
   def change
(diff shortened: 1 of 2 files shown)


db/migrate/20250127105807_add_graphql_optimisation_index_to_editions.rb
```diff
@@ -1,7 +1,9 @@
 class AddGraphqlOptimisationIndexToEditions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
```
yndajas, line 2: **suggestion:** might be worth explaining why this is needed? E.g. a link to this article: https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in

yndajas (approved):

PR #3115, opened 2025-01-30
Only retrieve change history from database if needed

brucebolt:
At the moment, we are retrieving the full change history for all documents from the datbase, whether it is needed or not.

This makes a change so the change history is only retrieved from the database if it has actually been requested by the client.

For the prime minister page, the number of database queries is reduced from 254 to 93.

[Trello card](https://trello.com/c/LorsXOIU)

### DIFF
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -135,7 +135,7 @@ module Types
 
       field :body, String
       field :brand, String
-      field :change_history, GraphQL::Types::JSON
+      field :change_history, GraphQL::Types::JSON, extras: [:parent]
       field :current, Boolean
       field :default_news_image, Image
       field :display_date, GraphQL::Types::ISO8601DateTime
@@ -158,6 +158,7 @@ module Types
     field :active, Boolean, null: false
     field :analytics_identifier, String
     field :base_path, String
+    field :change_history, GraphQL::Types::JSON
     field :content_id, ID
     field :current, Boolean
     field :description, String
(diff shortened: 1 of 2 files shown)


app/graphql/types/edition_type.rb
```diff
@@ -186,11 +187,15 @@ class WhipOrganisation < Types::BaseObject
...
+    def change_history
+      Presenters::ChangeHistoryPresenter.new(object).change_history
+    end
+
     def details
       Presenters::ContentTypeResolver.new("text/html").resolve(
         Presenters::DetailsPresenter.new(
           object.details,
-          Presenters::ChangeHistoryPresenter.new(object),
+          nil,
```
yndajas, line 198: **suggestion:** could we test this with some `Presenters::ChangeHistoryPresenter.new` called N times tests?

app/graphql/types/edition_type.rb
```diff
@@ -135,7 +135,7 @@ class WhipOrganisation < Types::BaseObject
 
       field :body, String
       field :brand, String
-      field :change_history, GraphQL::Types::JSON
+      field :change_history, GraphQL::Types::JSON, extras: [:parent]
```
yndajas, line 138: **suggestion:** it would be nice to see what effect this has on the resulting object. I'm not entirely sure what ["the previous `object` in the query"](https://graphql-ruby.org/fields/introduction#extra-field-metadata:~:text=parent%20(the%20previous%20object%20in%20the%20query)) is in this context

brucebolt: It returns the field of the same name from the previous object.  The current object is the details hash, so the previous object will be the edition from which the details were requested.

app/graphql/types/edition_type.rb
brucebolt: I've added `change_history` into the news articles integration test, since that document type is the only one we've migrated using this field.

app/graphql/types/edition_type.rb
brucebolt: It returns the field of the same name from the previous object.  The current object is the details hash, so the previous object will be the edition from which the details were requested.

PR #3120, opened 2025-02-04
(866) Add a pension content block

pezholio:
This adds a new `Pension` content block - this introduces a new concept of a block that has “embedded” objects. These objects MUST always have a title, the key of which is derived from the object’s title. I’ve made a Jsonnet util to support this, so we don’t have to repeat ourselves each time an embedded object is introduced. For example, the `details` of a simple pension block will look like this:

```
{
  "description": "Description goes here",
  "rates" :{
    "rate-1" :{
      "name": "Rate 1",
      "amount": "221.20",
      "cadence": "weekly",
      "description": "Your weekly pension amount"
    }
  }
}
```

It also introduces a non-standard `order` attribute. As we need to ensure all embedded objects  have a `name`, we have to merge the `properties` together with an object including `name`. This doesn’t preserve the order, so we keep have this additional field that we can then use in the Content Block Manager frontend to ensure the attributes appear in the order we specify, with the `name` as the first attribute.

### DIFF
--- a/content_schemas/allowed_document_types.yml
+++ b/content_schemas/allowed_document_types.yml
@@ -24,6 +24,7 @@
 - cma_case
 - content_block_email_address
 - content_block_postal_address
+- content_block_pension
 - complaints_procedure
 - completed_transaction
 - consultation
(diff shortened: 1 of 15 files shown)


Harriethw (approved):
looks good! the util thing is cool

PR #3121, opened 2025-02-05
Allow client to request only current role appointments

brucebolt:
The prime minister page does not need the former role appointments to render, so there is no need to retrieve them.

By adding an argument to the `role_appointments` link field, we can reduce the number of documents whose links need to be further expanded.

An index for editions with a current value in the details was added in e1b1955eb4e9e72b49674f72f690bb355c00f1c2.

[Trello card](https://trello.com/c/jDpZnpL0)


brucebolt:
I don't think this approach will work anyway, since Collections needs the value for `supports_historical_accounts` from the details to decide whether to show the past role holders.

PR #3142, opened 2025-02-14
content-modelling/ 880 - fix replacing embed codes in host documents

Harriethw:
https://trello.com/c/fP2L929o/901-bug-email-preview-in-mainstream-publisher-not-displaying-correctly

The helper method in Content Block Tools that finds
 content block embeds was recently set to
return mutliple references to one block [1] - this
then caused a knock-on effect for rendering blocks in
a document, because it would try to replace the
embed code multiple times, causing issues now that
the embed code is returned in the data attributes of
a span. We should only need to run `gsub` once to replace
all the embed codes for a reference, so I'm calling `uniq`
in the code that uses the Tool.

[1] https://github.com/alphagov/govuk_content_block_tools/pull/6

---

### DIFF
--- a/app/presenters/content_embed_presenter.rb
+++ b/app/presenters/content_embed_presenter.rb
@@ -52,7 +52,7 @@ module Presenters
       embedded_content_references = EmbeddedContentFinderService.new.find_content_references(content)
       return content if embedded_content_references.empty?
 
-      embedded_content_references.each do |content_reference|
+      embedded_content_references.uniq.each do |content_reference|
         embed_code = content_reference.embed_code
         embedded_edition = embedded_editions[content_reference.content_id]
         content = content.gsub(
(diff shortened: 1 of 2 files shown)


Harriethw:
also having to do the fix here https://github.com/alphagov/whitehall/pull/9937 
I feel like it's a bit of a shame we can't push the `uniq` back on the Tools repo, but the `EmbeddedContentFinderService` is also used to create the linked editions in `PutContent`, and we need to have multiple content IDs in there, because that's how we figure out the `count` in `GetHostContent` 🤯

pezholio (approved):

pezholio:
I wonder if the actual solution here is to put the `uniq` back in the gem and have a way to return the count for use when we actually want a count

PR #3160, opened 2025-02-25
Ipo finder nested facets

lauraghiorghisor-tw:
Adding some additional required fields for nested facets.

[Trello](https://trello.com/c/skH2B2fH/3373-nested-facets-finder-frontend)

### DIFF
--- a/content_schemas/dist/formats/finder/frontend/schema.json
+++ b/content_schemas/dist/formats/finder/frontend/schema.json
@@ -825,6 +825,14 @@
           "description": "A human readable label",
           "type": "string"
         },
+        "main_facet_label": {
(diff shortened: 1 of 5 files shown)


ChrisBAshton (approved):
I do wonder if we could make it clearer that the `main_facet_value` is the important bit for linking (otherwise my first instinct was "why do we have to define the sub_facet_key/name _and_ the main_facet_label/value?"). But couldn't think of a way of rephrasing the `description` to make it clearer 🤔

minhngocd:
> I do wonder if we could make it clearer that the `main_facet_value` is the important bit for linking (otherwise my first instinct was "why do we have to define the sub_facet_key/name _and_ the main_facet_label/value?"). But couldn't think of a way of rephrasing the `description` to make it clearer 🤔

Updated the description. Would that read better?

PR #3161, opened 2025-02-25
Remove sub_facets field from nested facets schema

lauraghiorghisor-tw:
We no longer send the `sub_facets`.

[Trello](https://trello.com/c/skH2B2fH/3373-nested-facets-finder-frontend)


minhngocd:
> Would be a good idea to refer to #3160 in the commit message so that future devs don't have to dig through GitHub histories to figure out what this replaces 👍

Done 👍

PR #3162, opened 2025-02-25
Add owning_organisations to ministers index links

richardTowers:
Currently we have to expand links from the minsiters index content item to every ministerial department, and from there:

- every minister in the department
  - every role the minister has
- every role in the department

We then use the department role ids to filter the list of roles for each minister, so we can show a list of departments, the department's ministers, and their roles within the department.

Unfortuanetly, the ordered_roles for a department includes _all_ the roles, not just ministerial roles. So there are a lot of them. The ministers index content item currently includes 1,745 roles:

    curl --silent 'https://www.gov.uk/api/content/government/ministers' | jq -r '.links.ordered_ministerial_departments[].links.ordered_roles[].title' | wc -l

The code in collections has to filter the roles for a minister based on this list of all the roles for the department:

https://github.com/alphagov/collections/blob/main/app/presenters/ministers_index_presenter.rb#L99

It would be simpler and more efficient if we could link directly from the role back to the departments the role belongs to. This information is already in the database, because organisations have ordered_role links to all the roles they own.

This commit introduces a reverse link for ordered_roles - owning_organisations. We can then use this in the expansion rules to include the owning_organisations for each role for each minister for each department.

There don't appear to be many tests for specific link expansion scenarios, but you can see this working by doing:

    Presenters::EditionPresenter.new(Edition.live.find_by(base_path: "/government/ministers")).expanded_links

in a rails console and looking for the owning_organisations links.

The ordered_roles link expansion is left in place for now, as we need to update collections to use the new links before we can remove the old ordered_roles expansion.

### DIFF
--- a/content_schemas/dist/formats/answer/frontend/schema.json
+++ b/content_schemas/dist/formats/answer/frontend/schema.json
@@ -111,6 +111,10 @@
           "description": "The organisation that published the original version of the page. Corresponds to the first of the 'Lead organisations' in Whitehall for the first edition, and is empty for all other publishing applications.",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
+        "owning_organisations": {
+          "description": "Link type automatically added by Publishing API",
+          "$ref": "#/definitions/frontend_links_with_base_path"
+        },
         "parent": {
           "description": "The parent content item.",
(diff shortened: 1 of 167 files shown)


yndajas (approved):
Congratulations on winning the content schema test lottery 🎉

PR #3166, opened 2025-02-28
Add document type labels to publication schema

ryanb-gds:
The [official documents](https://www.gov.uk/official-documents) and [research and statistics](https://www.gov.uk/search/research-and-statistics) finders render a human friendly version of the document type in the search result metadata.

Now that we are seeking to index all Whitehall documents via Publishing API, we need to make sure that we send these labels as part of the edition payload.

Link to Whitehall PR: https://github.com/alphagov/whitehall/pull/10001 

Trello: https://trello.com/c/upZnqIOL

### DIFF
--- a/content_schemas/dist/formats/publication/frontend/schema.json
+++ b/content_schemas/dist/formats/publication/frontend/schema.json
@@ -382,6 +382,10 @@
         "change_history": {
           "$ref": "#/definitions/change_history"
         },
+        "document_type_label": {
+          "description": "a human readable version of the document type",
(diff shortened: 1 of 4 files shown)


ChrisBAshton (approved):
I trust this PR and approve it - but it would be good to link to the Research and statistics finder (<https://www.gov.uk/search/research-and-statistics?content_store_document_type=upcoming_statistics&order=updated-newest>?) and also link to the code that defines the document type label (had a quick search and couldn't find it).

PR #3172, opened 2025-03-03
content-modelling/Handle incorrect content IDs found in embeds

Harriethw:
https://trello.com/c/S4xH4K03/878-debug-logging-and-error-handling-for-embed-codes

Previously, if an editor published an Edition with an embed code with an incorrect or missing content ID, then it would fail silently (on Mainstream) or show generic error page (on Whitehall). It seemed like the error for Mainstream wasn't propagating (possibly because it was being called in a Worker, so the Sidekiq job was just failing and retrying in background?).

This is a first pass at handling when we might have an incorrect Content ID, allowing the Edition to be published and the incorrect embed code rendered in plain text in the process, but also alerting Sentry so it will be easier to catch if this issue occurs.

The error handling was introduced [last year here](https://github.com/alphagov/publishing-api/pull/2813), before we started embedded Content Blocks in earnest and I think only Contacts were the embedded content being used? 

Is it better to allow the Editor to publish an incorrect block, when they can see that it's not being rendered and correct it, instead of either failing in the background or throwing an error preventing publication?  I'm not sure without touching more of the Mainstream architecture how we would throw a more helpful error....

Still TODO:

* Add the new errors to the Sentry rules to send them to our #content-modelling-dev channel

---

### DIFF
--- a/app/presenters/content_embed_presenter.rb
+++ b/app/presenters/content_embed_presenter.rb
@@ -59,10 +59,17 @@ module Presenters
       embedded_content_references.uniq.each do |content_reference|
         embed_code = content_reference.embed_code
         embedded_edition = embedded_editions[content_reference.content_id]
-        content = content.gsub(
-          embed_code,
-          get_content_for_edition(embedded_edition, embed_code),
-        )
+        if embedded_edition.present?
+          content = content.gsub(
+            embed_code,
+            get_content_for_edition(embedded_edition, embed_code),
+          )
+        else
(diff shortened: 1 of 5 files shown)


pezholio (approved):
Yeah, this sounds sensible to me. It's better for us to get some kind of error in the backend when this happens and the code to get ignored than the page to straight up not get published (especially as this could happen in a background job and not be immediately obvious to a user). I'd hope in "real life", anything like this would get picked up at the 2i process, but better to be safe than sorry!

PR #3192, opened 2025-03-10
Reduce dataloader queries

yndajas:
[Trello](https://trello.com/c/zMclxWKo/1589-use-lookahead-to-enable-selecting-subsets-of-columns-in-graphql)

This optimises(?) the database queries in `LinkedToEditionsSource` and `ReverseLinkedToEditionsSource`, reducing them down to one query each (in practice, one query per level of depth in the GraphQL query). This is a change en route to selecting a subset of columns - we'll open a separate pull request for that change

We're looking into the actual performance of this change at the minute but thought it was worth opening the pull request since the core bones are in place

---

### DIFF
--- a/app/graphql/sources/linked_to_editions_source.rb
+++ b/app/graphql/sources/linked_to_editions_source.rb
@@ -7,6 +7,10 @@ module Sources
     # rubocop:enable Lint/MissingSuper
 
     def fetch(editions_and_link_types)
+      all_selections = {
+        links: %i[link_type position],
+        documents: %i[content_id],
+      }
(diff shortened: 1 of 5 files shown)


brucebolt (commented):
The code changes all seem ok to me.

But I've run the code locally and queried the ministers index page, which gives the `role_appointments` links in a different order to those in the existing content item.  It could be the order has changed since my old dump of the database, but I'd suggest trying this code on integration to double check the order has stays the same when querying against the same database.

brucebolt (approved):

PR #3196, opened 2025-03-11
Subset graphql query selections

yndajas:


lib/graphql_selections.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def to_select_args
+    @tables_and_columns.transform_values(&:to_a)
+  end
+
+protected
+
+  def each(&block)
+    @tables_and_columns.each(&block)
+  end
```
brucebolt, line 104: I've been thinking about ways of implementing this without needing to create these methods.  I tried without using `Set` (using hashes and arrays instead) and came up with this which works for `QueryType`: https://github.com/alphagov/publishing-api/commit/6ccf0e4b0c31789271a0e5e449dcc22c8b679613.

This will make the code simpler by only using native Ruby methods.

brucebolt (commented):
Great work on figuring this out.

I've made a few suggestions to make the code easier to understand.  I'm particularly keen that we remove the added complexity in `GraphqlSelections` by creating our own `insert` and `merge` methods, which I think can be avoided by using a hash instead of a set.

app/graphql/types/query_type.rb
```diff
@@ -5,10 +5,29 @@ class QueryType < Types::BaseObject
...
-      Edition.where(content_store:).find_by(base_path:)
+    def edition(base_path:, content_store:, lookahead:)
+      selections = GraphqlSelections.with_root_edition_fields(
+        lookahead.selections.map(&:name),
+      )
+
+      selections.insert(:editions, %i[document_type])
+
+      if lookahead.selects?(:links)
+        if lookahead.selections.find { _1.name == :links }.selects?(:available_translations)
```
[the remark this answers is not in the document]
mike3985: Oh 😮 `selection` didn't produce the same results as the `find` when we tested this out in rails console... but we've made this change now and it seems to be working fine 🤷

lib/graphql_selections.rb
mike3985: We've made some changes to that object based on some of your concerns.

```ruby
selections.each do |table, columns|
  (all_selections[table] ||= []).append(*columns)
end
```

and

```ruby
(selections[:documents] ||= []).append(:content_id, :locale)
```

out there in the types and data loader sources code felt a little too hairy for me. For me, the object provides a smaller, cleaner interface and there's less chance of getting any of the details of that ^ code wrong if avoid repeating it.

We took your point about the Set... I personally hadn't actually realised that you could perform an `&` intersection on two vanilla Arrays 😊 _and_ I was relieved to be able to get rid of that `#transform_values(&:to_a)`. It wasn't essential to de-duplicate those column names, but it was nice to have that feature for free when we were using Sets... we've compromised and used an in-place `uniq!` 😬 

(We also took the opportunity while already editing the object and its call sites to shuffle some of those custom QueryType/root edition selections around, FWIW.)

What do you think? An improvement?

lib/graphql_selections.rb
```diff
@@ -0,0 +1,99 @@
...
+    redirects
+    rendering_app
+    routes
+    schema_name
+    state
+    title
+    update_type
+    updated_at
+    user_facing_version
+  ].freeze
```
brucebolt, line 33: Does `Edition.column_names` return these same values?  It would mean these values don't need to be hardcoded but would that mean we make another database query?

lib/graphql_selections.rb
brucebolt: Thanks for looking at this.  Your explanation for keeping the object mostly intact makes sense, especially keeping the complexity/logic out of the types and data loader sources.

brucebolt (approved):
Thanks for addressing the comments.  This all looks good to me.

lib/graphql_selections.rb
yndajas: Just tested this in a local Rails console:

```rb
edition_columns = [:content_id, :base_path]

Benchmark.bmbm do |x|
  x.report("constant") { 10000.times { ALL_EDITION_COLUMNS & edition_columns } }
   x.report("method") { 10000.times { Edition.column_names & edition_columns } }
end
```

Looks like the method version is about 60% slower, but we're talking 25 vs 40 milliseconds (on my local machine), so I'd be tempted to opt for the approach that keeps us in sync with the schema automatically (`Edition.column_names`)

This is a fairly representative result:

```
[#<Benchmark::Tms:0x0000ffff5a4cfe78
  @cstime=0.0,
  @cutime=0.0,
  @label="constant",
  @real=0.0025080839404836297,
  @stime=1.399999999995849e-05,
  @total=0.0025130000000002095,
  @utime=0.002499000000000251>,
 #<Benchmark::Tms:0x0000ffff5a4cfd88
  @cstime=0.0,
  @cutime=0.0,
  @label="method",
  @real=0.004154750029556453,
  @stime=0.0,
  @total=0.004160000000000608,
  @utime=0.004160000000000608>]
```

lib/graphql_selections.rb
```diff
@@ -0,0 +1,99 @@
...
+    @tables_and_columns[table].push(*columns).uniq!
+  end
+
+  def selects_from_table?(table)
+    @tables_and_columns[table].present?
+  end
+
+  def to_h
+    @tables_and_columns
+  end
```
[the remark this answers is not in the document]
yndajas: We talked about this and ended up thinking that we should actually prevent users from mutating the underlying hash stored in the `GraphqlSelections` instance without using the class's own API. To that end we've opted to keep it as a `#to_h` but with a `#deep_dup` - does that feel reasonable to you?

lib/graphql_selections.rb
yndajas: Bummer, we actually need to say `Edition.column_names.map(&:to_sym)` to intersect with the edition fields, which hits the performance quite substantially (10x slower than the constant):

```
[#<Benchmark::Tms:0x0000ffff8f63f508
  @cstime=0.0,
  @cutime=0.0,
  @label="constant",
  @real=0.002311042044311762,
  @stime=0.0,
  @total=0.0023169999999996804,
  @utime=0.0023169999999996804>,
 #<Benchmark::Tms:0x0000ffff9e7a7ad0
  @cstime=0.0,
  @cutime=0.0,
  @label="method",
  @real=0.01926020800601691,
  @stime=0.0,
  @total=0.019267000000000145,
  @utime=0.019267000000000145>]
```

yndajas: Tried looking at the response times locally with and without this change, and they're too random to confidently say this makes a difference in practice, so let's keep it simple for now and leave the minor optimisation as something we could revert to in future

PR #3198, opened 2025-03-11
Return locale when fetching host content

pezholio:
Trello card: https://trello.com/c/8SRwSAUh/968-bug-defer-payment-welsh-guide-preview-in-edit-journey-leads-to-server-error-published-state

When fetching host content, we will sometimes need to specify the locale (for example, when content is in Welsh). At the moment, if a document is only in Welsh, then we can’t generate a local preview in Content Block Manager.

We also need to update the `host-content` endpoint to optionally filter by locales, which will cover when a piece of content has multiple translations.

### DIFF
--- a/app/queries/get_host_content.rb
+++ b/app/queries/get_host_content.rb
@@ -64,9 +64,9 @@ module Queries
 
     ORDER_DIRECTIONS = %i[asc desc].freeze
 
-    attr_reader :target_content_id, :state, :order_field, :order_direction, :page, :per_page, :host_content_id
+    attr_reader :target_content_id, :state, :order_field, :order_direction, :page, :per_page, :host_content_id, :locale
 
-    def initialize(target_content_id, order_field: nil, order_direction: nil, page: nil, per_page: nil, host_content_id: nil)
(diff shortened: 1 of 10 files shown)


app/queries/get_host_content.rb
```diff
@@ -109,9 +112,9 @@ def clauses
       clauses = TABLES[:editions][:state].eq(state)
                                          .and(TABLES[:links][:link_type].eq(embedded_link_type))
                                          .and(TABLES[:links][:target_content_id].eq(target_content_id))
-      if host_content_id
-        clauses = clauses.and(TABLES[:documents][:content_id]).eq(host_content_id)
-      end
+
+      clauses = clauses.and(TABLES[:documents][:content_id]).eq(host_content_id) if host_content_id
```
callumknights, line 116: Can you explain why this `host_content_id` has moved? Wondering if that clauses assignment not longer being covered will cause issues

pezholio: Oh, I just changed it to an inline if to make the assignment a bit cleaner

callumknights: Oh yeah my bad, github review format tricking me, it's not the first time 🤦

callumknights (approved):
LGTM 👍

PR #3204, opened 2025-03-12
Remove S3 events import/export job

brucebolt:
This was switched off in https://github.com/alphagov/govuk-helm-charts/pull/2448 having been broken for many months prior to this.

Removing the code as it is completely unused at the moment.


docs/model.md
```diff
@@ -239,15 +239,6 @@ Events are used as a debugging and reference tool by developers of the
 Publishing API. As they generate large amounts of data the full details of
 them are not stored permanently.
 
-Events older then a month are archived to S3, you can import these events back
-into your local DB by running the rake tasks in lib/tasks/events.rake, after
-you set up the relevant ENV variables. For example if you want to find all the
-events that are relevant for a particular content id you can run:
-
-```sh
-rake 'events:import_content_id_events[a796ca43-021b-4960-9c99-f41bb8ef2266]'
```
yndajas, line 248: **suggestion (non-blocking):** the linked PR did turn off the latest related cron job (using `events:export_to_s3`), but is it worth also referencing the removal of the import task mentioned above, which was much earlier in 4335a751ff7004d2ae3cf444a1ef9df223cd666c?

yndajas, line 248: **suggestion (non-blocking):** should the following files (and related specs) be removed as part of this if we're no longer using them?

- https://github.com/alphagov/publishing-api/blob/main/lib/tasks/events.rake
- https://github.com/alphagov/publishing-api/blob/main/lib/events/s3_importer.rb
- https://github.com/alphagov/publishing-api/blob/main/lib/events/s3_exporter.rb

yndajas (approved):

docs/model.md
brucebolt: Hadn't spotted that rake task was removed.  I've added a link to https://github.com/alphagov/publishing-api/pull/947 in the commit message.

brucebolt: I think they should.  I'll do it in another PR.

yndajas (approved):

PR #3219, opened 2025-03-18
Add configuration for read replica

brucebolt:
The read replica is a separate deployment of this application, which connects to both the primary database (e.g. for writing user records after authentication) and the replica database (for all reads).

Toggling of the application is done by changing the `RAILS_ENV` environment variable from `production` to `production_replica`.

This was developed based on the guide "[Multiple Databases with Active Record](https://guides.rubyonrails.org/active_record_multiple_databases.html)".

Whilst authentication was removed from GraphQL requests in https://github.com/alphagov/publishing-api/pull/3220, I am leaving the connection to the primary database in the replica, since we may wish to authenticate other requests in the future.

Depends on https://github.com/alphagov/govuk-helm-charts/pull/3077.

[Trello card](https://trello.com/c/8QZKAla1)

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -9,34 +9,46 @@ class GraphqlController < ApplicationController
   skip_before_action :authenticate_user!, only: [:execute]
 
   def execute
-    variables = prepare_variables(params[:variables])
-    query = params[:query]
-    operation_name = params[:operationName]
-    context = {
-      # Query context goes here, for example:
-      # current_user: current_user,
-    }
-    result = PublishingApiSchema.execute(
-      query,
-      variables:,
-      context:,
-      operation_name:,
-    ).to_hash
+    execute_in_read_replica do
+      variables = prepare_variables(params[:variables])
+      query = params[:query]
+      operation_name = params[:operationName]
+      context = {
+        # Query context goes here, for example:
+        # current_user: current_user,
+      }
+      result = PublishingApiSchema.execute(
+        query,
+        variables:,
+        context:,
+        operation_name:,
+      ).to_hash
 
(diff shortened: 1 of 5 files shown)


app/controllers/graphql_controller.rb
```diff
@@ -9,35 +9,47 @@ class GraphqlController < ApplicationController
...
+      raise e unless Rails.env.development?
 
-    handle_error_in_development(e)
+      handle_error_in_development(e)
+    end
   end
 
 private
 
+  def execute_in_read_replica(&block)
```
yndajas, line 43: **thought:** I wonder if this is the write name if it only conditionally executes in the read replica. I can't think of a good alternative though!

yndajas (approved):

app/models/application_record.rb
```diff
@@ -1,3 +1,10 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  if Rails.env.production_replica?
+    connects_to database: {
+      writing: :primary,
```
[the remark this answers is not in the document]
brucebolt: I decided to leave this in place, to avoid us having to work out why authenticated requests aren't working in the future should be decide to add them in.

app/controllers/graphql_controller.rb
brucebolt: I did consider the name `execute_with_or_without_read_replica` but decided that was probably too confusing.

yndajas: `execute_one_way_or_another`

brucebolt: Hmm, I'll stick with the current name for now.  We can always rename it at some point in the future.

PR #3220, opened 2025-03-18
Do not authenticate GraphQL requests

brucebolt:
Publishing API is only available internally within our infrastructure and can only return  data we make available through that endpoint.

Therefore removing the authentication of read requests for GraphQL only. This matches [the same behaviour](https://github.com/alphagov/content-store/blob/2a77ee211152ff44b38613d6f61194ce93b01062/app/controllers/content_items_controller.rb#L2) we have in Content Store for returning content items.

With regards to the performance of the read replica, authenticating every request will result in data being read/written to the primary (non-replica) database for each request (e.g. user permissions).

[Trello card](https://trello.com/c/8QZKAla1)

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -6,6 +6,8 @@ class GraphqlController < ApplicationController
   # but you'll have to authenticate your user separately
   # protect_from_forgery with: :null_session
 
+  skip_before_action :authenticate_user!, only: [:execute]
+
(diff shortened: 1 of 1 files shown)


yndajas (approved):
As discussed, we might need to reconsider this in future if we end up serving the public Content API requests with GraphQL to avoid exposing draft content, but this seems reasonable for our current needs given Publishing API is only available internally

PR #3221, opened 2025-03-18
Update examples to remove rendering app from moved formats

KludgeKML:
- In general, the frontend examples (which are used in test suites) don't actually need to include rendering app. Some examples have it, some don't. So we can remove it from those examples where the rendering app has recently changed - we could alternatively have updated them, but since the info isn't actually necessary, it seems simpler to just remove it entirely.


deborahchua (approved):
LGTM

PR #3223, opened 2025-03-19
content-modelling/ refactor embedded editions

Harriethw:
https://trello.com/c/UwDgEDMN/970-bug-tech-debt-inefficient-embed-searching

When looking at optimising our embedded editions process, I noticed that we are using quite a complicated set of
db queries to get the embedded editions from a host edition - as the GraphQL team are looking to reduce queries [1] I thought I would take a look. 

This new  query I think improves things:
1. If I've worked it out right, this makes one less call to db
2. It's much simpler and I've removed the logic around locale and state - for embedded editions we only have `en` and we only ever want `published`editions. This might change in the future, but seems like the simpler we can keep things for now the better.

This will only affect calls where embed codes have been found, so won't make much of a difference now, but should still be a bit of an improvement if we ever scale to 100's of embeds!

[1] https://trello.com/c/ilVC6hCV

---

### DIFF
--- a/app/presenters/content_embed_presenter.rb
+++ b/app/presenters/content_embed_presenter.rb
@@ -43,8 +43,10 @@ module Presenters
         value.each do |nested_key, nested_value|
           value[nested_key] = convert_field(nested_value)
         end
-      else
+      when String
         render_embedded_editions(value)
+      else
+        value
       end
     end
 
--- a/app/presenters/content_embed_presenter.rb
+++ b/app/presenters/content_embed_presenter.rb
@@ -17,20 +17,8 @@ module Presenters
   private
 
     def embedded_editions
-      @embedded_editions ||= begin
-        target_content_ids = @edition
-         .links
-         .where(link_type: "embed")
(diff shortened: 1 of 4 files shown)


app/presenters/content_embed_presenter.rb
```diff
@@ -17,20 +17,8 @@ def render_embedded_content(details)
...
-        embedded_edition_ids = ::Queries::GetEditionIdsWithFallbacks.call(
-          target_content_ids,
-          locale_fallback_order: [@edition.locale, Edition::DEFAULT_LOCALE].uniq,
-          state_fallback_order: %w[published],
-        )
-
-        Edition.where(id: embedded_edition_ids).index_by(&:content_id)
-      end
+      @embedded_editions ||=
+        ::Queries::GetEmbeddedEditionsFromHostEdition.call(edition: @edition)
```
pezholio, line 21: Nice, big fan of this being pulled out into a query class

pezholio (approved):
LGTM - thanks for looking into this!

PR #3232, opened 2025-03-21
ADR-009 - change link_sets pkey to content_id

richardTowers:
I'm proposing an update to publishing-api's database schema, which will simplify link expansion.

✨  [Rendered](https://github.com/alphagov/publishing-api/blob/adr-009/docs/arch/adr-009-change-linksets-primary-key-to-content-id.md) ✨ 

At this point I'm just looking for feedback on whether or not this is a good idea, and whether I've missed anything in my description of the steps we'd need to take to make the change.

Let's leave "who's going to do the work" discussions until we've agreed whether or not it's a good idea in the first place (although I'd be happy to do a lot of the work myself).

### DIFF
--- a/docs/arch/adr-009-change-linksets-primary-key-to-content-id.md
+++ b/docs/arch/adr-009-change-linksets-primary-key-to-content-id.md
@@ -1,0 +1,298 @@
+# Decision Record: Change link_sets primary key from serial to content_id
+
+## Context
+
+Publishing API has the concept of "link sets".
+
+A link set comprises all the links belonging to a particular content_id. Links can have various link types and
+positions, but if they belong to the same content_id, they're part of the same link set.
+
+Link sets are exposed through the API as a resource - you can get the link set of a content_id
+using [GET /v2/links/:content_id](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#get-v2linkscontent_id),
+and you can update them
+using [PATCH /v2/links/:content_id](https://github.com/alphagov/publishing-api/blob/main/docs/api.md#patch-v2linkscontent_id).
+
+Because there's a potential race condition between getting and updating a link set, they have a lock version column. For
+example, two processes fetch a link set which looks like this:
+
+```json
+{
+  "links": {
+    "organisations": [
+      "591436ab-c2ae-416f-a3c5-1901d633fbfb"
+    ]
+  }
+}
+```
+
+And the first process attempts to add `492b07ba-1a83-4163-9d1f-37fa8a088b77` as an organisations link, and the second
+process attempts to add `489743df-25bd-45b3-b799-5bfcab131cce`, the request that comes later should fail because it's not
+updating the latest version of the resource.
+
+To enable this, link_sets have a stale_lock_version column, which is incremented on every update.
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


JonathanHallam (approved):
This all looks sensible to me, Thanks for laying out the plan as well 👍

richardTowers:
I'll draft the PRs to implement this before merging it, so that there's a better chance of getting it all done and not having an accepted-but-not-implemented ADR in the docs.

JonathanHallam:
> I'll draft the PRs to implement this before merging it, so that there's a better chance of getting it all done and not having an accepted-but-not-implemented ADR in the docs.

Can we save merging till the end of the week just to give the rest of the team a chance to have a look

brucebolt (approved):
This all makes sense to me and seems like a logical step in simplifying link expansion.

I think in the longer term, I'd like to explore the removal of link set links as a concept, and whether we could retain backward compatibility that mimicked link set links but stored as edition links in Publishing API's database.

docs/arch/adr-009-change-linksets-primary-key-to-content-id.md
```diff
@@ -0,0 +1,307 @@
...
+
+```sql
+-- 1. Drop the original primary key
+ALTER TABLE link_sets DROP CONSTRAINT link_sets_pkey
+
+-- 2. Rename existing index
+ALTER INDEX index_link_sets_on_content_id RENAME TO link_sets_pkey
+
+-- 3. Create new primary key using existing index
+ALTER TABLE link_sets ADD PRIMARY KEY USING INDEX link_sets_pkey
```
yndajas, line 275: **thought:** I got a little confused here about what this is doing that wouldn't be done in step 6, but I assume it's some fancy SQL stuff

yndajas (approved):
Seems sensible to me. I'm curious what led it to be architected this way initially, but I'm assuming it was old needs that no longer need to be met or just a best guess at the time, and not something more substantial that we're missing

docs/arch/adr-009-change-linksets-primary-key-to-content-id.md
richardTowers: Good question actually.

I _think_ there are two concepts of "primary key"

1) what Rails / ActiveRecord thinks is the primary key, which can be overridden in the model like:

```ruby
class Link < ApplicationRecord
  self.primary_key = 'content_id'
```

2) what the database actually uses as the primary key, i.e. the set of columns which have a primary key constraint on them

I'm assuming that we can change (1) in step 6 without `content_id` being an actual proper primary key in postgres land.

Once we've done that, I think we'll get all the performance benefits, and everything should work (hopefully), but the database is in a bit of a weird state if we've still got the old `id` column as the actual primary key, but rails using `content_id` as its "primary key".

I guess actually there's a missing step to remove this bit from step 6 once `content_id` is the true primary key, because rails should be able to work it out then.

```ruby
class Link < ApplicationRecord
  self.primary_key = 'content_id'
```

yndajas: Thanks for the explainer!

richardTowers:
Okay, we're going for it 🚀

PR #3241, opened 2025-03-28
Make the metadata block on specialist documents optional

GDSNewt:
These change is necessary to make the rendering the blue metadata box at the top of specialist documents dependent on a value defined in the schema of each finder.

**Associated PRs**

1. [Government-Frontend](https://github.com/alphagov/government-frontend/pull/3639)
2. [Specialist-Publisher](https://github.com/alphagov/specialist-publisher/pull/3076)

[Trello](https://trello.com/c/spUCgJtQ/3539-make-it-possible-to-hide-the-blue-box-on-specialist-documents)

### DIFF
--- a/content_schemas/examples/specialist_document/frontend/countryside-stewardship-grants.json
+++ b/content_schemas/examples/specialist_document/frontend/countryside-stewardship-grants.json
@@ -315,7 +315,8 @@
                 }
               ]
             }
-          ]
+          ],
+          "show_metadata_block": true
         },
         "links": {
         },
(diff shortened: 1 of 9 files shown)


content_schemas/examples/specialist_document/frontend/countryside-stewardship-grants.json
ChrisBAshton, line 1: Worth adding a link to https://trello.com/c/7vhL24Fk/840-update-government-frontend-tests-to-use-randomexample-generation-from-schemas-so-we-can-delete-specialist-document-examples-from in the commit message, to say that we have this backlog item that, if completed, would mean we don't need to update any examples?

ChrisBAshton (approved):
LGTM 👍

PR #3245, opened 2025-03-28
(ADR-009 PR 3) Populate link_set_content_id with rake task

richardTowers:
I considered two ways of doing this (see the individual commits). The way I settled on is much faster, but requires some SQL strings because ActiveRecord doesn't support `UPDATE / FROM` in postgres yet.

I've tested the rake task against my local database, and it chomps through at a reasonable rate - should take less than an hour I think.

### DIFF
--- a/lib/tasks/add_link_set_content_id_to_links.rake
+++ b/lib/tasks/add_link_set_content_id_to_links.rake
@@ -1,0 +1,19 @@
+desc "Add link_set_content_id to links"
+task :add_link_set_content_id_to_links, [:link_id] => [:environment] do |t, args|
+  link_id = args.fetch(:link_id, 0).to_i
+  link_set_links = Link.joins(:link_set).includes(:link_set).where(id: link_id..)
+  count = 0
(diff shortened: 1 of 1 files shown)


brucebolt (approved):
This looks fine to me ... but I wonder whether it'd be better as a migration, so it'll get run against other instances too (e.g. in development)?

richardTowers:
Hmm, maybe... I was worried it would be too slow as a migration. Is there some limit there? Or do they just run as long as the pod is up?

brucebolt:
> Hmm, maybe... I was worried it would be too slow as a migration. Is there some limit there? Or do they just run as long as the pod is up?

I think there may well be a time limit, as I recall a migration previously timing out 🤔

richardTowers:
I guess either way, everyone who wants to maintain a local database dump is going to have to run this long task (whether it comes from db:migrate, or if it's manual). I think I prefer the rake task, because it's resumable (if it times out, I can give it an id to start from).

PR #3253, opened 2025-04-03
Use content_id to associate actions with link sets instead of link_set_id

richardTowers:
Now that we have populated link_set_content_id and indexed it in all environments, we can start using it instead of link_set_id.

The actions table is the smallest and least risky of the changes we need to make. Previously actions related to link sets stored the content_id and the link_set_id, which is kind of redundant as they're both equivalent ways of referring to a link set. Now that we're removing the link_set_id in favour of content_id, we can only refer to the latter.

Unfortunately this means we have to give up the validation that an actions can't be for both an edition and a link_set. I don't think this is a huge concession though. It does mean if you look at an action that's not got anything to do with a link set, you'll be able to call `.link_set` on it, and get the link_set for the content_id of the edition. That feels okay to me though.

This will mean we can later remove the link_set_id column from actions, and eventually the id column from link_sets

### DIFF
--- a/app/models/action.rb
+++ b/app/models/action.rb
@@ -1,9 +1,8 @@
 class Action < ApplicationRecord
   belongs_to :edition, optional: true
-  belongs_to :link_set, optional: true
+  belongs_to :link_set, optional: true, foreign_key: :content_id, primary_key: :content_id
   belongs_to :event
 
-  validate :one_of_edition_link_set
   validates :action, presence: true
 
   def self.create_put_content_action(updated_draft, locale, event)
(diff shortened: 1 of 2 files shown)


mike3985 (approved):
This overall move didn't seem ideal at first glance. But since the one LinkSet-related type of action is distinguishable from the Edition-related types, we're not en route to losing any information. And since I'm struggling to find where any of these associations are even used (?), the convenience doesn't appear to be a big loss, either.

richardTowers:
Yeah, I'm not sure if they _are_ used. I think maybe they're there just in case we need to do some diagnostics in the rails console? So feels low risk to drop link_set_id.

PR #3255, opened 2025-04-03
Use content_id as primary key in link_set associations

richardTowers:
This changes the associations to join on link_set_content_id instead of joining through link_sets on link_set_id.

There are a couple of workarounds needed to ensure that link_set_id is still populated when links are created (either in patch link_set, or in the test factories). These should be removed when we're no longer using the link_set_id column.

Because content_id is now the pkey of LinkSet, we can't use LinkSet.last in the tests to get the most recent one anymore (it will get the one with the lexicographically last content_id, which is random). It's more explicit to look them up by content_id anyway.

This change will affect any queries using the associations. For example:

```
some_document.link_set_links
```

will now do

```
SELECT "links".* FROM "links" WHERE "links"."link_set_content_id" = ?
```

whereas before it would have done:

```
SELECT "links".* FROM "links"
INNER JOIN "link_sets" ON "links"."link_set_id" = "link_sets"."id"
WHERE "link_sets"."content_id" = ?
```

There are also several places where the code does the joins explicitly (e.g. `Document.joins(link_set: :links)`). In those cases we'll still join through the `link_sets` table following this PR, but we'll use:

```
INNER JOIN "links" ON "links"."link_set_content_id" = "link_sets"."content_id"
```

instead of

```
INNER JOIN "links" ON "links"."link_set_id" = "link_sets"."id"
```

We can clean these situations up in future PRs so we don't unnecessarily join through link_sets.

I've also added some scopes to Link, so we can get all the link_set_links or edition_links more conveniently. These will be used in future PRs.

### DIFF
--- a/spec/models/link_set_spec.rb
+++ b/spec/models/link_set_spec.rb
@@ -16,7 +16,16 @@ RSpec.describe LinkSet do
     end
 
     context "with only other links" do
-      let(:other_links) { [build(:link, link_set:)] }
+      let(:other_links) do
+        [
+          build(
+            :link,
+            link_set_content_id: link_set.content_id,
+            # TODO: ADR-009 - remove when we remove link_set_id
+            link_set_id: link_set.attributes["id"],
+          ),
+        ]
+      end
 
       it { is_expected.to be true }
     end
--- a/app/commands/v2/patch_link_set.rb
+++ b/app/commands/v2/patch_link_set.rb
@@ -19,6 +19,12 @@ module Commands
 
           payload_content_ids.uniq.each_with_index do |content_id, i|
             link_set.links.create!(
+              # link_set_id needs to be set explicitly, as active record is now
+              # using the link_set_content_id as the foreign key of the association.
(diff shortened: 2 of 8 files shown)


spec/models/link_set_spec.rb
```diff
@@ -16,7 +16,16 @@
...
+      let(:other_links) do
+        [
+          build(
+            :link,
+            link_set_content_id: link_set.content_id,
+            # TODO: ADR-009 - remove when we remove link_set_id
+            link_set_id: link_set.attributes["id"],
+          ),
+        ]
+      end
```
mike3985, line 28: This change and the spec example that it impacts are quite awkward

mike3985 (approved):

spec/models/link_set_spec.rb
richardTowers: Yeah. This should be temporary I think. Maybe I'm already at the point with one of my later stacked PRs where I could remove it... 🤔

PR #3256, opened 2025-04-03
Skip joining link_sets in graphql queries

richardTowers:
Now that the link_set_content_id column is populated in all environments, we can join directly from `links` to `documents` without the intermediate join through `link_sets`.

This is a small simplification, and the effect on performance is expected to be minimal. You can compare the query plans before and after here:

Before: https://www.pgexplain.dev/plan/673d9a6f-1df0-4a31-b993-9495c0ee1076
After: https://www.pgexplain.dev/plan/ab34f770-a24d-4792-b1a8-e3152e63eb30

Both take around 1ms to execute, and 2-5ms to plan. I think any actual difference in performance is smaller than the error bars on the measurements.

The code is slightly simpler though.

### DIFF
--- a/app/graphql/sources/person_current_roles_source.rb
+++ b/app/graphql/sources/person_current_roles_source.rb
@@ -6,12 +6,10 @@ module Sources
         .includes(
           document: {
             reverse_links: { # role -> role_appointment
-              link_set: {
-                documents: [
-                  :editions, # role_appointment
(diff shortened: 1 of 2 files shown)


brucebolt (approved):
Nice.  The queries are definitely easier to understand now we don't need that extra join.

PR #3258, opened 2025-04-03
Use link_set_content_id in Queries.Links

richardTowers:
This query is used in the link expansion code, so it is called a lot.

To build some confidence that this change doesn't break anything, I ran the link expansion of the ministers index page before and after this change, and recorded the `EXPLAIN (ANALYZE)` query plans.

To expand the ministers index page:

```ruby
e = Edition.live.joins(:document).find_by(base_path: "/government/ministers", document: { locale: "en" })
ep = Presenters::EditionPresenter.new(e)
result = ep.for_content_store(0)
File.write("before|after.json", result.to_json)
```

The before / after JSON results are identical, as they should be.

The query affected by this PR is made 457 times while expanding the ministers index page. For want of a better way of comparing the query plans, I've dumped them into this git repository: https://github.com/richardTowers/query-plan-comparisons

I've plotted the difference in planning and execution time here, and it looks like there's a fair improvement (although the old queries only took low numbers of ms, so it was already reasonably fast).

image

As an example, [plan 353 went from 9.712 ms execution time to 0.047 ms execution time](https://github.com/richardTowers/query-plan-comparisons/commit/d772b2e19865ed0f1b4def8a0adbb8cfe510a87d).

Using pgexplain to compare these:

- [before PR3258](https://www.pgexplain.dev/plan/c70fa026-42b7-4c1f-af0c-1ec130d778f5)
- [after PR3258](https://www.pgexplain.dev/plan/339fc7e0-c09a-4f8d-8cab-6d5d821dac3a)

Shows the main reason for the performance improvement. Before the change, postgres needed to look at three different indexes (`index_link_sets_on_content_id`, `index_links_on_link_set_id`, `index_links_on_link_type`) to find the nested links (i.e. find all the links with a particular content id and link type). After the change, it only needs to look at `index_links_on_link_set_content_id_and_link_type` because the content_id and the link_type are on the same table / in the same index.

| Before | After |
|----|----|
| image | image |

🚀

### DIFF
--- a/app/queries/links.rb
+++ b/app/queries/links.rb
@@ -104,7 +104,7 @@ module Queries
 
     def links_results
       Link
-        .joins(:link_set)
+        .link_set_links
         .where(where)
         .where.not(where_not)
         .order(link_type: :asc, position: :asc)
@@ -113,7 +113,7 @@ module Queries
 
     def where
       where = if mode == :from
-                { "link_sets.content_id": content_id }
+                { "links.link_set_content_id": content_id }
               else
                 { "links.target_content_id": content_id }
               end
@@ -133,7 +133,7 @@ module Queries
     end
 
     def link_content_id_field
-      mode == :from ? "links.target_content_id" : "link_sets.content_id"
+      mode == :from ? "links.target_content_id" : "links.link_set_content_id"
     end
 
     def group_results(results)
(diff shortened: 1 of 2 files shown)


callumknights (approved):
This looks like it's tested thoroughly, so I'm happy to approve this.

PR #3260, opened 2025-04-03
Use Renovate for Ruby version, GitHub Actions, and Docker images

yndajas:
[Trello](https://trello.com/c/KYb1Gz92/1653-use-renovate-for-ruby-version-maintenance)

Dependabot doesn't open PRs to upgrade Ruby. It would be good keep up to date where possible, and at least keep updates on our radar, so this sets up Renovate to open PRs to update Ruby. This also sets up Renovate to update GitHub Actions, but leaves Dependabot to manage updates to Ruby gems and Node packages

Renovate won't be enabled until govuk-helm-charts is updated: https://github.com/alphagov/govuk-helm-charts/pull/3111

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,7 +6,11 @@
     ":automergePr",
     ":automergeRequireAllStatusChecks"
   ],
-  "enabledManagers": ["ruby-version"],
+  "enabledManagers": [
+    "dockerfile",
+    "github-actions",
+    "ruby-version"
+  ],
   "labels": ["dependencies"],
   "packageRules": [
     {
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,25 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:best-practices",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:best-practices",
+    ":automergePatch",
+    ":automergePr",
+    ":automergeRequireAllStatusChecks"
+  ],
+  "enabledManagers": [
+    "docker-compose",
```
mike3985, line 10: I'm not sure we have anything Docker Compose-related inside the app's repository?

mike3985 (approved):
👍 other than the 1 comment, looks good, looking forward to seeing it in action

renovate.json
yndajas: Oh yeah, this should be "dockerfile"

yndajas: Oh no, I already included that. I'll remove this!

PR #3282, opened 2025-04-11
Produce datetimes from our GraphQL endpoint that look like Content API's

mike3985:
Content Store's datetimes are in GMT/BST. When they're stringified, its GMT ones end with timezone "+00:00" (and its BST ones end in "+01:00").

Until now GraphQL datetimes have been in UTC and when stringified they end with just a "Z".

We want to produce data as close to Content API's as is reasonable. This seems reasonable, right?

https://trello.com/c/ezCkqbaq/1665-make-graphql-version-of-world-index-the-same-as-the-regular-version

### DIFF
--- a/app/graphql/types/content_api_datetime.rb
+++ b/app/graphql/types/content_api_datetime.rb
@@ -1,0 +1,11 @@
+module Types
+  class ContentApiDatetime < Types::BaseScalar
+    def self.coerce_input(input_value, _context)
+      # N/A
+    end
+
+    def self.coerce_result(ruby_value, _context)
+      ruby_value.in_time_zone("Europe/London").iso8601
+    end
+  end
+end
--- a/spec/integration/graphql/generic_edition_spec.rb
+++ b/spec/integration/graphql/generic_edition_spec.rb
@@ -80,7 +80,7 @@ RSpec.describe "GraphQL" do
           base_path: "/my/withdrawn/edition",
           explanation: "for integration testing",
           document_type: "generic_type",
-          unpublished_at: "2024-10-27 17:00:00.000000000 +0000",
+          unpublished_at: Time.utc(2024, 10, 27, 17, 0, 0),
         )
       end
 
--- a/spec/integration/graphql/generic_edition_spec.rb
+++ b/spec/integration/graphql/generic_edition_spec.rb
@@ -2,7 +2,15 @@ RSpec.describe "GraphQL" do
   describe "generic edition" do
     before do
       document = create(:document, content_id: "d53db33f-d4ac-4eb3-839a-d415174eb906")
-      @edition = create(:live_edition, document:, document_type: "generic_type", base_path: "/my/generic/edition")
+      @edition = create(
+        :live_edition,
+        document:,
+        document_type: "generic_type",
+        base_path: "/my/generic/edition",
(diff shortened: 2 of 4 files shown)


app/graphql/types/content_api_datetime.rb
```diff
@@ -0,0 +1,11 @@
+module Types
+  class ContentApiDatetime < Types::BaseScalar
+    def self.coerce_input(input_value, _context)
+      # N/A
```
yndajas, line 4: **question:** are you saying this isn't needed because we don't take a `ContentApiDatetime` as an input value (argument?) anywhere?

app/graphql/types/content_api_datetime.rb
```diff
@@ -0,0 +1,11 @@
...
+    end
+
+    def self.coerce_result(ruby_value, _context)
+      ruby_value.in_time_zone("Europe/London").iso8601
+    end
```
yndajas, line 9: **suggestion:** I suppose readers can do their own research, but I wonder if it's useful to link to the docs that explain that [these are required methods when inheriting from Types::BaseScalar](https://graphql-ruby.org/type_definitions/scalars.html#custom-scalars)

spec/integration/graphql/generic_edition_spec.rb
```diff
@@ -80,7 +88,7 @@
           base_path: "/my/withdrawn/edition",
           explanation: "for integration testing",
           document_type: "generic_type",
-          unpublished_at: "2024-10-27 17:00:00.000000000 +0000",
+          unpublished_at: "2024-10-28 17:00:00.000000000 +0000",
```
yndajas, line 91: **suggestion:** maybe this should use `Time.utc` like the ones you added above? Seems like a nice way to distinguish between the specifically-formatted strings and the time-y things

spec/integration/graphql/roles_spec.rb
```diff
@@ -61,7 +61,7 @@
         schema_name: "role_appointment",
         details: {
           current: false,
-          started_on: Time.zone.local(2022, 10, 25),
+          started_on: Time.zone.local(2022, 11, 25),
           ended_on: Time.zone.local(2024, 7, 5),
```
yndajas, line 65: **question:** should these use `Time.utc` as well? Can't remember if we used local for a good reason

yndajas (approved):

app/graphql/types/content_api_datetime.rb
```diff
@@ -0,0 +1,11 @@
+module Types
```
yndajas, line 1: **thought:** should this commit message have a verb? 👀

spec/integration/graphql/generic_edition_spec.rb
mike3985: OK, I've done that in its own commit

spec/integration/graphql/roles_spec.rb
mike3985: Yep... Publishing API should always be UTC. And now that I've replaced it with `Time.utc`, it behaves the same.

I wonder if it took invoking that series of methods to summon the eventual expected string format, or something 🤷

app/graphql/types/content_api_datetime.rb
mike3985: Dunno, should it? 🙈

PR #3283, opened 2025-04-14
(553) Update embed codes to support Content ID aliases

pezholio:
Trello card: https://trello.com/c/fdJtXlmL/553-tech-spike-investigate-friendly-names-for-content-blocks-embed-codes

This updates the content block tools gem to support friendly names in embed codes, I've also had to tweak the embedded content finder a bit to ensure both UUIDs and friendly IDs are supported. We may change this in future, but for backwards compatibility, it seems sensible to have both.

The rest of the commits are mainly updates to tests to give us confidence that both methods work right the way through the flow.

### DIFF
--- a/app/services/embedded_content_finder_service.rb
+++ b/app/services/embedded_content_finder_service.rb
@@ -14,7 +14,8 @@ class EmbeddedContentFinderService
     when Hash
       value.map { |_, v| find_content_references(v) }.flatten
     when String
-      ContentBlockTools::ContentBlockReference.find_all_in_document(value)
+      content_references = ContentBlockTools::ContentBlockReference.find_all_in_document(value)
+      transform_aliases_to_content_ids(content_references)
     else
       []
     end
@@ -24,7 +25,7 @@ private
 
   def live_content_ids(content_references)
     found_editions = live_editions(content_references.uniq)
-    not_found_content_ids = content_references.map(&:content_id) - found_editions.map(&:content_id)
+    not_found_content_ids = content_references.map(&:identifier) - found_editions.map(&:content_id)
 
     if not_found_content_ids.any?
       GovukError.notify(CommandError.new(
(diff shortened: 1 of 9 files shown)


app/services/embedded_content_finder_service.rb
```diff
@@ -24,23 +25,44 @@ def find_content_references(value)
...
                           message: "Could not find any live editions for embedded content IDs: #{not_found_content_ids.join(', ')}",
                         ))
     end
-    content_references.map(&:content_id) - not_found_content_ids
+    content_references.map(&:identifier) - not_found_content_ids
+  end
+
+  def transform_aliases_to_content_ids(content_references)
+    embedded_aliases = content_references.select(&:identifier_is_alias?).map(&:identifier)
+    content_id_aliases = ContentIdAlias.where(name: embedded_aliases).map { |a| [a.name, a.content_id] }.to_h
```
Harriethw, line 41: I'm struggling a bit to understand the workflow of a ContentIdAlias...
Looking at `put_content`, [they are created when a param is passed in the `payload` ](https://github.com/alphagov/publishing-api/commit/15753423c1f19b279b7bfa450d0b6f24dc08c427)?  So would the `ContentIdAlias` be created/updated when a Block is published?

pezholio: And we create an alias using `friendly_id` here https://github.com/alphagov/whitehall/blob/a73ddeaad983bf3a2ac6282d72950083e420693c/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb#L11

app/services/embedded_content_finder_service.rb
```diff
@@ -24,23 +25,44 @@ def find_content_references(value)
...
+    content_references.map do |reference|
+      if reference.identifier_is_alias?
+        identifier = content_id_aliases[reference.identifier]
+        if identifier.nil?
+          GovukError.notify(CommandError.new(
```
Harriethw, line 46: minor/question: I wonder if we can't find the content_id if we don't return the Reference, as we can't do anything with it without the id? Or we strip them out where `identifier` is nil in `live_content_ids`?

app/services/embedded_content_finder_service.rb
```diff
@@ -24,23 +25,44 @@ def find_content_references(value)
...
+    not_found_content_ids = content_references.map(&:identifier) - found_editions.map(&:content_id)
 
     if not_found_content_ids.any?
       GovukError.notify(CommandError.new(
                           code: 422,
                           message: "Could not find any live editions for embedded content IDs: #{not_found_content_ids.join(', ')}",
                         ))
     end
-    content_references.map(&:content_id) - not_found_content_ids
+    content_references.map(&:identifier) - not_found_content_ids
```
Harriethw, line 36: minor/not really related but just trying to see any way to simplify! could we just return `found_editions.map(&:content_id)` here? and assign that earlier up to a variable so that we don't have to do the map twice?

Harriethw (approved):
🙌

app/services/embedded_content_finder_service.rb
pezholio: Good idea! 👍

app/services/embedded_content_finder_service.rb
pezholio: We need to return the identifiers rather than the found editions because we need to return duplicates if the same content block is referenced twice (see https://github.com/alphagov/publishing-api/pull/3129). That said, there's definitely scope to improve the logic, so I've added another commit. Will deploy to integration first to check it out, but it should be good 👍

PR #3287, opened 2025-04-15
Add document type to prometheus label for graphql requests

JonathanHallam:
This should allow us to build a funky table of average response times by content type or something else fun like that 

Trello - https://trello.com/c/d0zL17OH/1662-add-content-type-specific-metrics-to-dashboard

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -24,6 +24,8 @@ class GraphqlController < ApplicationController
         operation_name:,
       ).to_hash
 
+      set_prometheus_labels(result.dig("data", "edition"))
+
       if result.key?("errors")
         logger.warn("GraphQL query result contained errors: #{result['errors']}")
       else
@@ -75,5 +77,14 @@ private
     logger.error error.backtrace.join("\n")
 
     render json: { errors: [{ message: error.message, backtrace: error.backtrace }], data: {} }, status: :internal_server_error
+  end
+
+  def set_prometheus_labels(edition)
+    prometheus_labels = request.env.fetch("govuk.prometheus_labels", {})
+
+    request.env["govuk.prometheus_labels"] = prometheus_labels.merge(
+      document_type: edition["document_type"],
+      schema_name: edition["schema_name"],
+    )
   end
 end
--- a/spec/integration/graphql/generic_edition_spec.rb
+++ b/spec/integration/graphql/generic_edition_spec.rb
@@ -81,6 +81,24 @@ RSpec.describe "GraphQL" do
       expect(parsed_response).to eq(expected)
     end
 
+    it "sets the document type as a prometheus label" do
+      post "/graphql", params: {
+        query:
+          "{
+            edition(base_path: \"/my/generic/edition\") {
+              ... on Edition {
+                title
+                document_type
+                schema_name
+              }
+            }
+          }",
+      }
+
+      expect(request.env["govuk.prometheus_labels"][:document_type]).to eq(@edition.document_type)
+      expect(request.env["govuk.prometheus_labels"][:schema_name]).to eq(@edition.schema_name)
+    end
+
     context "when there is a withdrawn notice" do
       before do
         create(


callumknights (approved):
Nice 👍

app/controllers/graphql_controller.rb
```diff
@@ -76,4 +78,11 @@ def handle_error_in_development(error)
 
     render json: { errors: [{ message: error.message, backtrace: error.backtrace }], data: {} }, status: :internal_server_error
   end
+
+  def set_prometheus_labels(document_type)
+    prometheus_labels = request.env.fetch("govuk.prometheus_labels", {})
+    request.env["govuk.prometheus_labels"] = prometheus_labels.merge(
+      document_type: document_type,
```
yndajas, line 85: **suggestion:** not sure if you saw my [earlier comment on the branch](https://github.com/alphagov/publishing-api/commit/520121efe8b8ff9bbfa4fdeb91060221b88a4778#r155391037), but might we need schema name as well like with Government Frontend in order to group things like news articles, which have multiple document types? Document types and schema names apparently have a many to many relationship, and I suspect we mostly care about the schema name groupings

spec/integration/graphql/generic_edition_spec.rb
```diff
@@ -13,72 +13,90 @@
...
-                publishing_request_id
-                publishing_scheduled_at
-                rendering_app
-                scheduled_publishing_delay_seconds
-                schema_name
-                updated_at
-                withdrawn_notice {
-                  explanation
-                  withdrawn_at
+    context "when things are generic" do
```
yndajas, line 16: **question:** does this context add much value given we're already in a `describe "generic edition"` block?

app/controllers/graphql_controller.rb
JonathanHallam: I did not! Excellent shout!

spec/integration/graphql/generic_edition_spec.rb
```diff
@@ -81,6 +81,24 @@
...
 
+    it "sets the document type as a prometheus label" do
+      post "/graphql", params: {
+        query:
+          "{
+            edition(base_path: \"/my/generic/edition\") {
+              ... on Edition {
+                title
+                document_type
+                schema_name
```
yndajas, line 92: **issue:** this does bring in a new requirement that frontend queries request these two fields, otherwise these Prometheus labels will be `nil` (like you were seeing with the failing test before)

We could:
- try to throw the fields into the query somehow, but it's stringified and using GraphQL syntax so probably not simple to parse into an easily-manipulatable object
- execute a second query either via Active Record or using [multiplex](https://graphql-ruby.org/queries/multiplex.html), but to get the base path we'd still need to parse the query string, and this might introduce more database queries?
- create our own `PublishingApiSchema.execute` definition (we currently just inherit `GraphQL::Schema.execute`) to do one of the above options or something else
- rely on this being the case with all queries - we've already moved that way with [roles](https://github.com/alphagov/collections/commit/1ce63afa11aefe6e02c46bcd93f98c6931f7ea83) and the [world index](https://github.com/alphagov/collections/commit/6303986a70c23761f391b9a6cc39352c2dddbbb2), and the [news article query](https://github.com/alphagov/frontend/blob/main/app/queries/graphql/edition_query.rb) also requests these fields. The [ministers index query](https://github.com/alphagov/collections/blob/main/app/queries/graphql/ministers_index_query.rb) doesn't, and it [might not be on our radar to change that](https://trello.com/c/HcOHeixw/1663-make-graphql-version-of-ministers-index-the-same-as-the-regular-version) (have mentioned this to Mike - this might be an omission). This feels like the most viable option, but it does introduce a bit of a loose frontend dependency

yndajas (commented):
I think this is probably fine and the most sensible option, but there is a bit of a loose dependency worth noting

spec/integration/graphql/generic_edition_spec.rb
JonathanHallam: Yeah i think the last option is the best. And while yes it creates the dependency I'm not too worried. We'd see in our graphs if there were any that were missing. Also because we're doing this to see the worst performing things as long as it's missing for fast things it doesn't matter. I mean that's a really horrible approach to take but I think it works

PR #3289, opened 2025-04-16
Add `reshuffle` to `details` for GraphQL queries

brucebolt:
The `reshuffle` field is needed in https://github.com/alphagov/collections/pull/4053, therefore we need to be able to return that value in GraphQL queries.

This value allows us to show a 'reshuffle in progress' message on the ministers index page.

[Trello card](https://trello.com/c/HcOHeixw)

### DIFF
--- a/app/graphql/types/edition_type.rb
+++ b/app/graphql/types/edition_type.rb
@@ -148,6 +148,7 @@ module Types
       field :logo, Logo
       field :political, Boolean
       field :privy_counsellor, Boolean
+      field :reshuffle, GraphQL::Types::JSON
       field :role_payment_type, String
       field :seniority, Integer
       field :started_on, Types::ContentApiDatetime


app/graphql/types/edition_type.rb
```diff
@@ -148,6 +148,7 @@ class WhipOrganisation < Types::BaseObject
       field :logo, Logo
       field :political, Boolean
       field :privy_counsellor, Boolean
+      field :reshuffle, GraphQL::Types::JSON
```
yndajas, line 151: **question:** not sure if I'm reading the [schema](https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/ministers_index.jsonnet#L9-L11) right, but will this always be an object with a `message` key with a string value? If so, is it worth modelling it as that rather than any JSON?

yndajas (approved):

app/graphql/types/edition_type.rb
brucebolt: I did consider doing that, but thought it unnecessary for such a simple field.  I'm not really sure where to draw the line on how detailed we should implement fields like this?

PR #3290, opened 2025-04-16
Remove accidental n+1 loading link_sets

richardTowers:
When I removed the `includes ... link_set` from the query we use to get all the roles, I didn't realise we were explicitly calling .link_set on the roles below. This caused an n+1, where we'd issue a query to get the link set for every one of the roles. I only tested the main query, so I didn't spot this regression.

Calling the .strict_loading method on the relation will ensure that we don't accidentally do this again in future:

https://guides.rubyonrails.org/active_record_querying.html#strict-loading


yndajas (approved):
Thanks!

PR #3294, opened 2025-04-17
Only return published role appointments on ministers index page

brucebolt:
We are currently returning unpublished role appointments attached to a published role on the ministers index page (when rendered using GraphQL), resulting in ministers appearing to have roles that they do not hold.

For example, Angela Rayner has a current but unpublished role appointment which we are currently returning:

```
 #<Edition:0x0000ffff8f3c62e0
  id: 14474060,
  title: "The Rt Hon Angela Rayner MP - Secretary of State f...",
  public_updated_at: "2024-07-05 14:39:57.000000000 +0000",
  publishing_app: "whitehall",
  update_type: "major",
  phase: "live",
  analytics_identifier: nil,
  created_at: "2024-07-05 14:39:57.381491000 +0000",
  updated_at: "2024-07-10 12:33:27.619460000 +0000",
  document_type: "role_appointment",
  schema_name: "role_appointment",
  first_published_at: "2024-07-05 14:39:57.433063000 +0000",
  last_edited_at: "2024-07-05 14:39:57.386136000 +0000",
  state: "unpublished",
  user_facing_version: 1,
  base_path: nil,
  content_store: "live",
  document_id: 1602112,
  description: nil,
  publishing_request_id: "21-1720190397.253-10.13.22.63-2739",
  major_published_at: "2024-07-05 14:39:57.433063000 +0000",
  published_at: "2024-07-05 14:39:57.433063000 +0000",
  publishing_api_first_published_at: "2024-07-05 14:39:57.433063000 +0000",
  publishing_api_last_edited_at: "2024-07-05 14:39:57.386136000 +0000",
  auth_bypass_ids: [],
  details: {"current"=>true, "started_on"=>"2024-07-05T00:00:00+01:00", "person_appointment_order"=>1},
  routes: [],
  redirects: [],
  last_edited_by_editor_id: nil,
  rendering_app: nil>]
```

Adding an extra constraint here to ensure we only retrieve the published role appointments.

[Trello card](https://trello.com/c/HcOHeixw)

### DIFF
--- a/app/graphql/sources/person_current_roles_source.rb
+++ b/app/graphql/sources/person_current_roles_source.rb
@@ -19,7 +19,10 @@ module Sources
           document: { locale: "en" },
           reverse_links: { link_type: "role" },
           source_documents: { locale: "en" }, # role_appointment Documents
-          editions_documents: { document_type: "role_appointment" }, # role_appointment Editions
+          editions_documents: { # role_appointment Editions
+            document_type: "role_appointment",
+            state: "published",
+          },
           link_set_links: { target_content_id: person_content_ids, link_type: "person" },
         )
         .where("editions_documents.details ->> 'current' = 'true'") # editions_documents is the alias that Active Record gives to the role_appointment Editions in the SQL query
--- a/spec/graphql/sources/person_current_roles_source_spec.rb
+++ b/spec/graphql/sources/person_current_roles_source_spec.rb
@@ -16,4 +16,36 @@ RSpec.describe Sources::PersonCurrentRolesSource do
       expect(request.load).to eq([role_1, role_3])
     end
   end
+
+  it "does not return draft roles for a person" do
+    person_1 = create_person("Person 1")
+    published_role = create_role("Role")
+    draft_role = create_role("Draft role")
+    non_current_draft_role = create_role("Non-current draft role")
+    appoint_person_to_role(person_1, published_role)
+    appoint_person_to_role(person_1, draft_role, state: "draft")
+    appoint_person_to_role(person_1, non_current_draft_role, state: "draft", current: false)
+
+    GraphQL::Dataloader.with_dataloading do |dataloader|
+      request = dataloader.with(described_class).request(person_1.content_id)
+
+      expect(request.load).to eq([published_role])
+    end
+  end
+
+  it "does not return unpublished roles for a person" do
+    person_1 = create_person("Person 1")
+    published_role = create_role("Role")
+    unpublished_role = create_role("Unpublished role")
+    non_current_unpublished_role = create_role("Non-current unpublished role")
+    appoint_person_to_role(person_1, published_role)
+    appoint_person_to_role(person_1, unpublished_role, state: "unpublished")
+    appoint_person_to_role(person_1, non_current_unpublished_role, state: "unpublished", current: false)
+
+    GraphQL::Dataloader.with_dataloading do |dataloader|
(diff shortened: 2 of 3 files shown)


app/graphql/sources/person_current_roles_source.rb
```diff
@@ -23,6 +23,7 @@ def fetch(person_content_ids)
           link_set_links: { target_content_id: person_content_ids, link_type: "person" },
         )
         .where("editions_documents.details ->> 'current' = 'true'") # editions_documents is the alias that Active Record gives to the role_appointment Editions in the SQL query
+        .where("editions_documents.state = 'published'")
```
yndajas, line 26: **suggestion:** can this go into the main `.where` starting on line 17? The one above on line 27 is only needed because it's reading JSON, I think

spec/graphql/sources/person_current_roles_source_spec.rb
```diff
@@ -16,4 +16,36 @@
...
   end
+
+  it "does not return draft but current roles for a person" do
+    person_1 = create_person("Person 1")
+    role_1 = create_role("Role 1")
+    role_2 = create_role("Role 2")
+    role_3 = create_role("Role 3")
+    appoint_person_to_role(person_1, role_1)
+    appoint_person_to_role(person_1, role_2, current: false)
+    appoint_person_to_role(person_1, role_3, state: "draft")
```
yndajas, line 27: **suggestion:** is the first role current and live, the second non-current but live and the third current but draft? Might be good to label the roles with some of these details, because within this test I don't know the default behaviour of `create_role` or `appoint_person_to_role`

mike3985: Yeah, the default is **current** and **published**.

I think the difficulty is that those attributes belong to the role appointment and not to the role itself, so the inclusion of those details in the arguments to the helper method and not in the name of the roles reflects the data model more honestly (therefore, possibly more helpfully 🤷).

On reflection I think those defaults are very reasonable expectations to have about the role appointment that the helper method is creating... so I have taken your suggestion and renamed the roles, but mostly to reflect their deviation from those defaults :+1:

In fact, I've made some other changes to these spec examples. The preexisting example above tests published role appointments that're current/not current, so I've changed the setup of these new examples to include the states being tested +/- current. Does this makes sense?

app/graphql/sources/person_current_roles_source.rb
mike3985: Done

spec/graphql/sources/person_current_roles_source_spec.rb
```diff
@@ -16,4 +16,36 @@
...
+    appoint_person_to_role(person_1, non_current_draft_role, state: "draft", current: false)
+
+    GraphQL::Dataloader.with_dataloading do |dataloader|
+      request = dataloader.with(described_class).request(person_1.content_id)
+
+      expect(request.load).to eq([published_role])
+    end
+  end
+
+  it "does not return unpublished roles for a person" do
```
yndajas, line 36: **suggestion:** given these two tests are identical (I think?), I wonder if it's worth merging them (if having two assertions doesn't feel horrible) or removing the excluded but not described role from each?

yndajas (approved):

spec/graphql/sources/person_current_roles_source_spec.rb
yndajas: Ah, they aren't identical. The differences are just hard to spot!

PR #3297, opened 2025-04-17
Change the org name earch ype for UKHSA

lauraghiorghisor-tw:
We're chaning the organisation name field type from text to identifiers in search-api. This typically requires a reindex; in order to avoid that, we're renaming the field.

[Trello](https://trello.com/c/zubgVL1V/3569-specialist-finder-for-ukhsa-data-releases)


minhngocd (approved):
Commit message has typos that probably should be fixed

PR #3311, opened 2025-04-30
Send Content Block update jobs to the HIGH_QUEUE

pezholio:
We’ve had a lot of experiences with content block updates getting stuck behing bulk update jobs, which is causing confusion for users when carrying out user research, as well as causing intermittent failure with e2e tests.

Because content block updates are a different type of update to other types of dependency resolution updates (for example, listing pages), it’s more important that they’re updated in a timely manner.


yndajas (approved):
Might be good to explain what makes them higher priority rather than just "different" in the commit message, but in principle I think this is fine, given the traffic should be relatively low?

> Because content block updates are a different type of update
to other types of dependency resolution updates (for example,
listing pages), it’s more important that they’re updated in
a timely manner.

PR #3319, opened 2025-05-02
Correctly handle unpublished editions in GraphQL

brucebolt:
For all unpublished editions, we were previously returning the full edition even if it had been unpublished.  Additionally, we were including all unpublished linked editions within links.

This PR addresses these issues:
1. Ensures we return the information needed by the frontend application to present the unpublished edition correctly to the user and not return the edition itself (unless withdrawn).
2. Removes all unpublished (including withdrawn) editions from links, unless they are one of a specified list of types.  This matches [the existing behaviour](https://github.com/alphagov/publishing-api/blob/43f5d1f937d6986edcdac920c4542395d5e392be/lib/link_expansion.rb#L100-L107) for link expansion and can be seen by looking at a content item (e.g. [this page](https://www.gov.uk/api/content/government/collections/guidance-notes-for-regulated-immigration-advisers) which has a withdrawn edition in the `documents` links, but this is not included in the content item).

Note: the `substitute` unpublishing type does not need to be covered as these are never presented to users and cannot be requested by frontend applications.

[Trello card](https://trello.com/c/oD9MDupC)

### DIFF
--- a/spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
+++ b/spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
@@ -42,4 +42,48 @@ RSpec.describe Sources::ReverseLinkedToEditionsSource do
       expect(request.load).to eq([source_edition_1, source_edition_2])
     end
   end
+
+  context "when the linked item is unpublished" do
+    %w[children parent related_statistical_data_sets].each do |link_type|
+      it "includes unpublished links when they are of the permitted type #{link_type}" do
+        target_edition = create(:edition, content_store: "live")
+
+        link_set_linked_edition = create(:withdrawn_unpublished_edition, content_store: "live")
+        link_set = create(:link_set, content_id: link_set_linked_edition.content_id)
+        create(:link, link_set:, target_content_id: target_edition.content_id, link_type:)
+
+        edition_linked_edition = create(:withdrawn_unpublished_edition,
+                                        content_store: "live",
+                                        links_hash: {
+                                          link_type => [target_edition.content_id],
+                                        })
+
+        GraphQL::Dataloader.with_dataloading do |dataloader|
+          request = dataloader.with(described_class, content_store: target_edition.content_store).request([target_edition, link_type])
+
+          expect(request.load).to eq([link_set_linked_edition, edition_linked_edition])
+        end
+      end
+    end
+
+    it "does not include unpublished links when they are of another type" do
+      target_edition = create(:edition, content_store: "live")
+
+      link_set_linked_edition = create(:withdrawn_unpublished_edition, content_store: "live")
+      link_set = create(:link_set, content_id: link_set_linked_edition.content_id)
+      create(:link, link_set:, target_content_id: target_edition.content_id, link_type: "test_link")
+
+      create(:withdrawn_unpublished_edition,
+             content_store: "live",
+             links_hash: {
+               "test_link" => [target_edition.content_id],
+             })
+
+      GraphQL::Dataloader.with_dataloading do |dataloader|
+        request = dataloader.with(described_class, content_store: target_edition.content_store).request([target_edition, "test_link"])
+
+        expect(request.load).to eq([])
+      end
+    end
+  end
 end
--- a/app/graphql/sources/reverse_linked_to_editions_source.rb
+++ b/app/graphql/sources/reverse_linked_to_editions_source.rb
@@ -49,6 +49,8 @@ module Sources
       ).order("editions.id")
 
       all_editions.each_with_object(link_types_map) { |edition, hash|
+        next if edition.state == "unpublished" && %w[children parent related_statistical_data_sets].exclude?(edition.link_type)
+
         hash[[edition.target_content_id, edition.link_type]] << edition
       }.values
     end


app/graphql/sources/linked_to_editions_source.rb
```diff
@@ -78,6 +78,8 @@ def fetch(editions_and_link_types)
         .order(link_type: :asc, position: :asc)
 
       all_editions.each_with_object(link_types_map) { |edition, hash|
+        next unless %w[children parent related_statistical_data_sets].include?(edition.link_type) || edition.state != "unpublished"
```
yndajas, line 81: **suggestion:** although this mirros the link expansion code, I wonder if it's worth flipping this since `edition.state != "unpublished"` is easier to parse and gives useful context for the other side of the expression?

```rb
next unless edition.state != "unpublished" || %w[children parent related_statistical_data_sets].include?(edition.link_type)
```

---

Same for reverse links

yndajas, line 81: **note:** probably doesn't matter, but the link expansion equivalent uses an array of symbols whereas this uses an array of strings for the allowed link types

spec/graphql/sources/linked_to_editions_source_spec.rb
```diff
@@ -105,4 +105,54 @@
...
     end
   end
+
+  context "when the linked item is unpublished" do
+    %w[children parent related_statistical_data_sets].each do |link_type|
+      it "includes unpublished links when they are of the permitted type #{link_type}" do
+        target_edition_1 = create(:edition, content_store: "live")
+        target_edition_2 = create(:withdrawn_unpublished_edition, content_store: "live")
+        target_edition_3 = create(:edition, content_store: "live")
+        target_edition_4 = create(:withdrawn_unpublished_edition, content_store: "live")
```
yndajas, line 115: **suggestion:** more of a personal preference, but I'd maybe label these along the lines of:
- `edition_linked_edition`
- `withdrawn_edition_linked_edition`
- `link_set_linked_edition`
- `withdrawn_link_set_linked_edition`

A bit long, but I find it easier to understand the intention, follow along, and check for code correctness with the more explicit labels

---

Similar for reverse links/source editions

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
```diff
@@ -42,4 +42,48 @@
...
   end
+
+  context "when the linked item is unpublished" do
+    %w[children parent related_statistical_data_sets].each do |link_type|
+      it "includes unpublished links when they are of the permitted type #{link_type}" do
+        target_edition = create(:edition, content_store: "live")
+
+        source_edition_1 = create(:withdrawn_unpublished_edition, content_store: "live")
+        link_set = create(:link_set, content_id: source_edition_1.content_id)
+        create(:link, link_set: link_set, target_content_id: target_edition.content_id, link_type:)
```
yndajas, line 53: **nitpick:** `link_set: link_set` -> `link_set:`

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
```diff
@@ -42,4 +42,48 @@
...
+             links_hash: {
+               "test_link" => [target_edition.content_id],
+             })
+
+      GraphQL::Dataloader.with_dataloading do |dataloader|
+        request = dataloader.with(described_class, content_store: target_edition.content_store).request([target_edition, "test_link"])
+
+        expect(request.load).to eq([])
+      end
+    end
```
yndajas, line 87: **question:** should there be an edition link in this one too?

yndajas (approved):
Excuse the number of comments, but nothing major here, so I'm approving!

app/graphql/types/query_type.rb
```diff
@@ -8,7 +8,37 @@ class QueryType < Types::BaseObject
...
+        .find_by(base_path:)
+
+      return unless edition
+
+      if edition.unpublishing && !edition.unpublishing.withdrawal?
+        unpublishing_data = {
+          "type" => edition.unpublishing.type,
+        }
+
+        if edition.unpublishing.type == "gone"
```
yndajas, line 23: **thought:** one thing I forgot to mention: you could `case` statement this, depending on your palate for case statements

app/graphql/types/query_type.rb
```diff
@@ -8,7 +8,21 @@ class QueryType < Types::BaseObject
...
     def edition(base_path:, content_store:)
-      Edition.where(content_store:).find_by(base_path:)
+      edition = Edition
+        .includes(:unpublishing)
+        .where(content_store:)
+        .find_by(base_path:)
+
+      return unless edition
+
+      if edition.unpublishing && !edition.unpublishing.withdrawal?
```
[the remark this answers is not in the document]
brucebolt: I did originally write this, but then added an explicit return instead as that felt clearer to me.

spec/integration/graphql/generic_edition_spec.rb
```diff
@@ -99,6 +99,41 @@
...
+          expect(parsed_response.dig(:data, :edition)).to be_nil
+        end
+
+        it "includes the unpublishing type and other data in the error" do
+          parsed_response = JSON.parse(response.body).deep_symbolize_keys
+          expect(parsed_response.dig(:errors, 0, :message)).to eq("Edition has been unpublished")
+          expect(parsed_response.dig(:errors, 0, :extensions, :type)).to eq("gone")
+          expect(parsed_response.dig(:errors, 0, :extensions, :alternative_path)).to eq(edition.unpublishing.alternative_path)
+          expect(parsed_response.dig(:errors, 0, :extensions, :explanation)).to eq(edition.unpublishing.explanation)
+          expect(parsed_response.dig(:errors, 0, :extensions, :public_updated_at)).to eq(edition.unpublishing.unpublished_at.in_time_zone("Europe/London").iso8601)
```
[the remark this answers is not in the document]
brucebolt: Yes, I'll do that.  This definitely looks messy now I'm looking at it again.

app/graphql/sources/linked_to_editions_source.rb
brucebolt: Good idea.  I've swapped them both round.

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
brucebolt: We used to have a rubocop rule for this.  Have updated as that doesn't appear to be in place anymore.

app/graphql/types/query_type.rb
brucebolt: I think a case here is probably tidier and easier to understand, so I've put that in place.

spec/graphql/sources/linked_to_editions_source_spec.rb
brucebolt: Have updated the names in both the direct and reverse link specs.

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
brucebolt: There is (lines 76-80 in the snippet), but is hasn't got a variable name as the variable never gets reused (so rubocop requires the assignment to be removed).

app/graphql/sources/linked_to_editions_source.rb
yndajas: Or even this, to avoid the unless not

```rb
next if edition.state == "unpublished" && %w[children parent related_statistical_data_sets].exclude?(edition.link_type)
```

(TIL `exclude?` is a thing in Rails)

yndajas: Oh and so is `in?` 📖

```rb
edition.link_type.in?(%w[children parent related_statistical_data_sets])
```

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
yndajas: Huh

app/graphql/sources/linked_to_editions_source.rb
brucebolt: Nice.  I'll go for `exclude?`.

PR #3321, opened 2025-05-02
ADR009 Part 7 - drop link_set_id indexes and foreign key constraints

richardTowers:
As per [ADR009 Part 7](https://github.com/alphagov/publishing-api/blob/main/docs/arch/adr-009-change-linksets-primary-key-to-content-id.md#7---drop-indexes-and-constraints-on-link_set_id), we can remove the unused indexes and foreign key constraints on link_set_id, now that we're using link_set_content_id everywhere.

I've done a quick check to confirm that the indexes are no longer being used by running `rake pg_extras:index_info` and comparing the count of index scans for the ones I'm removing over a ~40 minute period. It hasn't increased at all in that time, which supports my already high level of confidence that these indexes are no longer used. I'll check the current numbers against fresh values next week, just to be doubly sure. Current production counts:

```
% k exec deploy/publishing-api -- rake pg_extras:index_info 2>/dev/null | awk -F '|' '/link_set_id/ { print $2 $6 }'
 index_links_on_link_set_id                                6094270031
 index_links_on_link_set_id_and_target_content_id          10190437328
 index_links_on_link_set_id_and_link_type                  1535748454
 ```

Only a few more steps after this one 😅

### DIFF
--- a/db/migrate/20250502153736_remove_indexes_from_links.rb
+++ b/db/migrate/20250502153736_remove_indexes_from_links.rb
@@ -1,0 +1,7 @@
+class RemoveIndexesFromLinks < ActiveRecord::Migration[8.0]
+  def change
+    remove_index :links, name: "index_links_on_link_set_id"
+    remove_index :links, name: "index_links_on_link_set_id_and_target_content_id"
+    remove_index :links, name: "index_links_on_link_set_id_and_link_type"
+  end
+end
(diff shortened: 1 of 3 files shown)


richardTowers:
Moving to draft, as I haven't tested that the remove_indexes migration works in its new reversible form (thanks rubocop). I'll need to get a fresh database dump to test that I think. Next week.

richardTowers:
Tested running and rolling back the migrations locally on a fresh copy of the DB.

If we do need to roll them back, I think the remove foreign key one might need special attention (by default they're added and validated at the same time, which locks the table, so it's best to do validate: false, but I haven't done that in the removal). Very unlikely we'll need to do that though.

I rechecked the index scan counts and they haven't moved since Friday, so I'm very confident they're no longer being used.

PR #3330, opened 2025-05-07
ADR-009 - Remove the link_set_id column

richardTowers:
Now that we've ignored the link_set_id column, it's safe to remove it.

The safety_assured block tells the strong_migrations gem that we've done the steps needed to make this migration safe, so it should allow it to proceed.


yndajas (approved):
I am assured of our safety

PR #3339, opened 2025-05-13
(CM-64) Add new fields to contact

pezholio:
This adds some additional embedded objects to the spike contact schema, as well as adding an optional `order` param to the `embedded_object` util, to allow us to specify the order fields appear in.


content_schemas/formats/shared/utils/content_block_utils.jsonnet
```diff
@@ -11,7 +11,7 @@
                 type: "string"
               }
             } + properties,
-            order: ["title"] + std.objectFields(properties),
```
Harriethw, line 14: if I understand right, then we have two sources of truth for the order of fields, the schema and our config? Just wondering if we keep the ordering in one place, e.g. in our config as it's kind of presentation logic? Unless `order` is something the schema should defs know about?

pezholio: Yeah, I realised that after pushing this. It would probably be best to dump this entirely and have the order as a config thing. Whaddyareckon?

Harriethw: > It would probably be best to dump this entirely and have the order as a config thing. Whaddyareckon?

yah i think this would make sense for me :)

Harriethw (approved):

PR #3343, opened 2025-05-14
(CM-97) Fetch organisations from documents with link set links, as well as edition links

pezholio:
We now want to fetch linked Organisations where the link is a link set link (in the case of Mainstream docs) or an edition link.


Harriethw (approved):
good sleuthing! 
Did you find in your travels any documentation on the Link Sets/where or why they would be used over regular Links?  [I found this but not much info there](https://docs.publishing.service.gov.uk/repos/publishing-api/model.html#linkset)

pezholio:
From previous conversations, I think LinkSets are deprecated, and it's recommended to use edition links instead. I don't think there's any documentation to that effect though.

PR #3382, opened 2025-06-05
Add govuk:content-has-history meta tags to false positive list in graphql diff script

callumknights:
A seperate bug is causing these meta-tags to be applied where they shouldn't be, so they need to be left out of the diff script until that is fixed.

Trello: https://trello.com/c/XqtBMIiR/1729-add-the-govukcontent-has-history-meta-tag-to-the-comparison-script-so-we-dont-get-a-load-of-false-positives-when-comparing-docum

### DIFF
--- a/script/diff_graphql/functions.sh
+++ b/script/diff_graphql/functions.sh
@@ -25,6 +25,7 @@ function curl_and_strip_hashes() {
         -e 's/nonce="[^"]{22}=="/nonce="HASH=="/g' \
         -e 's/ (aria-labelledby|for|id)="([^"]+)-[a-z0-9]{8}"/ \1="\2-HASH"/g' \
         -e 's/<meta name="govuk:updated-at" content=".*">/<meta name="govuk:updated-at" content="TIMESTAMP">/' \
+        -e '/<meta name="govuk:content-has-history" content=".*">/d' \
     > "$output_path"
 }
 


brucebolt (commented):
Whilst this is running, I get the error `sed: 1: "s/<meta name="govuk:con ...": unescaped newline inside substitute pattern` printed to the output.

Do I need a specific version of sed?

callumknights:
Nope, this is my fault and I didn't notice the error is amongst the other logs. Now I'm puzzle over where I've left a newline in

brucebolt:
This is the (rather unhelpful) error that sed returns if you don't terminate the regex.  For a substitution (with the `s` at the start), you'll need three occurrences of `/` for sed to stop complaining.

yndajas:
> This is the (rather unhelpful) error that sed returns if you don't terminate the regex. For a substitution (with the `s` at the start), you'll need three occurrences of `/` for sed to stop complaining.

Yeah this syntax isn't quite right for the `s` command, but you probably want to delete these lines with `d` (after the match, and no `s`) rather than substitute in this case too avoid a diff of just line breaks

https://www.gnu.org/software/sed/manual/html_node/sed-commands-list.html

callumknights:
🙌 Thank you @yndajas! 🙌

brucebolt (approved):

PR #3384, opened 2025-06-05
Get GraphQL links that match primary locale or "en"

mike3985:
This reapplies a new-and-improved version of the final commit from https://github.com/alphagov/publishing-api/pull/3371. We decided to revert that commit this morning because of an issue it introduced to our live GraphQL-backed pages.

The issue resulted from my slapdash partitioning of `row_number()`. Linked-to Editions were being compared and scored against each other when they should've been in separate windows, which meant that Editions we actually wanted were inadvertently excluded from the database query's results.

https://trello.com/c/Dqvtk09M/1668-fix-diff-we-only-return-the-english-links-even-if-a-translation-is-requested

### DIFF
--- a/app/graphql/sources/linked_to_editions_source.rb
+++ b/app/graphql/sources/linked_to_editions_source.rb
@@ -3,7 +3,8 @@ module Sources
     # rubocop:disable Lint/MissingSuper
     def initialize(content_store:, locale:)
       @content_store = content_store.to_sym
-      @locale = locale
+      @primary_locale = locale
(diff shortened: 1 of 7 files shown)


mike3985:
I've tested this branch in integration (i.e. for diffing Content Store pages with GraphQL ones). This new version of the code resolves those issues with the ministers index, those 2 role pages and all the rest of the examples that Ynda found yesterday (minus the ones that're also different in production).

PR #3385, opened 2025-06-06
Add a new special route for HMPO webchat

GDSNewt:
[Adds a new special route](https://github.com/alphagov/publishing-api/blob/main/docs/admin-tasks.md#publishing-special-routes) for the HMPO webchat view, this is required as the publishing app that generated this route previously(contacts) is being retired. 

It may be the case that we can remove this special route eventually, should we render this page through another publishing app. 

[Trello](https://trello.com/c/QrArPjwt)

### DIFF
--- a/lib/data/special_routes.yaml
+++ b/lib/data/special_routes.yaml
@@ -162,6 +162,12 @@
     :parent:
       - "db95a864-874f-4f50-a483-352a5bc7ba18"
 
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):

ryanb-gds (approved):

JonathanHallam (approved):
👍

PR #3423, opened 2025-06-26
fix: make the destroy work in batches

KludgeKML:
- this is 1.4 million records, too big to load in one gulp


mike3985 (approved):
Ah, good point.

I wonder if the batch size could safely be a lot bigger, though? I don't have any recent experience of using it in practice but I do recall the default being 1000

KludgeKML:
> Ah, good point.
> 
> I wonder if the batch size could safely be a lot bigger, though? I don't have any recent experience of using it in practice but I do recall the default being 1000

It probably could (for instance, the one-off job ran okay with 5,000 records), but honestly the drag is probably the individual destroys rather than the batching. I'd rather be a bit conservative here.

PR #3442, opened 2025-07-09
(CM-148) Include content block's changes in host change notes

pezholio:
This makes some updates to the downstream payload so when a content block is updated with a `major` update type,  we can trigger email alerts for dependent content.

We've had to make a few changes here, namely:

- Adding the change notes for dependent content blocks to the `EditionPresenter` payload
- Removing the recently added `content_block` update type (Additional work in Whitehall here https://github.com/alphagov/whitehall/pull/10405)
- Changing the `public_updated_at` value to be the `public_updated_at` of a content block, if the edition is republished by a content block

I've also added an ADR to document all this - would appreciate a look over by some Publishing Platform devs too, to ensure I'm on the right track 👍

### DIFF
--- a/app/downstream_payload.rb
+++ b/app/downstream_payload.rb
@@ -1,10 +1,11 @@
 class DownstreamPayload
-  attr_reader :edition, :payload_version, :draft
+  attr_reader :edition, :payload_version, :draft, :triggered_by_edition
 
-  def initialize(edition, payload_version, draft: false)
+  def initialize(edition, payload_version, draft: false, triggered_by_edition: nil)
     @edition = edition
     @payload_version = payload_version
     @draft = draft
+    @triggered_by_edition = triggered_by_edition
   end
 
   delegate :state, to: :edition
@@ -41,7 +42,7 @@ private
   end
 
   def content_presenter
-    @content_presenter ||= Presenters::EditionPresenter.new(edition, draft:)
+    @content_presenter ||= Presenters::EditionPresenter.new(edition, draft:, triggered_by_edition:)
   end
 
   def redirect_presenter
--- a/app/presenters/change_history_presenter.rb
+++ b/app/presenters/change_history_presenter.rb
@@ -7,39 +7,44 @@ module Presenters
     end
 
     def change_history
-      details[:change_history] || presented_change_notes
+      SymbolizeJSON.symbolize(
+        change_notes.as_json,
+      )
     end
 
   private
 
-    def details
-      SymbolizeJSON.symbolize(edition.details)
-    end
-
-    def presented_change_notes
-      SymbolizeJSON.symbolize(
-        change_notes
-          .pluck(:note, :public_timestamp)
-          .map { |note, timestamp| { note:, public_timestamp: timestamp } }
-          .as_json,
-      )
-    end
-
     def change_notes
-      ChangeNote
-        .joins(:edition)
-        .where(editions: { document: })
-        .where("user_facing_version <= ?", version_number)
-        .where.not(public_timestamp: nil)
-        .order(:public_timestamp)
+      all_change_notes
+        .map { |h| format_change_note(h) }
+        .sort_by { |h| Time.zone.parse(h[:public_timestamp]) }
+        .reverse
     end
 
-    def version_number
-      edition.user_facing_version
+    def format_change_note(history)
+      {
+        note: history[:note],
+        public_timestamp: convert_timestamp_to_utc(history[:public_timestamp]).to_s,
+      }
     end
 
-    def document
-      edition.document
+    def all_change_notes
+      query_result + details_change_notes
+    end
+
+    def details_change_notes
+      @details_change_notes ||= edition.details.fetch(:change_history, [])
+    end
+
+    def query_result
+      Presenters::Queries::ChangeHistory.new(edition, include_edition_change_history: details_change_notes.blank?)
+                                        .call
+                                        .pluck(:note, :public_timestamp)
+                                        .map { |note, timestamp| { note:, public_timestamp: timestamp } }
+    end
+
+    def convert_timestamp_to_utc(timestamp)
+      timestamp.is_a?(String) ? Time.zone.parse(timestamp).utc : timestamp.utc
     end
   end
 end
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -25,9 +25,10 @@ module Presenters
       withdrawn
     ].freeze
 
-    def initialize(edition, draft: false)
+    def initialize(edition, draft: false, triggered_by_edition: nil)
       @edition = edition
       @draft = draft
+      @triggered_by_edition = triggered_by_edition
     end
 
     def for_content_store(payload_version)
@@ -53,6 +54,7 @@ module Presenters
         .merge(document_supertypes)
         .merge(withdrawal_notice)
         .merge(publishing_request_id)
+        .merge(public_updated_at)
     end
 
     def expanded_links
@@ -65,7 +67,7 @@ module Presenters
 
   private
 
-    attr_reader :draft, :edition
+    attr_reader :draft, :edition, :triggered_by_edition
 
     def auth_bypass_ids
       return {} unless draft
@@ -169,6 +171,12 @@ module Presenters
       else
         {}
       end
+    end
+
+    def public_updated_at
+      return {} unless triggered_by_edition&.is_content_block?
+
+      { public_updated_at: triggered_by_edition.to_h[:public_updated_at] }
     end
   end
 end
--- a/docs/arch/adr-010-add-content-block-information-to-the-downstream-payload.md
+++ b/docs/arch/adr-010-add-content-block-information-to-the-downstream-payload.md
@@ -1,0 +1,35 @@
+# Decision Record: Include Content Block Change Notes in Downstream Payloads
+
+## Context
+
+The content modelling team needs email alerts to be triggered when a **major update** is made to a **content block**, 
+especially when that update affects other pieces of content that embed it.
+
+This is challenging because changes to dependent content happen indirectly - when the embedding content is republished, 
+it pulls in the updated content block. However, to users subscribed to email alerts for that content, these changes 
+are invisible since no alert is triggered (even though the content has effectively changed).
+
+To fix this, we need to ensure that change notes from content blocks are sent to the 
+[email-alert-service](https://github.com/alphagov/email-alert-service), so that subscribers receive notifications when 
+embedded content is updated.
+
+## Solution
+
+We will update the `EditionPresenter` to include **change notes from embedded content blocks**, in addition to the 
+change notes for the content itself.
+
+This involves:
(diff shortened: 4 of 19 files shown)


edavey (approved):
Thanks for the clearer presentation. I was able to follow along and learn a lot in the process. I've made a few minor suggestions which might add clarity in places. Where I've got the wrong end of the stick, apologies and please ignore!

docs/arch/adr-010-add-content-block-information-to-the-downstream-payload.md
```diff
@@ -0,0 +1,32 @@
+# Decision Record: Include Content Block Change Notes in Downstream Payloads
+
+## Context
+
+The content modelling team needs email alerts to be triggered when a **major update** is made to a **content block**, 
+especially when that update affects other pieces of content that embed it.
+
+This is challenging because changes to dependent content happen indirectly - when the embedding content is republished, 
+it pulls in the updated content block. However, to users subscribed to email alerts for that content, these changes 
+are invisible since no alert is triggered (even though the content has effectively changed).
```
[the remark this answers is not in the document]
pezholio: Yeah, exactly, and also, they won't really care if it's a content block that's been updated. All they care about is the fact that the content's changed.

callumknights (dismissed):
We've been through this as a team, but have only had a chance to look at the first few commits. We'll have a look at the rest once you've had a chance to address our current comments.

app/presenters/change_history_presenter.rb
```diff
@@ -7,39 +7,20 @@ def initialize(edition)
     end
 
     def change_history
-      details[:change_history] || presented_change_notes
```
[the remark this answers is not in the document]
pezholio: It looks like for that, we're removing any change notes without timestamps to appease the schema. Looking at https://github.com/alphagov/publishing-api/blob/5b88b2d8a61d522d36a6b8d5e70a396bb8d52258/app/commands/v2/publish.rb#L47 I can see that change notes get created from the change history within the details hash, so these should match

app/downstream_payload.rb
```diff
@@ -1,10 +1,11 @@
 class DownstreamPayload
-  attr_reader :edition, :payload_version, :draft
+  attr_reader :edition, :payload_version, :draft, :triggered_by_edition
 
-  def initialize(edition, payload_version, draft: false)
+  def initialize(edition, payload_version, draft: false, triggered_by_edition: nil)
```
yndajas, line 4: **issue:** we think this should come before the previous commit where it's being used

app/presenters/change_history_presenter.rb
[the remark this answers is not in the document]
pezholio: Ah fun. In that case, do you think it'd be better to go back to the original behaviour, but merge the change note hash and the query result, and converting the timestamps to UTC in the process?

app/presenters/edition_presenter.rb
```diff
@@ -25,9 +26,10 @@ class EditionPresenter
       withdrawn
     ].freeze
 
-    def initialize(edition, draft: false)
+    def initialize(edition, draft: false, triggered_by_edition: nil)
```
yndajas, line 29: **issue:** we think this should come before the previous commit where it's being used

app/presenters/edition_presenter.rb
```diff
@@ -170,5 +173,11 @@ def publishing_request_id
         {}
       end
     end
+
+    def public_updated_at
+      edition_to_use = triggered_by_edition&.is_content_block? ? triggered_by_edition : edition
+
+      { public_updated_at: edition_to_use.to_h[:public_updated_at] }
```
yndajas, line 180: **suggestion:** instead of exclude this via `NON_PRESENTED_PROPERTIES`, could we continue defaulting to including the field from the edition and then only overwrite by exception? This feels a little simpler to us

```rb
      return {} unless triggered_by_edition&.is_content_block?

      { public_updated_at: triggered_by_edition[:public_updated_at] }
```

spec/integration/content_block_publish_spec.rb
```diff
@@ -4,11 +4,11 @@
   let(:document_type) { "content_block_pension" }
   let!(:content_block_superseded_edition) { create(:edition, document: content_block_document, state: "superseded", content_store: nil, user_facing_version: 1, document_type:) }
   let!(:content_block_live_edition) { create(:edition, document: content_block_document, state: "published", content_store: "live", user_facing_version: 2, document_type:) }
-  let(:content_block) { create(:draft_edition, update_type:, document: content_block_document, user_facing_version: 3, document_type:) }
+  let(:content_block) { create(:draft_edition, update_type:, document: content_block_document, user_facing_version: 3, document_type:, public_updated_at: 1.minute.ago) }
 
   let!(:change_note) { create(:change_note, note: "Some note goes here", edition: content_block, created_at: 1.day.ago) }
 
-  let(:dependent_content) { create_list(:edition, 2, state: "published", content_store: "live") }
+  let(:editions_with_content_blocks) { create_list(:edition, 2, state: "published", content_store: "live", public_updated_at: 5.days.ago) }
```
yndajas: **issue:** looks like there's a `let` rename here that should have been fixed up to the earlier commit where it's introduced?

spec/integration/content_block_publish_spec.rb
```diff
@@ -0,0 +1,86 @@
+RSpec.describe "Content Block Publication" do
```
yndajas, line 1: **nitpick:** would be good for the describe and the file name to match

docs/arch/adr-010-add-content-block-information-to-the-downstream-payload.md
```diff
@@ -0,0 +1,33 @@
...
+- Including these change notes in the payload alongside those of the host document
+
+## Consequences
+
+Currently, the `EditionPresenter` returns the `change_notes` from the `details` hash of the edition if they are present.
+
+To ensure consistency and accuracy, we will now fetch all change notes directly from the Publishing API’s database when 
+presenting the edition. This also addresses discrepancies in time zones - specifically, Whitehall's Edition#details 
+uses local time, whereas the Publishing API's presenter queries use UTC. By fetching from the Publishing API's db, we 
+ensure all change notes are timestamped in UTC.
```
yndajas, line 33: **note:** the outcome of Bruce's earlier comment about change histories might affect this consequence

spec/presenters/change_history_presenter_spec.rb
```diff
@@ -1,83 +1,79 @@
...
-    it "omits change notes that don't have a public timestamp" do
-      create(:change_note, edition:, note: "with-timestamp", public_timestamp: 1.day.ago)
-      create(:change_note, edition:, note: "without-timestamp", public_timestamp: nil)
-      expect(subject.map { |item| item[:note] }).to eq %w[with-timestamp]
-    end
+    context "when an edition already has a change history" do
+      let(:one_day_ago) { 1.day.ago }
+      let(:two_days_ago) { 2.days.ago }
 
-    context "multiple editions for a single content id" do
```
brucebolt, line 50: Are these tests (for multiple editions of a single content ID) still relevant and need to be included?

spec/presenters/change_history_presenter_spec.rb
```diff
@@ -1,83 +1,79 @@
...
-    end
 
-    it "orders change notes by public_timestamp (ascending)" do
-      [1, 3, 2].to_a.each do |i|
-        create(:change_note, edition:, note: i.to_s, public_timestamp: i.days.ago)
+      it "returns the change notes from the database" do
+        expect(subject).to eq([
+          { public_timestamp: change_notes[0].public_timestamp.as_json, note: change_notes[0].note },
+          { public_timestamp: change_notes[1].public_timestamp.as_json, note: change_notes[1].note },
+          { public_timestamp: change_notes[2].public_timestamp.as_json, note: change_notes[2].note },
```
brucebolt, line 28: When I ran this, all the change notes had the same timestamp and note text (a blank string).  Maybe these need changing (perhaps by updating the factory to use a sequence or something like that on the note text) so each note is slightly different, in order to test the ordering.

brucebolt, line 28: In this test, the timestamp is being rendered `as_json`, whereas in the other tests it is `.utc.to_s`.  Do these return different things?

app/presenters/change_history_presenter.rb
```diff
@@ -7,39 +7,48 @@ def initialize(edition)
...
     end
 
-    def version_number
-      edition.user_facing_version
+    def query_result
+      Presenters::Queries::ChangeHistory.new(edition, include_edition_change_history: fallback_to_query_only?)
+                                        .call
+                                        .pluck(:note, :public_timestamp)
+                                        .map { |note, timestamp| { note:, public_timestamp: timestamp } }
     end
```
brucebolt, line 48: Could this be simplified to something like this (which removes the ternary, a method and avoids converting the times to UTC twice)?

```rb
  def change_notes
      all_change_notes
        .map { |h| format_change_note(h) }
        .sort_by { |h| Time.zone.parse(h[:public_timestamp]) }
        .reverse
    end

    def format_change_note(history)
      {
        note: history[:note],
        public_timestamp: convert_timestamp_to_utc(history[:public_timestamp]).to_s,
      }
    end

    def all_change_notes
      query_result + details_change_notes
    end

    def details_change_notes
      @details_change_notes ||= edition.details.fetch(:change_history, [])
    end

    def query_result
      Presenters::Queries::ChangeHistory.new(edition, include_edition_change_history: details_change_notes.any?)
                                        .call
                                        .pluck(:note, :public_timestamp)
                                        .map { |note, timestamp| { note:, public_timestamp: timestamp } }
    end
```

spec/presenters/change_history_presenter_spec.rb
pezholio: As above, this is handled in the query - there's no database activity in this test now

brucebolt (commented):
Thanks for making all these changes @pezholio 🙇.

Happy to approve once the failing test has been resolved.

pezholio:
@brucebolt Should be all sorted now! 👍

spec/presenters/change_history_presenter_spec.rb
```diff
@@ -1,83 +1,87 @@
...
-          :live_edition,
-          document:,
-          details:,
-          user_facing_version: 2,
-        )
+      let(:details) do
+        { change_history: [
+          { public_timestamp: one_day_ago.in_time_zone("GMT").to_s, note: "note 1" },
+          { public_timestamp: two_days_ago.in_time_zone("GMT").to_s, note: "note 2" },
+        ] }
```
brucebolt, line 50: Should these use the `utc` method, to remain consistent with the other tests (instead of `in_time_zone("GMT")`)?

brucebolt (approved):
Thanks!  Just left one more small comment, but otherwise looks good to go.

pezholio:
@brucebolt @callumknights @yndajas Thanks for the review - have confirmed this is all working via a manual deploy, so merging now 🙌

PR #3468, opened 2025-07-21
(CM-248) Remove default number from the Relay UK field

pezholio:
This removes the default telephone number from the Relay UK telephone number field, as well as adding some validation to ensure the field is populated if the user opts to show the Relay UK details.


edavey (approved):
👍 thks for including the link to the JSON Schema if/then/else documentation

PR #3508, opened 2025-08-11
temp: add three special routes that have been removed but not correctly unpublished

KludgeKML:
- Once they have been unpublished, we will remove these lines.

/robots.txt has been provided directly by nginx since: https://github.com/alphagov/govuk-helm-charts/pull/179
/account/cookies-and-feedback has not been supported in frontend since:
https://github.com/alphagov/frontend/pull/3225
/sign-in/first-time has not been supported in frontend since:
https://github.com/alphagov/frontend/pull/3251

### DIFF
--- a/lib/data/special_routes.yaml
+++ b/lib/data/special_routes.yaml
@@ -316,3 +316,21 @@
     - :base_path: "/chat/support"
     - :base_path: "/chat/privacy"
     - :base_path: "/chat/accessibility"
+
+# These are being added solely so they can be unpublished correctly, after which they'll be removed
+
+- :base_path: "/robots.txt"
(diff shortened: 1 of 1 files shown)


KludgeKML:
> Approving, but with a suggestion that there ought to be a rake task added to Publishing API, for this sort of situation 🤔 would avoid us having to add and remove routes via PRs.
> 
> Though I suppose an audit trail is helpful.

I sort of agree, but also this is an unusual situation in which we would just have had one PR way back in history (remove the routes after republish) if someone had done it right first time. So it's more of a process thing - how can we make sure that people unpublish the route correctly before they remove the route?

PR #3514, opened 2025-08-12
Added new properties to history page schema [WHIT-2268]

ryanb-gds:
We intend to begin publishing history pages from Whitehall again soon, so we will need an up to date Publishing API schema that allows us to publish a sidebar image and the headers from the govspeak text in a structured way.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2268


lauraghiorghisor-tw (approved):
LGTM

PR #3530, opened 2025-08-15
Normalise keys in generate_random_edition

callumknights:
Found while working on a flaky test fix, confusing and results in inconsistent keys in the final random content hash. 

Needed fixing in advance of the flaky fix anyway, but belonged in a separate PR.

### DIFF
--- a/spec/support/random_content_helpers.rb
+++ b/spec/support/random_content_helpers.rb
@@ -4,14 +4,14 @@ module RandomContentHelpers
   def generate_random_edition(base_path)
     GovukSchemas::RandomExample.for_schema(publisher_schema: "generic") do |content|
       content.merge(
-        base_path:,
-        update_type: "major",
+        "base_path" => base_path,
+        "update_type" => "major",
 
-        title: "Something not empty", # TODO: make schemas validate title length
-        routes: [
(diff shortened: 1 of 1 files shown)


spec/support/random_content_helpers.rb
```diff
@@ -12,7 +12,7 @@ def generate_random_edition(base_path)
           { path: base_path, type: "prefix" },
         ],
         redirects: [],
-      )
+      ).stringify_keys
```
mike3985, line 15: If the non-string keys are all being introduced right here in this `#merge` argument, could we call `#stringify_keys` on just the argument? Or, even better I guess, just rewrite these keys as strings?

```ruby
content.merge(
  content.merge(
    "base_path" => base_path,
    "update_type" => "major",

    "title" => "Something not empty", # TODO: make schemas validate title length
    "routes" => [
      { "path" => base_path, "type" => "prefix" },
    ],
    "redirects" => [],
  )
)
```

callumknights: Sorted, tried this before and made a silly mistake which stopped it working, hence the pivot to `stringify_keys`, my mistake.

PR #3537, opened 2025-08-19
[PP-7303] Break up expectations in flaky test

brucebolt:
This test is flaky as the values in `available_translations` are unordered and therefore can be returned in an unpredictable order.

Breaking up the single hash expectation into individual matchers, so we can use a less restrictive (i.e. unordered) `match_array` matcher for the `available_translations`.

I've deviated from the "one expectation per test" principle here, since we're testing the same JSON response in all the expectations and the GraphQL query could prove to be expensive if repeated multiple times.

I did consider making the order of `available_translations` consistent, but decided against this as the existing tests related to this link all expected the results in a non-specific order.

### DIFF
--- a/spec/integration/graphql/roles_spec.rb
+++ b/spec/integration/graphql/roles_spec.rb
@@ -159,79 +159,72 @@ RSpec.describe "GraphQL" do
           }",
       }
 
-      expected = {
-        data: {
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Looks good to me!

PR #3542, opened 2025-08-21
Avoid generating document_type=redirect editions

richardTowers:
These are rare (1/191), but because we put different messages on the queue for redirect document types they upset (at least) the message queue tests.

I've no reason to think that `gone` or `vanish` document types are handled specially, but they smell funny so I'm removing them too.

There are plenty of other document types that aren't weird, so patching the schema to remove the "special" types before using it to generate editions feels pragmatic.


callumknights (approved):
_chef kiss_ 🤌 👨‍🍳

PR #3546, opened 2025-08-26
[PP-7319] 'Handle' GraphQL query errors

yndajas:
If an invalid GraphQL query is executed against the schema, we get an object containing errors back. For example, if we request a field that's not defined on one of our GraphQL types (like Edition or Details), we get undefined field errors

We previously weren't checking for this anywhere, so we'd end up returning the error object with a 200 response from the GraphqlController. Hopefully we wouldn't roll out `GraphqlController#live_content` support for schemas with invalid queries, but this change catches any such cases by raising instead of returning a non-content item object

### DIFF
--- a/app/services/graphql_content_item_service.rb
+++ b/app/services/graphql_content_item_service.rb
@@ -1,4 +1,6 @@
 class GraphqlContentItemService
+  class QueryResultError < StandardError; end
+
   attr_reader :query_result
 
   def initialize(query_result)
@@ -6,6 +8,10 @@ class GraphqlContentItemService
   end
 
   def process
+    if error_messages.present?
+      raise QueryResultError, error_messages.join("\n")
+    end
+
     unpublishing || edition
   end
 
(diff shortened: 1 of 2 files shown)


app/services/graphql_content_item_service.rb
```diff
@@ -23,4 +29,12 @@ def unpublishing
       &.find { _1["message"] == "Edition has been unpublished" }
       &.[]("extensions")
   end
+
+  def error_messages
+    return @error_messages if defined?(@error_messages)
```
mike3985, line 34: What's the early exit for that `@error_messages ||=` wouldn't do?

app/services/graphql_content_item_service.rb
```diff
@@ -1,3 +1,5 @@
+class GraphqlContentItemError < StandardError; end
```
mike3985, line 1: This exception type should be declared inside the service class's namespace since it belongs to it

app/services/graphql_content_item_service.rb
```diff
@@ -6,6 +8,10 @@ def initialize(query_result)
   end
 
   def process
+    if error_messages.present?
+      raise GraphqlContentItemError, error_messages.join("\n")
```
mike3985, line 12: Yeah, this is the sort of "handling" that I personally had in mind 👍

app/services/graphql_content_item_service.rb
yndajas: If there are no errors it would be `nil` which doesn't work with `||=`

PR #3547, opened 2025-08-26
[PP-9319] Validate GraphQL response against content schemas

yndajas:
... and update the GraphQL queries to address validation errors

A few notes:
- This introduces a bit of metaprogramming, but that does mean we get validation out of the box for any new queries
- All frontend schemas require links via `SchemaGenerator::FrontendSchemaGenerator#required`. Some schemas, like world index, don't have any custom links. Without a links field, the content item wouldn't be valid against the schema. With a links field, GraphQL requires you to select one of its properties. I considered keeping a list of schemas without custom links and ignoring their missing links field errors, but given Content Store does return a couple of link types for the world index (the only example we have), I've opted instead to add the most common of those link types to its query
- I've included a hash of schema-specific field values to include in the factory-generated edition. We need these overrides to ensure the edition is valid before we retrieve and process it with our GraphQL code

### DIFF
--- a/app/graphql/queries/ministers_index.graphql
+++ b/app/graphql/queries/ministers_index.graphql
@@ -3,6 +3,7 @@ query ministers_index($base_path: String!) {
     ... on MinistersIndex {
       base_path
       content_id
+      description
       document_type
       first_published_at
       locale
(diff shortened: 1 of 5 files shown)


richardTowers (approved):
Nice! Personally I like the `.each` loop for the tests.

PR #3555, opened 2025-08-27
PP-7281 script/build_graphql_query

mike3985:
There are some big diffs here but I'm hoping the commits are clear enough when looked at individually.

Changes:
* Introduce a script for building a GraphQL query for a given content type
* Make some changes to our GraphQL types to address issues highlighted by the newly-built queries
* Make some changes to content schemas to address issues discovered the same way

### DIFF
--- a/lib/graphql_query_builder.rb
+++ b/lib/graphql_query_builder.rb
@@ -1,0 +1,255 @@
+class GraphqlQueryBuilder
+  MAX_LINK_DEPTH = 5
+
+  def initialize(schema_name)
+    @schema_name = schema_name
+    @content_item = GovukSchemas::RandomExample.for_schema(
+      frontend_schema: @schema_name,
+      strategy: :one_of_everything,
+    )
+  end
+
+  def build_query
+    parts = [
+      "query #{@schema_name}($base_path: String!) {",
+      "  edition(base_path: $base_path) {",
+      "    ... on #{edition_type_or_subtype} {",
+      build_fields(@content_item, indent: 6),
+      "    }",
+      "  }",
+      "}",
+    ]
+
+    parts.join("\n")
+  end
+
+private
+
+  def build_fields(data, indent: 2, link_path: [])
+    fields = data.sort_by(&:first).flat_map do |entry|
+      case entry
+      in ["withdrawn", *]
+        nil
+      in [String, {} | []]
+        nil
+      in ["details" | "withdrawn_notice" => key, Hash => hash]
+        [
+          "#{key} {",
+          hash.map { |hash_key, _| "  #{hash_key}" },
+          "}",
+        ]
+      in ["links", Hash => links]
+        [
+          "links {",
+          links.map { |link_key, array| build_links_query(link_path + [link_key.to_sym], array) }.compact,
+          "}",
+        ]
+      in [String => key, String | Numeric | true | false | nil]
+        key
+      end
+    end
+    fields.compact.join("\n").indent(indent)
+  end
+
+  def is_reverse_link_type?(link_path)
+    # if "role_appointments" is at the root of the link_path, it's not the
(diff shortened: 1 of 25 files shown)


lib/graphql_query_builder.rb
```diff
@@ -0,0 +1,351 @@
...
+          "#{key} {",
+          build_fields(value, link_path:),
+          "}",
+        ]
+      in [String => key, Hash => value]
+        [
+          "#{key} {",
+          build_fields(value, link_path:),
+          "}",
+        ]
```
richardTowers, line 58: Note from mobbing: now that we're not doing any recursion into the details fields, we think we might be able to get rid of these catch-all patterns, perhaps replacing them with special cases for the few other top-level fields in content items that have complex values (e.g. withdrawn_notice) - we don't think there are many.

lib/graphql_query_builder.rb
```diff
@@ -0,0 +1,351 @@
...
+    link_type = link_path.last
+
+    ExpansionRules.is_reverse_link_type?(link_type)
+  end
+
+  def build_links_query(link_path, links)
+    link_type = link_path.last
+
+    document_types = if is_reverse_link_type?(link_path)
+                       flip_reversed_link_types = ExpansionRules.reverse_to_direct_link_type(link_type)
```
richardTowers, line 80: https://youtu.be/cjIvu7e6Wq8

lib/graphql_query_builder.rb
mike3985: You were right, (obviously aside from "details" and "links") "withdrawn_notice" is the only non-scalar at the top level.

Since withdrawn_notice is only one level deep, I've adapted the existing details code to support both. If that's more confusing than it needed to be, I'll split 'em up.

```ruby
in ["details" | "withdrawn_notice" => key, Hash => hash]
  [
    "#{key} {",
    hash.map { |hash_key, _| "  #{hash_key}" },
    "}",
  ]
```

richardTowers: That looks good to me, nice one!

lib/graphql_query_builder.rb
```diff
@@ -0,0 +1,255 @@
...
+             document_types.map { |document_type|
+               ExpansionRules.expand_fields({ document_type: }, link_type:, draft: false)
+             }
+               .each_with_object(example_from_schema) do |item, example|
+                 # merge the example link (which is part of @content_item)
+                 # into the nil'd-out item from ExpansionRules#expand_fields
+                 # ...but ensure that any changes are updated on the example so
+                 # that we can access the finished result later if we want
+                 example.replace(item.deep_stringify_keys.deep_merge(example))
+               end
```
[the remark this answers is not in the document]
richardTowers: Ah, not it wouldn't be because we're assigning the result to `link`, not `example_from_schema`.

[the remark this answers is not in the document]
mike3985: You're right that they both do the same thing. I've heard the statement "`reduce` makes things impossible to understand" so many times in recent years that I was scrabbling around for an alternative to that, though. Let me run it by somebody else from the team?

richardTowers (approved):
Overall looks good to me.

It's a helper script, rather than a piece of production code, so I think it's forgivable that we haven't written any tests for it. I've left a few comments, but I'm happy for you to take them or leave them - nothing showstopping.

lib/graphql_query_builder.rb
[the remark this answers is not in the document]
mike3985: We swapped it for an `#inject` 😅 Not purely to be awkward.

I've also removed any extra effort that was being made to update `@content_item`, but turned a blind eye to any incidental mutation that might still be happening. (FWIW this was a hangover from a version that rolled schema validation into the script)

PR #3575, opened 2025-09-08
[PP7327] Restructure script folder

yndajas:
This was getting a bit inconsistent and unclear with the expansion in GraphQL-related scripts


script/live_content/build_query
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
```
mike3985, line 1: I'm not sure building the query is related to "live_content". It might just be related to building a GraphQL query?

script/live_content/diff
```diff
@@ -8,7 +8,7 @@ require "uri"
 # Using a list of base paths, render each page with Content Store and Publishing
 # API's GraphQL endpoint as the data source and diff the results.
 
-OUTPUT_DIR = "tmp/diff_graphql/content_items".freeze
```
mike3985, line 11: I originally used "live_content" in this script's name to refer to the "graphql#live_content" controller action. I wonder if it would've made more sense to've referred to "Content API" or "content item", though?

mike3985 (commented):
I've left a couple of comments, but it isn't easy to comment on file path changes in Github, so I hope it's clear what I'm talking about 😅

script/live_content/diff_frontend
```diff
@@ -3,7 +3,7 @@
 # Diff a rendered page in production when the data source is Content Store vs
 # when the data source is Publishing API's GraphQL endpoint.
 
-source script/diff_graphql/functions.sh
+source script/live_content/diff_frontend_functions.sh
```
mike3985, line 6: the inclusion of "frontend" in the name of these diff scripts is a nice improvement, by the way, but as with the other files, I think I made a mistake calling owt "live_content" in the first place so to me it'd be a shame to expand its use!

mike3985 (approved):
It could actually just be that I don't like the "live content" name much in general 🤔 I take your point(s) about most of these scripts' relevance to that endpoint, so let's do it

PR #3587, opened 2025-09-09
[PP-7327] Deep transform edition in content item service

yndajas:
This does two things:
- deeply prunes all properties with empty values (excluding false)
- deeply transforms keys to strings. We found that GraphQL uses symbol keys for hashes when using the `GraphQL::Types::JSON` type, whereas everything else uses string keys. The key format should be consistent, so we fix that here

### DIFF
--- a/app/services/graphql_content_item_service.rb
+++ b/app/services/graphql_content_item_service.rb
@@ -18,9 +18,25 @@ class GraphqlContentItemService
 private
 
   def edition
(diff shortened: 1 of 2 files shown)


callumknights:
I appear to have jumped the shark a bit there 🙈

yndajas:
Replaced by schema-aware compaction (merged) and the upcoming deep stringify and sort (#3602)

PR #3598, opened 2025-09-12
[PP-7356] Compact only optional fields according to the schema

richardTowers:
Rather than using `Hash#compact!` to remove fields with nil values from the GraphQL response, this uses `Hash#filter` to remove fields with nil values _only if they are optional according to the schema_.

This should mean that `live_content` GraphQL responses will be more similar to content-store (which often omits keys rather than including them with null values), and more likely to be valid against the content schemas.

I considered trying to also validate whether required fields are allowed to be null, and logging a warning if we produce an invalid response (because we've included a required field which is not allowed to be null as null). Unfortunately, this requires a deeper understanding of JSON schema, which makes the code a little complicated - it didn't feel like it was worth going to that much trouble just to log a warning.

It would also be good to remove `links { }` results that are empty arrays, as that's a common difference between content-store and the GraphQL backed `live_content` endpoint - I'll save that for a separate PR though.

I did a fair bit of refactoring on the way to this PR, so you may find the journey is easier to follow reviewing commit by commit.

### DIFF
--- a/lib/graphql/content_item_compactor.rb
+++ b/lib/graphql/content_item_compactor.rb
@@ -20,6 +20,6 @@ class Graphql::ContentItemCompactor
 private
 
   def compact_non_required_fields(hash, required_fields)
-    hash.reject { |key, value| value.nil? && !required_fields.include?(key) }
+    hash.reject { |key, value| value.nil? && required_fields.exclude?(key) }
   end
 end


lib/graphql/content_item_compactor.rb
```diff
@@ -1,7 +1,28 @@
...
+      compact_response["details"] = compact_non_required_fields(details, required_details_fields)
+    end
+
+    compact_response
+  end
+
+private
+
+  def compact_non_required_fields(hash, required_fields)
+    hash.reject { |key, value| value.nil? && !required_fields.include?(key) }
```
yndajas, line 23: **suggestion:** given this is Rails, we could use `#in?` or `#exclude?` to avoid the negation

brucebolt (approved):
This all looks good to me.

Thanks also for the detailed commit message explaining why you used `fetch` instead of `dig`, as I would've otherwise left a comment about that!

PR #3602, opened 2025-09-15
[PP-7357] Deep stringify and sort

yndajas:
This is done in Content Store: https://github.com/alphagov/content-store/pull/1100

Without this, we can get diffs like "Applies to England and Wales" in Content Store and "Applies to Wales and England" due to the properties of details["national_applicability"] being in a different order to how Content Store returns them. This likely only affects fields with a JSON type where we can't specify the order of fields (via selections)

We need to stringify the keys before sorting - otherwise symbols and strings would be sorted separately. We wanted to stringify the keys anyway for consistency...

TODO: stringify in one commit, sort in the next
TODO: check if Richard's intelligent compact traversal mechanism could be used for sorting and stringifying without the external library

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,3 +52,5 @@ group :development, :test do
   gem "timecop"
   gem "webmock", require: false
 end
+
+gem "deepsort", "~> 0.5.0"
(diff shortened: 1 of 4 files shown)


yndajas:
Replaced by #3618

PR #3614, opened 2025-09-19
[PP-7366] Deterministically order all linked editions

yndajas:
Even though we order links by link type and position everywhere, we still have instances of links being ordered differently in our GraphQL dataloaders than in the queries used by link expansion and dependency resolution. This can cause both JSON and HTML diffs when comparing GraphQL and Content Store

The reason this is possible is that links often have a position of 0 (about 80% of links, in fact) and in most queries we aren't specifying how to order links with matching types and position. In such cases the way Postgres plans the queries impacts the order in which we return links, likely down to a difference in choice of index. The queries are very distinct, so this is unsurprising

We think it's a good idea to have a deterministic ordering strategy in general, and doing so will allow us to reduce what we consider to be noise in the diffs (if two links have a position of 0, it shouldn't matter which way they're ordered, but consistency is preferable)


spec/graphql/sources/person_current_roles_source_spec.rb
```diff
@@ -17,6 +17,49 @@
...
+
+    # appoint person to role
+    role_appointment = create(
+      :live_edition,
+      document_type: "role_appointment",
+      details: { current: true },
+    )
+    link_set = create(:link_set, content_id: role_appointment.content_id)
+    create(:link, link_type: "person", target_content_id: person.content_id, link_set:)
+    create(:link, position: 2, link_type: "role", target_content_id: role_2.content_id, link_set:)
```
yndajas, line 54: **note:** this is reassigning variables. Maybe we can iterate over an array of [role, position]s

app/graphql/sources/linked_to_editions_source.rb
```diff
@@ -116,7 +120,7 @@ def fetch(editions_and_link_types)
           SQL
         )
         .where(editions: { row_number: 1 })
-        .order(link_type: :asc, position: :asc)
+        .order(link_type: :asc, position: :asc, link_id: :desc)
```
yndajas, line 123: **note:** in the commit message, I said:

> The reason this is possible is that links often have a position of 0 (about 80% of links, in fact) and in most queries we aren't specifying how to order links with matching types and position.

The highlighting of the proportion of zero-positioned links was based on some flawed thinking. My assumption was that position would be included as a field for each link in the payload, and zero was a default meaning "no specified position". In reality the position is set in Publishing API based on the order of the links array (see `Commands::V2::PatchLinkSet` and `Commands::V2::PutContent`)

Does the issue of multiple linked editions of the same type and position only present itself for reverse links then?

PR #3630, opened 2025-10-02
(CM-360) Add default value for Contact Block Opening Hours

pezholio:
[Jira card](https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-360)

This adds a default value for Opening Hours to serve as a prompt for users to edit when they click to show opening hours. This will function like a placeholder (but without the associated UX nightmare that placeholders provide), and we expect users to edit this when they see it. We'll observe how users get on with it and will review going forward.

### DIFF
--- a/content_schemas/dist/formats/content_block_contact/notification/schema.json
+++ b/content_schemas/dist/formats/content_block_contact/notification/schema.json
@@ -549,7 +549,8 @@
                   },
                   "properties": {
                     "opening_hours": {
-                      "type": "string"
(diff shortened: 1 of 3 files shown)


pezholio:
The schemas are built via the `rake build_schemas` task, so we only need to change the `jsonnet` file - see https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md

PR #3803, opened 2026-01-08
[PP-7469] Introduce `$with_drafts` variable for GraphQL queries and `/graphql/draft` endpoint to test it

mike3985:


spec/presenters/queries/content_item_presenter_spec.rb
```diff
@@ -120,6 +121,7 @@
           update_type: "major",
           first_published_at:,
           public_updated_at:,
+          user_facing_version: 1,
```
yndajas: **question:** I'm assuming all of these need to be 1 for some reason?

app/graphql/types/query_type.rb
```diff
@@ -4,14 +4,32 @@ module Types
...
+      argument :with_drafts, Boolean, required: false, default_value: false
     end
 
-    def edition(base_path:, content_store:)
+    def edition(base_path:, with_drafts:)
+      all_states = if with_drafts
+                     %i[draft published unpublished]
+                   else
+                     %i[published unpublished]
+                   end
```
yndajas: **question:** I've only really been thinking about the state fallback in terms of links. What's the use case for falling back when requesting a draft root edition?

spec/graphql/sources/linked_to_editions_source_spec.rb
```diff
@@ -53,27 +53,77 @@
...
+          end
+        end
+      end
+
+      context "when with_drafts=false" do
+        it "doesn't return links to drafts" do
+          target_edition_1 = create(:draft_edition, title: "edition 1, draft")
+          target_document = create(:document)
+          target_edition_2 = create(:live_edition, title: "edition 2, live", document: target_document)
+          create(:draft_edition, title: "edition 3, draft", document: target_document)
```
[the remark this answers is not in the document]
mike3985: Yeah, I haven't figured out a clear way to do this test setup. That document is actually associated with both of those live and draft target editions, but it seems like it's not doing a good job at readability

spec/graphql/sources/reverse_linked_to_editions_source_spec.rb
```diff
@@ -475,7 +569,7 @@
       end
     end
 
-    context "when the Edition is live" do
+    context "when requested with with_drafts=false" do
```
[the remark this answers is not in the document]
mike3985: Ah, I was sure there were no interesting scenarios in that combination. But if you mean it'd be good just to check that the combination works, yes, I'll add some!

spec/presenters/queries/content_item_presenter_spec.rb
mike3985: I think they're all because of how their specific expectations are defined. I could've changed the `expected_output` etc, but doing this explicitly seemed reasonable.

app/graphql/types/query_type.rb
yndajas: Probably worth documenting some of this in a commit

app/graphql/sources/queries/linked_to_editions.sql
```diff
@@ -28,19 +28,28 @@ edition_linked_editions AS (
...
-  ORDER BY links.id ASC, is_primary_locale DESC
+  ORDER BY
+    links.id ASC,
+    is_primary_locale DESC,
+    CASE editions.state
+      WHEN 'draft' THEN 0
+      WHEN 'published' THEN 1
+      WHEN 'unpublished' THEN 2
+      ELSE 3
+    END
```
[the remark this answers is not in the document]
yndajas: Presumably I was talking about this: https://github.com/alphagov/publishing-api/blob/36994484fabeb9c98aa8c5d10cb007c195ea3f63/app/queries/get_edition_ids_with_fallbacks.rb#L13-L14

spec/integration/link_expansion_precedence_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+              locale: test_case.root_locale,
+            ),
+            link_kind => test_linked_editions.map do
+              [
+                {
+                  link_type: test_case.link_type,
+                  target_content_id: it.content_id,
+                },
+              ]
+            end,
```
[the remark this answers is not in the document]
yndajas: let's create two link kind arrays above instead, then map over both under the appropriate keys here

.github/workflows/ci.yml
```diff
@@ -90,6 +90,84 @@ jobs:
...
+
+  test-graphql-precedence-different-content-id:
+    name: "Test GraphQL link expansion precedence (different content ID)"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Postgres
+        id: setup-postgres-graphql-precedence-different-content-id
+        uses: alphagov/govuk-infrastructure/.github/actions/setup-postgres@main
+        with:
+          POSTGRES_DB: postgres-parity
```
yndajas: Curious that this can use the same database(s) as the previous job but they need to use different ones than the main tests. I think the latter is because when we run `db:parallel:create` it can't duplicate the database that's likely already in use by the main test job, but I'm surprised there's no race condition with these two jobs that both need the separate multiple databases

PR #3806, opened 2026-01-09
WIP - test that graphql queries select all top level fields from thei…

richardTowers:
…r json schemas

... this is an idea of a way we could enforce that the GraphQL queries and the JSON schemas don't drift apart. It just checks the fields we select in the graphql query are all the same ones specified in the JSON schema (at the top level, and in details - not messing with links yet).


spec/graphql/queries/graphql_queries_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+             end
+
+    @top_level_graphql_query_fields = fields.map(&:name).to_set
+    @details_graphql_query_fields = fields.find { _1.name == "details" }&.selections&.map(&:name)&.to_set
+  end
+end
+
+RSpec.describe "GraphQL queries" do
+  Dir[Rails.root.join("app/graphql/queries/*.graphql")].each do |path|
+    /(?<schema_name>[^\/]+)[.]graphql\Z/ =~ path
```
richardTowers: 😈

brucebolt:

PR #3860, opened 2026-02-03
Make "Read more" link text optional, and start sending "parent" link, for Topical Event About pages [WHIT-3341]

ChrisBAshton:
We're likely to drop this altogether in the move to config-driven topical events.

Needed in conjunction with https://github.com/alphagov/whitehall/pull/11431

Jira: https://gov-uk.atlassian.net/browse/WHIT-3341

---

### DIFF
--- a/content_schemas/dist/formats/topical_event_about_page/frontend/schema.json
+++ b/content_schemas/dist/formats/topical_event_about_page/frontend/schema.json
@@ -273,7 +273,6 @@
     "details": {
       "type": "object",
       "required": [
-        "read_more",
         "body"
       ],
       "additionalProperties": false,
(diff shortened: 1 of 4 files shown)


TonyGDS (approved):
LGTM

I noted there were a few description attributes deleted, why was that?

Also with our current parent/child edition solution, the LinkSet links would apply to the live edition so at the moment we must assume the `parent` is going to be fixed.

ChrisBAshton:
> I noted there were a few description attributes deleted, why was that?

I noticed that too and am not sure tbh!

> Also with our current parent/child edition solution, the LinkSet links would apply to the live edition so at the moment we must assume the `parent` is going to be fixed.

In cases where there is both a parent edition link and a parent linkset link, [the edition link value takes priority](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#put-content-edition-links). But it's kind of moot here because we expect a child to always have the same parent. 👍

PR #4018, opened 2026-04-16
[PP-7588] Limit base paths to 512 bytes

yndajas:
Requires https://github.com/alphagov/govuk_schemas/pull/185

Search API has stopped indexing content with base paths over 512 bytes:
https://github.com/alphagov/search-api/commit/5ed2e5547fef3b662a2ae82ca6590f65334eb755. We want content to be searchable, so this ensures all new content meets this requirement. There are currently four live editions which would become invalid with this change, but they're all unpublished. This adds validations via the Edition and PathReservation models as well as the content schemas

Closes #4023


app/models/edition.rb
```diff
@@ -136,6 +137,10 @@ def user_facing_version_cannot_change
     errors.add(:user_facing_version, "has changed") if persisted? && user_facing_version_changed?
   end
 
+  def base_path_not_too_long
+    errors.add(:base_path, "over 512 bytes") if base_path.bytesize > 512
```
brucebolt, line 141: This should probably be greater than or equal to 512, as that's what [Search API is rejecting](https://github.com/alphagov/search-api/blob/56c2c8c327d50a402898ca5bcf8e0a9dcd5fc9e2/lib/analytics/ga4_import/page_view_consolidator.rb#L13).

yndajas: Oh yeah. That seems like a mistake given the "max" wording, and the commit message saying "values _exceeding_ 512 bytes can be safely treated as invalid". I'll check with the owning team

yndajas: Fixed in https://github.com/alphagov/search-api/commit/824953a52bb756ab8f6606b45e30a69ffadec538

brucebolt (approved):

yndajas:
[The Pact and content schema tests pass with the govuk_schemas update but without the schema change](https://github.com/alphagov/publishing-api/pull/4023), so I think the apps just need updating so they can handle the new use of `allOf`

PR #4025, opened 2026-04-20
[PNP-9943] Add plan_for_change_landing_page schema/document type

KludgeKML:
Adds a `plan_for_change_landing_page` schema_name and document_type as basically a direct copy of the existing `landing_page` schema which is effectively squatting a schema name we'd like to use and is only used for one specific type of page.

Related Content Data API PR: (tests here may not pass until that is merged?) https://github.com/alphagov/content-data-api/pull/2501

JIRA Card: https://gov-uk.atlassian.net/browse/PNP-9943


content_schemas/examples/plan_for_change_landing_page/frontend/plan_for_change_landing_page.json
```diff
@@ -0,0 +1,351 @@
+{
```
deborahchua, line 1: Could we include a link to the original schema this has been copied from in the commit description?

KludgeKML: Good thinking! Have added audit trail links to the schema, example, and graphQL (in the relevant commit) copied from.

deborahchua (approved):
LGTM

lauraghiorghisor-tw (approved):
This is a plain copy-paste of the previous `landing_page`, right? LGTM if the agreed naming is `plan_for_change_landing_page` 👍  

Two things:
- odd that the generated schema files are ever so slightly different in length for some of the generated schemas (since the jsonnet is the same)
- there is also a mention of landing_page in `publishing_api_schema_test_cases.yaml` though I am not entirely sure what that does

KludgeKML:
@lauraghiorghisor-tw 

> * odd that the generated schema files are ever so slightly different in length for some of the generated schemas (since the jsonnet is the same)

Interesting - where are you seeing that? Looking at my local copies, the files appear to be the same length (in lines, obviously in characters the plan_for_change ones are slightly bigger because of the additional name).

> * there is also a mention of landing_page in `publishing_api_schema_test_cases.yaml` though I am not entirely sure what that does

It looks like that's a special case of benchmarks which you run when you have a local database you can draw from that includes published pages, so I'm inclined not to touch it for the moment (although when all the PfC pages have been republished as the new schema type, then we could update that).

PR #4045, opened 2026-05-07
[PP-7705] Add information about old locks to Sidekiq UI

yndajas:
This adds a 'tab' to the Sidekiq web UI which provides:
- a table listing locks whose digests were created at least the default lock TTL ago (one hour at the time of writing)
- links from that table to retry set information on each digest (the locks usually if not always hang around because of working their way through retries)

It also records various recipes I noted down while working on this and which might prove useful to future Sidekiq investigators

### Old locks

image

### With no data

image

### Old lock information for a specific digest

image

### DIFF
--- a/app/views/sidekiq/old_locks.html.erb
+++ b/app/views/sidekiq/old_locks.html.erb
@@ -1,0 +1,35 @@
+<header>
+  <h3>Old locks</h3>
+</header>
+
+<p>
+  By 'old locks', we mean those that were created at least the default TTL ago
+  (<%= @default_ttl %>).
+</p>
+
+<% if @locks.any? %>
+    <div class="table_container">
+    <table class="table table-striped table-bordered table-hover">
+      <thead>
+        <tr>
+          <th>Digest</th>
+          <th>State</th>
+          <th>Since</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @locks.each do |lock| %>
+          <tr class="lock-row">
+            <td><%= lock[:digest] %></td>
+            <td><%= lock[:state] %></td>
+            <td><%= _safe_relative_time(lock[:created_at]) %></td>
+          </tr>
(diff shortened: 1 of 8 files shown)


yndajas:
Probably worth noting the pattern here is broadly based on https://github.com/mhenrixon/sidekiq-unique-jobs/blob/v8.x/lib/sidekiq_unique_jobs/web.rb

See also: https://github.com/sidekiq/sidekiq/blob/main/examples/webui-ext/lib/sidekiq-redis_info/web.rb

Edit: added to the first commit message

app/views/sidekiq/old_locks.html.erb
```diff
@@ -0,0 +1,26 @@
+<header>
+  <h3>Old locks</h3>
+</header>
+
+<% if @locks.any? %>
```
nacnudus, line 10: It might reassure a user to see a message like "There are no old locks at the moment", if there aren't any.

app/views/sidekiq/old_locks.html.erb
```diff
@@ -0,0 +1,26 @@
+<header>
+  <h3>Old locks</h3>
+</header>
```
nacnudus, line 3: The definition of "old" from the commit message would be useful here, especially if the default TTL value was dynamically included.

> locks whose digests were created at least 1 hour ago (the default lock TTL)

nacnudus (approved):
Looks good. I've suggested adding some explanatory text.

yndajas:
@nacnudus I've made a few changes including your suggestions, in case you want to take another glance. Diff (including updates from main - you can ignore all the content schemas stuff): https://github.com/alphagov/publishing-api/compare/e4e48a2115637c1e941ffcc3698b14915dfbe7bb..c7c8fd658d660340f9aa7aa0d0cb9e2398173a42

PR #4052, opened 2026-05-14
[PP-7606] Add error codes to 422 responses

AgaDufrat:
### Changes proposed
- 422 error responses from the API now include `error_code` by adding support for it in `BaseCommand` and `CommandError`
- relevant validation errors (`ActiveRecord::RecordInvalid`) include codes in `errors.details`
- Fallback behaviour:
  - explicit codes are preserved
  - missing mappings default to generic `:validation_failed` in relevant models
- `CommandError` validates that 422 errors always include a valid `error_code` and only predefined error codes are allowed (which acts as lightweight documentation of possible codes)
- To avoid breaking existing behaviour while gradually enforcing consistency unknown or missing codes:
    - notify the app team via Sentry in production
    - raise `ArgumentError` in test/development This avoids 

### Why

Publishing applications (e.g. Manuals Publisher) currently rely on regex parsing of the human-readable message field returned by Publishing API to determine the underlying validation failure to then display it nicely to the users in the UI.
Providing stable validation error codes will:
- Enable publishing apps to implement reliable validation error handling
- Remove the need for regex-based parsing of error messages
- Improve developer experience for publishing app teams
- Reduce accidental breaking changes caused by message copy updates
- Establish a clearer API contract for validation errors.

#### Notes
- Includes some cleanup:
  - Removed redundant `GovukStatsd` usage from CommandError and BaseCommand
  - Removed redundant path reservation test
- Not all validation errors are exposed via the API, so requiring explicit codes everywhere would create unnecessary maintenance overhead.
- Error code format was picked by a popular vote but it was 50/50 between snake case and upper snake case.
- Review note/question: `ActiveRecordError` store the values under `code` key, while higher level classes and response use `error_code`. We may want to keep it consistent. 
- Publishing app teams should add/update Pact tests once they start relying on the error codes

### Depends on
- https://github.com/alphagov/gds-api-adapters/pull/1395

### DIFF
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -6,16 +6,14 @@ module Commands
       logger.debug "#{self} called with payload:\n#{payload}"
 
       response = EventLogger.log_command(self, payload) do |event|
-        PublishingAPI.service(:statsd).time(name.gsub(/:+/, ".")) do
-          new(
-            payload,
-            event:,
-            downstream:,
-            callbacks:,
-            nested:,
-            **options,
-          ).call
-        end
+        new(
+          payload,
+          event: event,
+          downstream: downstream,
+          callbacks: callbacks,
+          nested: nested,
+          **options,
+        ).call
       end
 
       execute_callbacks(callbacks) unless nested
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -82,7 +82,7 @@ module Commands
           },
         }
 
-        raise_command_error(409, "Conflict", fields, friendly_message:)
+        raise_command_error(409, "Conflict", fields, error_code: :conflict, friendly_message:)
       end
 
       current_version
--- a/app/commands/base_command.rb
+++ b/app/commands/base_command.rb
@@ -47,15 +47,18 @@ module Commands
     def self.raise_validation_command_error(error)
       errors = error.record.errors
       full_message = errors.full_messages.join(", ")
+      error_code = errors.count == 1 ? errors.first.details[:code] : :multiple_validation_errors
 
       raise CommandError.new(
         code: 422,
+        error_code: error_code,
         message: full_message,
         error_details: {
           error: {
             code: 422,
             message: full_message,
-            fields: errors.to_hash,
+            error_code: error_code,
+            fields: errors.details,
           },
         },
       )
@@ -85,9 +88,10 @@ module Commands
       current_version
     end
 
-    def raise_command_error(code, message, fields, friendly_message: nil)
+    def raise_command_error(code, message, fields, error_code: nil, friendly_message: nil)
       raise CommandError.new(
         code:,
+        **(code == 422 ? { error_code: error_code } : {}),
         message:,
         error_details: {
           error: {
--- a/app/commands/v2/unpublish.rb
+++ b/app/commands/v2/unpublish.rb
@@ -44,7 +44,7 @@ module Commands
 
       def raise_invalid_unpublishing_type
         message = "#{unpublishing_type} is not a valid unpublishing type"
-        raise_command_error(422, message, { fields: {} })
+        raise_command_error(422, message, { fields: {} }, error_code: :type_invalid)
       end
 
       def edition
@@ -54,7 +54,7 @@ module Commands
       def validate_allow_discard_draft
         if payload[:allow_draft] && payload[:discard_drafts]
           message = "allow_draft and discard_drafts cannot be used together"
-          raise_command_error(422, message, { fields: {} })
+          raise_command_error(422, message, { fields: {} }, error_code: :conflicting_unpublishing_flags)
         end
       end
 
@@ -79,7 +79,7 @@ module Commands
             )
           else
             message = "Cannot unpublish with a draft present"
-            raise_command_error(422, message, { fields: {} })
+            raise_command_error(422, message, { fields: {} }, error_code: :cannot_unpublish_with_draft)
           end
         end
       end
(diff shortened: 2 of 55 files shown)


app/commands/base_command.rb
```diff
@@ -5,18 +5,16 @@ class BaseCommand
...
+    response = EventLogger.log_command(self, payload) do |event|
+      new(
+        payload,
+        event: event,
+        downstream: downstream,
+        callbacks: callbacks,
+        nested: nested,
+        **options,
+      ).call
+    end
```
nacnudus: nitpick: this indentation is fixed in 422a48483ba241d4b0a88977b93e1321f52174a0. It would be neater to be in this commit.

brucebolt (commented):
I haven't really looked at the code, but just a random thought: should the [error handling documentation](https://github.com/alphagov/publishing-api/blob/main/docs/publishing-application-examples.md#error-handling) also be updated as part of this PR?

spec/models/path_reservation_spec.rb
```diff
@@ -119,8 +121,42 @@
...
+      expect {
+        described_class.reserve_base_path!("not-a-valid-path", "publisher")
+      }.to raise_error(ActiveRecord::RecordInvalid) { |exception|
+             expect(exception.record.errors.details[:base_path]).to include(
+               error: "is not a valid absolute URL path", code: :absolute_path_invalid,
+             )
+           }
+    end
+
+    it "raises an error with :base_path_too_long error code when base_path is not absolute" do
```
nacnudus:
```suggestion
    it "raises an error with :base_path_too_long error code when base_path is too long" do
```

app/validators/routes_and_redirects_validator.rb
```diff
@@ -50,24 +50,24 @@ def check_redirects?(record)
   def must_have_unique_paths(record, routes, redirects)
     paths = routes.map { |r| r[:path] }
     unless paths == paths.uniq
-      record.errors.add(:routes, "must have unique paths")
+      record.errors.add(:routes, "must have unique paths", code: :routes_not_unique)
```
nacnudus:
```suggestion
      record.errors.add(:routes, "must have unique paths", code: :route_paths_not_unique)
```

app/validators/routes_and_redirects_validator.rb
```diff
@@ -50,24 +50,24 @@ def check_redirects?(record)
...
     end
 
     paths += redirects.map { |r| r[:path] }
     unless paths == paths.uniq
-      record.errors.add(:redirects, "must have unique paths")
+      record.errors.add(:redirects, "must have unique paths", code: :redirects_not_unique)
```
nacnudus:
```suggestion
record.errors.add(:redirects, "must have unique paths", code: :redirect_paths_not_unique)
```

AgaDufrat:
> I haven't really looked at the code, but just a random thought: should the [error handling documentation](https://github.com/alphagov/publishing-api/blob/main/docs/publishing-application-examples.md#error-handling) also be updated as part of this PR?

Absolutely! Thanks for the reminder. It slipped my mind while plumbing the depths of Rails error magic.

spec/support/routes_and_redirects_validator.rb
```diff
@@ -283,6 +283,7 @@
         ["foo/bar", "/url with spaces", "fdjkdfjkljsdaf"].each do |destination|
           edition.redirects = [{ path: "#{subject.base_path}/foo", type: "exact", destination: }]
 
+          pp subject.errors
```
nacnudus: todo: remove debugging statements

app/commands/v2/unpublish.rb
```diff
@@ -5,7 +5,7 @@ def call
         validate
 
         previous.supersede if previous_edition_should_be_superseded?
-        transition_state
+        unpublish
```
nacnudus: question: what does this change entail?

nacnudus (commented):
This looks really good so far. I've reviewed up to and including 422a48483ba241d4b0a88977b93e1321f52174a0. It's easy to follow, and has been applied consistently. I've left a few small comments and questions.

app/controllers/application_controller.rb
```diff
@@ -16,9 +16,11 @@ class ApplicationController < ActionController::API
   def parameter_missing_error(error)
     error = CommandError.new(
       code: 422,
+      error_code: :parameter_missing,
       error_details: {
         error: {
           code: 422,
+          error_code: :parameter_missing,
```
nacnudus: suggestion: a more specific error code `parameter_missing_or_invalid`, so that it isn't confused with `get_linked.rb`, where missing parameters are handled differently from invalid ones. Unless there's a reason to use the same code.

spec/support/routes_and_redirects_validator.rb
AgaDufrat: rubocop, you had one job 😆

app/commands/v2/unpublish.rb
AgaDufrat: Good spot. I initially thought that the `valid_unpublishing_type?` validation could be handled by the [Unpublishing model](https://github.com/alphagov/publishing-api/blob/551813df20593ac663861c6eeaef682edb6a812b/app/models/unpublishing.rb#L17) but the valid types are different. I forgot to unpick this and it's not covered by tests 😬

app/controllers/application_controller.rb
AgaDufrat: Great suggestion – more descriptive.

nacnudus (approved):

PR #4070, opened 2026-05-21
[PP-7750] Do not retry downstream jobs on specific failures

brucebolt:
We currently raise an exception for certain failures in the `DownstreamService`. This leads to the job being added to the retry queue and the lock (from sidekiq-unique-jobs) being reapplied.

However the job will never succeed because the conditional will always raise that same error, unless the document gets updated between retries to take it out of the current state. If the document gets updated, we attempt to schedule another job but that fails because of the lock that has been readded.

The retries occur over a period of 20 days. This means that following an update to the document that changes its state so it can be sent downstream, we won't actually exectute the job for some time, possibly days. Resulting in the document not being updated in Content Store immediately.

Changing the logic here to rescue from those exceptions and log the error instead of failing. This will mean the job doesn't get added to the retry queue and won't remain locked. This means if the document is updated so it can be sent downstream, a new job can be scheduled and the document sent to Content Store immediately.

### DIFF
--- a/app/sidekiq/downstream_discard_draft_job.rb
+++ b/app/sidekiq/downstream_discard_draft_job.rb
@@ -22,7 +22,7 @@ class DownstreamDiscardDraftJob
     update_expanded_links
 
     enqueue_dependencies if update_dependencies
-  rescue DiscardDraftBasePathConflictError => e
+  rescue DiscardDraftBasePathConflictError, DownstreamDraftExistsError, DownstreamInvalidStateError => e
     logger.warn(e.message)
   end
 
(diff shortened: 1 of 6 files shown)


yndajas (approved):
Per Slack, it would be good to confirm we have a replicable bad bulk republish to test this against before merging, but the code looks good

brucebolt:
> Per Slack, it would be good to confirm we have a replicable bad bulk republish to test this against before merging, but the code looks good

I'll republish an entire document type that previously had some failures. `national_statistics` was one of them.

PR #4075, opened 2026-05-29
Navigation spike - expanded reverse links [WHIT-3342]

ChrisBAshton:
Jira: https://gov-uk.atlassian.net/browse/WHIT-3342

---

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #4076, opened 2026-05-29
Navigation spike - combo [WHIT-3342]

ChrisBAshton:
Cherry-picking the best of https://github.com/alphagov/publishing-api/pull/4074 and https://github.com/alphagov/publishing-api/pull/4075

Jira: https://gov-uk.atlassian.net/browse/WHIT-3342

---

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

### DIFF
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -71,6 +71,12 @@ module_function
     ministerial: :ministers,
   }.freeze
 
+  # These expand every member of the reciprocal direct link type from
+  # the source link set, instead of only mirroring the edition being expanded.
+  REVERSE_LINKS_WITH_MEMBER_EXPANSION = {
+    document_collections: :documents,
+  }.freeze
+
   # These fields are required by the frontend_links definition
   MANDATORY_FIELDS = %i[
     content_id
@@ -227,6 +233,10 @@ module_function
 
   def reverse_link_type(link_type)
     REVERSE_LINKS[link_type.to_sym]
+  end
+
+  def member_expansion_direct_link_type(reverse_link_type)
+    REVERSE_LINKS_WITH_MEMBER_EXPANSION[reverse_link_type.to_sym]
   end
 
   def reverse_to_direct_link_type(link_type)
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -112,6 +112,7 @@ module_function
   PERSON_FIELDS_WITH_IMAGE = (DEFAULT_FIELDS + details_fields(:image, :privy_counsellor)).freeze
   ROLE_FIELDS = (DEFAULT_FIELDS + details_fields(:body, :role_payment_type)).freeze
   ROLE_APPOINTMENT_FIELDS = (DEFAULT_FIELDS + details_fields(:started_on, :ended_on, :current, :person_appointment_order)).freeze
+  SHARED_NAVIGATION_FIELDS = (DEFAULT_FIELDS + details_fields(:menu_items)).freeze
   STEP_BY_STEP_FIELDS = (DEFAULT_FIELDS + [%i[details step_by_step_nav title], %i[details step_by_step_nav steps]]).freeze
   STEP_BY_STEP_AUTH_BYPASS_FIELDS = (STEP_BY_STEP_FIELDS + %i[auth_bypass_ids]).freeze
   TAKE_PART_PAGE_FIELDS = (DEFAULT_FIELDS + %i[description details]).freeze
@@ -167,6 +168,8 @@ module_function
         fields: CONTACT_FIELDS },
       { document_type: :content_block_pension,
         fields: CONTENT_BLOCK_PENSION_FIELDS },
+      { document_type: :document_collections, # TODO: this would be shared_navigation
+        fields: SHARED_NAVIGATION_FIELDS },
       { document_type: :topical_event,
         fields: DEFAULT_FIELDS },
       { document_type: :organisation,
--- a/lib/expansion_rules.rb
+++ b/lib/expansion_rules.rb
@@ -168,7 +168,7 @@ module_function
         fields: CONTACT_FIELDS },
       { document_type: :content_block_pension,
         fields: CONTENT_BLOCK_PENSION_FIELDS },
-      { document_type: :document_collections, # TODO: this would be shared_navigation
+      { document_type: :navigation, # TODO: this would be shared_navigation
         fields: SHARED_NAVIGATION_FIELDS },
       { document_type: :topical_event,
         fields: DEFAULT_FIELDS },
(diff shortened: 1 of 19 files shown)


spec/lib/link_expansion_spec.rb
```diff
@@ -59,6 +59,76 @@
...
+          links_hash: {
+            documents: [
+              first_document_content_id,
+              second_document_content_id,
+            ],
+          },
+        )
+      end
+
+      it "expands all member documents under the collection" do
```
lauraghiorghisor-tw, line 112: I am not sure I am reading this test correctly. Can I check my understanding?

We want to expand the links on a parent or child document, so we show smth like:
```
{
  details: {},
  links: {
    shared_navigation: [
        {
            content_id: <>,
            links: {
              documents: [
                {
                  content_id: <self/this LP>,
                  title: <>,
                  base_path: <>  
                },
                { <some other doc in nav> }
              ]#end documents
            }#end shared navigation links
        }#end shared navigation item -> can we have multiple?
    ]#end shared navigation array
  }# end links
}
```
Is that right? The test feels like we are expanding the links on the navigation (collection) itself.

lauraghiorghisor-tw: The other bit I am struggling with, and maybe it's worth writing this out explicitly in the spike output - is on whose publication/put-content/both call we are doing the expansion work - a landing page's or the navigation's?

lib/link_expansion.rb
```diff
@@ -97,6 +103,28 @@ def auto_reverse_link(node)
       end
   end
 
+  def document_collection_member_links(collection_content_id)
```
lauraghiorghisor-tw, line 106: I know it's just a spike - would be nice to write this in a more readable way when we implement, the code duplication between this and the logic in the method that calls it, is a bit dizzying.

lib/link_expansion.rb
```diff
@@ -82,11 +82,17 @@ def auto_reverse_link(node)
       return {}
     end
 
+    reverse_link_type = node.link_types_path.first
```
lauraghiorghisor-tw, line 85: When we need to support multiple nesting, do we mix in the recursive logic as well?

lib/expansion_rules.rb
```diff
@@ -71,6 +71,12 @@ def details_fields(*fields)
     ministerial: :ministers,
   }.freeze
 
+  # These expand every member of the reciprocal direct link type from
+  # the source link set, instead of only mirroring the edition being expanded.
+  REVERSE_LINKS_WITH_MEMBER_EXPANSION = {
```
lauraghiorghisor-tw, line 76: So we need this because if we used the normal rule, the expanded nav links in a LP content item would include that single LP content item rather than all?

Just looking at a typical collection https://www.integration.publishing.service.gov.uk/api/content/government/statistics/dcms-economic-estimates-gva-2024-provisional.

Can this also be articulated in a bit more detail in the spike output in layman's terms?

lib/expansion_rules.rb
```diff
@@ -71,6 +71,12 @@ def details_fields(*fields)
...
+    document_collections: :documents,
```
lauraghiorghisor-tw, line 77: Why is the order of the key and the value reversed here - compared to the old collections? 🤔

lauraghiorghisor-tw (commented):
LGTM but a bit difficult to wrap my head around some bits - left some comments.

It would be nice to have in the spike output (Jira card or even better, an outcome section on the confluence decision log):
- A little flow/diagram/stickies/something outlining the exact interaction with pub api, the payloads in whitehall, and the eventual representation in content store after link expansion; using the last version of the terminology (I think your card comment has the latest?)
- Tag Sally about the terminology - I am sure she'd have thoughts.
- A more detailed explanation/ the why of some of the decisions around the new expansion rule and why current rules are insufficient.

spec/lib/link_expansion_spec.rb
lauraghiorghisor-tw: And also here, if we need to support non-child documents in the nav, and a document can be added to multiple navs, this `shared_navigation` link on the doc page would actually be an array of more than one - that would be terribly confusing for figuring out what nav to render. If we ever wanted to inject some nav on non-lp (read: semantic) content types - which honestly, let's not 😅 but I may have read somewhere on some doc that we would consider that.

ChrisBAshton: Hopefully this is clearer in the revamped spike and PR description over at #4085 😅 

As for the `shared_navigation` being an array - yes, that's a bit uncomfortable. Publishing API convention seems to be that these links are always arrays.

We could consider adding a `maxItems: 1` rule, as is done for [primary_publishing_organisation](https://github.com/alphagov/publishing-api/blob/156bd274ad6c954a634b6ce629c35328255d636c/content_schemas/formats/shared/whitehall_edition_links.jsonnet#L9-L12)?

lib/link_expansion.rb
ChrisBAshton: Nope - as the links are flat. The nesting (if we do ever do that) would be in the `details.menu_items` JSON structure, which isn't touched by link expansion / dependency resolution.

lib/expansion_rules.rb
ChrisBAshton: Attempted to explain this in the cleaned up spike at #4085.

lib/expansion_rules.rb
ChrisBAshton: Good question 🤔 Only a spike. But agree for the real implementation it would be nice for this to be consistent with everywhere else in the file.

ChrisBAshton:
Superseded by #4085

TonyGDS:
LGTM, I assume we'd need a new link type like `navigations -> navigation_documents` for the actual implemenation

lib/expansion_rules.rb
lauraghiorghisor-tw: I thought maybe the key implies some precedence like doing the expansion for a content item containing links like the key, or something. 

But I guess the expansion applies both ways for reverse links? In practice, you wouldn't send those payloads both ways as that would defy the point of the reverse link?

spec/lib/link_expansion_spec.rb
lauraghiorghisor-tw: I get it now about the behaviour we had on collections vs this.

- Test still reads weird to me 😅
- I like the maxItems idea 👌


On the point about the array actually having multiple navs for a linked page that is not a landing page (child) type - I am thinking we wouldn't enable rendering of any nav in the FE for our old types, so we simply wouldn't consume that expanded links information. We could just use the `parent` link solution for breadcrumbs if we have to include such pages in nav (end of Q2 concern).

PR #4077, opened 2026-06-02
[PP-7768] Add collector to export Sidekiq lock metrics

brucebolt:
We recently added a dashboard that allows us to see old `sidekiq-unique-jobs` locks that have not been removed after the default expiry time (currently one hour). See https://github.com/alphagov/publishing-api/pull/4045.

This adds a collector that pushes the count of the same into Prometheus.

This metric is used to add a Slack alert in https://github.com/alphagov/govuk-helm-charts/pull/4291.


AgaDufrat (approved):
LGTM and follows existing patterns.

PR #4098, opened 2026-06-10
[DO NOT MERGE] - ADR-014 end state

richardTowers:
Because I can't push branches to alphagov anymore (😭), I've had to create the intermediate PRs on my fork.

We have:

- [The initial PR which adds the ADR](https://github.com/alphagov/publishing-api/pull/4097)
- [Second PR to add BFS implementations of link expansion / dependency resolution](https://github.com/richardTowers/publishing-api/pull/2)
- [Third PR to switch to BFS by default](https://github.com/richardTowers/publishing-api/pull/3)
- [Fourth PR to remove the old code](https://github.com/richardTowers/publishing-api/pull/4)
- [Fifth PR to do some final cleanup](https://github.com/richardTowers/publishing-api/pull/5)

This represents the entire stack of changes as one PR, so you can see overall summaries in terms of lines added / removed, or look through the commits one by one if you prefer. It would be reckless to merge this in one go though - you should pause after the second PR and do some manual testing that the implementations behave the same as each other.


brucebolt:

PR #4099, opened 2026-06-11
Log extra context on Timeout::Error in QueuePublisher 

AgaDufrat:
We’ve seen some intermittent `Timeout::Error errors` raised by exchange.wait_for_confirms which are hard to reproduce.

We want to collect more data when it happen next time, without suppressing the error.  This will lets us see, in the logs, exactly what the state of the connection, channel and broker was at the moment of failure, which can help
pinpoint whether the issue is:
- AMQP connection dropped
- Broker failover
- Network issues
- Something wrong in the code’s connection management

Relevant incident report: https://docs.google.com/document/d/1gUOMQ3kEEYh9IhHEyWohfR67SVFbp8f6rBjqt6pKeDk/edit

### DIFF
--- a/lib/queue_publisher.rb
+++ b/lib/queue_publisher.rb
@@ -74,9 +74,7 @@ private
       success = exchange.wait_for_confirms
       event_type = routing_key.split(".").last
 
-      if success
-        PublishingAPI.service(:statsd).increment("message-sent.#{event_type}")
-      else
+      unless success
         GovukError.notify(
           PublishFailedError.new("Publishing message failed"),
           level: "error",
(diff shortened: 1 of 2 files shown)


lib/queue_publisher.rb
```diff
@@ -85,6 +86,21 @@ def publish_message(routing_key, message_data, options = {})
           },
         )
       end
+    rescue Timeout::Error => e
+      Rails.logger.error(
+        e,
+        extra: {
+          routing_key: routing_key,
+          event_type: event_type,
```
yndajas: **suggestion (non-blocking):** these could probably use the shorthand

```rb
          routing_key:,
          event_type:,
```

yndajas (approved):

AgaDufrat:
I added a rescue to log context in the `ensure` block as well.

PR #4100, opened 2026-06-12
[PP-7732] Allow skipping schema validation in patch links

yndajas:
There are some instances where we want to allow patch links requests without validating against the schema. One example is when an edition's schema name/document type has changed and a publishing app needs to remove links of types not accepted by the new schema. The canonical way of removing all links of a given type is to include in the request's links hash a key of the link type pointing to an empty array. But including keys of unsupported link types is considered invalid by the schema validator. Though the `validate_schema` method only appears to warn when validation fails, we've had reports that the request is actually rejected in such cases. We therefore want to add the ability to skip validation on request

Initially I just used the `payload[:validate_schema]` value directly and gave it a default in the `V2::LinkSetsController` (5e82971fe17eddbe2d288311d7f6d6533952bfdf; like we do with the `payload[:bulk_publishing]` value, but `Commands::V2::PatchLinkSet` is called in other places and we want the default to be `true` regardless of where it's called (unlike `payload[:bulk_publishing]`, which we want to be set intentionally everywhere: https://github.com/alphagov/publishing-api/pull/3861). This therefore
fully manages the defaulting to `true` behaviour within the command itself

Dropped commits:
- e61ffc1af655c0d4146039491b953140685537fb
- 5e82971fe17eddbe2d288311d7f6d6533952bfdf

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -155,7 +155,7 @@ presented edition and [warnings](#warnings).
 - `update_type` *(optional)*
   - Accepts: "major", "minor", "republish"
   - It is acceptable to send a "minor" update for the first ever draft
-- `bulk_publishing` *(optional, default: false)*
+- `bulk_publishing` *(required)*
   - Set this to true when making multiple requests. Publishing API will use a
     lower priority queue to avoid delays to standard publishing activity.
 
@@ -525,10 +525,10 @@ in the request is preserved.
     }
   }
 ```
+- `bulk_publishing` *(required)*
+  - Set this to true when making multiple requests. Publishing API will use a lower priority queue to avoid delays to standard publishing activity.
 - `previous_version` *(optional, recommended)*
   - Used to ensure that we are updating the current version of the link set.
-- `bulk_publishing` *(optional, default: false)*
-  - Set this to true when making multiple requests. Publishing API will use a lower priority queue to avoid delays to standard publishing activity.
 - `validate_schema` *(optional, default: true)*
   - Set to false to skip validation of links against the content schema for
     the current payload. After changing a document's document type, allows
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -529,6 +529,11 @@ in the request is preserved.
   - Used to ensure that we are updating the current version of the link set.
 - `bulk_publishing` *(optional, default: false)*
   - Set this to true when making multiple requests. Publishing API will use a lower priority queue to avoid delays to standard publishing activity.
+- `validate_schema` *(optional, default: true)*
+  - Set to false to skip validation of links against the content schema for the
+    latest edition. After changing a document's document type, allows for
+    removal of links of types unsupported.
```
brucebolt: Nitpick: I think the wording `latest edition` is a bit misleading here as linkset links don't have editions. Maybe this could be reworded to:

"Set to false to skip validation of links against the content schema for the current payload. This can be used prior to changing a document's document type in the next edition."

brucebolt (commented):
Looks good to me.  I've just added a minor comment to the docs.

Will hold off from merging until the Whitehall team have confirmed they are happy with the change.

docs/api.md
yndajas: I didn't look into the method in detail, but that's referring to the way the schema name is derived per the code below. It probably is a little confusing though, so feel free to amend!

https://github.com/alphagov/publishing-api/blob/f1b5c2b0a34a0b81e6017caed3d6ac81481efedc/app/commands/v2/patch_link_set.rb#L164-L168

yndajas: Though isn't the issue we're solving specifically when the document type has _already_ been updated? As as I understand it, the validations wouldn't fail before the document type has been updated

> This can be used _prior_ to changing a document's document type in the _next_ edition.

brucebolt: That's very true, I'd forgotten the ordering of events.

Maybe: "Set to false to skip validation of links against the content schema for the current payload. After changing a document's document type, allows for removal of links of types unsupported."?

yndajas:
Marked as draft while we await clarification from Whitehall on how the current code led to an error

PR #4117, opened 2026-07-07
[PP-7833] Disable GraphQL precedence tests

yndajas:
Per the comments, we're disabling these for now given our GraphQL functionality is not in active development and they are therefore wasting CI resources. This adds a check against GraphQL and link expansion files to remind us to re-enable them if we work on code that these tests are likely to exercise. This change can be reverted when that is the case


AgaDufrat (dismissed):
LGTM. 
I'd probably just remove them instead of leaving a long comment. We can always revert the commit and update the setup, if things change.

yndajas:
> LGTM. I'd probably just remove them instead of leaving a long comment. We can always revert the commit and update the setup, if things change.

Much cleaner!

AgaDufrat (approved):

PR #4141, opened 2026-08-03
[PP-7913] Do not send substitute message for blocking editions

brucebolt:
In b99667da09fa83642bc20625730eac8a05fa692f, some changes were made to broadcast a message to the message queue for substitute editions.

Within the `SubstitutionHelper`, we added a broadcast message after the `substitute` method is called on the "blocking editions".

However the `subsitute` method does not actually result in any changes being pushed downstream (i.e. to Content Store) but this broadcast message results in the document being removed from Search API (which indexes by base path, not Content ID).

It then means the document is present in Content Store, but has completely been removed from Search API v1 and cannot be found.

Removing this call, as it is not needed at this location.

There is [already code present](https://github.com/alphagov/publishing-api/blob/0674799792665b1ff12fd6d3cc8279a631beb2f5/app/downstream_payload.rb#L85) that broadcasts the substitute message at the correct time.

This resolves an issue where sometimes organisations are removed from Search API v1, despite having a live edition present in Content Store.

### DIFF
--- a/app/substitution_helper.rb
+++ b/app/substitution_helper.rb
@@ -30,7 +30,6 @@ module SubstitutionHelper
         discard_draft(blocking_edition, downstream, nested, callbacks)
       else
         blocking_edition.substitute
-        substitute_message(blocking_edition)
       end
     end
   end
--- a/spec/commands/v2/publish_spec.rb
+++ b/spec/commands/v2/publish_spec.rb
@@ -438,10 +438,6 @@ RSpec.describe Commands::V2::Publish do
 
       it "unpublishes the edition which is in the way" do
         expect(PublishingAPI.service(:queue_publisher)).to receive(:send_message).with(
-          hash_including(content_id: other_edition.content_id, document_type: "substitute"),
-          hash_including(event_type: "unpublish"),
-        )
-        expect(PublishingAPI.service(:queue_publisher)).to receive(:send_message).with(
           hash_including(content_id: draft_item.content_id),
           hash_including(event_type: "major"),
         )


app/substitution_helper.rb
```diff
@@ -30,7 +30,6 @@ def clear!(
         discard_draft(blocking_edition, downstream, nested, callbacks)
       else
         blocking_edition.substitute
-        substitute_message(blocking_edition)
```
yndajas, line 33: **question:** I'm not very familiar with substitutions so I'm struggling a bit for context here. Where does this helper get called and why is the presenter the right place for broadcasting the message instead (does it get called by a different command or later in the same command?)?

yndajas, line 33: **question:** do we still need the method definition if this call is removed? Is it used elsewhere?

yndajas, line 33: **suggestion:** 

> It then means the document is present in Content Store, but has completely been removed from Search API v1 and cannot be found.

Might be worth mentioning what the impact of that was - i.e. the missing organisation - and what would users see when?

> However the `subsitute` method does not actually result in any changes being pushed downstream

Per other comment, I'm a little out of context here, so I'm curious why it doesn't result in downstream changes - are they triggered later in the command?

brucebolt: This is still called from within the `Publish` command.

brucebolt: At the point of publishing a content item, we call the `SubstitutionHelper` to remove all published items of the same locale and base path from Publishing API's database.  In doing this, we look for both published and unpublished editions that match the same base path.

If we match published or unpublished editions (as in this case), which have to be of the type defined in `SUBSTITUTABLE_DOCUMENT_TYPES`, we call the `substitute` method on the model which marks the edition as `unpublished` with a type of `substitute`. But nothing gets sent downstream to Content Store as part of this.  The newest update to the edition will eventually get sent to Content Store and replace the old one.

However this line was broadcasting a message to the queue, telling Search API v1 to unpublish the base path.  This isn't correct, as we want the base path to continue to be indexed.  Whilst another request is made almost immediately afterwards to index the document again, the ordering of messages being processed in Search API v1 means the deletion sometimes happens last.

callumknights: Jumping in here as despite being the original author of this Substitution code, it's been a while and I'm struggling to remember the downstream implications. Is the basic problem here more the ordering of message processing by Search API V1. Cause in my head telling Search to unpublish is still correct, but that happens as part of the Presenter messaging, am I right in saying that the double application I added meant an extra call to Search to unpublish could happen later down the line, and sometimes be the last instruction? Resulting in the problem we've had recently?

brucebolt: I think when we're replacing a document with another of the same base path, we don't want to unpublish from Search API (or other message queue consumers) at all.  We just need to broadcast the change to the queue.

callumknights: Fantastic, thanks for the clarification.

yndajas (approved):

callumknights (approved):

PR #4128, opened 2026-07-21
Add validators to ensure base_path conforms to RFC-192

KludgeKML:
Uses the new GdsApi::Validators::BasePathValidator to check base_path validity on:
- Content put
- Path reservation
- Publishing intent put

...against [RFC-192 base path rules](https://github.com/alphagov/govuk-rfcs/blob/d4b4087ca64b522dc23afaee386265b211f542af/rfc-192-restrict-base-paths-to-a-smaller-character-set.md)

IMPORTANT - Do not merge before:
- [x] https://github.com/alphagov/gds-api-adapters/pull/1411 has been merged
- [x] a gem release including that code has happened 

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #4154, opened 2026-08-10
[PNP-10123] RFC-192 compliance

KludgeKML:
Adds RFC-192 compliance to publishing api.

This update checks against RFC-192 compliance in two places:
- In the put content command (where it is applied to all base paths except those on excluded content items - redirects, contacts, governments or content_blocks)
- In the path reservation class (where it is applied to all base paths except if the publishing app is short-url-manager).

This allows redirects to be created with non-RFC-192 characters in them (desirable), but blocks with a 422 any attempt to publish documents with a base_path present if it does not confirm to the RFC.

After merging this, publishing apps may (if they allow non-RFC-192 characters in their own slug/base_path creation) receive an error when pushing to publishing-api. Publishing apps should ensure they can either handle that error as gracefully as possible or use the same validation code internally (GdsApi::Validators::BasePathValidator).

https://gov-uk.atlassian.net/browse/PNP-10123

Previous PRs:
- https://github.com/alphagov/publishing-api/pull/4128
- https://github.com/alphagov/publishing-api/pull/4138 (reverting 4128 - revert occured because it blocked publishing content blocks, which at the time still had a path).

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

### DIFF
--- a/app/commands/v2/put_content_validator.rb
+++ b/app/commands/v2/put_content_validator.rb
@@ -1,6 +1,13 @@
 module Commands
   module V2
     class PutContentValidator
+      FORMATS_WITHOUT_BASE_PATH_VALIDATION = %w[
+        contact
+        content_block
+        government
+        redirect
+      ].freeze
+
       def initialize(payload, put_content)
         @payload = payload
         @put_content = put_content
@@ -9,6 +16,7 @@ module Commands
       def validate
         validate_schema
         validate_publishing_app
+        validate_base_path
       end
 
     private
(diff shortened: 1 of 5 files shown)


brucebolt (approved):
This looks good to me.  I wonder if a link to the previous PRs (where similar was added then reverted) would be useful in the PR description, so the history is easy to find?

PR #4168, opened 2026-08-18
[WHIT-3839] - Remove plan for change landing page schema

jamiestamp:
Follows on from the Whitehall PRs (alphagov/whitehall#11707, alphagov/whitehall#11708) which delete the Plan for Change landing page code and data.

Here we retire the no longer needed schema, following the [guidance for retiring a schema](https://docs.publishing.service.gov.uk/repos/publishing-api/content_schemas/retiring-a-schema.html):

- all documents of this type have been unpublished with redirects to the National Archives
- the format, its example, and its GraphQL query are removed, and the document type is dropped from `allowed_document_types.yml`

The commits are split so the first contains the hand-written files, and the second captures the generated files that change as a result of running `bundle exec rake build_schemas`.

Existing editions are left in the database, as with alphagov/publishing-api#3600 (`service_sign_in`). They are unpublished, so they are presented downstream via `RedirectPresenter` and never reference the retired schema.


ChrisBAshton (approved):
LGTM 👍 CI should pass when the dependent PRs are merged.

brucebolt (approved):

PR #4182, opened 2026-09-02
Add further schemas to list of schemas without a base path

brucebolt:
The list of schemas without a base path is incomplete, so updating it to include all schemas that currently do not have a base path.

This was obtained using the following query in Publishing API:

```rb
Edition.live.where(base_path: nil).pluck(:schema_name).uniq.sort
=> ["contact", "content_block", "external_content", "facet", "facet_group", "facet_value", "link_collection", "role", "role_appointment", "world_location"]
```

`facet_group` and `facet_value` have been excluded, as they do not have a schema, so will not be publishable anyway.  These are most likely orphaned documents.

These were missed in 819bb79d3997ee3c64d1080f85ebe17120ccd275, but inspired by https://github.com/alphagov/publishing-api/pull/4180 and https://github.com/alphagov/publishing-api/pull/4181.

### DIFF
--- a/app/commands/v2/put_content_validator.rb
+++ b/app/commands/v2/put_content_validator.rb
@@ -1,7 +1,7 @@
 module Commands
   module V2
     class PutContentValidator
-      FORMATS_WITHOUT_BASE_PATH_VALIDATION = %w[
+      SCHEMAS_WITHOUT_BASE_PATH_VALIDATION = %w[
         contact
         content_block
         external_content
@@ -61,7 +61,7 @@ module Commands
       end
 
       def validate_base_path
-        return if FORMATS_WITHOUT_BASE_PATH_VALIDATION.include?(payload[:schema_name])
+        return if SCHEMAS_WITHOUT_BASE_PATH_VALIDATION.include?(payload[:schema_name])
 
         base_path_validator = GdsApi::Validators::BasePathValidator.new(payload[:base_path])
         return if base_path_validator.valid?
--- a/app/commands/v2/put_content_validator.rb
+++ b/app/commands/v2/put_content_validator.rb
@@ -6,8 +6,11 @@ module Commands
         content_block
         external_content
         government
+        facet
+        link_collection
         redirect
         role
+        role_appointment
         world_location
       ].freeze
 


yndajas (commented):
I reverse checked the schema and found that some have examples with base paths, so we might need to think about whether those should still have validation:

```rb
candidate_schema = Edition.live.where(base_path: nil).pluck(:schema_name).uniq.sort
=> ["contact", "content_block", "external_content", "facet", "facet_group", "facet_value", "link_collection", "role", "role_appointment", "world_location"]

candidate_schema.filter { Edition.live.where(schema_name: it).where.not(base_path: nil).limit(1).any? }
=> ["contact", "content_block", "role", "world_location"]
```

Edit: quite possible that those examples are data issues

brucebolt:
> I reverse checked the schema and found that some have examples with base paths, so we might need to think about whether those should still have validation:
> 
> ```ruby
> candidate_schema = Edition.live.where(base_path: nil).pluck(:schema_name).uniq.sort
> => ["contact", "content_block", "external_content", "facet", "facet_group", "facet_value", "link_collection", "role", "role_appointment", "world_location"]
> 
> candidate_schema.filter { Edition.live.where(schema_name: it).where.not(base_path: nil).limit(1).any? }
> => ["contact", "content_block", "role", "world_location"]
> ```

Ooh, that's interesting.  I've looked in the schema JSON files and some of these have the base path marked as optional.

Will defer to @KludgeKML (as author of the validator) for an opinion on this.  Maybe we should validate these only if a base path is present in the payload?

yndajas:
If you limit it to published editions, the list gets smaller:

```rb
candidate_schema.filter { Edition.live.where(schema_name: it, state: "published").where.not(base_path: nil).limit(1).any? }
=> ["content_block", "role"]
```

brucebolt:
I _think_ content blocks definitely aren't supposed to have a base path (at least one isn't being added in [the presenter](https://github.com/alphagov/content-block-manager/blob/main/app/presenters/publishing_api/content_block_presenter.rb)).

Looks like [roles vary](https://github.com/alphagov/whitehall/blob/21e5f002cd439c1a41b35f8edac54371839f3adf/app/presenters/publishing_api/role_presenter.rb#L42-L52), depending on whether the role is ministerial or not.

KludgeKML:
> > I reverse checked the schema and found that some have examples with base paths, so we might need to think about whether those should still have validation:
> > ```ruby
> > candidate_schema = Edition.live.where(base_path: nil).pluck(:schema_name).uniq.sort
> > => ["contact", "content_block", "external_content", "facet", "facet_group", "facet_value", "link_collection", "role", "role_appointment", "world_location"]
> > 
> > candidate_schema.filter { Edition.live.where(schema_name: it).where.not(base_path: nil).limit(1).any? }
> > => ["contact", "content_block", "role", "world_location"]
> > ```
> 
> Ooh, that's interesting. I've looked in the schema JSON files and some of these have the base path marked as optional.
> 
> Will defer to @KludgeKML (as author of the validator) for an opinion on this. Maybe we should validate these only if a base path is present in the payload?

Content blocks definitely shouldn't have base paths now, but the other cases where base paths may or may not exist are tricky. I think in retrospect perhaps I made a mistep with requiring the base path no to be nil in the validator - I think it's created more problems than it solves. Am I right in thinking that nil base paths mean that an item will never get moved into content store?

brucebolt:
> Am I right in thinking that nil base paths mean that an item will never get moved into content store?

Yes, [that's correct](https://github.com/alphagov/publishing-api/blob/daee180d6f0cad6529834f3ca93474b74a1509b2/app/sidekiq/downstream_live_job.rb#L43).  No base path means we don't attempt to send the edition to Content Store.

brucebolt:
Closing, as this will be resolved by version 103.5.0 of gds-api-adapters.

PR #4193, opened 2026-09-09
[PP-7716] Pact v2 migration

geek4good:
Migrate pact verification tests to pact v2 ([Jira ticket](https://gov-uk.atlassian.net/browse/PP-7716))

Upgrade pact from v1.x to v2.x and rewrite the tests for the new Rust-core based API.

- include Pact tests in the CI run
- bump pact to ~> 2.0 and add pact-ffi (v2 verification runs on the Pact Rust core via FFI) and the rspec meta-gem (pact 2 no longer declares it as a runtime dependency)
- replace the v1 DSL under spec/service_consumers with the v2 RSpec DSL under spec/pact
- move per-interaction setup into Pact.configuration global hooks so auth stubbing and database cleaning run for every interaction
- define the pact:verify task explicitly in the Rakefile (pact v2 no longer ships lib/pact/tasks.rb)

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -24,7 +24,7 @@ Pact.service_provider "Publishing API" do
       base_url = ENV.fetch("PACT_BROKER_BASE_URL", "https://govuk-pact-broker-6991351eca05.herokuapp.com")
       url = "#{base_url}/pacts/provider/#{url_encode(name)}/consumer/#{url_encode(consumer_name)}"
 
-      pact_uri "#{url}/versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'master'))}"
+      pact_uri "#{url}/versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'branch-main'))}"
     end
   end
 end


spec/service_consumers/pact_helper.rb
```diff
@@ -24,7 +24,7 @@
       base_url = ENV.fetch("PACT_BROKER_BASE_URL", "https://govuk-pact-broker-6991351eca05.herokuapp.com")
       url = "#{base_url}/pacts/provider/#{url_encode(name)}/consumer/#{url_encode(consumer_name)}"
 
-      pact_uri "#{url}/versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'master'))}"
+      pact_uri "#{url}/versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'branch-main'))}"
```
yndajas, line 27: **note:** in CI we set the environment variable to the correct value:
- https://github.com/alphagov/publishing-api/blob/46c156f217b55c2efe726cb7a306e10cdbfe7bad/.github/workflows/pact-verify.yml#L31-L34
- https://github.com/alphagov/publishing-api/blob/46c156f217b55c2efe726cb7a306e10cdbfe7bad/.github/workflows/pact-verify.yml#L65-L66

But good to fix this for local usage!

spec/pact/consumers/gds_api_adapters_spec.rb
```diff
@@ -0,0 +1,661 @@
...
+      WebMock.reset!
+      DatabaseCleaner.clean_with :truncation
+      GDS::SSO.test_user = FactoryBot.create(:user, permissions: %w[signin view_all])
+    end
+  end
+
+  http_pact_provider "Publishing API", opts: {
+    app: PactExampleOrgHost.new(Rails.application),
+    http_port: 9292,
+    pact_uri: ENV["PACT_URI"],
```
yndajas, line 37: **question:** I'm not very familiar with this config, but do we need to account for `ENV["PACT_URI"]` potentially being `nil`? https://github.com/alphagov/publishing-api/blob/46c156f217b55c2efe726cb7a306e10cdbfe7bad/spec/service_consumers/pact_helper.rb#L21

spec/pact/consumers/gds_api_adapters_spec.rb
```diff
@@ -0,0 +1,661 @@
...
+    broker_url: ENV.fetch("PACT_BROKER_BASE_URL", "https://govuk-pact-broker-6991351eca05.herokuapp.com"),
+    consumer_name: "GDS API Adapters",
+    consumer_version_selectors: [
+      { branch: ENV.fetch("PACT_CONSUMER_VERSION", "branch-main").delete_prefix("branch-") },
+    ],
+    log_level: :info,
+    fail_if_no_pacts_found: true,
+  }
+
+  provider_state "a publish intent exists at /test-intent" do
```
yndajas, line 47: **question:** I diffed the provider state code. It mostly looks the same except the following states are missing:
- "the content item bed722e6-db68-43e5-9079-063f623335a7 is at lock version 3"
- "the linkset for bed722e6-db68-43e5-9079-063f623335a7 is at lock version 3"
- "there is content with document_type 'taxon' for multiple publishing apps"

Are those intentional?

.github/workflows/ci.yml
```diff
@@ -83,12 +83,21 @@ jobs:
           TEST_DATABASE_URL: ${{ steps.setup-postgres.outputs.db-url }}
         run: bundle exec rake spec
 
+      # Pact v2 specs don't run during default RSpec runs, so the consumer
+      # pact must be generated explicitly before it can be uploaded
+      - name: Generate pact test artifact
```
yndajas, line 88: **suggestion (non-blocking):** I wonder if it would be slightly clearer to say "Verify consumer pacts" or "Run consumer Pact tests"? Given the main thing we want to do is confirm the pacts are met - the artifact is just there for inspection in GitHub when they aren't?

yndajas (commented):
Just a few small queries, but generally looks good. I'm not familiar enough with the v2 approach to be able to verify the more subtle differences in setup in the specs compared with v1, but the tests are passing and nothing looks unusual!

spec/pact/consumers/gds_api_adapters_spec.rb
geek4good: If `ENV["PACT_URI"]` is `nil`, it falls back to the pact broker. So, tests are still run.

spec/pact/consumers/gds_api_adapters_spec.rb
geek4good: I'm not sure how they slipped through the cracks, but thank you for double-checking and pointing out what I've missed. I've now added the previously three missing tests in fe05e77.

.github/workflows/ci.yml
geek4good: You're right. I was merely focused on the side effect, when naming it after what's actually happening makes more sense. Fixed in 1055997.

spec/pact/consumers/gds_api_adapters_spec.rb
yndajas: I used diffchecker.com again and it looks like some typos have crept in - `satele` and `escapte` in a few places - and `Content` became `content` in a few others 👀

yndajas:
Other than https://github.com/alphagov/publishing-api/pull/4193#discussion_r4084690959, looks good!

spec/pact/consumers/gds_api_adapters_spec.rb
geek4good: Oh my… How have those tests not failed?
Great idea to diff the old and new test files. 👍 

Anyway, it's fixed now.

lib/tasks/pact.rake
```diff
@@ -16,8 +16,25 @@ end
 
 require "rspec/core/rake_task"
 
-desc "Verify the GDS API Adapters pacts against this application"
-RSpec::Core::RakeTask.new("pact:verify_v2") do |task|
+desc "Verify the consumer pacts against this application"
```
brucebolt: Are the descriptions for these two rake tasks the right way round?

